PR #341, opened 2014-12-18
Reference-Sequence Conversion Improvements for Merging

tmooney:
* Pass BED "track", "browser", etc. lines unmodified.
* A reference can now indicate it is a combination of multiple other references.
* Remove redundant print in `genome feature-list merge` since it will be printed by the import command.


iferguson90:
+1

PR #348, opened 2014-12-18
Disable fillmd/BamQc for bsmap

iferguson90:
We were having trouble running `fillmd` on `bsmap` alignments. This disables `fillmd` and BamQc when the aligner is set to `bsmap`.

Would like input from @jasonwalker80 before merging.


acoffman:
+1

iferguson90:
This has been stewing for a while. :stew: Merging! :ramen:

PR #350, opened 2014-12-19
Update GMT::Bsmap::MethRatio

iferguson90:
Supersedes #244.

- [x] Pull test data from Bsmap
- [x] Update tests to run `methratio.py`
- [x] Verify that removing default version doesn't break anything


chrisamiller:
In response to our backchannel messages, I can't see any reason why removing the default version would break things. +1

lib/perl/Genome/Model/Tools/Bsmap/MethRatio.t
```diff
@@ -4,10 +4,90 @@ use strict;
...
+
+        my $fasta = Genome::Model::Build::ReferenceSequence->get(
+            name => $mapping{$ref_short_name}
+        )->cached_full_consensus_path('fa');
+
+        is($test_obj->_reference_fasta, $fasta, 'Got correct fasta object');
+    }
+}
+
+{
```
tmooney: Should these blocks be [subtests](http://search.cpan.org/~exodist/Test-Simple-1.001014/lib/Test/More.pm#subtest)?

acoffman:
+1 :shipit:

tmooney:
+1 :pineapple:

PR #355, opened 2014-12-19
clinseq-converge-stats update

gatoravi:
Add a new file to the converge stats command, this is the Stats.tsv from summarizing SNVIndelReports. To help accomplish this make base quality and mapping quality a parameter of Base.pm under Converge.


acoffman:
+1

PR #357, opened 2014-12-19
iterate_file_lines() 

brummett:
Add iterate_file_lines() to Genome::Sys.  This method is an abstraction to work through a file one line at a time.

Convert Genome::Model::Build::Command::DetermineError to use it.

There are probably other places it could be used.


acoffman:
+1

PR #360, opened 2014-12-22
Bwamem stream

tabbott:
Add the bwamem-stream aligner module which can potentially be a lot faster and use a lot less /tmp space than the bwamem aligner.


tabbott:
Looks like there is some more wiring to do with bedtools_version; my test build failed in the merge step due to not specifying bedtools_version when trying to look up the alignment result. Investigating...

iferguson90:
+1
If test builds are succeeding I think we should merge this before it gets any larger.

tabbott:
This has been open long enough to accumulate some minor conflicts. I'll rebase onto master, resolve them, and push to master to avoid a merge commit. Closing.

PR #366, opened 2014-12-24
Minor PolymuttDenovo Filter Improvements

tmooney:
More cleanup on the side of the ongoing SR project... replace a few `print STDERR` lines and an instance of `cp` in backticks.


acoffman:
+1

PR #369, opened 2014-12-29
Don't generate file summaries if we can't reach the target directory.

tmooney:
Otherwise we end up with summaries for whatever arbitrary directory we happened to be in.


acoffman:
+1

PR #372, opened 2014-12-31
get_alignments() defers to existing method to get each alignment.

tmooney:
The other method handles segment information, so it should work in more situations.


acoffman:
+1

PR #373, opened 2014-12-31
Remove WithSavedResults from Genome::Db::Cosmic::Command::Import

tmooney:
As part of the ongoing SoftwareResult project `WithSavedResults` must be changed.  Since this importer is non-functional (`execute()` ends in `die()`), I propose to remove its inheritance instead of updating it to the new interface.

[My preferred course of action would be to merge PR #207, thus removing this module completely.]


acoffman:
+1

jweible:
+1

PR #374, opened 2014-12-31
Add functionality to compare Genome::Process output files

susannasiebert:
This is a first step towards tackling https://jira.gsc.wustl.edu/browse/AT-680. This PR implements a compare_output subroutine for Genome::Process that would give a similar output to G::M::B::compare_output.

### DIFF
--- a/lib/perl/Genome/Process.pm
+++ b/lib/perl/Genome/Process.pm
@@ -600,7 +600,7 @@ sub unique_results {
 sub result_with_label {
     my ($self, $label) = validate_pos(@_, 1, 1);
 
-    my @results = grep {$_->users(label => $label)} $self->unique_results;
+    my @results = grep {my @users = $_->users(label => $label); scalar(@users) > 0} $self->unique_results;
     if (scalar(@results) != 1) {
         die sprintf("Found (%d) results with label (%s), but expected only one, they are: %s",
             scalar(@results), $label, pp(map {$_->id} @results));
--- a/lib/perl/Genome/Process.pm
+++ b/lib/perl/Genome/Process.pm
@@ -630,5 +630,31 @@ sub result_is_on_cle_disk_group {
     return Genome::Disk::Group::Validate::GenomeDiskGroups::is_cle_disk_group_name($disk_group_name);
 }
 
+sub compare_output {
+    my ($self, $other_process_id) = @_;
+    my $process_id = $self->id;
+
+    unless (defined ($other_process_id)) {
+        die $self->error_message('Require process ID argument!');
+    }
+    my $other_process = Genome::Process->get($other_process_id);
+    unless ($other_process) {
+        die $self->error_message('Could not get process for ID (%s)', $other_process_id);
+    }
+
+    unless ($self->class eq $other_process->class) {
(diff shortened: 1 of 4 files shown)


lib/perl/Genome/Process.pm
```diff
@@ -630,5 +630,31 @@ sub result_is_on_cle_disk_group {
...
+    unless (defined ($other_process_id)) {
+        die $self->error_message('Require process ID argument!');
+    }
+    my $other_process = Genome::Process->get($other_process_id);
+    unless ($other_process) {
+        die $self->error_message('Could not get process for ID (%s)', $other_process_id);
+    }
+
+    unless ($self->class eq $other_process->class) {
+        die $self->error_message('Processes (%s) annd (%s) are not of the same type', $process_id, $other_process_id);
```
nnutter: Typo, "annd".

lib/perl/Genome/Process.pm
```diff
@@ -600,7 +600,7 @@ sub unique_results {
 sub result_with_label {
     my ($self, $label) = validate_pos(@_, 1, 1);
 
-    my @results = grep {$_->users(label => $label)} $self->unique_results;
+    my @results = grep {my @users = $_->users(label => $label); scalar(@users) > 0} $self->unique_results;
```
nnutter: Another way to force list context is to assign to `()`, for example,

```
my @results = grep { scalar(() = $_->users(label => $label)) > 0 } $self->unique_results;
```

nnutter: However, it can probably be shortened to just,

```
my @results = grep { () = $_->users(label => $label) } $self->unique_results;
```

lib/perl/Genome/VariantReporting/Process/CreateReport.pm
```diff
@@ -83,6 +84,38 @@ sub create_disk_allocation {
...
+        my $other_output_dir = $other_process->output_directory($report);
+        my $comparison = File::DirCompare->compare(
+            $other_output_dir,
+            $output_dir,
+            sub {
+                my ($other_file, $file) = @_;
+                if (! $file) {
+                    $diffs{$other_file} = sprintf(
+                        'no file %s found for process %s and report %s with output directory %s',
+                        File::Spec->abs2rel($other_file, $other_output_dir), $self->id, $report, $output_dir
```
nnutter: This module should `use File::Spec`.

lib/perl/Genome/VariantReporting/Process/CreateReport.pm
```diff
@@ -83,6 +84,38 @@ sub create_disk_allocation {
     return $rv;
 }
 
-
+sub _compare_output_files {
+    my ($self, $other_process) = @_;
+
+    my %diffs;
```
nnutter: It seems more intuitive to me for `%diffs` to be keyed off the subpath (common path) of the two paths, e.g. `File::Spec->abs2rel($other_file, $other_output_dir)` rather than the absolute paths of either `$file` or `$other_file`.  Alternatively you might just store them in an array if they keys don't really matter at all.

susannasiebert:
This PR was superseded by https://github.com/genome/genome/pull/391

PR #375, opened 2014-12-31
Convert GMT::Htseq::Count to a G:C::DelegatesToResult

tmooney:
Between this and #373 no classes will be left using `Genome::Command::WithSavedResult` so it can be removed.

### DIFF
--- a/lib/perl/Genome/Model/Tools/Htseq/Count.pm
+++ b/lib/perl/Genome/Model/Tools/Htseq/Count.pm
@@ -4,8 +4,7 @@ use warnings;
 use Genome;
 
 class Genome::Model::Tools::Htseq::Count {
-    is => 'Genome::Command::WithSavedResults',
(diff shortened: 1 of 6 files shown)


acoffman:
+1

lib/perl/Genome/Model/Tools/Htseq/Count.pm
```diff
@@ -71,260 +69,31 @@ class Genome::Model::Tools::Htseq::Count {
...
-    else {
-        die "no _htseq_stranded_param implementation for version " . $self->result_version;
-    }
+sub result_class {
+    return __PACKAGE__ . '::Result';
 }
 
-sub _merge_v2 { _merge_v1(@_) }
-sub _merge_v1 {
+sub input_hash {
```
tmooney: That sounds good to me.  I'll merge this PR and open a new one that moves it up and removes other subclass implementations where they're not needed.

PR #376, opened 2014-12-31
Query reference alignment build directly for inputs.

tmooney:
The model inputs might have changed from what's recorded on the build as we progress.  The builds then cannot be trusted.


acoffman:
+1

ebelter:
+1

PR #377, opened 2015-01-02
Merged the TCGA importer into the basic importer. More PRs to come to smooth out the interface.

ebelter:


lib/perl/Genome/InstrumentData/Command/Import/Basic.pm
```diff
@@ -160,26 +161,23 @@ sub _resolve_original_format {
 sub _resolve_instrument_data_properties {
     my $self = shift;
 
-    my $properties = {};
-    if ( $self->instrument_data_properties ) {
-        my $helpers = Genome::InstrumentData::Command::Import::WorkFlow::Helpers->get;
-        $properties = $helpers->key_value_pairs_to_hash( $self->instrument_data_properties );
-        return if not $properties;
+    my $class = 'Genome::InstrumentData::Command::Import::WorkFlow::ResolveInstDataProperties';
+    if ( $self->import_source_name =~ /^CgHub$/i ) {
```
nnutter: Should this special case be documented on the `import_source_name` `doc` string?

lib/perl/Genome/InstrumentData/Command/Import/Basic_cghub.t
```diff
@@ -0,0 +1,82 @@
...
+
+my @instdata =  $cmd->_new_instrument_data;
+is(@instdata, 1, 'create instrument data');
+my $instdata = $instdata[0];
+
+# Bam
+my $bam_path = $instdata->bam_path;
+ok(-s $bam_path, "bam exists");
+my $expected_bam_path = File::Spec->join($data_dir, 'expected.bam');
+is(File::Compare::compare($bam_path, $expected_bam_path), 0, 'bam matches');
```
nnutter: I was going to suggest `Genome::Utility::Test::compare_ok` but it looks like you can't use that on non-text files.

lib/perl/Genome/InstrumentData/Command/Import/Basic.pm
ebelter: Yes. I am working on documenting the whole process. Getting these properties from a DL'd metadata file is from the TCGA importer. I hope to have a better system of creating work flows for importing. For now, the TCGA importer needs to go away. It allows importing of bams that are unusable by the ref align pipeline.

ebelter: You are correct - I am updating the doc for this property.

lib/perl/Genome/InstrumentData/Command/Import/Basic_cghub.t
nnutter: I don't know if comparing BAM files is already a common practice or not but comparing compressed files may not work since the files could differ bit-for-bit while still having the same content.  We may need to compare them in SAM format.  It looks like `File::Compare::compare` would support this by opening a handle to the BAM, piping through `samtools view`, and passing in the handles instead of the paths.  Again, I don't know if this is actually necessary.

ebelter: The best practice wold be to uncompress the file [or convert to text (sam)], then compare. This is done in multiple areas. If you'd like, I can make a G::T method to do this and update all the tests in an additional PR.

lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/ResolveInstDataPropertiesFromCgHub.pm
```diff
@@ -0,0 +1,166 @@
...
+use Genome::Model::Tools::CgHub::Query;
+require File::Spec;
+require File::Temp;
+use Path::Class;
+
+class Genome::InstrumentData::Command::Import::WorkFlow::ResolveInstDataPropertiesFromCgHub { 
+    is => 'Genome::InstrumentData::Command::Import::WorkFlow::ResolveInstDataProperties',
+};
+
+sub metadata_file {
```
nnutter: I think the names for `metadata_file` and `_resolve_metadata_file` could use clarification since both are doing logic to find a metadata file.

lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/ResolveInstDataPropertiesFromCgHub.pm
```diff
@@ -0,0 +1,166 @@
...
+    my $self = shift;
+    my @sources = $self->sources;
+    my $source_dir = Path::Class::file( $sources[0] )->dir;
```
nnutter: Should we check all `@sources` dirs?

nnutter:
+1

lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/ResolveInstDataPropertiesFromCgHub.t
```diff
@@ -0,0 +1,67 @@
...
+        source_md5 => '1fb1b9a352c04623ec1422689754e83a',
+        source_size => 45644,
+    },
+    'resolved_instrument_data_properties',
+);
+
+# ERRORS to test
+# required args not in metadata and not given
+# target region
+# feature list
```
tmooney: Are these comments suggestions for future tests?

tmooney:
+1

nnutter:
This is pending documentation being added to Confluence.

ebelter:
CI, test this.

ebelter:
Still needing documentation completed, but it will be done by the time this goes live.

PR #378, opened 2015-01-04
Make library_names and sample_names manditory inputs

davidlmorton:
Some care has to be taken for the SoftwareResult subclasses (reports)
since they have to be listed as 'input' or 'param' or the SoftwareResult
infrastructure will yell about ambiguous properties.

### DIFF
--- a/lib/perl/Genome/VariantReporting/Command/Wrappers/ModelPair.t
+++ b/lib/perl/Genome/VariantReporting/Command/Wrappers/ModelPair.t
@@ -14,10 +14,21 @@ use File::Slurp qw(read_file);
 use Genome::VariantReporting::Command::Wrappers::TestHelpers qw(get_build compare_directories);
 use Genome::VariantReporting::Framework::TestHelpers qw(test_dag_xml);
 use Genome::Utility::Test qw(compare_ok);
+use Sub::Install qw(reinstall_sub);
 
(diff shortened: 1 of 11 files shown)


lib/perl/Genome/VariantReporting/Command/Wrappers/ModelPair.t
```diff
@@ -14,11 +14,22 @@ use File::Slurp qw(read_file);
 use Genome::VariantReporting::Command::Wrappers::TestHelpers qw(get_build compare_directories);
 use Genome::VariantReporting::Framework::TestHelpers qw(test_dag_xml);
 use Genome::Utility::Test qw(compare_ok);
+use Sub::Install qw(reinstall_sub);
 
 my $pkg = "Genome::VariantReporting::Command::Wrappers::ModelPair";
 
 use_ok($pkg);
 
+reinstall_sub({
```
nnutter: It doesn't really matter since the scope ends up being the same in this case but `Sub::Override` is probably a slightly better fit for overriding subroutines in tests.

nnutter:
We ended up with millions of results with incomplete metadata because the default behavior of UR was not to yell about it.

nnutter:
I don't know whether these should be `input` or `param`.  Is there a meaningful distinction between the two types?

nnutter:
+1

tmooney:
+1

PR #381, opened 2015-01-05
A default implementation of G:C:DelegatesToResult->input_hash()

tmooney:
As suggested in #375, use the new implementation of `input_hash` from `Genome::Model::Tools::Htseq::Count` as the default for all `Genome::Command::DelegatesToResult`.


acoffman:
+1

davidlmorton:
+1

PR #383, opened 2015-01-05
Fixes for HtseqCount in RnaSeq; Final Removal of WithSavedResults and BuildStepWrapper.

tmooney:
This follows up from #373 and #375.  No more magic results and commands are generated in `Genome.pm`.

(This should also address the RnaSeq model test failure; it includes changes that would've been better included in #375.)


acoffman:
+1

`28 additions and 761 deletions` :tada:

davidlmorton:
+1 So glad to see the magic go away.

tmooney:
Merging early to fix RnaSeq model test.

PR #384, opened 2015-01-06
Tweaks to DetermineError

tmooney:
*We're after the very last `die()`-ish message in the log file.
*Use `above` so we test the current directory instead of stable.


acoffman:
+1

davidlmorton:
+!

iferguson90:
+1

PR #385, opened 2015-01-06
add baseq, mapq parameters for docm bamrc.

gatoravi:
control the mapping quality and base quality parameters for
bam-readcounting for DOCM sites. We need to be as sensitive as possible
so set these to the absolute minimum.


tmooney:
+1

gatoravi:
merged since this is a minor update to get build-tests passing.

PR #386, opened 2015-01-06
Add "scope" parameter to Genome:Sys->lock_resource

mark-burnett:
While integrating Nessy, we discovered that `lock_resource` was being used for both site-wide locks and for machine-local locks (specifically in `Genome::Model::Build::ReferenceSequence`).  The machine-local locks were not putting machine-specific information in the resource, rather they were assuming that their resource name was being used as a file path and were specifying a non-network directory for the locks.

When we turned on Nessy in the CI server, these locks were suddenly competing at a whole-site level.  @nnutter suggested that we address this by adding a `scope` parameter to `lock_resource`.

This PR adds that functionality along with two initial scopes: `site` and `host`.  This PR also adds the environment variable `GENOME_SYS_LOCK_DEFAULT_SCOPE` for configuring the default scope (which should normally be `site`).  We think this will prove useful for testing, as we can set that to a value like `process` and potentially avoid switching resource names based on `UR_DBI_NO_COMMIT` as is [sometimes done](https://github.com/genome/genome/blob/master/lib/perl/Genome/SoftwareResult.pm#L855).

### DIFF
--- a/lib/perl/Genome/Sys/Lock.pm
+++ b/lib/perl/Genome/Sys/Lock.pm
@@ -159,7 +159,7 @@ sub with_default_lock_resource_args {
 
     $args{block_sleep} = 60 unless defined $args{block_sleep};
     $args{max_try} = 7200 unless defined $args{max_try};
-    $args{scope} = 'site' unless defined $args{scope};
+    $args{scope} = $ENV{GENOME_SYS_LOCK_DEFAULT_SCOPE} unless defined $args{scope};
     $args{wait_announce_interval} = 0 unless defined $args{wait_announce_interval};
 
     return %args;
--- a/lib/perl/Genome/Sys/Lock.pm
+++ b/lib/perl/Genome/Sys/Lock.pm
@@ -67,6 +67,16 @@ sub lock_resource {
     my $class = shift;
     my %args = with_default_lock_resource_args(@_);
 
+    if (exists $RESOURCE_LOCK_SCOPE{$args{resource_lock}}
+        && $RESOURCE_LOCK_SCOPE{$args{resource_lock}} ne $args{scope}) {
+        Carp::confess(sprintf("Attempted to lock the resource (%s) in "
+                . "scope (%s) while it has an existing lock in scope (%s).  "
+                . "Locking in multiple scopes is not currently supported.",
+                $args{resource_lock},
+                $RESOURCE_LOCK_SCOPE{$args{resource_lock}},
+                $args{scope}));
+    }
+
     my @locks;
     my $unwind = sub {
         for my $pair (@locks) {
--- a/lib/perl/Genome/Sys/Lock.pm
+++ b/lib/perl/Genome/Sys/Lock.pm
@@ -62,6 +62,7 @@ then C<lock()> will C<croak()>.
 
 =cut
 
+my %RESOURCE_LOCK_SCOPE;
 sub lock_resource {
     my $class = shift;
     my %args = with_default_lock_resource_args(@_);
@@ -75,7 +76,7 @@ sub lock_resource {
         }
     };
 
-    for my $backend (backends()) {
+    for my $backend (backends($args{scope})) {
         my @lock_args = $backend->translate_lock_args(%args);
         my $lock = $backend->lock(@lock_args);
         if ($lock) {
@@ -88,13 +89,15 @@ sub lock_resource {
         }
     }
 
-    if (all { $_->has_lock($args{resource_lock}) } backends()) {
+    if (all { $_->has_lock($args{resource_lock}) } backends($args{scope})) {
         Genome::Utility::Instrumentation::increment('genome.sys.lock.lock_resource.consistent');
     } else {
         Genome::Utility::Instrumentation::increment('genome.sys.lock.lock_resource.inconsistent');
     }
 
     $class->_cleanup_handler_check();
+
+    $RESOURCE_LOCK_SCOPE{$args{resource_lock}} = $args{scope};
 
     return $args{resource_lock};
 
@@ -119,12 +122,16 @@ sub unlock_resource {
     my $class = shift;
     my %args = @_;
 
+    my $scope = $RESOURCE_LOCK_SCOPE{$args{resource_lock}};
+
     my $rv = 1;
-    for my $backend (backends()) {
-        my @unlock_args = $backend->translate_unlock_args(%args);
-        my $unlocked = $backend->unlock(@unlock_args);
-        if (is_mandatory($backend)) {
-            $rv = $rv && $unlocked;
+    for my $backend (backends($scope)) {
+        if ($backend->has_lock($args{resource_lock})) {
+            my @unlock_args = $backend->translate_unlock_args(%args);
+            my $unlocked = $backend->unlock(@unlock_args);
+            if (is_mandatory($backend)) {
+                $rv = $rv && $unlocked;
+            }
         }
     }
 
@@ -140,8 +147,10 @@ individually.
 =cut
 
 sub release_all {
-    for my $backend (backends()) {
-        $backend->release_all();
+    for my $scope (scopes()) {
+        for my $backend(backends($scope)) {
+            $backend->release_all();
+        }
     }
 }
 
@@ -161,27 +170,49 @@ sub is_mandatory {
     return $backend->is_mandatory();
 }
 
-my @backends;
+my $backends = {};
 sub backends {
-    return @backends;
+    my $scope = shift;
+
+    if (!defined($scope)) {
+        return %$backends;
+
+    } elsif ($scope eq 'any') {
+        return map {backends($_)} scopes();
+
+    } elsif (exists $backends->{$scope}) {
+        return @{$backends->{$scope}};
+
+    } else {
+        Carp::confess(sprintf(
+                'Unknown locking scope (%s) -- expected something in [%s].',
+                $scope, join(' ', scopes())));
+    }
 }
 
-sub add_backend {
-    my ($class, $backend) = @_;
-    push @backends, $backend;
+sub scopes {
+    return keys %$backends;
 }
 
 sub clear_backends {
-    @backends = ();
+    $backends = {};
 }
 
 sub set_backends {
     my $class = shift;
+    my %new_backends = @_;
 
     $class->clear_backends();
-    for my $backend (@_) {
-        $class->add_backend($backend);
+    for my $scope (keys %new_backends) {
+        for my $backend (@{$new_backends{$scope}}) {
+            $class->add_backend($scope, $backend);
+        }
     }
+}
+
+sub add_backend {
+    my ($class, $scope, $backend) = @_;
+    push @{$backends->{$scope}}, $backend;
 }
 
 my $_cleanup_handler_installed;
--- a/lib/perl/Genome/Sys/Lock.pm
+++ b/lib/perl/Genome/Sys/Lock.pm
@@ -159,7 +159,7 @@ sub with_default_lock_resource_args {
 
     $args{block_sleep} = 60 unless defined $args{block_sleep};
     $args{max_try} = 7200 unless defined $args{max_try};
-    $args{scope} = $ENV{GENOME_SYS_LOCK_DEFAULT_SCOPE} unless defined $args{scope};
+    $args{scope} = 'site' unless defined $args{scope};
     $args{wait_announce_interval} = 0 unless defined $args{wait_announce_interval};
 
     return %args;
--- a/lib/perl/Genome/Sys/Lock.pm
+++ b/lib/perl/Genome/Sys/Lock.pm
@@ -4,7 +4,7 @@ use strict;
 use warnings;
 
 use Carp qw(carp croak);
-use Genome::Sys::FileLock;
+use Genome::Sys::Lock::FileBackend;
 use List::MoreUtils qw(all);
 
 =item lock_resource()
@@ -156,7 +156,7 @@ sub with_default_lock_resource_args {
     return %args;
 }
 
-my @backends = ('Genome::Sys::FileLock');
+my @backends = ('Genome::Sys::Lock::FileBackend');
 
 sub is_mandatory {
     my $backend = shift;
--- a/lib/perl/Genome/Sys/Lock.pm
+++ b/lib/perl/Genome/Sys/Lock.pm
@@ -171,9 +171,17 @@ sub add_backend {
     push @backends, $backend;
 }
 
-sub remove_backend {
-    my ($class, $backend) = @_;
-    @backends = grep { $_ ne $backend } @backends;
+sub clear_backends {
+    @backends = ();
+}
+
+sub set_backends {
+    my $class = shift;
+
+    $class->clear_backends();
+    for my $backend (@_) {
+        $class->add_backend($backend);
+    }
 }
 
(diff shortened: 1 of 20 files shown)


lib/perl/Genome/Sys/Lock.pm
```diff
@@ -151,30 +159,63 @@ sub with_default_lock_resource_args {
...
+    for my $scope (keys %new_backends) {
+        for my $backend (@{$new_backends{$scope}}) {
+            $class->add_backend($scope, $backend);
+        }
+    }
+}
+
+sub add_backend {
+    my ($class, $scope, $backend) = @_;
+    if ($scope eq 'any') {
```
nnutter: Should this be more like,

```
if (! grep { $scope eq $_ } qw(site host)) {
```

?

lib/perl/Genome/Sys/Lock/NessyBackend.pm
```diff
@@ -178,6 +181,32 @@ sub min_timeout {
...
 
+sub _translate_resource {
+    my ($self, $resource) = @_;
+
+    my @subpath_parts;
+    my $resource_is_subpath = 1;
+
+    my @base_parts = File::Spec->splitdir($ENV{GENOME_LOCK_DIR});
+    my @resource_parts = File::Spec->splitdir($resource);
+    my $zip_iter = List::MoreUtils::each_array(@base_parts, @resource_parts);
```
nnutter: :heart: Perl Magic.

lib/perl/Genome/Sys/Lock/NessyBackend.pm
```diff
@@ -178,6 +181,32 @@ sub min_timeout {
     return 5;
 }
 
+sub _translate_resource {
```
nnutter: This might work as well:

```
sub _translate_resource {
    my ($self, $resource) = @_;
    my $genome_lock_dir = Path::Class::Dir->new($ENV{GENOME_LOCK_DIR});
    my $resource_file = Path::Class::File->new($resource);
    my $resource = $genome_lock_dir->contains($resource_file)
                 ? $resource_file->relative($genome_lock_dir)->stringify
                 : $resource_file->stringify;
    return $resource;
}
```

lib/perl/Genome/Sys/Lock.pm
```diff
@@ -151,30 +159,63 @@ sub with_default_lock_resource_args {
 
     $args{block_sleep} = 60 unless defined $args{block_sleep};
     $args{max_try} = 7200 unless defined $args{max_try};
+    $args{scope} = $ENV{GENOME_SYS_LOCK_DEFAULT_SCOPE} unless defined $args{scope};
```
nnutter: We can't put `GENOME_SYS_LOCK_DEFAULT_SCOPE` into environment until LIMS is on a snapshot with it so I think we also have to have a hard-coded default?  Plus, anyone that doesn't start a new shell won't have `GENOME_SYS_LOCK_DEFAULT_SCOPE`.

lib/perl/Genome/Sys/Lock.pm
```diff
@@ -97,6 +97,8 @@ sub lock_resource {
 
     $class->_cleanup_handler_check();
 
+    $RESOURCE_LOCK_SCOPE{$args{resource_lock}} = $args{scope};
```
nnutter: What happens if the resource is locked at different scopes?

acoffman:
I love the idea of scopes on locks. I'm not sure how I feel about having number of scopes be tied to number of backends though. It seems weird to add two File backends (and maybe eventually two Nessy backends?) two facilitate two scopes. It seems like the backend is concerned with the actual storage and implementation mechanisms of the locking (is it Nessy? File system, etc), and that shouldn't be tied to the scoping.

I could be mistaken, but as best as I can tell, this doesn't actually give us scopes for locks if they use the same storage right? For instance, if I add a two FileBackends with different scopes, then request a resource with one of the scopes (that happens to be network accessible), I've locked it in both scopes. Presumably it would be the same with Nessy (though not dependent on network accessible files) right? It seems like this relies on the same assumption that is currently being made (that local File locks will point to non-network mounted disks).

I'm not sure its better, but my gut reaction is that Backends should continue to be ignorant of scopes and that perhaps `Sys::Lock` should modify the requested resource name with the scope somehow rather than storing a different backend per-scope. This would require tracking which scopes had been seen or pre-registering valid scopes to allow for the `any` query that currently works. It would also break existing locks which sucks, so I'm not sure its the right answer.

lib/perl/Genome/Sys/Lock.pm
mark-burnett: Oops.  For a while, I was treating the scope `any` as a special scope, so I needed to check for its validity.  I think we should just remove the check and leave it to configuration to decide what the scopes are.

lib/perl/Genome/Sys/Lock.pm
mark-burnett: I set a `default_value` sub in Env/GENOME_SYS_LOCK_DEFAULT_SCOPE.pm.  Does that not make it optional?

lib/perl/Genome/Sys/Lock.pm
mark-burnett: That is currently unsupported, but could be a feature add.  I'm not sure what the use case would be right now.

lib/perl/Genome/Sys/Lock.pm
nnutter: Ah, yeah, forgot that.

lib/perl/Genome/Sys/Lock.pm
nnutter: This is basically the same concern Adam expressed.  I'm still trying to think about it.

mark-burnett:
@acoffman Overall, I think I'm hearing you say that you consider some of the behavior of scope to be surprising.

One thing that might address this would be to make `G:S:L:FileBackend` a proper class, and instantiate it with parameters controlling where it is allowed to make locks.  That way, you could force `host` and `site` scoped file locks to be put in different places (for example have white/black lists of directories where locks may be placed with those backends).

It would be relatively easy to `die` if there is already a lock for the existing resource in `lock_resource`, so that might make the situation more clear to users.

Locking a particular resource at multiple scopes seems a bit dubious to me off hand.  For example locking resource `FOO` at the `host` scope on multiple machines should be allowed, but locking `FOO` at the `site` scope, should probably require a `host`-scoped lock on every single host.  Locking `FOO` in the backends associated with the `site` scope, doesn't (maybe can't?) accomplish that.  I think doing this would require backends for the `site` scope to be aware of all the backends in the `host` scope for each machine.

Finally, the current system does not support the concept of locking the same resource multiple times (regardless of parameters sent to `lock_resource`), so I don't consider it a feature regression.  It may still be smart to implement some additional error checking, though.

I think that the backends are currently ignorant of scopes, since the scope parameter is not passed into the backend (at least not intentionally).  Being able to lookup and use the various backends based on the scope parameter seems like a reasonable way to keep them ignorant of the scope.

Re: `host` vs. `site` scoped locks with Nessy
I think it would require modifications to `G:S:L:NessyBackend` to do something like you modifying the resource string based on parameters at instantiation -- perhaps by appending the hostname (as you suggested).  The initial plan was to not use Nessy for `host` scope (as reflected in the Site configuration), though it's pretty reasonable to do so with a modified resource.

It's possible that the changes discussed above to `FileBackend` and `NessyBackend` might be a better entry point for modifying the locking behavior for testing than creating a new kind of scope like `process` would be.

lib/perl/Genome/Sys/Lock/NessyBackend.pm
mark-burnett: Ah, I didn't think that `Path::Class` was installed -- I should have checked.  If `Path::Class` is preferred, I'm happy to update!

lib/perl/Genome/Sys/Lock.pm
mark-burnett: Yeah, I agree with the sentiment.  I'm glad this PR is generating some discussion about it!

mark-burnett:
@brummett I was thinking to leave the valid scopes to configuration, but it is probably better to limit it to a few valid scopes.

lib/perl/Genome/Sys/Lock/NessyBackend.pm
mark-burnett: Yeah, I only did it because it matches the example in the `List::MoreUtils` docs..

acoffman:
Thanks for the detailed response!

I think that perhaps your instance idea would work well. The crux of my uncomfortableness (I think) is that you're relying on the thing requesting the lock to understand the underlying mechanisms of the Backends (ie, local vs network storage for File). If I'm reading the code correctly, with the current implementation, if I request a `host` lock on `/gscmnt/some_shared_disk/resource` it will actually be globally locked, which seems to defeat the purpose of the scope. The onus is on the obtainer of the lock to know that a `host` lock needs to be on local disk. If that's the case (and I totally could be reading this wrong), I'm not sure what the scopes are buying us? Likewise, with the Nessy backend, it seems like (unless we run two Nessy servers), if I request `host` lock on a resource, it will lock that resource for all hosts right?

Also, I totally agree that the Backends are currently ignorant of the scopes (and that's a good thing), sorry it was unclear what I meant! I also agree that we probably don't need to worry about locking the same resource at multiple scopes.

mark-burnett:
@acoffman Currently, creating a lock in Nessy is global.  That's actually the initial problem we're trying to overcome in `Genome::Model::Build::ReferenceSequence`.


So far, I think the current proposed changes are:

- Fix the allowed list of scopes (currently only `site` and `host`).  Use constants instead of strings.
- Remove `GENOME_SYS_LOCK_DEFAULT_SCOPE`, since its utility seems questionable.
- Change `Genome::Sys::Lock::FileBackend` to be a normal class instead of a "static" one.
- Add parameters to `Genome::Sys::Lock::FileBackend` to allow white/black-listing of lock prefixes.
- Extend `Genome::Sys::Lock::NessyBackend` to have a new parameter that allows appending a suffix to the resource.
- Add `host` scoped instance of the `NessyBackend` with the hostname as the suffix.
- Replace usage of `File::Spec` with appropriate usage of `Path::Class` in `Genome::Sys::Lock::NessyBackend`.
- Prevent users from calling `lock_resource` on the same `resource_lock` in multiple scopes.

Does this seem complete and reasonable to everyone?

mark-burnett:
I think I left off some important configuration steps:

- Use a whitelist for the `FileBackend` in `host` scope of `/var/cache/tgisan`.
- Use a blacklist for the `Filebackend` in `site` scope of `/var/cache/tgisan`.

lib/perl/Genome/Sys/Lock/NessyBackend.pm
nnutter: If you think it is in fact equivalent then I think `Path::Class` is more terse/clear.  You may want to use `subsumes()` instead of `contains()`; `contains()` actually hits the filesystem.

acoffman:
That seems to address my concerns!

lib/perl/Genome/Sys/Lock.pm
nnutter: I've got myself worried again.  You should try running the ClinSeq model tests on this before merging; `/gsc/scripts/opt/genome/bin/model-test 5.10-clinseq-wer`.  Maybe, we should figure out a way to test the circular dependency in Perl tests so that's not necessary.

I'm thinking that have the `default_value` will be carried forward to wherever ClinSeq calls `illumina_info`, or whatever, which will then call into LIMS's Genome snapshot which won't recognize the variable and will crash.

mark-burnett: Yeah, I think this is possibly true.  I propose below that we eliminate the environment variable altogether.

Maybe we should change the check for `GENOME_*` environment variables into a warning instead of an error?

nnutter: I think, for now, I would vote for removing the environment variable and just hard-coding a default.  If we need to override the default for tests we can override a `default_scope()` method.

Going the warning route would require that change being a separate PR so it can trickle down to LIMS before this PR get deployed.  If we went the warning route I think I would vote more for providing `genome config validate` or something that would allow people to verify config changes independent of running any `genome` command.

mark-burnett: I agree that changing config validation does not belong in this PR.

I also think that a `genome config validate` command is a good idea, as it leaves the door open for more than environment config validation.

mark-burnett:
After some additional discussion with @nnutter, we came up with the following plan (feedback appreciated):

- Change `Genome::Sys::Lock::FileBackend` to be a normal class that puts its locks in a specified directory only (e.g. it prepends `$ENV{GENOME_LOCK_DIR}`).
- Define the following fixed scopes:
    - `site` with `NessyBackend` and `FileBackend` at `$ENV{GENOME_LOCK_DIR}`.
    - `tgisan` with `FileBackend` at `/var/cache/tgisan` -- this could end up being temporary.
    - `host` with `FileBackend` at `/var/lock/genome` -- optional, for future use.
- Update all calls of `lock_resource` to not prepend the resource with `ENV{GENOME_LOCK_DIR}`.  We only found about 30 of these in the code-base, which was fewer than I expected.  If there are locks being created in multiple directories, we can add more scopes for those directories.
- Make the valid scopes fixed constants.
- Make `site` the default scope (hardcoded).
- Remove `GENOME_SYS_LOCK_DEFAULT_SCOPE`.
- Prevent users from calling `lock_resource` on the same `resource_lock` in multiple scopes.

It seems like it is an optional change to make it possible to use Nessy as a `host` backend.  If we do that, I guess there's no point in using the `FileBackend` for that scope, though.

mark-burnett:
I have pushed some additional commits in attempt to address the concerns raised.  I am happy to close this PR and make a new, cleaner one if that is deemed useful.

The new changes follow the plan that @nnutter and I discussed with the following comments:
- I added an `unknown` scope, which is supported by a `FileBackend` at path `/`.  This is used in the `ReferenceSequence` build, which creates locks in its build directory.
- I have made `scope` a mandatory parameter to `lock_resource`.
- I did not add a `host` scope, or modify the `NessyBackend` to have an optional suffix parameter for resources (which is how I would probably implement a host scope on that backend).
- I did not update the valid scopes to use the `constant` module, because it seemed cumbersome at the call sites.  Is that a reasonable tradeoff, or should I make that change?

@acoffman @brummett @nnutter Would you please review?

brummett:
A better name for the "unknown" scope might be "abs-path", as it's locking a fully qualified pathname.  Since it's only used for build related locks, it could alternatively be called "build" as long as there was some mechanism for the locking code to know what build it was talking about.

nnutter:
> A better name for the "unknown" scope might be "abs-path", as it's locking a fully qualified pathname. Since it's only used for build related locks, it could alternatively be called "build" as long as there was some mechanism for the locking code to know what build it was talking about.

Despite the fact that they are absolute paths the scope really is unknown because those paths could be on local filesystem or on shared filesystem.  Another alternative might be "root" but I personally feel like "unknown" is accurate.

Perhaps relevant, I think the use of "unknown" is just a stopgap meant to keep this PR from growing more complicated.  We will need to migrate those "unknown" scope to either "site" or "host" but that will require many weeks to transition.

+1

acoffman:
+1

PR #387, opened 2015-01-06
New module Genome::Utility::File::DirCompare to compare contents of two directories

susannasiebert:
This new module Genome::Utility::File::DirCompare was adapted from File::DirCompare (http://search.cpan.org/~gavinc/File-DirCompare-0.7/DirCompare.pm). 
If two paths are both symlinks, this module will resolve the symlinks and do a file comparison on the targets. The previous behavior in File::DirCompare for two symlinks was to count them as different if they were pointing to different targets. With this change they will only be counted as different if the contents of the targets are different. This is the behavior we are desiring for a lot of variant reporting work.


nnutter:
It would probably be nice to have the original contents of `File::DirCompare` in the commit history so that the patch you are making is obvious.  I think you are also intending on submitting a patch upstream?  I would guess that upstream would want this feature added as an option.  It's a fairly standard concept to optionally follow symlinks, for example, `File::Find::Rule` enables this with,

```
{ follow => 1 }
```

This will also end up being useful in [Genome::Utility::Test::compare_dirs][1].

[1]: https://github.com/susannasiebert/genome/blob/dircompare/lib/perl/Genome/Utility/Test.pm#L221

lib/perl/Genome/Utility/File/DirCompare.pm
```diff
@@ -0,0 +1,289 @@
...
+            $sub->($f1, undef) unless $opts->{ignore_unique};
+        }
+        # Only in $dir2
+        elsif (! $d1{$f}) {
+            $sub->(undef, $f2) unless $opts->{ignore_unique};
+        }
+        # Item exists in both directories
+        else {
+            # Both symlinks
+            if (-l $f1 && -l $f2) {
```
nnutter: You can might be able to simplify this by always just using `Cwd::abs_path` on both paths.

lib/perl/Genome/Utility/File/DirCompare.pm
```diff
@@ -0,0 +1,289 @@
...
+                my $t1 = readlink $f1 or die $self->error_message("Cannot read symlink $f1: $!");
```
nnutter: What about symlinks to symlinks?  Also, if you worry about them do you want to be afraid of circular symlinks?

nnutter:
It would probably be nice to have the original contents of `File::DirCompare` in the commit history so that the patch you are making is obvious.  I think you are also intending on submitting a patch upstream?  I would guess that upstream would want this feature added as an option.  It's a fairly standard concept to optionally follow symlinks, for example, [File::Find][2] enables this with,

```
{ follow => 1 }
```

This will also end up being useful in [Genome::Utility::Test::compare_dirs][1].

[1]: https://github.com/susannasiebert/genome/blob/dircompare/lib/perl/Genome/Utility/Test.pm#L221
[2]: https://metacpan.org/pod/File::Find#follow

brummett:
This might be better handled by contacting the maintainer of File::DirCompare and getting your changes in upstream.

Alternatively, making a genome-vendor repo with the original sources, patching on top of that, and then making a new package from there.

susannasiebert:
@brummett It would be great to get this change into File::DirCompare itself but the last update was from 2 years ago and there is one open pull request from 2013 that hasn't been addressed so I'm not hopeful that it will happen. But I will make a PR and see where that goes.
I will also look into making File::DirCompare a genome vendor repo and going from there.

susannasiebert:
@nnutter You're right about putting the original code in a separate commit. I should've done that. It's probably a mood point if we're going to base this on a genome vendor repo instead. 
I plan on submitting this patch upstream.

lib/perl/Genome/Utility/File/DirCompare.pm
susannasiebert: I have a local fix for this which uses `File::Find` with `follow` enabled instead of readlink. I also added tests for "normal" symlinks, symlinks to symlinks, and circular symlinks. I didn't push the changes since I will close this PR and look into implementing this tool via the genome vendor route.

susannasiebert:
I'm closing this PR and will implement this tool through genome vendor instead, as Tony suggested.

susannasiebert:
With @tmooney's help I was able to use File::DirCompare in a way that resolves symlinks. It will not be necessary to package and patch File::DirCompare.

PR #388, opened 2015-01-06
Add tag parameter to AddConfigFile for AUE-383

jweible:
Allow tags to be stapled in here for convenience.

### DIFF
--- a/lib/perl/Genome/Config/AnalysisProject/Command/AddConfigFile.pm
+++ b/lib/perl/Genome/Config/AnalysisProject/Command/AddConfigFile.pm
@@ -26,7 +26,7 @@ class Genome::Config::AnalysisProject::Command::AddConfigFile {
     ],
     has_optional_input => [
         tag => {
-            is => 'Text',
+            is => 'Genome::Config::Tag',
             is_many => 1,
             doc => 'Optional tags to add',
         }
@@ -81,9 +81,7 @@ sub _mark_instrument_data_bridges {
 
 sub _apply_tags {
     my ($self, $profile_item) = @_;
-    return unless $self->tag;
-    for my $tag_name ($self->tag){
-        my $tag = Genome::Config::Tag->create(name => $tag_name);
+    for my $tag ($self->tag){
         $profile_item->add_tag($tag);
     }
     return 1;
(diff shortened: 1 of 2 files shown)


lib/perl/Genome/Config/AnalysisProject/Command/AddConfigFile.pm
```diff
@@ -70,4 +79,28 @@ sub _mark_instrument_data_bridges {
     return 1;
 }
 
+sub _apply_tags {
+    my ($self, $profile_item) = @_;
+    return unless $self->tag;
+    for my $tag_name ($self->tag){
+        my $tag = Genome::Config::Tag->create(name => $tag_name);
```
tmooney: `Genome::Config::Tag` is a table that holds a small, canonical set of tags.  The type of the `tag` parameter should be `Genome::Config::Tag` to save looking up or creating them here.

lib/perl/Genome/Config/AnalysisProject/Command/AddConfigFile.pm
```diff
@@ -70,4 +79,28 @@ sub _mark_instrument_data_bridges {
     return 1;
 }
 
+sub _apply_tags {
+    my ($self, $profile_item) = @_;
+    return unless $self->tag;
```
tmooney: It would be good to have the test try using more than one tag.  This `return` isn't strictly necessary since the `for` loop will execute zero times if there are no tags.

lib/perl/Genome/Config/AnalysisProject/Command/AddConfigFile.pm
```diff
@@ -70,4 +79,28 @@ sub _mark_instrument_data_bridges {
...
+    for my $tag_name ($self->tag){
+        my $tag = Genome::Config::Tag->create(name => $tag_name);
+        $profile_item->add_tag($tag);
+    }
+    return 1;
+}
+
+sub __errors__ {
```
tmooney: Did you mean to add `__errors__` in this commit? (The base class should take care of this.)

jweible: Fixed

lib/perl/Genome/Config/AnalysisProject/Command/AddConfigFile.pm
jweible: Fixed

lib/perl/Genome/Config/AnalysisProject/Command/AddConfigFile.pm
jweible: Fixed

jweible:
Pushed changes to address comments from tmooney

tmooney:
+1

acoffman:
+1

PR #389, opened 2015-01-06
added option to change indel size and define reference transcript version

clu76:


lib/perl/Genome/Model/Tools/Validation/ProcessSomaticValidation.pm
```diff
@@ -619,6 +635,11 @@ sub execute {
   my $ref_seq_build = Genome::Model::Build->get($ref_seq_build_id);
   my $ref_seq_fasta = $ref_seq_build->full_consensus_path('fa');
   my $annotation_build_name = $model->annotation_build->name;
+  if(defined $self->reference_transcripts){
+      print STDERR "Model's annotation build overriden. Using " . $self->reference_transcripts . "\n";
```
tmooney: "overridden" is missing a `d`.

Ideally each of these `print STDERR` in this tool would be replaced with `$self->status_message()` or `$self->warning_message()`.  This one could end up something like:
```perl
$self->status_message("Model's annotation build overridden. Using %s", $self->reference_transcripts);
```

nnutter:
Does [Genome/Model/Tools/Somatic/ProcessSomaticVariation.pm](https://github.com/genome/genome/blob/7be06303ddc9ac2d695ca29533f1d251fcab7609/lib/perl/Genome/Model/Tools/Somatic/ProcessSomaticVariation.pm#L585) need the same change?

chrisamiller:
@nnutter the other tool already has the reference-transcripts option, but yes, it'd be good to add the indel-size param there as well. @clu76, can you make that change too, please?

nnutter:
All comments will be simple to address.  I repeated similar feedback so that it will be easy to find during iteration.  I would be happy to pair up to work through remaining feedback.

lib/perl/Genome/Model/Tools/Validation/ProcessSomaticValidation.pm
nnutter: I think if this were to come up in the future this should be suggested as a separate PR.  I think the argument would be that it would be better to replace `print` with `*_message` but local consistency suggests that should be done in bulk and do to the scope of that it should be a separate PR.

susannasiebert:
@chrisamiller does G::M::SomVar::Command::CreateReport also need to be updated?

chrisamiller:
I think for the standard reports we still want indels to be displayed as is, so it's not necessary to port this at present.  Thanks for asking!

PR #390, opened 2015-01-07
InstData Import: use names instead of library object for steps interacting

ebelter:
with gmt picard fastq to sam

### DIFF
--- a/lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/SraToBam.pm
+++ b/lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/SraToBam.pm
@@ -21,9 +21,13 @@ class Genome::InstrumentData::Command::Import::WorkFlow::SraToBam {
             is => 'Text',
             doc => 'Path of the SRA.',
         },
-        library => {
-            is => 'Genome::Library',
-            doc => 'The library name to use and then derive the read group name.',
+        sample_name => {
+            is => 'Text',
+            doc => 'The sample name to use in the RG header.',
+        },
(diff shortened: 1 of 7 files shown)


lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/SraToBam.pm
```diff
@@ -288,7 +292,9 @@ sub convert_fastq_to_bam {
         my $fastq_to_sam = Genome::Model::Tools::Picard::FastqToSam->create(
             fastq           => "$unaligned_fastq",
             output          => "$unaligned_bam",
-            sample_name     => "$sample_name",
+            sample_name => $self->sample_name,
+            library_name => $self->library_name,
+            read_group_name => $self->library_name,
```
acoffman: I see that this class doesn't have a `$self->read_group_name`, so I'm assuming the re-use of `library_name` is intentional here?

ebelter: Yes - the library_name here goes into the LB tag in the RG header. SM is sample_name. Ex:

@RG	ID:__TEST_SAMPLE__-extlibs	LB:__TEST_SAMPLE__-extlibs	SM:__TEST_SAMPLE__

acoffman: Cool, makes sense to me!

acoffman:
+1

tmooney:
+1

PR #391, opened 2015-01-07
Add functionality to compare Genome::Process output files

susannasiebert:
This PR supersedes https://github.com/genome/genome/pull/374.

This is a first step towards tackling https://jira.gsc.wustl.edu/browse/AT-680. This PR implements a compare_output subroutine for Genome::Process that would give a similar output to G::M::B::compare_output.

In this PR Genome::Process::_compare_output_files handles the comparison instead of being abstract and deferring the implementation to the child classes. It utilizes Genome::Process::symlink_results to symlink all the results of a process into one directory. Genome::Process::_compare_output_directories handles the actual comparison of those directories.

### DIFF
--- a/lib/perl/Genome/Process.pm
+++ b/lib/perl/Genome/Process.pm
@@ -676,51 +676,48 @@ sub _compare_output_directories {
         sub {
             my ($other_file, $file) = @_;
             my $template = 'no %s %s found for process %s';
-            if (! $file) {
-                if (-f $other_file) {
-                    $diffs{abs_path($other_file)} = sprintf(
+            my $other_target = defined($other_file)? abs_path($other_file) : undef;
+            my $target = defined($file) ? abs_path($file) : undef;
+            if (! $target) {
+                if (-f $other_target) {
+                    $diffs{$other_target} = sprintf(
                         $template,
                         'file', File::Spec->abs2rel($other_file, $other_output_dir), $self->id
                     );
                 }
-                elsif (-d $other_file) {
+                elsif (-d $other_target) {
                     $diffs{File::Spec->abs2rel($other_file, $other_output_dir)} = sprintf(
                         $template,
                         'directory', File::Spec->abs2rel($other_file, $other_output_dir), $self->id
                     );
                 }
-            } elsif (! $other_file) {
-                if (-f $file) {
-                    $diffs{abs_path($file)} = sprintf(
+            } elsif (! $other_target) {
+                if (-f $target) {
+                    $diffs{$target} = sprintf(
                         $template,
                         'file', File::Spec->abs2rel($file, $output_dir), $other_process->id
                     );
                 }
-                elsif (-d $file) {
+                elsif (-d $target) {
                     $diffs{File::Spec->abs2rel($file, $output_dir)} = sprintf(
                         $template,
                         'directory', File::Spec->abs2rel($other_file, $other_output_dir), $self->id
                     );
                 }
             } else {
-                #Resolve possible symlinks
-                my $other_target = abs_path($other_file);
-                my $target = abs_path($file);
-                if (-l $file || -l $other_file) {
-                    if (-d $target && -d $other_target) {
-                        my %additional_diffs = $self->_compare_output_directories($target, $other_target, $other_process);
-                        %diffs = (%diffs, %additional_diffs);
-                        return;
-                    }
-                    elsif (-f $target && -f $other_target && !compare($target, $other_target)) {
-                        #Files are in fact the same - do nothing
-                        return;
-                    }
+                if (-d $target && -d $other_target) {
+                    my %additional_diffs = $self->_compare_output_directories($target, $other_target, $other_process);
+                    %diffs = (%diffs, %additional_diffs);
                 }
-                $diffs{$other_target} = sprintf(
-                    'files are not the same (diff -u %s %s)',
-                    $target, $other_target
-                );
+                elsif (-f $target && -f $other_target && !compare($target, $other_target)) {
+                    #Files are in fact the same - do nothing
+                }
+                else {
+                    $diffs{$other_target} = sprintf(
+                        'files are not the same (diff -u %s %s)',
+                        $target, $other_target
+                    );
+                }
             }
         },
     );
--- a/lib/perl/Genome/Process.pm
+++ b/lib/perl/Genome/Process.pm
@@ -679,31 +679,13 @@ sub _compare_output_directories {
             my $other_target = defined($other_file)? abs_path($other_file) : undef;
             my $target = defined($file) ? abs_path($file) : undef;
             if (! $target) {
-                if (-f $other_target) {
-                    $diffs{$other_target} = sprintf(
-                        $template,
-                        'file', File::Spec->abs2rel($other_file, $other_output_dir), $self->id
-                    );
-                }
-                elsif (-d $other_target) {
-                    $diffs{File::Spec->abs2rel($other_file, $other_output_dir)} = sprintf(
-                        $template,
-                        'directory', File::Spec->abs2rel($other_file, $other_output_dir), $self->id
-                    );
-                }
+                my $type = ( -f $other_target ? 'file' : 'directory' );
+                my $rel_path = File::Spec->abs2rel($other_file, $other_output_dir);
+                $diffs{$rel_path} = sprintf($template, $type, $rel_path, $self->id);
             } elsif (! $other_target) {
-                if (-f $target) {
-                    $diffs{$target} = sprintf(
-                        $template,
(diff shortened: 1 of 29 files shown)


lib/perl/Genome/Process.pm
```diff
@@ -630,5 +633,97 @@ sub result_is_on_cle_disk_group {
...
+    unless ($self->class eq $other_process->class) {
+        die $self->error_message('Processes (%s) and (%s) are not of the same type', $process_id, $other_process_id);
+    }
+
+    my %diffs = $self->_compare_output_files($other_process);
+
+    return %diffs;
+}
+
+sub _compare_output_files {
```
nnutter: Why does this method have to exist?  A code comment might be in order depending on the answer.

lib/perl/Genome/Process.pm
```diff
@@ -630,5 +633,97 @@ sub result_is_on_cle_disk_group {
...
+    my ($self, $output_dir, $other_output_dir, $other_process) = @_;
+
+    my %diffs;
+    File::DirCompare->compare(
+        $other_output_dir,
+        $output_dir,
+        sub {
+            my ($other_file, $file) = @_;
+            if (! $file) {
+                if (-f $other_file) {
```
nnutter: You can deduplicate these lines a bit by having 'file' and 'directory' be a string passed in to a common `sprintf()` template.

lib/perl/Genome/Process.pm
susannasiebert: It's recursively calling itself on line 710.

lib/perl/Genome/Process.pm
```diff
@@ -630,5 +633,97 @@ sub result_is_on_cle_disk_group {
...
+                elsif (-d $file) {
+                    $diffs{File::Spec->abs2rel($file, $output_dir)} = sprintf(
+                        'no directory %s found for process %s',
+                        File::Spec->abs2rel($other_file, $other_output_dir), $self->id
+                    );
+                }
+            } else {
+                #Resolve possible symlinks
+                my $other_target = abs_path($other_file);
+                my $target = abs_path($file);
```
nnutter: I think you have to check that `$target` and `$other_target` are defined.  I seem to recall `abs_path()` returning `undef` under certain conditions; maybe when the target doesn't exist.

lib/perl/Genome/Process.pm
nnutter: Sorry, I was more meaning why the body of this method exists.  Why are we making temp paths and symlinking?

susannasiebert: The child classes don't have a single output directory themselves. They merely keep track of the software results they create. The symlink command takes care to aggregate all of the output in one location. E.g., for Genome::VariantReporting::Process::CreateReport, there is one software result for each report. The symlink_results method in that class symlinks all of the files for those software results into one output_directory of your choosing.

nnutter: Ah ok.  If I knew more about VariantReporting that probably would have been obvious.  Thanks for teaching me.

lib/perl/Genome/Process.pm
susannasiebert: I doesn't look like this behavior is documented and we don't seem to usually check for this case in our code base, for better or for worse. I'm trying to figure out how to best handle it since it seems to me that if one or the other target is not defined, we should count that as a diff. If we handle it this way than there will be a lot of code duplication between how to handle symlink targets and how to handle files. Maybe I just need to move the resolving of symlinks all the way to the top. Anyway, I will work on.

nnutter: It looks like `abs_path` will return empty string when an intermediate directory does not actually exist.  It only relates to symlinks in that the rule applies to the symlink's target.

Here's an example of a non-existant intermediate directory in path,

```
$ perl -e 'use strict; use warnings; use v5.10; use Cwd qw(abs_path); say q(real path = ) . abs_path(q(/Users)); say q(unreal file = ) . abs_path(q(/Users/foo)); say q(unreal directory (foo) = ) . abs_path(q(/Users/foo/bar))'
real path = /Users
unreal file = /Users/foo
unreal directory (foo) = 
```

So a symlink can be broken and still return an `abs_path` if it follow this rule, i.e. a symlink pointing to `/Users/foo` returns `/Users/foo` but a symlink pointing to `/Users/foo/bar` return an empty string.

susannasiebert: I just pushed a commit to address this issue.

lib/perl/Genome/Process.pm
```diff
@@ -630,5 +633,95 @@ sub result_is_on_cle_disk_group {
...
+    my %diffs;
+    File::DirCompare->compare(
+        $other_output_dir,
+        $output_dir,
+        sub {
+            my ($other_file, $file) = @_;
+            my $template = 'no %s %s found for process %s';
+            my $other_target = defined($other_file)? abs_path($other_file) : undef;
+            my $target = defined($file) ? abs_path($file) : undef;
+            if (! $target) {
```
nnutter: If I can assume that...

1. the mixing of relative and absolute paths in the keys of `%diffs` was unintentional.
1. the use of `$self->id` in the `!$other_target` was unintentional.

Then the following ~25 lines can be simplified to,

```
if (! $target) {
    my $type = ( -f $other_target ? 'file' : 'directory' );
    my $rel_path = File::Spec->abs2rel($other_file, $other_output_dir);
    $diffs{$rel_path} = sprintf($template, $type, $rel_path, $self->id);
} elsif (! $other_target) {
    my $type = ( -f $target ? 'file' : 'directory' );
    my $rel_path = File::Spec->abs2rel($file, $output_dir);
    $diffs{$rel_path} = sprintf($template, $type, $rel_path, $other_process->id);
}
```

lib/perl/Genome/Process.pm
```diff
@@ -630,5 +633,95 @@ sub result_is_on_cle_disk_group {
...
+                    my %additional_diffs = $self->_compare_output_directories($target, $other_target, $other_process);
+                    %diffs = (%diffs, %additional_diffs);
+                }
+                elsif (-f $target && -f $other_target && !compare($target, $other_target)) {
+                    #Files are in fact the same - do nothing
+                }
+                else {
+                    $diffs{$other_target} = sprintf(
+                        'files are not the same (diff -u %s %s)',
+                        $target, $other_target
```
nnutter: I think this can, and possibly should, be `$file` and `$other_file`.

lib/perl/Genome/Process.pm
```diff
@@ -630,5 +633,95 @@ sub result_is_on_cle_disk_group {
...
+
+    my %diffs;
+    File::DirCompare->compare(
+        $other_output_dir,
+        $output_dir,
+        sub {
+            my ($other_file, $file) = @_;
+            my $template = 'no %s %s found for process %s';
+            my $other_target = defined($other_file)? abs_path($other_file) : undef;
+            my $target = defined($file) ? abs_path($file) : undef;
```
nnutter: I like this solution to the `abs_path()` returning `undef` problem!

lib/perl/Genome/Process.t
```diff
@@ -77,5 +78,83 @@ is_deeply([$p->results], [$result],
...
+        is(scalar(keys %diffs), $test_info->{diff_count}, 'Number of differences as expected');
+        if (scalar(keys %diffs) > 0) {
+            like((values %diffs)[0], qr/$test_info->{diff_message}/, 'Diff message as expected');
+        }
+    };
+}
 
 done_testing();
+
+sub compare {
```
nnutter: This `sub` appears unused.  Can it be removed?

lib/perl/Genome/Process.pm
susannasiebert: 1. It was intentional as I was trying to mirror the behavior of G::M::B which uses absolute paths in the keys. However, using absolute paths won't work for directories since `$output_dir` and `$other_output_dir` are only temporary; hence the usage of relative paths in this case. Your solution is ok, if we agree that we don't need to mirror G::M::B. I'm fine with that but the keys might need to change again in the future depending on how this tool will be used when we're trying to make build-like jenkins tests for processes.
2. Yes, that was a copy-and-paste mixup.

susannasiebert:
@nnutter : I pushed two commits to fix the remaining issues.

nnutter:
+1 assuming you are choosing to disagree about the [filenames][1] to use when the files are not the same.

[1]: https://github.com/genome/genome/pull/391#discussion_r22732822

susannasiebert:
No, you're right. I just forgot about it. Thanks for the reminder. It's now fixed.

nnutter:
+1

davidlmorton:
+1

PR #392, opened 2015-01-07
Test Factory for Genome::Process

tmooney:


lib/perl/Genome/Process.pm
```diff
@@ -227,6 +227,11 @@ sub _schedule_process {
 
 sub create {
     my $class = shift;
+
+    if ($class->__meta__->is_abstract) {
+        return $class->SUPER::create(@_); #UR will re-call with the concrete class.
+    }
+
```
ebelter: I dislike the check. Why not use the subclass in the test factory to create the test object?

tmooney: This is the same check that is in [Model](https://github.com/genome/genome/blob/105151eab532ea717ea86ef7a4f9c4da98ed1adc/lib/perl/Genome/Model.pm#L476), [ProcessingProfile](https://github.com/genome/genome/blob/105151eab532ea717ea86ef7a4f9c4da98ed1adc/lib/perl/Genome/ProcessingProfile.pm#L86), [SoftwareResult](https://github.com/genome/genome/blob/105151eab532ea717ea86ef7a4f9c4da98ed1adc/lib/perl/Genome/SoftwareResult.pm#L263), and others.  I believe it is correct practice for all abstract UR classes that override `create`, and I advocate for its existence even if the Test Factory is not added.

The reason I made the test factory this way was so a specific other subclass of Process could be easily chosen if desired.

ebelter: I seen it before, and believe it to do the 'right' thing. I dislike it. Your reason is reasonable :octocat:

ebelter:
+1

acoffman:
+1

PR #393, opened 2015-01-07
Add bam_header and comparison_flagstat as required parameters

dufeiyu:
Make bam_header an optional parameter.

### DIFF
--- a/lib/perl/Genome/InstrumentData/AlignmentResult/Command/RecreatePerLaneBam.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult/Command/RecreatePerLaneBam.pm
@@ -35,7 +35,10 @@ class Genome::InstrumentData::AlignmentResult::Command::RecreatePerLaneBam {
         bam_header => {
             is  => 'FilePath',
             doc => 'The path of bam header used to reheader per lane bam',
-            is_optional => 1,
+        },
+        comparison_flagstat => {
+            is  => 'FilePath',
+            doc => 'The path of flagstat file that is used to compare with that of recreated bam',
         },
     ],
     has_transient_optional => [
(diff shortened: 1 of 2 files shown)


lib/perl/Genome/InstrumentData/AlignmentResult/Command/RecreatePerLaneBam.pm
```diff
@@ -32,6 +32,11 @@ class Genome::InstrumentData::AlignmentResult::Command::RecreatePerLaneBam {
             doc => 'Picard version to replace bam header',
             valid_values => [Genome::Model::Tools::Picard->available_picard_versions],
         },
+        bam_header => {
```
nnutter: It might be reasonable to make `bam_header` required and remove the heuristic for finding one via `_output_dir`.

lib/perl/Genome/InstrumentData/AlignmentResult/Command/RecreatePerLaneBam.pm
```diff
@@ -156,14 +168,17 @@ sub _revert_markdup {
...
 
 sub _reheader_bam {
-    my ($self, $temp_bam) = @_;
-
-    my @header_files = glob($self->_output_dir ."/*.header");
-    unless (@header_files and @header_files == 1) {
-        die $self->error_message("Failed to get 1 header file from original per lane bam location");
+    my ($self, $temp_bam, $bam_header) = @_;
+
+    unless ($bam_header) { #find bam header from planned places
```
nnutter: If this is kept it would probably be clearer to extract this heuristic as a `_default_bam_header()` method or something.

lib/perl/Genome/InstrumentData/AlignmentResult/Command/RecreatePerLaneBam.pm
ebelter: I agree. I don't know the context this is going to be called in, but it seems reasonable that the header location logic live somewhere else.

dufeiyu:
@nnutter @ebelter  I agree that bam_header should be set as required not optional to be less heuristic and clearer to others. The logic to get heuristic header file should go to the code calling this one. I also extracted another similar one to be a new parameter: comparison_flagstat. Fix is done.

nnutter:
+1

ebelter:
+1

PR #394, opened 2015-01-07
Handle bam path urls in igv sessions correctly

susannasiebert:
See https://jira.gsc.wustl.edu/browse/AT-374.

### DIFF
--- a/lib/perl/Genome/Model/Tools/Analysis/DumpIgvXmlMulti.pm
+++ b/lib/perl/Genome/Model/Tools/Analysis/DumpIgvXmlMulti.pm
@@ -98,8 +98,7 @@ XML
 
     my $i=0;    
     for($i=0;$i<@bams;$i++){
-        my $uri = URI->new($bams[$i]);
-        my $path = $uri->scheme ? $bams[$i] : abs_path($bams[$i]);
+        my $path = resolve_bam_path($bams[$i]);
         $header .= <<"XML";
      <Resource path="$path" relativePath="false"/>
 XML
@@ -125,7 +124,7 @@ XML
     my $panels;
 
     for($i=0;$i<@bams;$i++){
-        my $path = abs_path($bams[$i]);
+        my $path = resolve_bam_path($bams[$i]);
         my $label = $labels[$i];
         my $cov = $path . "_coverage";
 
@@ -173,6 +172,24 @@ XML
     
 return 1;
     
+}
+
+sub resolve_bam_path {
+    my $bam = shift;
+    my $uri = URI->new($bam);
+    my $path;
+    if (defined($uri->scheme)) {
+        if ($uri->scheme eq 'file') {
+            $path = abs_path($uri->path);
+        }
+        else {
+            $path = $uri->as_string;
+        }
+    }
+    else {
+        $path = abs_path($bam);
+    }
+    return $path;
 }
 
 
--- a/lib/perl/Genome/Model/Tools/Analysis/DumpIgvXmlMulti.pm
+++ b/lib/perl/Genome/Model/Tools/Analysis/DumpIgvXmlMulti.pm
@@ -75,7 +75,7 @@ sub execute {
     my @review_bed_files = map { abs_path($_) }$self->review_bed_files;
     my $reference_name = $self->reference_name;
 
-    my @bams = split(/\,/,$self->bams);
+    my @bams = map { resolve_bam_path ($_) } split(/\,/, $self->bams);
     my @labels = split(/\,/,$self->labels);
 
     unless (@bams == @labels){
@@ -98,7 +98,7 @@ XML
 
     my $i=0;    
     for($i=0;$i<@bams;$i++){
-        my $path = resolve_bam_path($bams[$i]);
+        my $path = $bams[$i];
         $header .= <<"XML";
      <Resource path="$path" relativePath="false"/>
 XML
@@ -124,7 +124,7 @@ XML
     my $panels;
 
     for($i=0;$i<@bams;$i++){
-        my $path = resolve_bam_path($bams[$i]);
+        my $path = $bams[$i];
         my $label = $labels[$i];
         my $cov = $path . "_coverage";
 
--- a/lib/perl/Genome/Model/Tools/Analysis/DumpIgvXmlMulti.pm
+++ b/lib/perl/Genome/Model/Tools/Analysis/DumpIgvXmlMulti.pm
@@ -7,6 +7,7 @@ use Genome;
 use Command;
 use IO::File;
 use Cwd qw(abs_path);
+use URI::URL;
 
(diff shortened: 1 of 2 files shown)


lib/perl/Genome/Model/Tools/Analysis/DumpIgvXmlMulti.pm
```diff
@@ -7,6 +7,7 @@ use Genome;
 use Command;
 use IO::File;
 use Cwd qw(abs_path);
+use URI::URL;
```
nnutter: I am giving in to sharing a personal preference... maybe even a rant...

It really bugs me that Perl modules export things by default.  I would much rather see things imported explicitly (and not imported if unused).  I make certain exceptions, e.g. `Test::More`, but they are very few.  For example, `URI::URL` exports `url()` by default.  Even if `url()` was being used it would not necessarily be obvious where it came from (more so with more generic method names like, `move`).

I try to do,

```
use URI::URL qw();
```

The `qw()` is intentional as opposed to, for example, `""`.  The reason is that an empty list is different than passing an empty string through and I think empty list is most correct.

Another alternative is,

```
require URI::URL;
```

However, in addition to not importing default exported methods it doesn't call `import()` at all and delays loading the module until runtime.  There are definitely cases where delaying loading until runtime is desired; it's maybe even a better default for performance reasons.  I don't know that not calling `import()` at all causes problems.

So I don't know whether `use Foo qw()` is a better default practice than `require Foo` but in my opinion either is better than `use Foo`.

P.S. maybe we should also, or maybe alternatively, use `namespace::autoclean`.

nnutter: I guess in this case you really should just be doing `use URI`.

lib/perl/Genome/Model/Tools/Analysis/DumpIgvXmlMulti.pm
```diff
@@ -97,7 +98,8 @@ XML
 
     my $i=0;    
     for($i=0;$i<@bams;$i++){
-        my $path = abs_path($bams[$i]);
+        my $uri = URI->new($bams[$i]);
+        my $path = $uri->scheme ? $bams[$i] : abs_path($bams[$i]);
```
nnutter: `has_recognized_scheme` might be more correct than `scheme` but I don't think it actually matters because of the constructor that is used.

`file:///some/path` is a valid URI.  Would you want to `abs_path` the path in that case or leave it alone?

lib/perl/Genome/VariantReporting/Command/Wrappers/IgvSession.pm
```diff
@@ -72,8 +72,7 @@ sub bams {
 
 sub bam_paths {
     my $self = shift;
-    #return join(',', map {File::Spec->join($ENV{GENOME_SYS_SERVICES_FILES_URL}, $_)} values %{$self->bams});
-    return join(',', values %{$self->bams});
+    return join(',', map {File::Spec->join($ENV{GENOME_SYS_SERVICES_FILES_URL}, $_)} values %{$self->bams});
```
nnutter: It would be slightly more portable, I think, to use `URI` to do this rather than `File::Spec`.l

lib/perl/Genome/Model/Tools/Analysis/DumpIgvXmlMulti.pm
susannasiebert: Calling `abs_path` on a URI returns undef.

susannasiebert: It looks like we're using an old version of URI (1.52) that doesn't support `has_recognized_scheme`.

nnutter: > Calling `abs_path` on a URI returns undef.

You would call it on the URI's path and only for file schemed URI if you were to care about it in that case.

nnutter:
I noticed in the failed test that the diff has,

```
<Resource path="file://fake/path1" relativePath="false"/>
```

I think that `path` is incorrect and should have three slashes, i.e. `file:///fake/path1`.  I'm not sure where that's coming from.

susannasiebert:
Looks like File::Spec->join is not doing the correct thing anyway. Using URI instead fixes that. I will push a new commit shortly.

lib/perl/Genome/Model/Tools/Analysis/DumpIgvXmlMulti.pm
susannasiebert: For some reason I'm not able to reproduce this behavior of abs_path anymore. I will do some more testing but you're right that this needs some more refined behavior.

susannasiebert:
@nnutter: I pushed a few commits that address your comments.

lib/perl/Genome/VariantReporting/Command/Wrappers/IgvSession.pm
```diff
@@ -72,8 +73,10 @@ sub bams {
 
 sub bam_paths {
     my $self = shift;
-    #return join(',', map {File::Spec->join($ENV{GENOME_SYS_SERVICES_FILES_URL}, $_)} values %{$self->bams});
-    return join(',', values %{$self->bams});
+    return join(',', map {
+        my $u = URI->new($_, $ENV{GENOME_SYS_SERVICES_FILES_URL});
```
nnutter: `$ENV{GENOME_SYS_SERVICES_FILES_URL}` doesn't need to show up twice.  Instead it can be either:

```
my $u = URI->new_abs($_, $ENV{GENOME_SYS_SERVICES_FILES_URL});
$u->as_string;
```

or:

```
my $u = URI->new($_);
$u->abs($ENV{GENOME_SYS_SERVICES_FILES_URL})->as_string;
```

lib/perl/Genome/Model/Tools/Analysis/DumpIgvXmlMulti.pm
```diff
@@ -97,7 +98,7 @@ XML
 
     my $i=0;    
     for($i=0;$i<@bams;$i++){
-        my $path = abs_path($bams[$i]);
+        my $path = resolve_bam_path($bams[$i]);
```
nnutter: This is not very important but it looks like you could just call `resolve_bam_path` once up around [line 78][1], for example,

```
my @bams = map { resolve_bam_path ($_) } split(/\,/, $self->bams);
```

[1]: https://github.com/susannasiebert/genome/blob/igv/lib/perl/Genome/Model/Tools/Analysis/DumpIgvXmlMulti.pm#L78

susannasiebert:
@nnutter: A commit to address your last two comments has been pushed

nnutter:
+1

davidlmorton:
+1

PR #395, opened 2015-01-07
Proposed Fixes for RT#103295

nnutter:
This is trying to correct the problem that region_of_interest_set_bed_file
writes to a build's data directory but is not run by the build.  This is a
common problem and this only covers when tiny occurence of it.

It also tries to address the fact that the failure left an empty file in place
and that this empty file was passed as an argument to downstream operations.

See [RT#103295][1] for full details.

/cc @jasonwalker80 @malachig 

[1]: https://rt.gsc.wustl.edu/Ticket/Display.html?id=103295

### DIFF
--- a/lib/perl/Genome/Model/Tools/CopyNumber/Cnmops.pm
+++ b/lib/perl/Genome/Model/Tools/CopyNumber/Cnmops.pm
@@ -124,7 +124,8 @@ sub get_ROI_bed{
   if($self->roi_bed) {
     return $self->roi_bed;
   } elsif($refalign->last_succeeded_build->region_of_interest_set_bed_file) {
-    return $refalign->last_succeeded_build->region_of_interest_set_bed_file;
+    my $bed_path = Genome::Sys->create_temp_file_path();
+    return $refalign->last_succeeded_build->region_of_interest_set_bed_file($bed_path);
   } else {
     die $self->error_message("Unable to find ROI_set_bed file for " . $refalign->name);
   }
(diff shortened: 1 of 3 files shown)


lib/perl/Genome/Model/Tools/CopyNumber/Cnmops.pm
```diff
@@ -124,7 +124,8 @@ sub get_ROI_bed{
   if($self->roi_bed) {
     return $self->roi_bed;
   } elsif($refalign->last_succeeded_build->region_of_interest_set_bed_file) {
-    return $refalign->last_succeeded_build->region_of_interest_set_bed_file;
+    my $bed_path = Genome::Sys->create_temp_file_path();
+    return $refalign->last_succeeded_build->region_of_interest_set_bed_file($bed_path);
```
tmooney: This will end up possibly generating the BED file twice--once in the refcov directory (in the `elsif` condition) and once in the temp file (here).

nnutter: Thanks, pushed a fix.

tmooney:
+1

malachig:
@gatoravi not sure if practical since you are out of town, but if possible could you take a look at these commits?  They are relevant to a clin-seq step where we do cnv analysis using cnmops with exome data.

gatoravi:
+1 for the clinseq change.

malachig:
Tests were passing, and everyone is quite impatient for the results of this fix, so I went ahead and merged.  Hope this is okay.  Thanks @nnutter for looking at this and @tmooney and @gatoravi for the review.

PR #396, opened 2015-01-08
Speedup CopyCat Annotation Lister

tmooney:
The lister by default shows (and therefore "hints" at all the properties of the SoftwareResult, which triggers the giant non-performant query on SoftwareResult with joins on the Inputs and Params.  This change means more (but smaller) DB queries, which reduces the runtime of the command from several minutes to under three seconds.

### DIFF
--- a/lib/perl/Genome/Model/Tools/CopyCat/List.pm
+++ b/lib/perl/Genome/Model/Tools/CopyCat/List.pm
@@ -10,6 +10,9 @@ class Genome::Model::Tools::CopyCat::List {
             is_constant => 1,
             value => 'Genome::Model::Tools::CopyCat::AnnotationData',
         },
+        show => {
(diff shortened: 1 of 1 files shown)


tmooney:
@chrisamiller Are there any other columns you would like to see by default?  (The `--show` parameter can still be used to display any others.)

chrisamiller:
That looks perfect. +1

nnutter:
Is the `_hint_string()` really helping?  With default show, but without the `hint_string()` override, the lister does 4 queries with the first having 2 joins.  While with the `hint_string()` override it does 13 queries.

tmooney:
@nnutter The `_hint_string` doesn't have much effect on the new default show, but it does dramatically improve listers with assorted custom shows (including the original default show).

nnutter:
Alright, +1.

PR #397, opened 2015-01-09
Updating SraToBam.pm to use new input params. 

dummey:


apipe-tester:

nnutter:
CI, test this once.

nnutter:
@dummey, I'm assuming this crashed because of `$self->library`?  Should the doc string for `library_name` or `sample_name` be updated as well?  They both mention being used in the RG headers.  Otherwise, +1 if tests pass.

tmooney:
+1

PR #398, opened 2015-01-09
Fix a bug on comparing flagstat

dufeiyu:
Expect one line diff on duplicates between extracting bam flagstat and comparison flagstat. Expect no diff between reverting Markdup bam flagstat and comparison flagstat.The comparison flagstat is always from the original per lane bam that is not markdupped. The unit test is updated accordingly.

### DIFF
--- a/lib/perl/Genome/InstrumentData/AlignmentResult/Command/RecreatePerLaneBam.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult/Command/RecreatePerLaneBam.pm
@@ -97,12 +97,15 @@ sub execute {
     $self->debug_message('Extract read group bam');
     $self->_extract_readgroup_bam($temp_bam);
 
-    $self->debug_message('Compare flagstat');
-    $self->_compare_flagstat($temp_bam, $flagstat);
+    $self->debug_message('Compare flagstat first time expect one diff');
+    $self->_compare_flagstat($temp_bam, $flagstat, 'diff');
 
     $self->debug_message('Revert bam Markdup tag');
     $self->_revert_markdup($temp_bam, $no_markdup_bam);
     
+    $self->debug_message('Compare flagstat second time expect no diff');
+    $self->_compare_flagstat($no_markdup_bam, $flagstat);
+
     $self->debug_message('Reheader bam');
     $self->_reheader_bam($no_markdup_bam, $bam_header);
     
@@ -136,7 +139,7 @@ sub _extract_readgroup_bam {
 
 
 sub _compare_flagstat {
-    my ($self, $temp_bam, $flagstat) = @_;
+    my ($self, $temp_bam, $flagstat, $diff) = @_;
 
     my $temp_flagstat = Genome::Sys->create_temp_file_path;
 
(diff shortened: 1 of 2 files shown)


gsanders:
+1

The only thing I'd possibly change is the way you do the comparison. compare_ok has a lot of options for doing comparisons and expecting differences. There may be a way to use this instead of your specialized diff. However, your diff is fine too because it is very specific (expecting one line), probably more specific than you can be with compare_ok.

lib/perl/Genome/InstrumentData/AlignmentResult/Command/RecreatePerLaneBam.pm
```diff
@@ -136,7 +139,7 @@ sub _extract_readgroup_bam {
 
 
 sub _compare_flagstat {
-    my ($self, $temp_bam, $flagstat) = @_;
+    my ($self, $temp_bam, $flagstat, $diff) = @_;
```
nnutter: I think that this whole "diff" thing should just be a separate function.  However, if you keep it like this then I think calling `$diff` something else, maybe `$ignore_duplicates`, would be a better description.

lib/perl/Genome/InstrumentData/AlignmentResult/Command/RecreatePerLaneBam.pm
```diff
@@ -150,8 +153,17 @@ sub _compare_flagstat {
         die $self->error_message("Fail to run flagstat on $temp_bam");
     }
     
-    unless (compare($temp_flagstat, $flagstat) == 0) {
-        die $self->error_message("The bam flagstat from the extracting is different from the comparison flagstat $flagstat");
+    if ($diff) { #expect only one line diff on duplicates
```
nnutter: Possible alternatives implementations...

1) Might be better because it leaves parsing of the flagstat file, however trivial it may be, in the place designated for parsing flagstat files.

```
use List::MoreUtils qw(all uniq);
...
my %flagstat_data = Genome::Model::Tools::Sam::Flagstat->parse_file_into_hashref($flagstat);
my %temp_flagstat_data = Genome::Model::Tools::Sam::Flagstat->parse_file_into_hashref($temp_flagstat);
delete $flagstat_data{reads_marked_duplicates};
delete $temp_flagstat_data{reads_marked_duplicates};
my @keys = uniq keys(%flagstat_data), keys(%temp_flagstat_data);
unless (all { $flagstat_data{$_} eq $temp_flagstat_data{$_} } @keys) {
    die ...
}
```

2) This technique would allow `_compare_flagstat()` to take `$comparator` as an argument just like `compare()` can.

```
my $comparator = sub { $_[0] !~ /duplicates/ && $_[0] ne $_[1] };
$self->_compare_flagstat($temp_bam, $flagstat, $comparator);
...
sub _compare_flagstat {
    my ($self, $temp_bam, $flagstat, $comparator) = @_;
    ...
    my @compare_args = ($temp_flagstat, $flagstat);
    if ($comparator) {
        push @compare_args, $comparator;
    }
    ...
    unless (compare(@compare_args) == 0) {
        die ...
    }
}
```

dufeiyu:
@nnutter  thanks for the detailed suggestion. I like the alternative 1 and implemented it into the latest commit.

nnutter:
+1

PR #399, opened 2015-01-09
Don't die if no allele frequency information can be found.

davidlmorton:
It is possible to have cases where there is no allele frequency
information because the dbsnp vcf has it listed as '.'.  Another way
this can happen is if dbsnp doesn't have any information about the
allele at all, but does have information about another allele in that
vcf line.

### DIFF
--- a/lib/perl/Genome/VariantReporting/Suite/Joinx/Dbsnp/DbsnpInterpreter.pm
+++ b/lib/perl/Genome/VariantReporting/Suite/Joinx/Dbsnp/DbsnpInterpreter.pm
@@ -43,8 +43,9 @@ sub _interpret_entry {
     my $caf = $parser->process_entry($entry);
     my %return_values;
 
+    my $has_tag = $entry->info($parser->TAG);
     for my $variant_allele (@$passed_alt_alleles) {
-        if (defined($entry->info('CAF')) && !defined($caf->{$variant_allele})) {
+        if ($has_tag && !defined($caf->{$variant_allele})) {
             # This may be fine, but I want to hear about it.  I think it can
             # happen in two ways:
             #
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/VariantReporting/Suite/Joinx/Dbsnp/DbsnpInterpreter.pm
```diff
@@ -44,16 +44,18 @@ sub _interpret_entry {
...
-        if (!defined $caf) {
-            $return_values{$variant_allele}->{allele_frequency} = undef;
-        }
-        elsif (defined $caf->{$variant_allele}) {
-            $return_values{$variant_allele}->{allele_frequency} = $caf->{$variant_allele};
-        }
-        else {
-            die sprintf("No allele frequency for allele ($variant_allele) given in caf: %s",
-                Data::Dumper::Dumper($caf));
+        if (defined($entry->info('CAF')) && !defined($caf->{$variant_allele})) {
```
tmooney: The call to `info()` could happen outside of the loop.

tmooney: Also, would it be better to use `$parser->TAG` instead of hard-coding `'CAF'` here?

lib/perl/Genome/VariantReporting/Suite/Joinx/Dbsnp/DbsnpInterpreter.pm
```diff
@@ -43,17 +43,20 @@ sub _interpret_entry {
     my $caf = $parser->process_entry($entry);
     my %return_values;
 
+    my $has_tag = $entry->info($parser->TAG);
     for my $variant_allele (@$passed_alt_alleles) {
-        if (!defined $caf) {
```
nnutter: Was this case impossible?  If not then the new code will emit warnings in the two places `$caf->{$variant_allele}` is called.

susannasiebert:
Is this interpreter actually being used in the wild? There is also the CafInterpreter which looked to me like it was superseding the DbsnpInterpreter and I was planning on removing the DbsnpInterpreter in an upcoming PR.

davidlmorton:
@susannasiebert and I discussed this, and we're going to delete this interpreter in leu of the other existing interpreters that overlap functionally.

https://github.com/genome/genome/pull/418

PR #400, opened 2015-01-10
Added CustomReport

davidlmorton:
This report lets you specify the header columns in the plan file.


tmooney:
+1.

Is there a way to specify a list of parameters in the plan? That might be better than splitting on `,`, although if the labels definitely won't contain commas this syntax is probably simpler.

davidlmorton:
There is a way but the order cannot be guaranteed for is_many properties.

acoffman:
+1

Could you use an ARRAY type rather than an is-many property? Don't know if that preserves order or not.

davidlmorton:
It does preserve the order, but then it cannot be persisted in the DB :(

acoffman:
Exciting! Commas it is then :+1:

brummett:
You could keep them ordered if the Header Labels were instances of another class with text and a rank.  Then, they could be ordered by rank and saved in the DB.

davidlmorton:
Apipe-ci says this passes tests.

PR #401, opened 2015-01-12
Implement a custom sorting algorithm for vaf headers

susannasiebert:
For CLE in particular we want the vaf columns to be sorted in a way that makes intuitively sense to the analysts, which doesn't follow an alphabetical order. This sorting algorithm return the vaf columns in the desired order. Specifically, we want the vaf columns to be sorted by the vaf fields in the order of ref_count, var_count, vaf and we want the columns to be sorted by the CLE sample labels in the order of Normal, Discovery, Followup. Please see @jasonwalker80's comment on https://jira.gsc.wustl.edu/browse/AT-686 and the added test file in this PR for an example of the desired sort order.


lib/perl/Genome/VariantReporting/Suite/BamReadcount/VafInterpreterHelpers.pm
```diff
@@ -74,16 +74,79 @@ sub many_samples_field_descriptions {
...
+    my %params        = @_;
+    my $headers       = $params{headers};
+    my $sample_names  = $params{sample_names};
+    my $library_names = $params{library_names};
+
+    my @annotated_headers;
+    for my $header (@$headers) {
+        my %annotated_header = (
+            header => $header,
+            library => 0,
```
tmooney: Are default values for `sample` and `vaf_priority` also needed here?

susannasiebert: No. The way that it's set up, `create_sample_specific_field_names` and `create_library_specific_field_names` (the subroutines that create the original, un-annotated headers), will always return strings that include one of the vaf fields as well as one of the sample names.

lib/perl/Genome/VariantReporting/Suite/BamReadcount/VafInterpreterHelpers.t
```diff
@@ -0,0 +1,131 @@
...
+    $ENV{UR_DBI_NO_COMMIT} = 1;
+    $ENV{UR_USE_DUMMY_AUTOGENERATED_IDS} = 1;
+}
+
+use strict;
+use warnings;
+
+use above "Genome";
+use Test::More;
+use List::Util qw(shuffle);
```
davidlmorton: I'm not sure about using shuffle here.  Perhaps if we repeated the test a number of times we could guarantee that we didn't just accidentally pass it.  Does shuffle guarantee that the order isn't coincidentally the same as the input order?

nnutter: Shuffling isn't really important to unsort a sorted list but for fun:

```
sub unsort {
    my @sorted = @_;
    my $middle = int(@sorted / 2);
    my @left = splice(@sorted, 0, $middle);
    return shuffle(@sorted), shuffle(@left);
}
```

susannasiebert:
@brummett @davidlmorton I pushed some commits to address your comments

lib/perl/Genome/VariantReporting/Suite/BamReadcount/VafInterpreterHelpers.pm
tmooney: Fair enough! Maybe this subroutine should die if they're not present as expected.

davidlmorton:
+1 with defaults for sample and vaf_priority

susannasiebert:
@tmooney @davidlmorton : I implemented defaults for sample and vaf_priority

tmooney:
+1.  (I prefer @nnutter's `unsort` algorithm to the non-deterministic `while` approach, but if we're concerned about `shuffle` I'd rather permute the list and try them all.)

susannasiebert:
I decided to go with @nnutter's unsort approach.

PR #402, opened 2015-01-12
Some minor changes to the variant caller interpreter

susannasiebert:
- List the valid_callers explicitly in the plan files instead of having a default
- Add GatkSomaticIndel and Pindel to the list of valid_callers in the indel plan files

### DIFF
--- a/lib/perl/Genome/VariantReporting/plan_files/gold_germline_report_indels.yaml
+++ b/lib/perl/Genome/VariantReporting/plan_files/gold_germline_report_indels.yaml
@@ -64,6 +64,10 @@ reports:
                 normal_sample_names:
                     - normal
             'variant-callers':
+                valid_callers:
+                    - VarscanSomatic
+                    - Sniper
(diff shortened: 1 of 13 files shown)


tmooney:
+1

lib/perl/Genome/VariantReporting/plan_files/gold_germline_report_indels.yaml
```diff
@@ -64,6 +64,10 @@ reports:
                 normal_sample_names:
                     - normal
             'variant-callers':
+                valid_callers:
+                    - VarscanSomatic
+                    - Sniper
+                    - Strelka
```
davidlmorton: I think you missed the Indel callers here.

susannasiebert: Fixed.

davidlmorton:
+1

PR #405, opened 2015-01-13
Shellpipes

tabbott:
I need to make some commands involving bash pipes. This PR generalizes the code for dealing with pipe exit codes found in gmt bwa run-mem into its own (reusable) module.

### DIFF
--- a/lib/perl/Genome/Sys/ShellPipeline.pm
+++ b/lib/perl/Genome/Sys/ShellPipeline.pm
@@ -1,4 +1,4 @@
-package Genome::Model::Tools::ShellPipeline;
+package Genome::Sys::ShellPipeline;
 
 use strict;
 use warnings;
@@ -10,7 +10,7 @@ use File::Spec qw();
 use List::AllUtils qw(any);
 use File::Slurp qw(read_file write_file);
 
-class Genome::Model::Tools::ShellPipeline {
+class Genome::Sys::ShellPipeline {
     is => "UR::Object",
     attributes_have => [qw(is_input is_output is_optional)],
     has_input => {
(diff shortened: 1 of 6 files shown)


tmooney:
I would put something like this under `Genome::Sys` rather than `Genome::Model::Tools`, but otherwise +1.

tabbott:
All suggestions taken.

acoffman:
+1

lib/perl/Genome/Sys/ShellPipeline.pm
```diff
@@ -0,0 +1,190 @@
...
+    chmod oct(755), $script_path;
+
+    try {
+        Genome::Sys->shellcmd(cmd => $script_path);
+    }
+    catch {
+        croak $self->error_message($_);
+    }
+    finally {
+        $self->_check_pipestatus($pipestatus_path);
```
nnutter: Since `_check_pipestatus` wants to throw exceptions I'm not sure this is a safe use of `finally`.  From the docs:

> Furthermore exceptions in finally blocks are not trappable and are unable to influence the execution of your program. This is due to limitation of DESTROY-based scope guards, which finally is implemented on top of. This may change in a future version of Try::Tiny.

I think `_check_pipestatus` can just be moved inside the `try` block?

brummett: Moving ```_check_pipestatus()``` into the try block is what I was trying to suggest.  Then, that catch is all you need to handle errors.

tabbott: Moving inside the try block does not get me what I want, which is to have check_pipestatus whether or not shellcmd throws. Finally does this, but as you have shown, throwing from DESTROY based scope guards is bad. I guess I can buffer errors and throw them later.

brummett:
+1

PR #406, opened 2015-01-13
Fix allele translations for bam readcount

susannasiebert:
The current translation doesn't work for indels at arbitrary positions in the reference. We just assume that the indel is after the first reference allele. This PR will fix that and return the correct translation an long indels at arbitrary positions in the reference.


davidlmorton:
+1

davidlmorton:
I'd like to have feedback from @tabbott on this, but @ernfrid has been looking at real data run off of this change and seems satisfied, so we might merge before we get that feedback.

davidlmorton:
Also, apipe-ci tested this and it succeeded, so I'm not sure why it is still 'Waiting to hear about'

gsanders:
I think I'd like to see some comments in translate_pure_indels. I believe I understand what you were doing but it left me scratching my head a little.

I think a simple example would perhaps help?

susannasiebert:
I will add some comments. For some examples on what this code is supposed to do, with have some test cases that might clear things up a bit.

gsanders:
Looks awesome! +1

tabbott:
@davidlmorton 

This is how I think about the problem (in Perl). It performs a little better on my machine:

http://codepad.org/n6ZjezBF

This will get you the smallest representation of REF and ALT (short of splitting into multiple smaller variants which would involve dynamic programming).

If you do that, then the you can easily (and correctly I might add) check if something is a pure INS or pure DEL by testing if the REF or ALT are empty (respectively).

susannasiebert:
@tabbott : I made a new Jira issue for implementing your suggestion https://jira.gsc.wustl.edu/browse/AT-725

PR #409, opened 2015-01-14
Make sample/library name_labels transient

davidlmorton:
HASH type properties cannot be persisted in the DB.  It would be nice to have these values persisted though because if they change, then the Report that gets generated may be different.

I'd like to use UR::Value::JSON for these, but it seems that doesn't "just work".  Perhaps a discussion with @brummett is in order...


acoffman:
+1 :rocket:

PR #412, opened 2015-01-15
Make library_names optional for VafInterpreter

susannasiebert:
We use the VafInterpreter without library_names in the
MinCoverageObservedInterpreter and the MaxVafObservedInterpreter so it needs to be optional.

### DIFF
--- a/lib/perl/Genome/VariantReporting/Suite/BamReadcount/MaxVafObservedInterpreter.pm
+++ b/lib/perl/Genome/VariantReporting/Suite/BamReadcount/MaxVafObservedInterpreter.pm
@@ -56,7 +56,6 @@ sub _interpret_entry {
         for my $sample_name ($self->$sample_name_accessor) {
             my $interpreter = Genome::VariantReporting::Suite::BamReadcount::VafInterpreter->create(
                 sample_name => $sample_name,
-                library_names => [],
             );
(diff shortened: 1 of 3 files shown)


mkiwala:
+1

tmooney:
+1

davidlmorton:
I think that the `VafInterpreter` is pulling double duty and should probably be refactored into a base-class (that doesn't inherit from `WithManyLibraryNames`) and a derived class (that does).  That way we can reuse the base-class and keep the `library_names` a required property.

There may be other options here too, such as refactoring the interpret_entry out into functions that could be used by the other vaf interpreters so they wouldn't need to create any `VafInterpreter` objects.

acoffman:
+1 

I don't know enough about the VariantReporting framework to actually have an opinion, but it does seem odd that something that "isa" `Genome::VariantReporting::Framework::Component::WithManyLibraryNames`. Though if I had more context, it could make total sense.

susannasiebert:
@davidlmorton Yeah, I wasn't totally happy with this fix either.I like your solution better and will work on implementing it.

susannasiebert:
After talking to @davidlmorton in person we decided to move ahead with this PR so that we can get out a new CLE snapshot, but make a Jira issue to improve this further in a different PR.

PR #416, opened 2015-01-15
AnP: do some validation for subject mapping imports

ebelter:


lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Import/SomaticValidation.pm
```diff
@@ -86,6 +82,24 @@ sub execute {
...
+    if ( $line->{normal_sample} eq $line->{tumor_sample} ) {
+        die $self->error_message('Same sample given for normal/tumor: %s', $line->{normal_sample});
+    }
+
+    my %subject_common_names_seen;
+    for my $subject_label (@subject_labels) {
+        my $subject_identifier = $line->{$subject_label};
+        next unless $subject_identifier;
+        my $subject_common_name = $self->_create_subject($mapping, $subject_label, $subject_identifier);
+        $subject_common_names_seen{$subject_common_name.'_sample'}++;
```
tmooney: Is `%subject_common_names_seen` being used anywhere?

ebelter: Nope - relic of more checking. Will remove.

tmooney:
@acoffman reminded me that, in the single-sample case, a normal may end up in the tumor_sample position.  It's only potentially wrong if it's paired with something else.

acoffman:
So, I actually am not a huge fan of the extra validation.

I think that adding conditional code that makes assumptions about what sorts of experiments we want to run and how they should be set up is the path back to something like AQID. I think that we should continue to validate that we're given real samples, but otherwise accept what's passed in, assuming the analyst knew their experimental design better than we do.

I realize that opens the doors for mistakes, and other people probably feel differently. At its core however, I think the AnP Config system is there to allow analysts to set up and configure projects that they want to run and stay out of the way otherwise. I think that getting back in the business of making assumptions about how things should be set up or run is a slippery slope.

If others disagree, I'm willing to lose on this. I just figured I'd offer my two cents.

ebelter:
It doesn't make assumptions, it enforces the setup of the SomVal model. Maybe it's definition is inadequate for are experiments.

ebelter:
That said - I do not really care about the extra validation. I am fixing some flipped normal/tumor samples and was trying to prevent this from happening in the future.

jasonwalker80:
I'm all for preventing users from doing the wrong thing, but there way too many edge cases that encoding validation or enforcement of sample meta-data properties is a nightmare.  The rules for validation could change on a per-project basis.  The fact that we already have normal samples being entered into a tumor_sample column (ex. Germline with SomVal) is confusing.  My vote is to avoid validation that makes assumptions about what the user intended.

ebelter:
I will close this PR then.

PR #417, opened 2015-01-15
bsmap alignment status and convertion rate from chrM and lambda

litd:


lib/perl/Genome/Model/Tools/Bsmap/MethCalcConversionRate.pm
```diff
@@ -0,0 +1,263 @@
...
+
+sub help_brief {
+    "calculate methylation conversion using Mitochondrial DNA or spiked lambda"
+}
+
+sub help_detail {
+    "calculate methylation conversion using Mitochondrial DNA or spiked lambda"
+}
+
+sub bs_rate {
```
nnutter: It is a _Perl Best Practice_ to "Always unpack `@_` first" rather than using `@_` throughout a subroutine.  This is to better document what the expected inputs are as well as to avoid accidentally changing one of the inputs.

lib/perl/Genome/Model/Tools/Bsmap/MethCalcConversionRate.pm
```diff
@@ -0,0 +1,263 @@
...
+
+class Genome::Model::Tools::Bsmap::MethCalcConversionRate {
+    is => 'Command',
+    has => [
+        snvs_file => {
+            is => 'String',
+            is_optional => 1,
+            doc => 'Use snvs.hq file to calculate methylation conversion',
+        },
+        model_id => {
```
nnutter: It's generally true that tools in `Genome::Model::Tools` are supposed to be file-based and not interact with the Genome database.  So taking a model ID as an input is breaking that convention.

If that convention needs to be broken then this can take a model as an input instead of an ID by changing its type from 'String' to 'Genome::Model' and removing the `Genome::Model->get()` call on line 127.

lib/perl/Genome/Model/Tools/Bsmap/MethCalcConversionRate.pm
```diff
@@ -0,0 +1,263 @@
...
+    my $methreads = 0;
+
+    open (my $fh, '<', $_[0]) or die("Could not open snvs file.");
+    while (my $line = <$fh>) {
+        if ( $count == 0 ) {
+            $count = 1;
+            next;
+        }
+        chomp($line);
+        my @F = split("\t",$line);
```
nnutter: Parsing of files should generally not be done ad-hoc like this.  We have parser library for dealing with tab-delimited data.  It may also be beneficial to abstract parsing of this type of file so that it can be re-used.

lib/perl/Genome/Model/Tools/Bsmap/MethCalcConversionRate.pm
```diff
@@ -0,0 +1,263 @@
...
+        }
+    }
+
+    if (defined($model_id)){
+        # get model IDs
+        my $model = Genome::Model->get($model_id);
+        # get the bam paths of the last succeeded build
+        my $dir = $model->last_succeeded_build->data_directory;
+        # flagstat 
+        my @flagstat = glob("$dir/alignments/*flagstat");
```
nnutter: It is better to ask the build for its flagstat file that to reach in to the filesystem.  For example, `Genome::Model::Build::ReferenceAlignment::whole_rmdup_bam_flagstat_file()`.

lib/perl/Genome/Model/Tools/Bsmap/MethCalcConversionRate.pm
```diff
@@ -0,0 +1,263 @@
...
+        my (@field, $total, $duplicates, $mapped, $properly);
+        for my $flagstat (@flagstat) {
+            if (-s "$flagstat"){
+                if (defined($output_file)){
+                    open($cfile, ">$output_file") || die "Can't open output file for writing.\n";
+                    print $cfile "\nMethylation alignment status:\n";
+                    print $cfile $flagstat, "\n";
+                    open (my $fh, '<', $flagstat) or die("Could not open flagstat file.");
```
nnutter: See `Genome::Model::Tools::Sam::Flagstat::parse_file_into_hashref()`.

lib/perl/Genome/Model/Tools/Bsmap/MethCalcConversionRate.t
```diff
@@ -0,0 +1,46 @@
...
+use Genome::Utility::Test qw(compare_ok abort);
+use Test::More tests => 8;
+
+eval {
+    # create a class instance
+    my $class = 'Genome::Model::Tools::Bsmap::MethCalcConversionRate';
+    use_ok($class);
+
+    # check test data files
+#    my $data_dir = Genome::Utility::Test->data_dir($class, 'v2');
```
nnutter: Just delete unused code instead of commenting it out.

lib/perl/Genome/Model/Tools/Bsmap/MethCalcConversionRate.pm
```diff
@@ -0,0 +1,263 @@
...
+                $self->error_message("can't find lambda snvs file");
+            }
+        }
+    }
+1;
+
+}
+1;
+
+# methratio 2.74
```
nnutter: Unneeded comments?

lib/perl/Genome/Model/Tools/Bsmap/MethCalcConversionRate.pm
```diff
@@ -0,0 +1,263 @@
...
+        }
+        print $cfile "Meth reads\t=\t", $methreads, "\n";
+        print $cfile "Total reads\t=\t", $totalreads, "\n";
+        if ($totalreads != 0) {
+            print $cfile "Bisulfite conversion (%)\t=\t", 100-($methreads/$totalreads*100), "\n\n";
+        }
+        close $cfile;
+    }
+    else {
+        print "Meth reads\t=\t", $methreads, "\n";
```
nnutter: This could be de-duplicated with the above block by always printing to a handle.  Just default to a handle to stdout if `$_[2]` is undefined.

lib/perl/Genome/Model/Tools/Bsmap/MethCalcConversionRate.pm
```diff
@@ -0,0 +1,263 @@
...
+        die("Must provide snvs file OR model ID.\n");
+    }
+
+    my $cfile;
+
+    if (defined($snvs_file)){
+        # snvs
+        if (-s "$snvs_file"){
+            if (defined($output_file)){
+                open($cfile, ">$output_file") || die "Can't open output file for writing.\n";
```
nnutter: These two cases would also get de-duplicated if `bs_rate()` always wrote to a handle and default to a handle to stdout.

lib/perl/Genome/Model/Tools/Bsmap/MethCalcConversionRate.pm
```diff
@@ -0,0 +1,263 @@
...
+                    print "Methylation conversion based on lambda:\n";
+                    bs_rate($lambda_hq, "lambda");
+                }
+            }
+            else {
+                $self->error_message("can't find lambda snvs file");
+            }
+        }
+    }
+1;
```
nnutter: This should be indented.  I would also use an explicit `return`.

lib/perl/Genome/Model/Tools/Bsmap/MethCalcConversionRate.pm
```diff
@@ -0,0 +1,263 @@
...
+                    }
+                    close($flagstat);
+                }
+            }
+            else {
+                $self->error_message("can't find flagstat file");
+            }
+        }
+        # mt snvs.hq
+        my @mt_hq = glob("$dir/variants/snv/meth-ratio-*/MT/snvs.hq");
```
nnutter: You could loop over these special chromosome cases to deduplicate the code.

nnutter:
I would be happy to pair with you sometime if you want to work on this.  Just contact me on instant messaging or email.

lib/perl/Genome/Model/Tools/Bsmap/MethCalcConversionRate.pm
chrisamiller: The purpose of this is to give some QC metrics to production on bisulfite data runs. It probably wouldn't be out of place to have this run automatically on every bisulfite model and output into the build dir. I have no idea how to accomplish that though, given that bisulfite is just a refalign model with a specific PP, and that bamQC is disabled for these PPs. If someone more familiar with that corner of the code wants to weigh in, it'd be appreciated.  (@jwalker?)

jasonwalker80: The command operations will need to run should go under Genome::Model::ReferenceAlignment::Command.  There are currently a handful of QC commands located there.  All of the tool can be moved or you could refactor the tool to operate on files alone and the G:M:RA:Command would call the tool after gathering the required file inputs.  There are several QC tools that would benefit from automation.  For now, the command should work to serve operations needs.

lib/perl/Genome/Model/ReferenceAlignment/Command/BsmapMethCalcConversionRate.pm
```diff
@@ -0,0 +1,100 @@
...
+
+use strict;
+use warnings;
+use Genome;
+use Genome::Model::Tools::Bsmap::MethCalcConversionRate;
+
+class Genome::Model::ReferenceAlignment::Command::BsmapMethCalcConversionRate {
+    is => 'Command::V2',
+    has => [
+        model => {
```
tmooney: Just to note... If this will be added to the pipeline, it will need to accept a specific build instead of a model.  (From the command line it could still be passed a model ID, and it will prompt the user to select one of the builds.)

tmooney:
+1, this looks good.

It might be possible to simplify the printing code further with something like
```perl
sub print_statistic {
  my $cfile = shift;
  my $label = shift;
  my $value = shift;
  print $cfile $label, "\t=\t", $value, "\n";
}
```
so each line from the flagstat becomes something like
```perl
print_statistic($cfile, 'Total read', $flagstat_data->{total_reads});
```

nnutter:
@litd, I realized we forgot to fix the mixing of tabs and spaces.  I'll help you setup Vim to use spaces and we can covert the tabs to spaces.  (I think spaces are more common.)

lib/perl/Genome/Model/ReferenceAlignment/Command/BsmapMethCalcConversionRate.pm
```diff
@@ -0,0 +1,100 @@
...
+    "calculate methylation conversion using Mitochondrial DNA or spiked lambda"
+}
+
+sub execute {
+    my $self = shift;
+    my $output_file = $self->output_file;
+
+	my $cfile;
+	if ($output_file) {
+		open($cfile, '>', $output_file) or die;
```
susannasiebert: We generally prefer to use the subroutines provided in Genome::Sys for things like these. It's a good package to familiarize yourself with as it provides a lot of functionality. Here you can use Genome::Sys->open_file_for_writing. It already does the error handling for you.

susannasiebert:
+1 after the remaining small issues are addressed. Looks good, though!

litd:
Thanks. I changed most suggestions.

PR #420, opened 2015-01-21
create /tmp path if NE for bwa mem

tabbott:
because some tools just really insist on a prefix existing :p

### DIFF
--- a/lib/perl/Genome/Model/Tools/Bwa/RunMem.pm
+++ b/lib/perl/Genome/Model/Tools/Bwa/RunMem.pm
@@ -403,7 +403,13 @@ sub _sort_commands {
     my $comp_level = 0;
     my $threads = $self->num_threads;
     my $per_thread_mem = $self->_sort_memory_per_thread_mb;
-    my $tmp_path = File::Spec->catfile($self->temp_dir, "samtools-sort");
+    my $tmp_path = $self->temp_dir;
+
+    if (!$tmp_path) {
+        $tmp_path = Genome::Sys->create_temp_file_path;
+    }
+
+    my $tmp_path = File::Spec->catfile($tmp_path, "samtools-sort");
 
     $self->debug_message("max per-thread sorting memory is ${per_thread_mem}MB");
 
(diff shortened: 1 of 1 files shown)


ernfrid:
+1

We need to expedite a snapshot of this code once this is merged.

lib/perl/Genome/Model/Tools/Bwa/RunMem.pm
```diff
@@ -403,7 +403,10 @@ sub _sort_commands {
     my $comp_level = 0;
     my $threads = $self->num_threads;
     my $per_thread_mem = $self->_sort_memory_per_thread_mb;
-    my $tmp_path = File::Spec->catfile($self->temp_dir, "samtools-sort");
+    my $tmp_path = $self->temp_dir;
+
+    $tmp_path = Genome::Sys->create_temp_file_path unless $tmp_path;
```
tmooney: Would `Genome::Sys->create_temp_directory` be more appropriate here?

jasonwalker80: +1, given the priority.  Many X Ten validation alignments are failing because of this.  Can we merge ASAP unless there are objections?

My assumption is that one can define temp_dir with a network disk location or a defined output location to monitor the temp usage.  I'm also assuming that in our pipelines no temp_dir is defined so `Genome::Sys->create_temp_file_path` turns out to be the method called.

tmooney: There's a temporary directory [here](https://github.com/genome/genome/blob/56113b35d095ec66cfba3117ce51aec94311872b/lib/perl/Genome/InstrumentData/AlignmentResult/BwamemStream.pm#L138) already.

Perhaps that one should be passed along?

ernfrid: I was wondering that as well. I think this code is still necessary based on the documentation though.

jasonwalker80: @tmooney, Can we merge this PR for the tool change and make a new one to pass the temp_scratch_directory from the AlignmentResult to the tool?

tmooney: I've opened #422 which contains a version of this PR that I would +1.

That said, this PR looks like it will work as-is if you choose to merge it.

PR #421, opened 2015-01-21
Flexbar --(no-)length-dist parameter

tmooney:
The parameter changed between the two flexbar versions.

It looks like some other parameters changed, too, which this command handled by having both and cautioning the user to use one or the other.  It seemed really confusing to me to end up with
```
--length-dist
--nolength-dist
--no-length-dist
--nono-length-dist
```
half of which work with one version and half with another, so I took a different approach here.  If that's against the spirit of SX, I could switch it to the above.

### DIFF
--- a/lib/perl/Genome/Model/Tools/Sx/Trim/Flexbar.pm
+++ b/lib/perl/Genome/Model/Tools/Sx/Trim/Flexbar.pm
@@ -18,7 +18,7 @@ class Genome::Model::Tools::Sx::Trim::Flexbar {
             is => 'Boolean',
             is_optional => 1,
             default => 0,
-            doc => 'Remove the reverse compl;ement of the single adapter.',
+            doc => 'Remove the reverse complement of the single adapter.',
         },
         no_length_dist => {
             is => 'Boolean',
@@ -66,12 +66,12 @@ sub _cmd_properties {
         adapter_min_overlap => {
             is => 'Text',
             is_optional => 1,
-            doc => ' Minimum overlap of adapter and read in base pairs.',
+            doc => 'Minimum overlap of adapter and read in base pairs.',
         },
         adapter_threshold  => {
             is => 'Number',
             is_optional => 1,
-            doc => 'Allowed mismatches and indels per 10 bases for adapter',
+            doc => 'Allowed mismatches and indels per 10 bases for adapter.',
         },
         adapter_trim_end => {
             is => 'Text',
@@ -81,7 +81,7 @@ sub _cmd_properties {
         adapter_tail_length => {
             is => 'Number',
             is_optional => 1,
-            doc => ' Number of bases for tail trim-end types, default: adapter length',
+            doc => 'Number of bases for tail trim-end types, default: adapter length',
         },
         adapter_no_adapt => {
             is => 'Boolean',
--- a/lib/perl/Genome/Model/Tools/Sx/Trim/Flexbar.pm
+++ b/lib/perl/Genome/Model/Tools/Sx/Trim/Flexbar.pm
@@ -247,6 +247,18 @@ sub _env_var_for_version {
     return $env_var;
 }
 
+sub _resolve_length_dist_parameter {
+    my $self = shift;
+
+    my $no_length_dist = $self->no_length_dist;
+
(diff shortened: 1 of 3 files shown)


brummett:
+1

nnutter:
Side comment, I linked the [GitHub Cheat Sheet][1] on `se-lunch`.  This PR is an example of where one of the [tips][2], to ignore whitespace with `?w=1`, really helps.  Just add `?w=1` to the URL when viewing _Files changed_.

[1]: https://github.com/tiimgreen/github-cheat-sheet
[2]: https://github.com/tiimgreen/github-cheat-sheet#ignore-whitespace

lib/perl/Genome/Model/Tools/Sx/Trim/Flexbar.pm
```diff
@@ -281,6 +282,18 @@ sub _resolve_adapters {
     return 1;
 }
 
+sub _resolve_length_dist_parameter {
+    my $self = shift;
+
+    my $no_length_dist = $self->no_length_dist;
+
+    if ( $self->version >= 230 ) {
```
nnutter: The other special cases are checking `$self->version eq 230`.  I would put a small vote towards consistency and keeping this method next to the others that are also handling special cases for 230.

tmooney: I've moved this one up by the others.

I made mine a `>=` because there is a newer release of `flexbar` available, and it also follows the newer parameter convention. I've changed the others where appropriate.  (The environment variable looks version specific.)

On another note, the two installed versions are `/usr/bin/flexbar` and `/gsc/bin/flexbar`.  The public versions are all 2.x whereas we have integer version numbers.  Currently `230` is `2.4` and `229` is `2.21`.  This seems suspicious.

tmooney:
@nnutter By the way, thanks for the `?w=1` tip--that does make this one a lot easier to read!

nnutter:
+1 but I think @ebelter should look this over.

ebelter:
What triggered this pull request? Are these parameters not correct in the versioned commands of flexbar: gmt sx trim flexbar v2x4 & v2x21 ? 

I apologize for the confusion on the commands and paramters. Flexbar has a sordid history, and  changes it's parameters, which leads to confusion. I implemented the versioned commands in the flexbar directory. The reason for the main flexbar comand existence is it is referenced by several processing profiles. So we could not change and point it to a 'versioned' command.

Versions, starting with the oldest.
cmd => gmt sx trim flexbar v2x21
version => 2.21
notes=> This version is actually 2.21 + a few commits because of a bug. They fixed the bug, but did not release a new version. We used commit 229 that had the bug fix. 

cmd => gmt sx trim flexbar 
version => 230 [improperly versioned/packaged]
notes => This 'version' is not a proper version and is not packaged properly. I do not know what the real version is.  It is deployed to /gsc/bin. This is from the days before code review.

cmd => gmt sx trim flexbar v2x4
version => 2.4
This is the latest version we have, it is the true version. This is the one that should be used.There is antother, 2.5, but no one has requested that yet.

tmooney:
I was lead to make these changes after finding failures in the the Analysis Project in [RT#103581](https://rt.gsc.wustl.edu/Ticket/Display.html?id=103581).  Is it possible to update the processing-profile to something equivalent that works?

ebelter:
Yes. instdata supports putting an entire sx command into the read_trimmer_name. To use the new version of flexbar, put this command as the read_trimmer_name and do not include read trimmer params or version.

gmt sx trim flexbar v2x4 --max-uncalled 250 --min-read-length 25 --adapter-trim-end RIGHT --threads 4 --adapters /gscmnt/sata849/info/far_flexbar_adapter_sequence/adapters.may-14-2014.fasta

That said, this namespace is messed up. If you think the PR brings clarity to it - I am for it. If it makes it more confusing, then we shouldn't merge it. Up to you.

tmooney:
I've added another commit that adds the `length_dist` parameter to the `V2x4` module.  I think I prefer keeping the changes to the base module since we have existing processing profiles that expect it to work.

It sounds like we might be able to backfill processing profiles to use the replacement modules and then get rid of this jumbled one with odd versioning.

PR #422, opened 2015-01-21
create /tmp path if NE for bwa mem and pass along the AR's scratch directory

tmooney:
This is #420 with a few additional changes.


indraniel:
+1: _(For immediate needs.)_

Although inside `Genome::InstrumentData::AlignmentResult::BwamemStream`, wouldn't it be consistent to add the newly created `$tmp_path` here to the the temporary files queue:

```perl
$self->add_to_temporary_input_files_queue($tmp_path)
```
?

That way we can ensure that all the temprorary files created during alignment are removed after the `_run_aligner` subroutine ends.

tmooney:
The GMT doesn't know about alignment results, so it could not add it to that queue.  However, as long as the existing scratch directory is passed in, no new temp directory will be created inside the GMT.

jasonwalker80:
+1, please merge. close #420

tabbott:
Just saw the snapshot, thanks Tom.

PR #423, opened 2015-01-22
Change cascading behavior of Analysis Project delete().

tmooney:
* Refuse to delete if models or instrument data are assigned.
* Cascade to delete any existing timeline events.

I tested this by deleting @chrisamiller's long-standing Analysis Project he's wanted deleted.  (There were previously other issues with deleting it that were resolved by earlier schema changes.)

### DIFF
--- a/lib/perl/Genome/Config/AnalysisProject.pm
+++ b/lib/perl/Genome/Config/AnalysisProject.pm
@@ -95,14 +95,9 @@ sub delete {
         die $self->error_message($msg, $self->__display_name__, 'instrument data');
     }
 
-    eval {
-        my @events = Genome::Timeline::Event::AnalysisProject->get(analysis_project => $self);
-        for ($self->config_items, $self->subject_mappings, @events) {
-            $_->delete();
-        }
-    };
-    if(my $error = $@) {
-        die($error);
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Config/AnalysisProject.pm
```diff
@@ -86,23 +86,25 @@ sub __display_name__ {
...
-        for ($self->analysis_project_bridges) {
-            $_->delete();
-        }
-        for ($self->subject_mappings) {
+        my @events = Genome::Timeline::Event::AnalysisProject->get(analysis_project => $self);
+        for ($self->config_items, $self->subject_mappings, @events) {
             $_->delete();
         }
     };
     if(my $error = $@) {
```
nnutter: I realize this is not changing it seems like the `eval` and this could just be removed.

nnutter:
:+1: regardless of `eval` comment.

lib/perl/Genome/Config/AnalysisProject.pm
tmooney: Good point; that is pretty useless.  Removed!

acoffman:
+1 :rocket:

PR #425, opened 2015-01-22
InstData Import: unzip fastqs before running picard sam to fastq; up picard

ebelter:
version to 1.113

### DIFF
--- a/lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/FastqsToBam.pm
+++ b/lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/FastqsToBam.pm
@@ -5,6 +5,9 @@ use warnings;
 
 use Genome;
 
+use Archive::Extract;
+$Archive::Extract::PREFER_BIN = 1;
+$Archive::Extract::DEBUG = 1;
 use Data::Dumper 'Dumper';
 require File::Basename;
 require List::Util;
@@ -44,6 +47,9 @@ sub execute {
     my $self = shift;
     $self->debug_message('Fastqs to bam...');
 
+    my $unarchive_if_necessary = $self->_unarchive_fastqs_if_necessary;
+    return if not $unarchive_if_necessary;
+
     my $fastq_to_bam_ok = $self->_fastqs_to_bam;
     return if not $fastq_to_bam_ok;
 
@@ -54,6 +60,40 @@ sub execute {
     return if not $cleanup_ok;
 
     $self->debug_message('Fastqs to bam...done');
+    return 1;
+}
+
+sub _unarchive_fastqs_if_necessary {
+    my $self = shift;
+    $self->debug_message('Unarchive fastqs if necessary...');
+
+    my @new_fastq_paths;
+    for my $fastq_path ( $self->fastq_paths ) {
+        if ( $fastq_path !~ /\.gz$/ ) {
+            $self->debug_message('Unarchive not necessary for '.$fastq_path);
+            push @new_fastq_paths, $fastq_path;
+            next;
+        }
+        my $unarchived_fastq_path = $fastq_path;
+        $unarchived_fastq_path =~ s/\.gz$//;
+        $self->debug_message('Unarchiving: '.$fastq_path);
+        my $rv = eval{ Genome::Sys->shellcmd(cmd => "gunzip $fastq_path"); };
+        #my $extract_ok = $extractor->extract(to => $unarchived_fastq);
+        if ( not $rv or not -s $unarchived_fastq_path ) {
+            #my $error_message = $extractor->error;
+            #$self->error_message($error_message) if defined $error_message;
+            #$self->error_message($error_message) if defined $error_message;
+            $self->error_message('Failed to unarchive!');
+            return;
+        }
+        #my $unarchived_files = $extractor->files;
+        $self->debug_message("Unarchived fastq: $unarchived_fastq_path");
+        push @new_fastq_paths, $unarchived_fastq_path;
+        unlink $fastq_path;
+    }
+    $self->fastq_paths(\@new_fastq_paths);
+
+    $self->debug_message('Unarchive fastqs if necessary...');
     return 1;
 }
 
@@ -71,6 +111,7 @@ sub _fastqs_to_bam {
         sample_name => $self->sample_name,
         library_name => $self->library_name,
         read_group_name => $self->library_name,
+        use_version => '1.113',
     );
     if ( $fastqs[1] ) {
         $self->debug_message("Fastq 2: $fastqs[1]");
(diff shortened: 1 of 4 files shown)


lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/FastqsToBam.pm
```diff
@@ -57,6 +63,40 @@ sub execute {
     return 1;
 }
 
+sub _unarchive_fastqs_if_necessary {
+    my $self = shift;
+    $self->debug_message('Unarchive fastqs if necessary...');
+
+    my @new_fastq_paths;
+    for my $fastq_path ( $self->fastq_paths ) {
+        if ( $fastq_path !~ /\.gz$/ ) {
```
acoffman: Would it make more sense to shell out to unix `file` here? You'd get something back like `gzip compressed data` if it was in fact zipped. That seems more reliable than just checking the extension. It does seem like more overhead though. Your call!

nnutter: We could probably use [File::Type][1]; it's already packaged, `libfile-type-perl`.

[1]: https://metacpan.org/pod/File::Type

tmooney:
Here's the diff I'm seeing in the test failures.  (This one is from `Basic_archive.t`):
```
$ diff <(samtools view -h /tmp/temp9v3QJ/info/instrument_data/imported/55b723f87e54493cbdd63efa95dbd3a0/all_sequences.bam) <(samtools view -h /gscmnt/gc7007/info/genome/test_inputs/Genome-InstrumentData-Command-Import/fastq/v2/all_sequences.bam)
1c1
< @HD    VN:1.4    SO:queryname
---
> @HD    VN:1.0    SO:queryname
```

lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/FastqsToBam.pm
```diff
@@ -57,6 +63,40 @@ sub execute {
...
+            $self->debug_message('Unarchive not necessary for '.$fastq_path);
+            push @new_fastq_paths, $fastq_path;
+            next;
+        }
+        my $unarchived_fastq_path = $fastq_path;
+        $unarchived_fastq_path =~ s/\.gz$//;
+        $self->debug_message('Unarchiving: '.$fastq_path);
+        my $rv = eval{ Genome::Sys->shellcmd(cmd => "gunzip $fastq_path"); };
```
ebelter: I originally did use it, except it ignored the Archive::Extract::PREFER_BIN=1 and sucked a fastq into memory, killing the process. Hence shelling the shelling out. I forgot to remove it.

lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/FastqsToBam.pm
ebelter: @acoffman True, it is not the most reliable. If the file is gzipped and the extension is not .gz then gunzip will not work unless it is written to STDOUT. This just the simplest, and what Archive::Extract (I was originally going to use it) does.

@nnutter I am for that, but File::Type is not deployed for 5.10 on our system. We would need to hold this PR to do the testing and PR required to add this package. To note, the XTen imports are waiting on this PR.

nnutter: > To note, the XTen imports are waiting on this PR.

You don't need to hold the PR.

lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/FastqsToBam.pm
```diff
@@ -57,6 +61,42 @@ sub execute {
...
+            next;
+        }
+        $self->debug_message('Unarchiving: '.$fastq_path);
+        try {
+            Genome::Sys->shellcmd(cmd => "gunzip $fastq_path");
+        }
+        catch {
+            $self->error_message($_) if $_;
+            $self->error_message('Failed to gunzip fastq!');
+            return;
```
nnutter: This is just returning from the `catch` block.  For example,

```
$ perl -MTry::Tiny -e 'use strict; use warnings; my $s = sub { print qq(entering\n); try { die q(foo) } catch { return }; print qq(exiting\n) }; $s->()'
entering
exiting
```

lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/FastqsToBam.pm
```diff
@@ -57,6 +61,42 @@ sub execute {
...
+        };
+        my $unarchived_fastq_path = $fastq_path;
+        $unarchived_fastq_path =~ s/\.gz$//;
+        $self->debug_message("Unarchived fastq: $unarchived_fastq_path");
+        if ( not -s $unarchived_fastq_path ) {
+            $self->error_message('Unarchived fastq does not exist!');
+            return;
+        }
+        push @new_fastq_paths, $unarchived_fastq_path;
+        unlink $fastq_path;
```
nnutter: If we are going to delete things should we delete them as a commit observer?  Do we need to be deleting the original files?

lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/FastqsToBam.pm
```diff
@@ -83,11 +124,18 @@ sub _fastqs_to_bam {
...
-    my $execute_ok = eval{ $cmd->execute; };
-    if ( not $execute_ok or not -s $output_bam_path ) {
-        $self->error_message($@) if $@;
+    try {
+        $cmd->execute;
+    }
+    catch {
+        $self->error_message($_) if $_;
         $self->error_message('Failed to run picard fastq to sam!');
         return;
```
nnutter: Another return that I suspect isn't doing what it's believed to be doing.

lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/FastqsToBam.pm
ebelter: Should I just remove the try tiny and let the the shellcmd die?

lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/FastqsToBam.pm
ebelter: Yes. But, these are not the 'originals', they are /tmp copies. To save space, when a step finishes, it removes input files. That should be delegated to another place.

lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/FastqsToBam.pm
nnutter: That's more of a preference choice.

In both examples I've noticed you could do:

```
my $success = try {
    # same try block
} 
catch {
    # same catch block
}
return unless $success;
```

ebelter: Done: https://github.com/genome/genome/pull/430

PR #426, opened 2015-01-22
A few more fixes and a test for GMT::Htseq::Count::Result

tmooney:
Hot on the heels of #419 here are a few more fixes to the parallelization logic in this result.  Rather than hope it works, this time I added a test case to cover holes in the Htseq::Count test. :smile:


acoffman:
+1 :bank:

jweible:
+1

PR #427, opened 2015-01-22
add genome-test

nnutter:
CHANGELOG: Run tests now by running `genome-test`.  `genome-test` is
basically an alias to `genome-test-env test-tracker prove` that will
default to using the `--lsf` and `--git` options if no options are
specified.

### DIFF
--- a/bin/genome-test
+++ b/bin/genome-test
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+# delegate to local version if it is found
+if git rev-parse --is-inside-work-tree 1> /dev/null 2> /dev/null
+then
+    WORK_TREE="$(git rev-parse --show-toplevel)"
+    BIN="$(basename "$0")"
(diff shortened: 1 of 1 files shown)


tmooney:
Will `genome` need to be modified to delegate to `genome-test` as `genome test` as indicated in the changelog?

nnutter:
> Will `genome` need to be modified to delegate to `genome-test` as `genome test` as indicated in the changelog?

Derp, yeah, thanks.

nnutter:
Should I just keep it `genome-test` and modify the commit or actually make a subcommand `genome test`?

tmooney:
I would do the former.  It already matches all the other `genome-` commands, and going through `genome` itself adds a lot of complexity.

nnutter:
I've amended the commit message.

tmooney:
+1 :meat_on_bone:

brummett:
+1

PR #428, opened 2015-01-22
fix compile.t to not error on detached heads

nnutter:
Using '@{u}' assumes 1) that you're on a branch and 2) that that branch has an
upstream branch.  So if one was on a detached head (checking out a PR, etc.)
then this test would error and fail.

This changes compile.t to fallback on compile checking everything if it can't
guess a way to only check what has changed.


brummett:
+1
Now I won't have to ignore compile.t failing

tmooney:
+1 :ice_cream:

PR #429, opened 2015-01-22
Require a "Sponsor" and "Requestor" when getting or creating a SoftwareResult

tmooney:
When a SoftwareResult is found via `get_with_lock` or `get_or_create`, two specific users must be supplied: a `sponsor` and a `requestor`.  The `sponsor` is the entity accountable for resource usage for the result.  (There may be multiple sponsors for a result if it used by multiple projects.)  The `requestor` is the entity with processing underway that needed the result.

```perl
users => {
  sponsor => ,#either a Genome::Config::AnalysisProject or a Genome::Sys::User object
  requestor => ,#either a Genome::Model::Build or a Genome::Process object
  ... #additional label => user mappings as desired
}
```

This will create two `Genome::SoftwareResult::User` objects for the SoftwareResult, one with the label `sponsor` and one with either `shortcut` or `created` depending on if this was the requestor that originally created the result.  (Note that a requestor may end up as both if a subsequent step fetches a result that the same build earlier created.)

Additional users can be set at creation time by supplying additional labels and user objects in the hash.


lib/perl/Genome/Command/DelegatesToResult.pm
```diff
@@ -15,13 +15,21 @@ class Genome::Command::DelegatesToResult {
         },
     ],
     has_optional_input => [
+        sponsor => {
+            is => 'UR::Object',
+            doc => 'The responsible party for this result (defaults to the current user)',
+        },
+        requestor => {
```
nnutter: Why not make `requestor` required instead of manually implementing a similar check in `__errors__`?

lib/perl/Genome/Db/Ensembl/Command/Run/Base.pm
```diff
@@ -669,4 +692,16 @@ sub cache {
...
 
+sub _result_users {
+    my $self = shift;
+
+    my $analysis_build_anp = $self->analysis_build?
+        $self->analysis_build->model->analysis_projects : undef;
+
+    return {
+        requestor => ($self->analysis_build // $self->analysis_process),
+        sponsor => ($self->analysis_project // $analysis_build_anp // Genome::Sys->current_user),
```
nnutter: I've noticed several places using `//`?  Are you intentionally suggesting that zero values exist and are okay?

lib/perl/Genome/FeatureList.pm
```diff
@@ -421,11 +421,17 @@ sub generate_converted_bed_file {
     }
     my $original_md5 = Genome::Sys->md5sum($original_file_path);
 
+    my $result_users = delete($args{result_users});
+    $result_users->{'feature-list'} = $self;
+    $result_users->{requestor}    ||= $reference;
+    $result_users->{sponsor}      ||= Genome::Sys->current_user;
```
nnutter: Could having `sponsor` default to `current_user` result in incorrect resource tracking?

lib/perl/Genome/InstrumentData/AlignmentResult/WithIntermediateResults.pm
```diff
@@ -63,10 +76,20 @@ sub get_or_create_intermediate_result_for_params {
...
+
     my @iar_users = Genome::SoftwareResult::User->get(user => $self, label => intermediate_result_user_label);
     for my $iar_user ( @iar_users ) {
         my $iar = $iar_user->software_result;
         $iar_user->active(0);
+
+        if($result_users) {
+            for my $value (values %$result_users) {
+                my @using = $iar->users(user => $value, label => ['created', 'shortcut', 'sponsor']);
+                map $_->active(0), @using;
```
nnutter: Should the `active(0)` logic be in `delete`?

lib/perl/Genome/Model/Build/ReferenceSequence/Converter.pm
```diff
@@ -57,11 +57,20 @@ sub __errors__ {
     return @errors;
 }
 
+sub exists_for_references {
+    my $class = shift;
+    my $source_reference = shift;
+    my $destination_reference = shift;
+
+    my @ref = $class->_faster_get(source_reference_build => $source_reference, destination_reference_build => $destination_reference);
+    return @ref;
```
nnutter: Should this method return something besides the objects?

lib/perl/Genome/Model/Build/ReferenceSequence/Converter.pm
```diff
@@ -57,11 +57,20 @@ sub __errors__ {
     return @errors;
 }
 
+sub exists_for_references {
+    my $class = shift;
+    my $source_reference = shift;
+    my $destination_reference = shift;
+
+    my @ref = $class->_faster_get(source_reference_build => $source_reference, destination_reference_build => $destination_reference);
```
nnutter: Why `_faster_get`?

lib/perl/Genome/Model/Build/ReferenceSequence/Converter.pm
```diff
@@ -57,11 +57,20 @@ sub __errors__ {
...
+    return @ref;
+}
+
 sub convert_bed {
     my $class = shift;
     my ($source_bed, $source_reference, $destination_bed, $destination_reference) = @_;
 
-    my $self = $class->get_with_lock(source_reference_build => $source_reference, destination_reference_build => $destination_reference);
+    my $self = $class->_faster_get(source_reference_build => $source_reference, destination_reference_build => $destination_reference);
```
nnutter: Why `_faster_get`?

lib/perl/Genome/Db/Ensembl/Command/Run/Base.pm
tmooney: I make no claims as to the value; only that we accept it if it is defined.  I believe that `//` is more correct in these situations, as we're not performing a logical test of the values.  (You're right that a false-valued build or analysis project isn't something that is likely to exist.)

lib/perl/Genome/Model/PhenotypeCorrelation/Command/CaseControl/Unrelated.pm
```diff
@@ -297,7 +297,8 @@ sub _create_workflow {
                 work_dir => $vep_work_directory,
                 ensembl_annotation_build => $ensembl_annotation_build,
                 log_dir => $log_dir,
-
+                analysis_build => $self->build,
+                analysis_project => $self->build->model->analysis_projects,
```
nnutter: RHS can be a list.  Assigning to a scalar would trigger an exception if that's encountered.  As is it could add garbage to the inputs.

nnutter:
I'm taking a break from this about half-way through and I've really only skimmed the first half.

lib/perl/Genome/Model/PhenotypeCorrelation/Command/CaseControl/Unrelated.pm
tmooney: `->analysis_projects` is poorly named.  There is a unique constraint in the database such that a model can be linked to only one analysis project.

nnutter: `Genome::Model::analysis_projects` also labeled as [is_many](https://github.com/genome/genome/blob/master/lib/perl/Genome/Model.pm#L185).  As a separate PR should we deprecate the plural form and add a singular form, etc?

lib/perl/Genome/InstrumentData/AlignmentResult/WithIntermediateResults.pm
tmooney: `delete()` won't necessarily know which user relationships are safe to mark as inactive.  We're selectively marking specifically our own, but `delete()` will fail if any others remain.  This is an addition to the `active(0)` being set on line 84.

lib/perl/Genome/Model/PhenotypeCorrelation/Command/CaseControl/Unrelated.pm
tmooney: Sure!  Though maybe after this one is settled to avoid making it more complicated. :smile: 

(As an aside, if you press `y` after highlighting a line, Github will make an absolute reference to the line number at its last change revision so that the line-number link won't drift over time.)

lib/perl/Genome/Model/Build/ReferenceSequence/Converter.pm
tmooney: Cheating. Reference sequence converters are implemented as SRs but don't hold data and thus aren't useful to track with sponsors and requestors.  `_faster_get()` allows us to bypass the mandatory user relationships.  It's all contained within the class--calling `_faster_get` from the outside would be bad practice.

lib/perl/Genome/Model/Build/ReferenceSequence/Converter.pm
tmooney: I almost wrote
```perl
return @ref? 1 : 0;
```
or
```perl
return !!@ref;
```

That would stop someone from abusing the return value.  I could change it if that would be better.

lib/perl/Genome/Model/Build/ReferenceSequence/Converter.pm
tmooney: See [other comment](https://github.com/genome/genome/pull/429/files#r23571928).

lib/perl/Genome/Model/PhenotypeCorrelation/Command/CaseControl/Unrelated.pm
acoffman: I believe the `is_many` marking was a limitation of UR? I could be mistaken, but models have a bridge table to analysis projects (rather than a direct FK on the model table) and I thought at the time that UR could only handle such a relationship as `is_many` but I could be very wrong on that! If I'm wrong, I agree we should change it.

lib/perl/Genome/Command/DelegatesToResult.pm
tmooney: This is a conceit to the fact that the `requestor` is indirectly specified by the VariantReporting framework.  Those commands have an explicit process_id supplied which is then indirectly translated to the sponsor.  Ptero evaluates the presence (or not) of the input without first loading the Perl class, so it doesn't realize this relationship exists. Thus it bails out even though the value will actually be present when the Command is created.

lib/perl/Genome/FeatureList.pm
tmooney: These defaults are present so that the `genome feature-list dump-merged-list` command can still function without requiring the end user to say they're making a BED file for a particular build and AnP.

I don't think it's worse than setting the `requestor` to the reference sequence build, but I agree that both are not great.  Maybe we do force users to provide this info?

nnutter: Or maybe make dummy values that would allow these to be separated from cases where the sponsor is a person, etc.?

lib/perl/Genome/Model/Build/ReferenceSequence/Converter.pm
nnutter: For a separate PR, maybe we should rename `_faster_get()` to `lookup()`?  Doesn't seem like there's any reason why this method needs to be "private".

lib/perl/Genome/Model/Build/ReferenceSequence/Converter.pm
nnutter: I'm a fan of not returning an object for boolean sounding methods.

lib/perl/Genome/Model/Tools/DetectVariants2/Result/DetectionBase.pm
```diff
@@ -405,6 +406,7 @@ sub vcf_result_params {
         aligned_reads_sample => $aligned_reads_sample,
         ($control_aligned_reads_sample? (control_aligned_reads_sample => $control_aligned_reads_sample) : ()),
 
+        #users => $users, #Needed if get() inside get_vcf_result becomes get_with_lock()
```
nnutter: Will something meaningful alert a developer if this becomes relevant?  Comments don't seem like the right thing.

lib/perl/Genome/SoftwareResult.pm
```diff
@@ -110,10 +110,28 @@ sub _faster_get {
...
             # NOTE we do this so that get can be noisy when called directly.
             @objects = $class->SUPER::get(lookup_hash => $lookup_hash);
+            if (@objects > 1) {
+                die $class->error_message(
+                    'Multiple results found for lookup hash %s!',
+                    $lookup_hash
+                );
+            }
+
+            if (@objects) {
```
nnutter: I wonder if this should be implemented as an observer for any object retrieved from the database?

lib/perl/Genome/Command/DelegatesToResult.pm
nnutter: Maybe a comment is in order, I dunno.

lib/perl/Genome/SoftwareResult/User.pm
```diff
@@ -39,5 +43,86 @@ class Genome::SoftwareResult::User {
...
+
+sub _valid_types { return @{VALID_TYPES->{shift()}} }
+
+sub user_hash_for_build {
+    my $class = shift;
+    my $build = shift;
+
+    return {
+        requestor => $build,
+        sponsor   => $build->model->analysis_projects // Genome::Sys::User->get(username => $build->model->run_as),
```
nnutter: I don't know that one's more correct but why `$build->model->run_as` rather than `$build->run_by`?

lib/perl/Genome/SoftwareResult/User.pm
```diff
@@ -39,5 +43,86 @@ class Genome::SoftwareResult::User {
     doc => 'links a software result to other entities which depend on it',
 };
 
+sub with_registered_users {
+    my $class = shift;
+    my %params = Params::Validate::validate(
```
nnutter: I would personally vote for `allow_extra => 1` to be set here instead of for the whole module.

lib/perl/Genome/Model/Build/ReferenceSequence/Converter.pm
tmooney: Sounds good.  Updated!

lib/perl/Genome/Model/Tools/DetectVariants2/Result/DetectionBase.pm
tmooney: If someone tries to change it to `get_with_lock()` tests will break with an error from `Params::Validate`.  I could leave this comment out--it is a bit silly to add commented-out code.

lib/perl/Genome/SoftwareResult/User.pm
tmooney: I think `run_as` better captures the intent.  If, for example, someone runs a build of a model for debugging purposes, they're not a stakeholder in that result.

tmooney:
I have rebased in order to pull in the newly added `Htseq/Count/Result.t` so it could be updated.  (This test failed in Jenkins after the previous commit, since it was included in the merged build.)

lib/perl/Genome/Model/PhenotypeCorrelation/Command/CaseControl/Unrelated.pm
nnutter: genome/genome#437

lib/perl/Genome/Model/Tools/DetectVariants2/Result/DetectionBase.pm
nnutter: I'm not a fan of the comment but it's definitely good that this won't break silently.

lib/perl/Genome/SoftwareResult/User.pm
acoffman: I agree, I was under the impression from the documentation that it was only a module-specific setting. If that's not the case, I'll go ahead and make the change!

lib/perl/Genome/SoftwareResult.pm
acoffman: I'm not 100% sure I understand this suggestion, would you mind elaborating a bit?

lib/perl/Genome/SoftwareResult/User.pm
nnutter: Here is a link to the relevant [docs](https://metacpan.org/pod/Params::Validate#PER-INVOCATION-OPTIONS),

```perl
my %p = validate_with(
    params => \@_,
    spec   => {
        foo => { type    => SCALAR },
        bar => { default => 10 }
    },
    allow_extra => 1,
);
```

lib/perl/Genome/SoftwareResult.pm
nnutter: I am suggesting that we validate that `lookup_hash` is correct for any SR loaded from the database not just here in `_faster_get()`.  I think this could be done by creating an observer like,

```perl
Genome::SoftwareResult->add_observer(
    aspect => 'load',
    callback => \&Genome::SoftwareResult::validate_lookup_hash,
);
```

lib/perl/Genome/Model/Tools/DetectVariants2/Result/DetectionBase.pm
tmooney: Comment removed. Excelsior!

nnutter:
:pray:

lib/perl/Genome/SoftwareResult/User.pm
acoffman: This is fixed in the latest commit!

PR #434, opened 2015-01-27
Don't call execute() twice on the Command.

tmooney:


acoffman:
+1 :dog: 

This is another PR where adding `?w=1` to the end of the url makes the diff much easier.

ebelter:
:+1:

nnutter:
:+1:

PR #435, opened 2015-01-27
fix permissions on genome-test

nnutter:


acoffman:
+1

tmooney:
+1 :goat: Sorry we missed that before!

PR #436, opened 2015-01-27
Various Variant reporting IGV improvements

susannasiebert:
This PR addresses the following issues:
https://jira.gsc.wustl.edu/browse/AT-717
https://jira.gsc.wustl.edu/browse/AT-718
https://jira.gsc.wustl.edu/browse/AT-719

### DIFF
--- a/lib/perl/Genome/Model/Tools/Analysis/DumpIgvXmlBasic.pm
+++ b/lib/perl/Genome/Model/Tools/Analysis/DumpIgvXmlBasic.pm
@@ -1,0 +1,96 @@
+package Genome::Model::Tools::Analysis::DumpIgvXmlBasic;
+
+use strict;
+use warnings;
+
+use Genome;
+use Command;
+use IO::File;
+use Cwd qw(abs_path);
+use URI;
+use XML::Writer;
+
+class Genome::Model::Tools::Analysis::DumpIgvXmlBasic {
+    is => 'Command',
+    has => [
+        resource_files => {
+            type => 'HASH',
+            doc => 'A keyed list of files to display in the IGV session. Each key is a path to a file and the corresponding value is that label that should be displayed in the IGV session for that file.',
+        },
+        output_file => {
+            type => 'String',
+            is_optional => 0,
+            doc => 'Output XML file',
+        },
+        reference_name => {
+            type => 'String',
+            is_optional => 0,
+            doc => 'the name of the reference (in IGV) that the bams are aligned to. E.g. "b37" for build 37 or "reference"for our internal build36',
+        },
+    ]
+};
+
+sub help_brief {
+    "Makes an IGV session for review. Supports as many tracks as you want. Panels and track layout are not explicitly specified."
+}
+
+sub help_detail {
+    <<'HELP';
+This helps create IGV session files for manual reviewers
+HELP
+}
+
+sub execute {
+    my $self=shift;
+
+    my $output_fh = Genome::Sys->open_file_for_writing($self->output_file);
+    my $xml_writer = new XML::Writer(
+        OUTPUT => $output_fh,
+        NEWLINES => 1,
+        ENCODING => 'utf-8'
+    );
+
+    $xml_writer->xmlDecl('UTF-8', 'no');
+    $xml_writer->startTag(
+        'Session',
+        genome => $self->reference_name,
+        locus => '1:1-100',
+        version => '3',
+    );
+    $xml_writer->startTag('Resources');
+    while (my ($file, $label) = each %{$self->resource_files}) {
+        $xml_writer->emptyTag(
+            'Resource',
+            path => resolve_file_path($file),
+            relativePath => 'false',
+            name => $label,
+        );
+    }
+    $xml_writer->endTag('Resources');
+    $xml_writer->endTag('Session');
+    $xml_writer->end();
+    $output_fh->close();
+
+    return 1;
+}
+
+sub resolve_file_path {
+    my $file = shift;
+    my $uri = URI->new($file);
+    my $path;
+    if (defined($uri->scheme)) {
+        if ($uri->scheme eq 'file') {
+            $path = abs_path($uri->path);
+        }
+        else {
+            $path = $uri->as_string;
+        }
+    }
+    else {
+        $path = abs_path($file);
+    }
+    return $path;
+}
+
+
+1;
--- a/lib/perl/Genome/Model/Tools/Analysis/DumpIgvXmlBasic.pm
+++ b/lib/perl/Genome/Model/Tools/Analysis/DumpIgvXmlBasic.pm
@@ -26,7 +26,7 @@ class Genome::Model::Tools::Analysis::DumpIgvXmlBasic {
             type => 'String',
             is_optional => 0,
             doc => 'The name of the reference (in IGV) that the bams are aligned to.',
-            example_values => ['b37', 'reference']
+            example_values => ['b37', 'hg18', 'mm9']
         },
     ]
 };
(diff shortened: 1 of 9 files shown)


lib/perl/Genome/Model/Tools/Analysis/DumpIgvXmlBasic.pm
```diff
@@ -0,0 +1,96 @@
...
+        },
+        output_file => {
+            type => 'String',
+            is_optional => 0,
+            doc => 'Output XML file',
+        },
+        reference_name => {
+            type => 'String',
+            is_optional => 0,
+            doc => 'the name of the reference (in IGV) that the bams are aligned to. E.g. "b37" for build 37 or "reference"for our internal build36',
```
tmooney: There's a space missing in `"reference"for`.  (Also `build 37` vs. `build36` is inconsistent.)

These could also be listed as `example_values => ['b37', 'reference']`.

lib/perl/Genome/Model/Tools/Analysis/DumpIgvXmlBasic.pm
```diff
@@ -0,0 +1,96 @@
...
+    ]
+};
+
+sub help_brief {
+    "Makes an IGV session for review. Supports as many tracks as you want. Panels and track layout are not explicitly specified."
+}
+
+sub help_detail {
+    <<'HELP';
+This helps create IGV session files for manual reviewers
```
tmooney: Should the brief and the detail be switched? It's unusual for the "brief" version to be longer than the "detail" version.

lib/perl/Genome/Model/Tools/Analysis/DumpIgvXmlBasic.pm
```diff
@@ -0,0 +1,96 @@
...
+        OUTPUT => $output_fh,
+        NEWLINES => 1,
+        ENCODING => 'utf-8'
+    );
+
+    $xml_writer->xmlDecl('UTF-8', 'no');
+    $xml_writer->startTag(
+        'Session',
+        genome => $self->reference_name,
+        locus => '1:1-100',
```
tmooney: Does this tool have limitations on what references it accepts?  (What happens if the chosen reference doesn't have a chromosome `1` but instead, say, `chr1`?)

lib/perl/Genome/VariantReporting/Framework/Component/Report/MergedReport.pm
```diff
@@ -540,5 +541,25 @@ sub get_entry_source {
...
 
+sub category {
+    my $self = shift;
+
+    my @report_users = map { $_->users('label like' => 'report:%') } $self->report_results;
+    my $category;
+    for my $user (@report_users) {
+        if ($user->label =~ /report:(.*)/) {
+            my $metadata_json = $1;
+            my $m = from_json($metadata_json);
```
tmooney: I'm late to the game, but it seems like JSON would be better stored elsewhere (a SoftwareResult::Metric perhaps?) than in a user label.

davidlmorton: That is something we should discuss in more detail, I agree (on the face of it) that it seems better to have that information live in a Metric instead of a User label.  

Perhaps the `DelegatesToResult` class could take an argument that would create the metrics in almost the same manner as it takes a `user` and a `label` argument now.

We'll have to think about migration too if we decide to switch it.

lib/perl/Genome/Model/Tools/Analysis/DumpIgvXmlBasic.pm
susannasiebert: That is a very good question and I don't know the answer off-hand. For what it's worth, this tools is based on the other DumpIgvXml tools which also use 1:1-100.

susannasiebert:
@tmooney I pushed a commit to address your comments about the documentation.

lib/perl/Genome/Model/Tools/Analysis/DumpIgvXmlBasic.pm
susannasiebert: I did some trial runs on an example igv session xml using variations of b37/h18 and 1:1-100/ch1:1-100 and it seems to translate between the two without problems.

lib/perl/Genome/VariantReporting/Framework/Component/Report/MergedReport.pm
susannasiebert: I think we can agree that this is outside of the scope of this PR.

susannasiebert:
I pushed another commit to clean up the igv reference names since there were some inconsistencies/errors.

lib/perl/Genome/Model/Tools/Analysis/DumpIgvXmlBasic.pm
```diff
@@ -0,0 +1,97 @@
...
+
+use Genome;
+use Command;
+use IO::File;
+use Cwd qw(abs_path);
+use URI;
+use XML::Writer;
+
+class Genome::Model::Tools::Analysis::DumpIgvXmlBasic {
+    is => 'Command',
```
davidlmorton: Should this be Command::V2?

lib/perl/Genome/Model/Tools/Analysis/DumpIgvXmlBasic.pm
```diff
@@ -0,0 +1,97 @@
...
+    has => [
+        resource_files => {
+            type => 'HASH',
```
davidlmorton: Does Command / Command::V2 support HASH type arguments to be supplied via the command line?

davidlmorton:
+1 with Command -> Command:V2 change or some reason why it isn't changed. :)

lib/perl/Genome/Model/Tools/Analysis/DumpIgvXmlBasic.pm
susannasiebert: I honestly don't know. I think these igv tools are primarily called inside of Perl code, not on the command line. @tmooney do you know?

What's the alternative here if we can't use HASH?

lib/perl/Genome/Model/Tools/Analysis/DumpIgvXmlBasic.pm
susannasiebert: My bad. This is what happens when you just copy things from existing tools.

lib/perl/Genome/VariantReporting/Framework/Component/Report/MergedReport.pm
tmooney: It definitely is.  Should we make a JIRA issue about it?

lib/perl/Genome/Model/Tools/Analysis/DumpIgvXmlBasic.pm
susannasiebert: This is now a Command:V2

lib/perl/Genome/Model/Tools/Analysis/DumpIgvXmlBasic.pm
susannasiebert: This has been addressed by setting _is_hidden_in_docs to 1 which results in this command not showing up in the docs/autocomplete of the command line. People can still try to execute it on the command line but they would need to know that it exists.

lib/perl/Genome/VariantReporting/Framework/Component/Report/MergedReport.pm
susannasiebert: Yes. I made one here https://jira.gsc.wustl.edu/browse/AT-731

tmooney:
+1 assuming tests pass! :dolphin:

PR #437, opened 2015-01-28
make analysis_projects singular on models

nnutter:
Models can only belong to one `AnalysisProject` so the `analysis_projects` property
should be singular.  When originally implemented this was believe to not be
supported by UR.  This adds `analysis_project` and deprecates `analysis_projects`.

### DIFF
--- a/lib/perl/Genome/Config/AnalysisProject/Command/AddModel.t
+++ b/lib/perl/Genome/Config/AnalysisProject/Command/AddModel.t
@@ -29,7 +29,7 @@ my $cmd2 = $class->create(
 isa_ok($cmd2, $class, 'created command');
 ok($cmd2->execute, 'command succeeds when pending analysis project');
 for my $m (@models) {
-    is($m->analysis_projects, $analysis_project, 'analysis project assigned');
+    is($m->analysis_project, $analysis_project, 'analysis project assigned');
 }
 
 my $cmd3 = $class->create(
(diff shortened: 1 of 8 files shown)


chrisamiller:
Just a note to please send out an announcement about this change after merging. We use that accessor a lot, and I'll have some confused analysts if it changes without notice.  Thanks.

nnutter:
@chrisamiller, this should not affect the use of the old accessor.  However, a CHANGELOG will be added at merge to announce the deprecation of the old accessor.

ebelter:
+1

nnutter:
@davidlmorton, I changed `genome model build view`.  Do you want to take a look?

@tmooney and @acoffman, do you want to review?

acoffman:
+1 :baby_chick: 

Looks good to me!

tmooney:
+1 :cherries:

This change could also simplify [this loop](https://github.com/genome/genome/blob/01e8b6b840722f191f3f8171a368821272da059d/lib/perl/Genome/Config/Command/ConfigureQueuedInstrumentData.pm#L114) and change [this call](https://github.com/genome/genome/blob/01e8b6b840722f191f3f8171a368821272da059d/lib/perl/Genome/Config/Command/ConfigureQueuedInstrumentData.pm#L239) in CQID.  It also enables [this query](https://github.com/genome/genome/blob/01e8b6b840722f191f3f8171a368821272da059d/lib/perl/Genome/Config/AnalysisProject/Command/View.pm#L309) and the calls in [this test](https://github.com/genome/genome/blob/01e8b6b840722f191f3f8171a368821272da059d/lib/perl/Genome/Config/AnalysisProject/Command/AddModel.t) to be changed.

We could also make a follow-up PR for those.

nnutter:
Thanks @tmooney, I've implented the changes you've suggested.

PR #438, opened 2015-01-28
add per-model transactions in "Model Summary"

nnutter:
Our "Model Summary" command has been failing. genome/UR#56 addresses a severe
performance problem that occurs when there is "invalid data for save".  This PR
is trying to minimize the collateral damage that occurs under the same
scenario.


tmooney:
+1 :cool:

brummett:
+1

PR #439, opened 2015-01-29
Start of a System for Notifying

tmooney:
Whereas:
* We need a way to tell AnalysisProject stakeholders when their project is misconfigured without sending them repeat notices every five minutes when CQID runs.
* Some users don't care about "Build Initialized" or other mail the system sends.

Therefore:
* This PR offers the start of a standard way to notify someone of something.

This is incomplete as of yet, but I thought it'd be better to solicit feedback early before getting too far along (unlike certain other recent PRs I opened :smile:).

Future work:
* A possible way to subscribe to notifications about a subject that one wouldn't normally get.
* Sqitch migrations to back some of these new classes.

### DIFF
--- a/lib/perl/Genome/Notify.pm
+++ b/lib/perl/Genome/Notify.pm
@@ -1,0 +1,35 @@
+package Genome::Notify;
+
+use strict;
+use warnings;
+
+use Params::Validate qw();
+
+use Genome;
+
+class Genome::Notify {
+    is => 'UR::Singleton',
+};
+
+sub notify {
+    my $class = shift;
+    my %options = Params::Validate::validate(@_, {
+        subject => { isa => 'UR::Object' },
+        type    => { isa => 'Genome::Notify::NoticeType' },
+        header  => { type => Params::Validate::SCALAR },
+        body    => { type => Params::Validate::SCALAR },
+        target  => { isa => 'Genome::Sys::User' },
+    });
+
+    my $notice = Genome::Notify::Notice->issue(%options);
+    return unless $notice;
+
+    my $notifier = Genome::Notify::Preference->notifier_for_target(
+        type => $options{type},
+        target => $options{target},
+    );
+
+    $notifier->notify($notice);
+}
+
+1;
--- a/lib/perl/Genome/Notify.t
+++ b/lib/perl/Genome/Notify.t
@@ -1,0 +1,99 @@
+#!/usr/bin/env genome-perl
+
+use strict;
+use warnings;
+
+BEGIN {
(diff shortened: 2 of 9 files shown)


nnutter:
My current interpretation of the classes:

- `Genome::Notify` is an API to request notifications.
- `Genome::Notify::Notice` is a notification to be delivered.
- `Genome::Notify::NoticeType` is a way of categorizing notifications.
- `Genome::Notify::Preference` let's a user set a preferred notification method.

General Comments:

- The idea of having more than one receiver of log messages is like a notification scheme.  For example, `Genome::Logger` is setup so that that `info` and above messages go to the screen while `debug` and above go to `syslog`.  It make me wonder if notifications and logging are actually separate or not.
- Ensuring sending of notifications sounds like the work of a background work queue.
- Would notification replace notes?
- Notification could be delivered by email, chat bot, web browser, or RSS feed.  The API should not assume one in particular.  This might mean, for example, that the `body` of a notification should not depend on the `header` for context.
- Notifications seems potentially very complex.  For example, issue trackers (JIRA) and monitoring systems barely cope with the notification schemes people want.
- Notifications do not seem trivial.  For example, they do not seem more trivial than locking or test databases which both turned out more complicated than originally thought.
- Notifications are quite possibly relevant outside Genome.  If so, it should be designed as a standalone component.
- We may want the same consistency requirements for notifications that I think we want for a background work queue which to me has meant that submission should be done transactionally with the rest of the database changes.

tmooney:
That interpretation of the classes sounds right!

General Responses:

* The difference between logs and notices is that logs are internal notifications about the operation of the system, and that notices are external notifications about something in the world.  Perhaps that's not a useful (or even accurate) distinction, though.
* I agree, before this becomes used it belongs in a queue.  The main work thread doesn't care what is involved in notifying.
* I'm not sure how this would be used to replace notes.
* Supporting anything people might want does sound complex.  My not-very-secret mission is to provide a way to send e-mails to people when their AnPs are messed up without repeating that mail every five minutes.  Right now we do that by writing one by hand when we notice something is amiss.  A good question might be, "How far towards a general solution do we need to go to solve that problem?".  I thought a step away from a very specific implementation was good, but I doubt it's worth embarking on a yearlong project in order to solve that pain.  (The five-minute solution might be to send an e-mail to the user and then automatically put their project on Hold so it doesn't fire again.  Maybe we do that and forget this?)
* I agree that a  "complete" solution to Notifications would be better placed outside of Genome.

tmooney:
I'm going to close this until I can come back to it.

PR #440, opened 2015-01-29
Add option to copy tags when copying configuration.

tmooney:


nnutter:
It may be unwarranted or beyond the scope of this PR but the thing that immediately jumped out to me was that the object could be "taught" to copy itself like was recently done elsewhere, e.g. [UR::Object][1] and [Genome::Subject][2].

[1]: https://github.com/genome/UR/blob/62498ec2367de1d9a348b81a5d1965ab04597ad9/lib/UR/Object.pm#L31
[2]: https://github.com/genome/genome/blob/a3a1e68cacb3937e57ce443025c493c0eb563bd6/lib/perl/Genome/Subject.pm#L173

lib/perl/Genome/Config/AnalysisProject/Command/CopyConfig.pm
```diff
@@ -55,19 +62,31 @@ sub _copy_config_profile_items_to_project {
     my @config_profile_items = @_;
 
     for my $original_config_item (@config_profile_items) {
+        my $new_item;
         if ($original_config_item->analysis_menu_item) {
-            Genome::Config::Profile::Item->create(
+            $new_item = Genome::Config::Profile::Item->create(
                 analysis_project => $analysis_project,
                 analysis_menu_item => $original_config_item->analysis_menu_item,
                 status => $original_config_item->status,
```
nnutter: I'm curious if you could get an array of tags and just pass `tags => \@tags` into the create calls?  Seems like `Genome::Config::Profile::Item` [tags][1] is setup for that.

[1]: https://github.com/tmooney/genome/blob/7cfc7a28d99558f0c2457562effc7283317e3965/lib/perl/Genome/Config/Profile/Item.pm#L54

lib/perl/Genome/Config/AnalysisProject/Command/CopyConfig.t
```diff
@@ -51,6 +54,7 @@ my $target_project = Genome::Config::AnalysisProject->create(
 my $cmd = $class->create(
     to_project => $target_project,
     from_project => $source_project,
+    tags => 1,
```
nnutter: Seems like maybe it would be overkill but you might test that tags aren't copied in the opposite case and/or that the command works okay if no tags exist.

lib/perl/Genome/Config/AnalysisProject/Command/CopyConfig.pm
tmooney: That does sound better.  I'll change it!

lib/perl/Genome/Config/AnalysisProject/Command/CopyConfig.t
tmooney: I'd skipped it, since there's no code that only runs in the "no tags" case, but it is worthwhile to see that the flag is respected.  I'll add some more tests!

nnutter: I dunno if adding tests is beneficial or not; that's what I meant by "overkill".  I agree they don't seem necessary now.  I'm worried about future changes/growth but YAGNI... maybe.

tmooney: I reworked the test so the flags are parametrized.  That way it wasn't a bunch of extra coding to try a bunch of combinations.  It still runs quickly, so I'm inclined to err on the side of comprehensiveness.

tmooney:
Updated based on feedback!

nnutter:
:+1:

acoffman:
+1 :shipit:

PR #441, opened 2015-01-30
genome-env rewrite/refactor

nnutter:
This pulls in code that was in the apipe-ci repo, adds more control to the
constructed env through the use of options, and adds support for using a test
database.

### DIFF
--- a/bin/genome-env-next
+++ b/bin/genome-env-next
@@ -92,9 +92,10 @@ function main {
     log.info ''
     log.info "Genome Perl: $(genome-perl -e 'printf(qq(%vd\n), $^V)')"
     log.info "Genome Prove: $(genome-prove --version)"
-    log.info "UR: $(wtf UR | head -n 1)"
-    log.info "Workflow: $(wtf Workflow | head -n 1)"
-    log.info "Genome: $(wtf Genome | head -n 1)"
+    for MODULE in UR Workflow Genome
+    do
+        log.info "${MODULE}: $(genome-perl -M$MODULE -e "print \$INC{q(${MODULE}.pm)}, qq(\n)")"
+    done
     log.info "Test DB Snapshot: $SNAPSHOT_NAME"
     log.info "Test DB Name: $TESTDBSERVER_DB_NAME"
     log.info ''
@@ -392,30 +393,42 @@ function setup_env {
         SNAPSHOT_LIB="/gsc/scripts/opt/genome/current/user/lib/perl" \
         SNAPSHOT_BIN="/gsc/scripts/opt/genome/current/user/bin"
 
-    # remove /gsc/scripts/opt/genome/current/user/*
-    PERL5LIB="$(echo $PERL5LIB | tr : "\n" | grep -v "$SNAPSHOT_LIB" | tr '\n' : | sed 's/:$//')"
+    # remove deployed bin directory from PATH
     PATH="$(echo $PATH | tr : "\n" | grep -v "$SNAPSHOT_BIN" | tr '\n' : | sed 's/:$//')"
+    PATH=$UR_REPO/bin:$PATH
+    PATH=$WF_REPO/bin:$PATH
+    PATH=$BASE_DIR/bin:$PATH
+    export PATH
+
+    # remove deployed lib/perl directory from PATH
+    PERL5LIB="$(echo $PERL5LIB | tr : "\n" | grep -v "$SNAPSHOT_LIB" | tr '\n' : | sed 's/:$//')"
+
+    for MODULE in UR Workflow Genome
+    do
+        if genome-perl -M$MODULE -e 1 2> /dev/null
+        then
+            log.error "still found $MODULE after removing deployed snapshot"
+            exit 1
+        fi
+    done
 
     PERL5LIB=$UR_REPO/lib:$PERL5LIB
-    PATH=$UR_REPO/bin:$PATH
-
     PERL5LIB=$BASE_DIR/configuration-manager/lib:$PERL5LIB
+    PERL5LIB=$WF_REPO/lib:$PERL5LIB
+    PERL5LIB=$BASE_DIR/lib/perl:$PERL5LIB
+    export PERL5LIB
 
     GENOME_CONFIG_PATH=$BASE_DIR/gms-confg/etc/gms
 
-    PERL5LIB=$WF_REPO/lib:$PERL5LIB
-    PATH=$WF_REPO/bin:$PATH
-
-    PERL5LIB=$BASE_DIR/lib/perl:$PERL5LIB
-    PATH=$BASE_DIR/bin:$PATH
-
-    export PERL5LIB
-    export PATH
-
     set +o nounset
 
-    for MODULE in UR Workflow Genome; do
-        if wtf $MODULE | grep -q "$SNAPSHOT_LIB"; then log.info "$MODULE found in $SNAPSHOT_LIB! Aborting!" && exit 1; fi
+    for MODULE in UR Workflow Genome
+    do
+        if ! genome-perl -M$MODULE -e 1 2> /dev/null
+        then
+            log.error "failed to load $MODULE"
+            exit 1
+        fi
     done
 
     for BIN in ur workflow genome; do
--- a/bin/genome-env-next
+++ b/bin/genome-env-next
@@ -32,9 +32,9 @@ function help {
 
     -M    Disable migration of database.
 
-    -S    Disable configuration of database.
+    -D    Disable configuration of database.
 
-    -s db_snapshot_name
+    -d db_snapshot_name
           The name of the database snapshot to use instead of the default.
 
     -U    Disable configuration of UR.
@@ -57,7 +57,7 @@ function main {
         TESTDBSERVER_DB_PASS="mypassword"
 
     local DISABLE_DB DISABLE_MIGRATIONS DISABLE_UR DISABLE_WF
-    local SNAPSHOT_NAME TOP UR_REPO VERBOSE WF_REPO
+    local DB_SNAPSHOT_NAME TOP UR_REPO VERBOSE WF_REPO
 
     if $GIT rev-parse --is-inside-work-tree 1> /dev/null 2> /dev/null
     then
@@ -84,7 +84,7 @@ function main {
     do
         log.info "${MODULE}: $(genome-perl -M$MODULE -e "print \$INC{q(${MODULE}.pm)}, qq(\n)")"
     done
-    log.info "Test DB Snapshot: $SNAPSHOT_NAME"
+    log.info "Test DB Snapshot: $DB_SNAPSHOT_NAME"
     log.info "Test DB Name: $TESTDBSERVER_DB_NAME"
     log.info ''
 
@@ -109,7 +109,7 @@ function validate_bash {
 }
 
 function parse_opts {
-    DISABLE_DB="" DISABLE_MIGRATIONS="" DISABLE_UR="" DISABLE_WF="" SNAPSHOT_NAME="" UR_REPO="" WF_REPO=""
+    DISABLE_DB="" DISABLE_MIGRATIONS="" DISABLE_UR="" DISABLE_WF="" DB_SNAPSHOT_NAME="" UR_REPO="" WF_REPO=""
 
     local opts
     while getopts :hMSs:Uu:vWw: opts "$@"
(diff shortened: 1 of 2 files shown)


nnutter:
Eventually this will replace the existing `genome-env` and I'll remove the old sources from [genome/wugc-hudson](https://github.com/genome/wugc-hudson).

nnutter:
Fixed accidental commit of `bin/genome-env` changes.

For anyone reviewing, if the size or language bother you at least review the "interface", i.e. the `help` function.

bin/genome-env-next
```diff
@@ -0,0 +1,432 @@
+#!/bin/bash
+
+# if this was found via PATH as opposed to a path to executable
+THIS="$(basename "$0")"
```
tabbott: or, since we're explicitly using bash, ${0##*/} is ~~ the result of s!.*/!! on $0 (greedily eat everything until the last slash)
no need to change it, just sharing since you asked :)

bin/genome-env-next
```diff
@@ -0,0 +1,432 @@
...
+if test "$THIS" == "$0"
```
tabbott: If the script is found via $PATH, won't $0 be an absolute path? I'm reading this as a test to see if we were called with a path or not. I'm not sure if that's what will actually happen though (please correct me if I am wrong!). If you ran the command from the dir it was in as "bash genome-env-next" then you would get $0 == "genome-env-next". I'm not sure how else you'd get that though.

bin/genome-env-next
```diff
@@ -0,0 +1,432 @@
...
+then
+    # delegate to local version if it is found
+    if git rev-parse --is-inside-work-tree 1> /dev/null 2> /dev/null
```
tabbott: I spell this redirection as ... > /dev/null 2>&1. No problem though.

bin/genome-env-next
```diff
@@ -0,0 +1,432 @@
...
+    then
+        WORK_TREE="$(git rev-parse --show-toplevel)"
+        NEW_THIS="${WORK_TREE}/bin/${THIS}"
+        if test -x "$NEW_THIS"
+        then
+            echo "delegating to $NEW_THIS" 1>&2
+            exec "$NEW_THIS" "$@"
```
tabbott: After an embarrassing incident at a previous company, I always protect this type of pattern with a variable that can be tested by the re-execed process to make sure we don't keep re-execing forever.

bin/genome-env-next
```diff
@@ -0,0 +1,432 @@
...
+}
+
+function __quiet_git {
+    (
+        set +o errexit
+        OUT="$(git "$@" 2>&1)"
+        GIT_EXIT=$?
+        if test $GIT_EXIT -ne 0
+        then
+            echo -e "$OUT" 1>&2
```
tabbott: Why is __quiet_git making so much noise? Is the idea to keep it quiet on stdout?

bin/genome-env-next
```diff
@@ -0,0 +1,432 @@
...
+
+function main {
+    local TESTDBSERVER_DB_NAME TESTDBSERVER_DB_HOST TESTDBSERVER_DB_PORT \
+        TESTDBSERVER_DB_USER="genome" \
+        TESTDBSERVER_DB_PASS="mypassword"
+
+    local DISABLE_DB DISABLE_MIGRATIONS DISABLE_UR DISABLE_WF
+    local SNAPSHOT_NAME TOP UR_REPO VERBOSE WF_REPO
+
+    if git rev-parse --is-inside-work-tree 1> /dev/null 2> /dev/null
```
tabbott: Seems like this (and a few other git commands) test to learn about the environment we are executing in are being performed multiple times in multiple places throughout the code. I wonder if it would make more sense to pull these up closer to the top of the script and to do each test only once. If nothing else, it would make it easier to document each thing for git novices while also providing the opportunity to provide more descriptive names (e.g., $GIT_REPO_ROOT may be easier to parse than git rev-parse --show-toplevel for some).

bin/genome-env-next
```diff
@@ -0,0 +1,432 @@
...
+    shift $((OPTIND-1))
+
+    rebuild_meta_db
+    setup_test_db "$TOP"
+    setup_env "$TOP"
+
+    stderr ''
+    stderr "Genome Perl: $(genome-perl -e 'printf(qq(%vd\n), $^V)')"
+    stderr "Genome Prove: $(genome-prove --version)"
+    stderr "UR: $(wtf UR | head -n 1)"
```
tabbott: wtf (while possibly the best named program ever) is a wash u thing. it's also hard-coded to use /gsc/bin/perl, which may or may not make you happy. I wonder if something like a bash function along the lines of

genome-perl -M$module -e "print \$INC{'${module}.pm'}"

(perhaps with some other value of genome-perl) would be more appropriate.

bin/genome-env-next
```diff
@@ -0,0 +1,432 @@
...
+    stderr ''
+
+    if [ $# -gt 0 ]; then
+        echo "$@"... 1>&2
+        exec "$@"
+    else
+        if [ -n "$BASH" ]; then
+            echo 'Press <Ctrl>-D or type `logout` to return to normal environment.'
+            echo ''
+            if test -z "$GE_ENV_NAME"
```
tabbott: Curious why sometimes it's

if [ -[nz] ]

and other times:

if test -[nz]

is there a reason beyond "mood at the time it was written?"

bin/genome-env-next
```diff
@@ -0,0 +1,432 @@
...
+
+function set_defaults {
+    local TOP="$(git rev-parse --show-toplevel 2> /dev/null)"
+    local SUBMODULES
+
+    SUBMODULES=(gms-config configuration-manager)
+
+    if test -z "$SNAPSHOT_NAME"
+    then
+        SNAPSHOT_NAME="82f3e4d"
```
tabbott: Is this a commit hash prefix? Should this be stored as data somewhere else?

bin/genome-env-next
```diff
@@ -0,0 +1,432 @@
...
+        do
+            regen_db "$DB_TXT"
+        done
+    else
+        find "$PWD" -iname '*sqlite3-dump' -or -iname '*sqlite3n-dump' | while read DB_TXT; do
+            regen_db "$DB_TXT"
+        done
+    fi
+
+    echo ' done' 1>&2
```
tabbott: I'll comment on this only here, but I think a lot of the locations that match a search for >&2 could use the stderr function you defined instead.

bin/genome-env-next
nnutter: I used to do `if [ ... ]; then` but noticed the [git][] bash scripts, e.g. [git-rebase.sh][], do,

```
if test ...
then
    ...
fi
```

I don't know any reason to do one or the other except that I find the bash man page frustrating to search.  So I switch to `if test`.  I've also become more of a fan of keeping "statements" on separate lines so I don't do the `; then` anymore.

[git]: https://github.com/git/git
[git-rebase.sh]: https://github.com/git/git/blob/master/git-rebase.sh

bin/genome-env-next
nnutter: Good point.

bin/genome-env-next
nnutter: I'll take a pass through and look.  I agree with the sentiment.

bin/genome-env-next
nnutter: I already screwed that up once so sounds like a good idea.

bin/genome-env-next
nnutter: You're right, this isn't doing what I thought it was.  I'll have to re-think this.

I'm trying to make sure people use the "right" version but I'm bothered by this magic/redirection.

bin/genome-env-next
nnutter: Looks like we need an `*` in there.  I expect that the `basename` version might be more obvious to non-bash authors but it seems like avoiding another process might make it worth it.

bin/genome-env-next
nnutter: It's a template name used to create the ephemeral database that happens to be names after a commit hash.  It's definitely data.  I haven't decided where to store it yet.  It may need to be tightly coupled to the source revision.

bin/genome-env-next
tabbott: Indeed, who ate my star?

bin/genome-env-next
nnutter: Yeah, trying to make pretty output and `git` tends to print stuff on either stdout or stderr.

bin/genome-env-next
nnutter: I'll make a pass through to fix those.

bin/genome-env-next
nnutter: The infinite re-exec protection suggested below works to do what I wanted to do here as well, i.e. have an ability to disable the re-exec.

nnutter:
I've pushed new code.  Most of the prior feedback has been addressed.  I think the only feedback that hasn't really been addressed was the feedback about pulling "git awareness" up more.  I don't think there is much room to improve that as what remains mostly operates on multiple Git repos, i.e. `prepare_submodules`.

bin/genome-env-next
```diff
@@ -0,0 +1,452 @@
...
+    export GENOME_DS_GMSCHEMA_LOGIN=$TESTDBSERVER_DB_USER
+    export GENOME_DS_GMSCHEMA_AUTH=$TESTDBSERVER_DB_PASS
+
+    hash -r
+}
+
+THIS="$(basename "$0")"
+
+GIT="$(type -p git)" || fatal $DEP_ERR "git is required"
+SQITCH="$(type -p sqitch)" || fatal $DEP_ERR "sqitch is required"
```
tabbott: We should add sqitch to genome-snapshot-deps. I had to install it manually before the script would run.

bin/genome-env-next
```diff
@@ -0,0 +1,452 @@
...
+        NEW_THIS="${WORK_TREE}/bin/${THIS}"
+        if test -x "$NEW_THIS"
+        then
+            export STOP_GENOME_ENV_REDIRECT=1
+            exec "$NEW_THIS" "$@"
+        fi
+    fi
+fi
+
+set -o nounset
```
tabbott: Why doesn't this live with the other options at the top of the file? Also, there is at least one unset variable by default (line 98: GE_ENV_NAME: unbound variable)

bin/genome-env-next
```diff
@@ -0,0 +1,452 @@
...
+    if test "$#" -gt 0
+    then
+        log.info "$@"...
+        exec "$@"
+    else
+        if test -n "$BASH"
+        then
+            echo 'Press <Ctrl>-D or type `logout` to return to normal environment.'
+            echo ''
+            if test -z "$GE_ENV_NAME"
```
tabbott: set -o nounset makes this noisy.

bin/genome-env-next
```diff
@@ -0,0 +1,452 @@
...
+            then
+                GE_ENV_NAME="genome-env"
+            fi
+            unset PROMPT_COMMAND
+            export PS1="(${GE_ENV_NAME}) \w $ "
+            exec bash --norc --noprofile
```
tabbott: A lot of people probably have /gsc/bin/bash as the first bash in their path, but /bin/bash as their login shell. You tested $BASH, why not use that?

bin/genome-env-next
nnutter: Good point, I'll get it deployed; however, I think I'll do it via Puppet not `genome-snapshot-deps` since it's a development dependency only.

bin/genome-env-next
nnutter: Fixed.

bin/genome-env-next
nnutter: Fixed.

bin/genome-env-next
nnutter: I moved everything to the bottom of the file so I could use `fatal` (L425); (basically) everything above is inside functions.  I don't know of any way to check if a variable is unbound so I enable it after I've checked `GE_ENV_NAME` and `STOP_GENOME_ENV_REDIRECT`.

nnutter: FYI, I just renamed `STOP_GENOME_ENV_REDIRECT` to `GE_NO_REDIRECT`.

nnutter:
Pushed a bunch of new commits (starting at `839a4a9`).

nnutter:
I'm going to merge this soon; not because I think it's perfect but because I think it's good for a first iteration.  In the upcoming sprint I plan to add tests to help verify everything works correctly now and in the future.  Any remaining feedback would still be appreciated here or by email or in person.

PR #444, opened 2015-02-03
Add Common Reference Combination Cases in FeatureList Merge

tmooney:
Generally solving the problem of deciding what references are combinable would probably be better handled with a graph database of some sort, but this adds a few of the cases we're more likely to encounter (including one for which there is an open ticket presently).

Also in this PR... adding the same sponsor and requestor as discussed in [this thread](https://github.com/genome/genome/pull/429/files#r23570637).


acoffman:
+1

apregier:
+1

iferguson90:
+1 :boar:

PR #445, opened 2015-02-03
Correct Custom BSMap Resource Requirements

tmooney:
* Pass along the params so we get the correct thread count.
* Specify a job group for each aligner so we can separately throttle them to avoid monopolizing the queues.  (This only applies to ReferenceAlignment for now.)

### DIFF
--- a/lib/perl/Genome/InstrumentData/AlignmentResult/Bsmap.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult/Bsmap.pm
@@ -28,9 +28,10 @@ sub required_arch_os { 'x86_64' }
 # LSF resources required to run the alignment.
 sub required_rusage {
     my $self = shift;
+    my %params = @_;
     my $cores = 4; # default to 4, although decomposed_aligner_params should always include "-p N" even if the processing-profile does not
 
-    my $params = $self->decomposed_aligner_params();
+    my $params = $self->decomposed_aligner_params($params{aligner_params});
     if ($params =~ /-p\s+(\d+)/) {
         $cores = $1;
     }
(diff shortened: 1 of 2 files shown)


susannasiebert:
+1

iferguson90:
These changes seem reasonable. +1 assuming thread counts now work as intended.

lib/perl/Genome/Model/Event/Build/ReferenceAlignment/AlignReads.pm
```diff
@@ -47,7 +48,17 @@ sub bsub_rusage {
...
+    my $group = $self->_resolve_job_group;
+
+    return join(' ', $rusage, $group);
+}
+
+sub _resolve_job_group {
+    my $self = shift;
+
+    my $group_name = Genome::Utility::Text::sanitize_string_for_filesystem($self->results_class->__meta__->property(property_name => 'aligner_name')->default_value);
+    return sprintf('-g /align/%s', $group_name);
```
nnutter: Verified that `/align` is owned by `apipe-builder`,

```
$ bjgroup -s /align | awk '{print $10}'
OWNER
apipe-builder
```

lib/perl/Genome/InstrumentData/AlignmentResult/Bsmap.pm
```diff
@@ -28,9 +28,10 @@ sub required_arch_os { 'x86_64' }
 # LSF resources required to run the alignment.
 sub required_rusage {
     my $self = shift;
+    my %params = @_;
     my $cores = 4; # default to 4, although decomposed_aligner_params should always include "-p N" even if the processing-profile does not
 
-    my $params = $self->decomposed_aligner_params();
+    my $params = $self->decomposed_aligner_params($params{aligner_params});
```
nnutter: Looks like the argument will be ignored if `$self` is an instance.  I'm not sure where `required_rusage` is being called off the top of my head; maybe we should protect against `$self` being an instance or is it intentional?

tmooney: If `$self` is an instance it will have the correct params already attached.  This is to handle the case where it's being called as a class method.

nnutter:
:+1:

PR #447, opened 2015-02-04
New picard option

apregier:
This new option is needed for the X10 data; see AT-732
As an added sanity check that this will work, I made a somatic validation build a637dfde0ba24bf8acc90bf02da71296 that uses the new parameter in its alignment strategy.  It correctly submitted the option to the MarkDuplicates command line, and picard ran without complaint.

### DIFF
--- a/lib/perl/Genome/Model/Tools/Picard/MarkDuplicates.pm
+++ b/lib/perl/Genome/Model/Tools/Picard/MarkDuplicates.pm
@@ -55,7 +55,7 @@ class Genome::Model::Tools::Picard::MarkDuplicates {
         },
         read_name_regex => {
             is => 'Text',
-            doc => 'Set to null to turn off optical duplicate detection',
+            doc => "Set to 'null' to turn off optical duplicate detection",
             is_optional => 1,
         },
     ],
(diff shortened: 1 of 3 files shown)


lib/perl/Genome/Model/Tools/Picard/MarkDuplicates.pm
```diff
@@ -53,6 +53,11 @@ class Genome::Model::Tools::Picard::MarkDuplicates {
             doc => 'comment to include as a @CO in the BAM header',
             is_optional => 1,
         },
+        read_name_regex => {
+            is => 'Text',
+            doc => 'Set to null to turn off optical duplicate detection',
```
tmooney: I know that Perl has `undef

tmooney:
+1 :maple_leaf: assuming tests pass!

dufeiyu:
Can this be tested ?

lib/perl/Genome/Model/Tools/Picard/MarkDuplicates.pm
apregier: Fixed!

apregier:
Thanks @dufeiyu test added!

dufeiyu:
+1

PR #448, opened 2015-02-04
Trust an Allocation's Status When Deciding if Archived

tmooney:
When a previously-archived allocation is purged it will have status `purged` but a mount path pointing to the disk archive.  This leads to trouble when someone or something tries to unarchive this purged allocation.


nnutter:
:+1:

jweible:
:+1:

PR #449, opened 2015-02-04
protect against corrupted Workflow::Models

nnutter:
If _find_first_nondone_step_impl crashes it can leave the related
Workflow::Models in a corrupt state that ends up having "invalid data for
save".  The (side-effect) changes are probably a bug and not relevant to
the task at hand so it's safest just to rollback if
_find_first_nondone_step_impl crashes.

The "invalid data for save" was already being rolled back later when the
transaction fails to commit but by pre-emptively doing it here we can save
a ton of time due to the slow performance of the "invalid data for save"
code.

### DIFF
--- a/lib/perl/Genome/Model/Command/Admin/ModelSummary.pm
+++ b/lib/perl/Genome/Model/Command/Admin/ModelSummary.pm
@@ -296,16 +296,23 @@ sub previous_build {
 sub workflow_status {
     my $build = shift;
 
-    my %status = eval {
+    # The following is wrapped in a transaction and try to protect against
+    # "corrupt" Workflow::Models.
+    my $tx = UR::Context::Transaction->begin();
+    my %status = try {
         my $build_instance = $build->newest_workflow_instance;
-        my @child_instances = $build_instance->sorted_child_instances if $build_instance;
+        my @child_instances = $build_instance ? $build_instance->sorted_child_instances : ();
 
         my %status;
         for my $child_instance (@child_instances) {
             (my $name = $child_instance->name) =~ s/^[0-9]+\s+//;
             $status{$name} = $child_instance->status;
         }
+        $tx->commit();
         return %status;
+    } catch {
+        $tx->rollback();
+        return;
     };
 
     return %status;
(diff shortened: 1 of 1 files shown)


nnutter:
I'm also trying to see if I can improve the performance of "invalid data for save" and may also try to look into the side-effect changes but at the moment model-summary is taking three hours to run instead of three minutes so I'd like to get something out.

lib/perl/Genome/Model/Command/Admin/ModelSummary.pm
```diff
@@ -324,12 +326,32 @@ sub status_compare { # http://stackoverflow.com/q/540229
...
 sub find_first_nondone_step {
+    my $build = shift;
+
+    # The following is wrapped in a transaction and try to protect against
+    # "corrupt" Workflow::Models.
+    my $tx = UR::Context::Transaction->begin();
+    my $first_nondone_step = try {
+        my $wf = $build->newest_workflow_instance;
+        my $step =  _find_first_nondone_step_impl($wf);
+        $tx->commit();
```
tmooney: Does this trigger the `ERROR: Invalid data for save!` message here instead of later?

nnutter: In our current problem `_find_first_nondone_step_impl()` is crashing so `$tx->commit()` does not run.

nnutter:
I've been chasing rabbits... Tom's UR fix wasn't yet in master.   With the UR fix the performance of "invalid data for save" is amazingly better (1000x).  This PR may not be needed but I haven't decided whether it stand on its own merit or not.

nnutter:
This PR does not significantly improve performance but does avoid the noise caused by Workflow::Models that can no longer correctly load due to removed (command) classes.

tmooney:
+1 :palm_tree: Less noise is good!

brummett:
+1

Alternatively, it could have used UR::Context::Transaction::do().  Though I see that eval_or_do() doesn't properly protect $@.  I'll fix that.

PR #450, opened 2015-02-05
Refactor the variant reporting vaf interpreter

susannasiebert:
This PR is to clean up the vaf interpreters. The current situation is that there is one VafInterpreter that takes one sample and its libraries and calculates sample vaf and per-library vaf information for one sample. There is also a ManySamplesVafInterpreter that takes multiple samples and their libraries and calculates the sample vafs and per-library vafs for all of them, using the VafInterpreter under the hood.

This has caused two problems:
- Some additional interpreters use the VafInterpreter as well but aren't interested in the per-library vafs so the library names have to be optional
- We create a lot of throwaway VafInterpreter objects

In addition, there is currently no checking that the provided library names match the library names in the vcf entry that is being interpreted.

This PR makes the following main changes:
- It splits out the sample vaf interpretations from the per-library vaf interpretations, using separate interpreters
- It gets rid of the distinction between single sample and many sample vaf interpreters. It makes it so that these vaf interpreters are multi sample.
- We check that the provided libraries match the libraries in the vcf for the samples that we are interested in analyzing


lib/perl/Genome/VariantReporting/Suite/BamReadcount/MaxVafObservedInterpreter.pm
```diff
@@ -53,14 +53,16 @@ sub _interpret_entry {
     my @vaf_hash_names        = (\%normal_vafs,        \%tumor_vafs);
     my $it                    = each_array(@sample_name_accessors, @vaf_hash_names);
     while ( my ($sample_name_accessor, $vaf_hash_ref) = $it->() ) {
-        for my $sample_name ($self->$sample_name_accessor) {
+        if (defined($self->$sample_name_accessor)) {
```
tmooney: if `$sample_name_accessor` is returning a list, then it will always be `defined` in a scalar context.  (With no values it would return `0`.)

lib/perl/Genome/VariantReporting/Suite/BamReadcount/PerLibraryVafInterpreter.pm
```diff
@@ -0,0 +1,132 @@
...
+}
+
+sub available_libraries {
+    my ($self, $entry) = @_;
+
+    my %readcount_entries;
+    for my $sample_name ($self->sample_names) {
+        %readcount_entries = (%readcount_entries, %{$self->get_readcount_entries($entry, $sample_name)});
+    }
+    $DB::single=1;
```
tmooney: :mag_right: A breakpoint!

lib/perl/Genome/VariantReporting/Suite/BamReadcount/MaxVafObservedInterpreter.pm
susannasiebert: I fixed this by removing the `defined`

lib/perl/Genome/VariantReporting/Suite/BamReadcount/PerLibraryVafInterpreter.pm
susannasiebert: Breakpoint removed.

tmooney:
+1 :deciduous_tree:

(That `'A'` still makes me suspicious, but AT-733 is there for that!)

lib/perl/Genome/VariantReporting/Suite/BamReadcount/PerLibraryVafInterpreter.t
```diff
@@ -0,0 +1,158 @@
...
+        }
+    );
+
+    my $entry = create_long_deletion_entry();
+    my %result = $interpreter->interpret_entry($entry, ['GTATA']);
+    is_deeply(\%result, \%expected, "Values are as expected");
+};
+
+subtest "libraries don't match" => sub {
+    for my $incorrect_library_names ([qw(Solexa-135853)], [qw(Solexa-135854 Solexa-135853 Solexa-135852)], [qw(nonexistent_library)]) {
```
apregier: I don't understand why the first one of these is expected to fail - if you only want info for one of the libraries, could that be valid?

davidlmorton:
I am a little unsure of the `translate_ref_allele` function and its usage here (I know it wasn't changed in this PR).  We'll have to talk about this in person (and include allison), but it makes me nervous to see how we're just doing substr blindly based on `is_deletion` (ignoring the alt).  That was the issue with the vaf calculations before, so I just want to make sure its alright.

lib/perl/Genome/VariantReporting/Suite/BamReadcount/PerLibraryVafInterpreter.t
susannasiebert: It could be but the way that the interpreter was originally written it will always return the vaf fields for all of the sample's libraries. When the library_names were added they were only ever used to  determine that available fields. We never restricted the return values to those libraries only. So I added this check to bring the available fields  in line with the actual output of this interpreter. 

I'm totally open to changing this behavior but I think it would need to happen in a different PR.

susannasiebert:
@dmorton Tom Mooney brought up some similar concerns and I made https://jira.gsc.wustl.edu/browse/AT-733 to address them.

dmorton:
Wrong person

susannasiebert:
@davidlmorton I pushed a few commits to address your concerns.

apregier:
We will definitely want to communicate with users as this goes out since plan files will need to change.

dmorton:
I am not the person you referenced in your action item. Please stop using
me for whatever you are doing.

apregier:
@dmorton Sorry about that!  Your username is similar to one of our team members.  When we accidentally mentioned you, it added you as a participant in the pull request.  We will do our best to get the correct person in the future!  You can unsubscribe if you go to https://github.com/genome/genome/pull/450 and click the "Unsubscribe" button on the right hand side of the page.  I'm not sure we can remove you.

nnutter:
I'm going to try deleting "dmorton" comments to see if that removes him/her as a participant.

nnutter:
That worked.  Once references to "dmorton" were gone the participant was removed.

dmorton:
No problem. Will unsubscribe when I get home to my laptop.

No harm done, except that your messages didnt reach you team member.

Good luck with your project

nnutter:
@dmorton we figured out we can delete your comments and that removes you from the PR.  We'll do that now so this should be the last notification you get.

susannasiebert:
@apregier @davidlmorton Is there anything else you'd like me to do for this PR?

davidlmorton:
+1

PR #451, opened 2015-02-05
Properly scope the saved STDOUT and STDERR handles

brummett:
This should fix the two tests hanging on Jenkins:
    InstrumentData/Command/Import/Basic_sra.t
    InstrumentData/Command/Import/WorkFlow/SraToBam.t

The problem is that Genome::InstrumentData::Command::Import::WorkFlow::SraToBam::do_shellcmd_with_stdout()
is reading from a file of saved STDERR of a child process run with
Genome::Sys::shellcmd() and printing it to its own STDERR.  However, the
current process' STDERR is also printing to that same file.  The tests hang
because do_shellcmd_with_stdout() never gets to the end of the file (it's
filling it as it's reading).

I'm not sure what the _actual_ bug is, but this seems to fix it, while
removing global filehandle variables.  The strange part is that the tests ran fine
when I added seemingly unrelated prints inside the guard closures, and also ran
fine in debuggers.

### DIFF
--- a/lib/perl/Genome/Sys.pm
+++ b/lib/perl/Genome/Sys.pm
@@ -1493,20 +1493,20 @@ sub shellcmd {
                 my ($restore_stdout);
                 if ($redirect_stdout) {
                     no warnings 'once'; # OLDOUT is used only once, not
-                    open(OLDOUT, '>&STDOUT') || die "Can't dup STDOUT: $!";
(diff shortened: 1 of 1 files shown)


tmooney:
+1. :fishing_pole_and_fish: Yay removing global variables!

jasonwalker80:
+1

brummett:
Merging early to get tests passing again

PR #452, opened 2015-02-05
Relay child process' STDERR as error_message()s

brummett:
the old code wasn't doing what it thought it was doing


tmooney:
I don't know what the old code was trying to do, but +1! :crocodile:

acoffman:
+1 :feelsgood:

brummett:
The old code assumed the while would assign each thing returned by ```$err->getline``` to ```$_```, but that's not what happens at all.

nnutter:
:+1:

PR #455, opened 2015-02-05
Another try at fixing STDOUT and STDERR redirection

brummett:
This supercedes e4ae511.  The previous commit would have caused our
web server to wig out.  See the message in commit 0d56bbc for why.

### DIFF
--- a/lib/perl/Genome/Sys.pm
+++ b/lib/perl/Genome/Sys.pm
@@ -1490,24 +1490,30 @@ sub shellcmd {
         $t1 = time();
         my $system_retval;
         eval {
-                my ($restore_stdout);
+            # Use fork/exec here so we can redirect stdout and/or stderr in the child and
+            # not have to worry about resetting them after the child process is done.
+            # Note to the future: FCGI ties STDOUT and STDERR and does something with them
+            # that means you can't open() them with typeglobs or references.  See commit 0d56bbc
+
+            my $pid = fork();
+            if (!defined $pid) {
+                # error
+                die "Couldn't fork: $!";
+
+            } elsif ($pid) {
+                # parent
+                waitpid($pid, 0);
+                $system_retval = $?;
+                print STDOUT "\n"; # add a new line so that bad programs don't break TAP, etc.
+
+            } else {
+                # child
                 if ($redirect_stdout) {
-                    no warnings 'once'; # OLDOUT is used only once, not
-                    open(my $old_out, '>&', *STDOUT) || die "Can't dup STDOUT: $!";
                     open(STDOUT, '>', $redirect_stdout) || die "Can't redirect stdout to $redirect_stdout: $!";
-                    $restore_stdout = UR::Util::on_destroy(sub {
-                        open(STDOUT, '>&', $old_out);
-                    });
                 }
 
-                my ($restore_stderr);
                 if ($redirect_stderr) {
-                    no warnings 'once'; # OLDERR is used only once, not
-                    open(my $old_err, '>&', *STDERR) || die "Can't dup STDERR: $!";
                     open(STDERR, '>', $redirect_stderr) || die "Can't redirect stderr to $redirect_stderr: $!";
-                    $restore_stderr = UR::Util::on_destroy(sub {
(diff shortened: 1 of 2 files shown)


lib/perl/Genome/Sys.pm
```diff
@@ -1490,24 +1490,30 @@ sub shellcmd {
         $t1 = time();
         my $system_retval;
         eval {
-                my ($restore_stdout);
+            # Use fork/exec here so we can redirect stdout and/or stderr in the child and
+            # not have to worry about resetting them after the child process is done.
+            # Note to the future: FCGI ties STDOUT and STDERR and does something with them
+            # that means you can't open() them with typeglobs or references.  See commit 0d56bbc
+
+            my $pid = fork();
```
tmooney: Is it safe to `fork` here without using `UR::Context::Process->fork()`?

brummett: Maybe?

This process is going away one way or another, either through the exec or by dying.  We don't have to worry about setting up a new UR::Context::Process, or disconnect/reconnect DB handles.

brummett:
Added a new commit to make sure the child process goes away if the exec() fails

lib/perl/Genome/Sys.pm
nnutter: I think it should be.  This is just a special version of `system`.  The child is not a Perl process.

lib/perl/Genome/Sys.pm
```diff
@@ -1522,10 +1528,15 @@ sub shellcmd {
                 {   # POE sets a handler to ignore SIG{PIPE}, that makes the
                     # pipefail option useless.
                     local $SIG{PIPE} = 'DEFAULT';
-                    $system_retval = system('bash', '-c', "$shellopts_part $cmd");
+                    my @cmdline = ('bash', '-c', "$shellopts_part $cmd");
+                    exec(@cmdline)
+                        or do {
```
nnutter: I suppose the `or do` isn't even necessary.  The next two lines after `exec` could just be the `print` and `exit` since they will never run if the `exec` works.  I'm not sure if this idea is good or bad...

brummett: Not necessary, but it matches the pattern I try to use for these kinds of syscalls
```
function() or die '...';
```
Since there's 2 steps to "die" here, I put it in a do {} block.

brummett:
Two new commits added to fix the Sys-shellcmd.t test.  The new behavior about printing the extra newline actually changed between the new shellcmd() and the old.

acoffman:
+1

iferguson90:
Forking seems a lot cleaner. Cautious +1.

nnutter:
@apregier Yes, they will have one less newline but only in the case where stdout is redirected.  The newline was there because some programs did not have a terminating newline and that would break TAP.  When redirecting to a file TAP is probably not relevant and adding the newline there would require an additional process/fork.

apregier:
I guess since unit tests are passing, we can assume that nothing is depending on the extra newline.

brummett:
The only unit test that broke was the test for shellcmd() itself.
I suppose it could cause a diff in a test build file, if the source is a tool that prints to stdout, and the Genome code redirects its stdout to a file.

brummett:
Added another commit to preserve the old behavior of always adding an extra newline

apregier:
I wasn't necessarily advocating that we keep the old behavior, but it does seem like the safer route to go. +1

nnutter:
Test failure is external and will be addressed by genome/genome#459.

brummett:
Merged from master to get  4ac5110 in #459

nnutter:
> Merged from master to get 4ac5110 in #459

I wanted to share that you don't necessarily have to merge from master but could just "rebuild" the tests.  We run tests on a merge _to_ master that GitHub makes.  Theoretically they make a new merge commit when master updates but we've found that that isn't always the case.  If I can't figure out the reason GitHub sometimes doesn't update the merge then I'm thinking we should do the merge on apipe-ci so that the only time we have to merge master is to resolve conflicts.

PR #456, opened 2015-02-06
clin-seq docm report update

gatoravi:
This PR adds RNASeq read-counts to the DOCM report that is generated within ClinSeq. Currently only WGS and Exome counts are reported back.


lib/perl/Genome/Model/ClinSeq/Command/Converge/SnvIndelReport.pm
```diff
@@ -857,7 +860,8 @@ sub parse_read_counts{
       my $prefix = $samples{$sample_name}{prefix};
       my $coverage = $samples{$sample_name}{coverage};
       #don't apply min_coverage on rnaseq, the transcript might not be expressed.
-      $min_coverage_observed = $coverage if ($coverage < $min_coverage_observed and $prefix !~ /rnaseq/);
+      $min_coverage_observed = $coverage if ($coverage < $min_coverage_observed and
```
nnutter: Some people would prefer that complex conditionals not be used in postfix notation.

lib/perl/Genome/Model/ClinSeq/Util.pm
```diff
@@ -1226,7 +1227,8 @@ sub get_ref_align_builds{
 
       $ref_builds{$refalign_name}{type} = $build_type;
       $ref_builds{$refalign_name}{sample_name} = $subject_name;
-      $ref_builds{$refalign_name}{sample_name_build_type} = $subject_name . "_" . $build_type;
+      $ref_builds{$refalign_name}{sample_name_build_type} =
+        $subject_name . "_" . $subject_common_name . "_" . $subject_icn . "_" . $build_type;
```
tmooney: Is it okay to have multiple consecutive `_`s in these headers if the subject's or the individual's common names are not set?  (Currently this will produce warnings if they are not.  It would be better to either default to the empty string or omit them as preferred.)

This could use `join('_', ...)` instead a series of concatenations.

nnutter:
Looks like you need to merge master into your branch because GitHub says, "We can’t automatically merge this pull request."

```
# cd/checkout branch
git fetch
git merge origin/master
git push
```

nnutter:
:+1: if tests pass after merge.

tmooney:
Since there was a conflict in updating the blessed ClinSeq build, will you need to run another build after the merge to bless a combination of both updates?

gatoravi:
This is true. This took a while since my builds have been failing due to /tmp issues. I've updated the blessed-build now.

gatoravi:
merging !

PR #458, opened 2015-02-06
add scope argument to POD

nnutter:


acoffman:
+1

nnutter:
Test failure is external and will be addressed by genome/genome#459.

lib/perl/Genome/Sys/Lock.pm
```diff
@@ -17,6 +17,10 @@ Keyword Arguments:
 C<resource_lock> is the path to the file representing the lock for some
 resource.
 
+=item scope
+
+C<scope> is the scope to which the lock is bound.  See C<scope()> for valid scopes.
```
tmooney: Should this point to `scopes()` instead of `scope()`?

nnutter: Fixed.

tmooney:
+1 :penguin:

PR #459, opened 2015-02-06
new expected output due to bam-window update

nnutter:
new expected output due to bam-window update

### DIFF
--- a/lib/perl/Genome/Model/Tools/BamWindow.t
+++ b/lib/perl/Genome/Model/Tools/BamWindow.t
@@ -18,7 +18,7 @@ use File::Compare;
 my $class = 'Genome::Model::Tools::BamWindow';
 use_ok($class);
 
-my $test_version = 2;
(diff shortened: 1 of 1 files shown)


nnutter:
@tabbott @ernfrid @chrisamiller Tests broke when I merged genome/genome-snapshot-deps#12 to deploy `bam-window` v0.5.4.  Can you please verify the new output is correct?

```
$ cd $GENOME_TEST_INPUTS/Genome-Model-Tools-BamWindow && diff --brief -r v{2,3}
Files v2/output.txt and v3/output.txt differ
```

chrisamiller:
It looks correct to me. All columns have fewer overall reads, which is expected due to the added filtering in the default params.  I'll sign off on it.

nnutter:
Thanks @chrisamiller.  I'll go ahead and merge it to get tests passing.  If someone else sees a problem we can obviously still fix it.

PR #460, opened 2015-02-09
update genome-env-next default DB snapshot

nnutter:


nnutter:
Tests don't actually do anything yet and there's nothing really to review since this is `genome-env-next` and is only in "beta".

PR #461, opened 2015-02-09
move sqitch/gms to submodule sqitch/genome

nnutter:
No changes were made except for moving it to `sqitch/genome` instead of `sqitch/gms` to avoid problem that would cause when switch between commits before this one.


nnutter:
I don't think there's anything to review here.

tmooney:
:shipit:

PR #462, opened 2015-02-09
genome-env-next iteration

nnutter:
Some fixes and improvements for `genome-env-next`.


nnutter:
Reviewing individual commits probably makes the most sense.

/cc @tabbott

nnutter:
No test coverage for this but it's in "beta" because of that.

ebelter:
+1 Seems reasonable to me, but I am not completely familiar with the code.

apregier:
+1

PR #463, opened 2015-02-09
Recognize new date format

mkiwala:
Capturing this new date format seems more desirable than changing how
the logs are created for two reasons:
1) changing how the logs are created would require re-packaging and
re-deploying Flow
2) updating the recognizer to allow this date format allows us to
automatically determine historical errors

### DIFF
--- a/lib/perl/Genome/Model/Build/Command/DetermineError.pm
+++ b/lib/perl/Genome/Model/Build/Command/DetermineError.pm
@@ -23,6 +23,10 @@ our $WORKFLOW_DATE_AND_HOST = qr{
 our $PTERO_DATE = qr{
         \[(?<date>\d{4}/\d{2}/\d{2}\s\d{2}\:\d{2}\:\d{2}).\d+\]
     }x;
+our $FLOW_DATE = qr{
+        (?<date>\d{4}-\d{2}-\d{2}\s\d{2}\:\d{2}\:\d{2}).\d+
+    }x;
+
 our $ERROR_LOCATION = qr{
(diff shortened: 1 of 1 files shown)


nnutter:
:+1: I think that 2) is sufficient and that 1) is probably not.

brummett:
Does ```$FLOW_DATE``` need to get added to ```$EXCEPTION_FINDING_REGEX``` also?

mkiwala:
The `$EXCEPTION_FINDING_REGEX` appears to be tailored to Perl exceptions.  I think we'd have to do something different for python exceptions.

brummett:
The date part of each line is added by a log annotator.  In the old Workflow code, it was added by annotate-log.  I'm not sure how the dates get there in Ptero.

mkiwala:
@brummett -- Looks like [this](https://github.com/genome/flow-workflow/blob/52b2c5c3ba1b3f54c2b74f89ebe2a3dc83776653/bin/workflow-wrapper.pl#l24) could be the place.

brummett:
If that style date can ever appear before an exception, then ```$FLOW_DATE``` would need to be added to ```$EXCEPTION_FINDING_REGEX```

mkiwala:
@brummett -- I agree.  I added `$FLOW_DATE` to `$EXCEPTION_FINDING_REGEX`.

brummett:
+1

iferguson90:
+1 :sheep:

nnutter:
CI, test this.

PR #464, opened 2015-02-09
Make vaf interpreters transient optional parameters

susannasiebert:
We want to avoid creating throwaway vaf interpreter objects every time _interpret_entry is called. Instead this PR makes it so that the vaf interpreter(s) are made into transient optional parameters.

### DIFF
--- a/lib/perl/Genome/VariantReporting/Suite/BamReadcount/MaxVafObservedInterpreter.pm
+++ b/lib/perl/Genome/VariantReporting/Suite/BamReadcount/MaxVafObservedInterpreter.pm
@@ -23,6 +23,16 @@ class Genome::VariantReporting::Suite::BamReadcount::MaxVafObservedInterpreter {
             doc => 'A list of normal samples to use for calculation'
         },
     ],
+    has_transient_optional => [
+        tumor_vaf_interpreter => {
+            is => 'Genome::VariantReporting::Suite::BamReadcount::VafInterpreter',
+            is_structural => 1,
+        },
+        normal_vaf_interpreter => {
+            is => 'Genome::VariantReporting::Suite::BamReadcount::VafInterpreter',
+            is_structural => 1,
+        },
+    ],
     doc => 'Calculate the maximum vaf value observed between all the specified normal samples, and the maximum vaf value observed between all the tumor samples',
 };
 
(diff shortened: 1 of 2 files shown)


tmooney:
+1 :foggy:

It might be good to add a `doc` to the definition of `is_structural` in `Genome::VariantReporting::Framework::Component::Base`, but that's well outside of this PR.

lib/perl/Genome/VariantReporting/Suite/BamReadcount/MaxVafObservedInterpreter.pm
```diff
@@ -49,14 +59,13 @@ sub _interpret_entry {
...
+    my @sample_name_accessors     = qw/normal_sample_names    tumor_sample_names/;
+    my @vaf_hash_names            = (\%normal_vafs,           \%tumor_vafs);
+    my @vaf_interpreter_accessors = qw/normal_vaf_interpreter tumor_vaf_interpreter/;
+    my $it                    = each_array(@sample_name_accessors, @vaf_hash_names, @vaf_interpreter_accessors);
+    while ( my ($sample_name_accessor, $vaf_hash_ref, $vaf_interpreter_accessor) = $it->() ) {
         if ($self->$sample_name_accessor) {
-            my $interpreter = Genome::VariantReporting::Suite::BamReadcount::VafInterpreter->create(
-                sample_names => [$self->$sample_name_accessor],
-            );
+            my $interpreter = $self->$vaf_interpreter_accessor;
```
apregier: This could return undef, in which case the next line will throw an error.

apregier: I guess since you have the check on line 67, it shouldn't ever return undef.  It might be clearer to check whether the $vaf_interpreter_accessor call returns undef at this level, and let the subroutine check if the $sample_name_accessor is defined

apregier:
+1

lib/perl/Genome/VariantReporting/Suite/BamReadcount/MaxVafObservedInterpreter.pm
susannasiebert: I'm trying to avoid this by checking for `$self->$sample_name_accessor` on line 67. Do you think that's sufficient or should I add another check for the return value of `$self->$vaf_interpreter_accessor`?

susannasiebert: I would love to update this PR with a commit to implement your suggestion but I'm having disk issues.

PR #465, opened 2015-02-10
Only check that converter exists during ROI validation.

tmooney:
These should have been changed as part of #429 but were not.


susannasiebert:
+1 once tests pass

jweible:
:+1:

PR #466, opened 2015-02-10
Change how the PerLibraryvafInterpreter handles planned libraries

susannasiebert:
Instead of trying to check that the planned libraries match the libraries provided in the vcf, restrict the interpretations to the set of planned libraries.

### DIFF
--- a/lib/perl/Genome/VariantReporting/Suite/BamReadcount/PerLibraryVafInterpreter.pm
+++ b/lib/perl/Genome/VariantReporting/Suite/BamReadcount/PerLibraryVafInterpreter.pm
@@ -38,16 +38,6 @@ sub _interpret_entry {
     my $entry = shift;
     my $passed_alt_alleles = shift;
 
-    if ($self->library_names) {
-        my $available_libraries = $self->available_libraries($entry);
-        my $expected_libraries  = Set::Scalar->new($self->library_names);
-        unless ($available_libraries->is_equal($expected_libraries)) {
-            die $self->error_message(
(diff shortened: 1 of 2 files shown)


tmooney:
The previous version would have prevented someone from asking for a non-existent library, and it would still work if no libraries were specified.  Do we need to care about either of those cases now that we trust the supplied library list?

susannasiebert:
The goal of this iteration as well as the previous iteration was to consolidate the list of planned libraries with the libraries that are being output in the interpretations hash. The problem with enforcing that the list of planned libraries have to match the libraries in the vcf is that a lot of times we don't exactly know which libraries are in the vcf before processing it. Our wrappers that use this interpreter just put all the sample's libraries in the library_names list, even ones that aren't in the vcf. The goal of this PR is actually to become more lenient and allow non-existent libraries so that our wrappers still work. However, I think in the past we've entered a 'null character' in the interpretations hash when we didn't have a value for a field. This change doesn't do that (it would be undef). I will push a commit to fix that.

We do want to enforce libraries being specified because the planned libraries are supposed to be an indicator of what to expect in the interpretations output (via the field_descriptions and the available_fields). If you don't specify any libraries then we would expect that no interpretations field are returned. The planned libraries should dictate the output, not the libraries in the vcf.

tmooney:
Cool, +1! :last_quarter_moon_with_face:

apregier:
+1

susannasiebert:
CI, test this.

PR #468, opened 2015-02-11
Defer SR user creation to commit time and lock to prevent duplicates.

tmooney:
There was a race condition here where multiple processes thought they needed to create the sponsor relationship.  This adds locking around the user creation to avoid the duplicates.

The creation is delayed to a pre-commit hook to avoid needing to either commit the user relationship in a separate transaction or potentially hold the lock for a long time.

### DIFF
--- a/lib/perl/Genome/SoftwareResult/User.pm
+++ b/lib/perl/Genome/SoftwareResult/User.pm
@@ -5,6 +5,9 @@ use warnings;
 use Genome;
 use List::MoreUtils qw(any);
 use Params::Validate qw(:types);
+use Carp qw();
+
+use Genome::Utility::Text;
 
 class Genome::SoftwareResult::User {
     table_name => 'result.user',
@@ -96,14 +99,86 @@ sub _register_users {
     my $label = $newly_created ? 'created' : 'shortcut';
     $user_hash{$label} = $requestor;
 
+    my @all_params;
     while(my ($label, $object) = each %user_hash) {
         my %params = (
             label           => $label,
             user            => $object,
             software_result => $software_result,
         );
-       $class->get(%params) || $class->create(%params);
+        push @all_params, \%params;
     }
+
+    my $observer;
+    $observer = UR::Context->process->add_observer(
+        aspect => 'precommit',
+        callback => sub {
+            if($observer) {
+                $observer->delete;
+                $observer = undef;
+            } else {
+                Carp::confess 'observer triggered multiple times!';
+            }
+            my @locks;
+            for my $params (@all_params) {
+                push @locks, $class->_get_or_create_with_lock($params);
+            }
+            my $unlocker;
+            $unlocker = UR::Context->process->add_observer(
+                aspect => 'commit',
+                callback => sub {
+                    if($unlocker) {
+                        $unlocker->delete;
+                        $unlocker = undef;
+                    } else {
+                        Carp::confess 'unlocker triggered multiple times!';
+                    }
+
+                    for my $lock (@locks) {
+                        Genome::Sys->unlock_resource(resource_lock => $lock);
+                    }
+                }
+            );
+        }
+    );
+    unless($observer) {
+        die 'Failed to create observer';
+    }
+}
+
+sub _get_or_create_with_lock {
+    my $class = shift;
+    my $params = shift;
+
+    my $existing = $class->get(%$params);
+    return if $existing;
+
+    my $resource = $class->_resolve_lock_name($params);
+    my $lock = Genome::Sys->lock_resource(resource_lock => $resource, scope => 'site');
+    $class->_get_or_create($params);
+
+    return $lock;
+}
+
+sub _get_or_create {
+    my $class = shift;
+    my $params = shift;
+
+    my $self = $class->load(%$params);
+    unless($self) {
+        $self = $class->create(%$params);
+    }
+
+    return $self;
+}
+
+sub _resolve_lock_name {
+    my $class = shift;
+    my $params = shift;
+
+    return 'genome/software-result-user/' . Genome::Utility::Text::sanitize_string_for_filesystem(
+        join('_', $params->{label}, $params->{user}->id, $params->{software_result}->id)
+    );
 }
 
 sub _role_for_type {
--- a/lib/perl/Genome/SoftwareResult/User.t
+++ b/lib/perl/Genome/SoftwareResult/User.t
@@ -21,12 +21,16 @@ class SR_Test {
     },
 };
 
+sub SR_Test::create {
+    return shift->Genome::SoftwareResult::create(@_);
+}
+
 sub _newly_created_callback {
-    return (SR_Test->__define__(p1 => 'turkey'), 1);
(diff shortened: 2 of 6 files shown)


lib/perl/Genome/SoftwareResult/User.pm
```diff
@@ -96,16 +99,88 @@ sub _register_users {
...
+
+    my $observer;
+    $observer = UR::Context->process->add_observer(
+        aspect => 'precommit',
+        callback => sub {
+            if($observer) {
+                $observer->delete;
+                $observer = undef;
+            } else {
+                Carp::confess 'observer triggered multiple times!';
```
nnutter: These checks don't seem necessary.  Do you have some reason to believe they are?

tabbott: They were when we were committing in the callback, but maybe they aren't any more.

lib/perl/Genome/SoftwareResult/User.pm
```diff
@@ -96,16 +99,88 @@ sub _register_users {
...
+    $class->_get_or_create($params);
+
+    return $lock;
+}
+
+sub _get_or_create {
+    my $class = shift;
+    my $params = shift;
+
+    my $self = $class->load(%$params);
```
nnutter: Why `load`?

lib/perl/Genome/SoftwareResult/User.t
```diff
@@ -68,6 +72,7 @@ my $new_sr = $class->with_registered_users(
             callback => \&_newly_created_callback,
             %users_hash
         );
+UR::Context->commit();
```
nnutter: Do these need to be `UR::Context->process->commit()`?  In which case will these deadlock?

lib/perl/Genome/SoftwareResult/User.pm
tmooney: We try to `get` before locking in `get_or_create_with_lock`, so we need to check the database directly the second time.

lib/perl/Genome/SoftwareResult/User.t
nnutter: In other, maybe better words...

If you have two things trying to `_get_or_create_with_lock` in `precommit` and the lock isn't removed until `commit`.  Isn't that a deadlock?  Tests didn't catch it because they're calling commit on the wrong context?

lib/perl/Genome/SoftwareResult/User.pm
nnutter: If there's still an issue it seems like it's a bug in UR that would need to be fixed.  As is, it looks like you are worried about multi-threaded code or not trusting `$observer = undef`.

lib/perl/Genome/SoftwareResult/User.pm
nnutter: > There are only two hard things in Computer Science: cache invalidation and naming things.

lib/perl/Genome/SoftwareResult/User.t
tmooney: If two things in the same process are trying to get_or_create the same user, the second one should find the one created by the first one in the `get` before locking.

The observers are firing or we would not successfully find the user relationships in the test.  (For what it's worth we had STDERR prints in the second observer originally to make sure it fired.)  I thought `UR::Context->commit();` was on the process if you're not in a transaction.

lib/perl/Genome/SoftwareResult/User.pm
tmooney: The issue we had before is that, if `commit` is triggered while inside the callback, the remaining ones would be queued multiple times even though they are deleted after the first time through.  We no longer call `commit` inside the callback so that isn't an issue anymore.  This is now the equivalent of a `die('This should never happen!');`.

I'll take it out if it seems extraneous.

brummett: Sounds like UR observers could use a ```once => 1``` attribute

lib/perl/Genome/SoftwareResult/User.t
nnutter: Oh, yeah, that makes sense.  It seems like it would be good if the test checked that they were unlocked after commit.  If you extracted a lock method that allowed arguments to be passed to `Genome::Sys->lock_resource` you could pass `max_try => 1`.

lib/perl/Genome/SoftwareResult/User.pm
nnutter: I don't know what the right thing to do is.

nnutter: > Sounds like UR observers could use a `once => 1` attribute

In order to replace the enclosed observer delete pattern, sure, but not to work around issues when trying to `commit` in `precommit`.

tabbott: For the record, we were calling commit in a "commit" observer (not precommit).

lib/perl/Genome/SoftwareResult/User.t
```diff
@@ -21,12 +21,16 @@ class SR_Test {
     },
 };
 
+sub SR_Test::create {
```
tabbott: If all this does is call the parent constructor, do we really need it?

tmooney: Nope! Removed.

lib/perl/Genome/SoftwareResult/User.t
tmooney: I've added a test to check that the lock was unlocked.

nnutter:
:+1:

lib/perl/Genome/SoftwareResult/User.pm
```diff
@@ -96,16 +99,88 @@ sub _register_users {
...
+    my $params = shift;
+
+    my $existing = $class->get(%$params);
+    return if $existing;
+
+    my $resource = $class->_resolve_lock_name($params);
+    my $lock = Genome::Sys->lock_resource(resource_lock => $resource, scope => 'site');
+    $class->_get_or_create($params);
+
+    return $lock;
```
acoffman: Its a little odd to me that a method on `Genome::SoftwareResult::User` called `get_or_create_with_lock` returns a lock instance rather than a `G::SR::U` instance.

tmooney: > There are only two hard things in Computer Science: cache invalidation and naming things.

acoffman:
+1 :cool:

lib/perl/Genome/SoftwareResult/User.pm
nnutter: `_get_or_create_with_lock` could add the `commit` observer instead of letting the caller do that.  Then you wouldn't have to return the lock object.

tmooney: I merged this PR so that we could get the desired functional changes out there.  But I opened a new one, #475, to continue the discussion of how this might be better.

PR #469, opened 2015-02-11
need to install handlers before backends start locking

nnutter:
If you interrupted during first mid-`lock_resource` then locks would not be cleaned up because the cleanup handlers weren't installed yet.


brummett:
+1

That fixes another thing I noticed while looking at the Nessy lock bug

acoffman:
+1

PR #470, opened 2015-02-12
fix option spec

nnutter:
I let the option spec get out of sync again.  This fixes it for changes that
were previously made to the case statement:

- add the -m option
- remove the -v option


acoffman:
+1

PR #471, opened 2015-02-12
Variant reporting: fail vep annotation early if feature list contains invalid characters

susannasiebert:
Sometimes a feature list that is used for vep annotation in the variant reporting subsystem will contain special characters which might cause downstream tools to fail. E.g., semicolons cause errors for joinx vcf-merge if they aren't used to separate info fields. This change will check the feature lists supplied to the vep annotation and crash if it contains a special character.

Right now the only special character I'm aware of is semicolon but this code change allows for easy addition of other invalid characters.

### DIFF
--- a/lib/perl/Genome/FeatureList.pm
+++ b/lib/perl/Genome/FeatureList.pm
@@ -359,7 +359,7 @@ sub processed_bed_file {
     Genome::Sys->write_file($temp_file, $content);
     return $temp_file;
 }
-Memoize::memoize('processed_bed_file');
+Memoize::memoize('processed_bed_file', LIST_CACHE => 'MERGE');
 
 sub generate_merged_bed_file {
     my $self = shift;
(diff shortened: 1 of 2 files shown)


davidlmorton:
I understand why you want to do this, but I'm concerned that we're imposing a potentially substantial computational cost on everybody by doing this.  These bed files could be very large.  Also, the most common case is that there are no invalid characters, especially when we use the same `FeatureList` over and over.

Maybe we could cache this check on the `FeatureList` somehow.  Do `FeatureList` objects have metrics or notes?

susannasiebert:
I don't know about storing this on the feature list or not but I share your concerns. I was also thinking that the 'annotation' field often seems to be in the same format for every line. E.g., the problematic feature list I've encountered had semicolons on every single line. We might be be able to get away with only checking the first 10 lines or so.

davidlmorton:
Since we're only talking about failing earlier (with a clearer error message perhaps) than before, and not a correctness issue, I'm pretty happy with just checking the first n lines where n is small.

susannasiebert:
Ok, I will make that change. What changes would you like to see made to the error message?

davidlmorton:
I think that error message looks good.

susannasiebert:
I pushed a commit that changes the validation to only check the first 10 lines of a feature list.

acoffman:
+1 :dog:

tmooney:
+1 :snowflake:

(Also, yay changing that `,` to a `;`!)

PR #472, opened 2015-02-12
Deduplicate InstrumentData AlignmentBams commands

tmooney:
There were two commands that basically did the same thing, but with different algorithms for finding the BamQC (neither of which were quite right).  This combines the two and uses a (hopefully) better approach to getting the LaneQC.

Additionally the part used by ClinSeq is updated so that it actually returns one result per instrument data instead of squashing any that share a lane.  (ClinSeq was previously calling the version that gave the same result regardless of instrument data, so it didn't affect anything before.)


ebelter:
+1

apregier:
+1

tmooney:
I've included an updated blessed ClinSeq build to account for the differing directory structure and that the command finds different results now.

PR #474, opened 2015-02-12
BWAMEM AR: pass ref index users along to aligner index software result

ebelter:


lib/perl/Genome/InstrumentData/AlignmentResult/BwamemBase.pm
```diff
@@ -300,9 +300,10 @@ sub prepare_reference_sequence_index {
     Genome::Sys->create_symlink($refindex->reference_build->get_sequence_dictionary("sam"), $staging_dir ."/all_sequences.dict" );
 
     my $bwa_index = Genome::Model::Build::ReferenceSequence::AlignerIndex->get_or_create(
+        users              => [ $refindex->users ],
```
tmooney: `Genome::SoftwareResult::WithNestedResults` provides a `_user_data_for_nested_results()` accessor that returns a value ready-made to be passed here as the `users` parameter.

tmooney:
+1 :ear_of_rice:

jweible:
+1

PR #475, opened 2015-02-12
Tweak internals of SR User locking

tmooney:
It was suggested that `_get_or_create_with_lock` was a confusing name for a thing that returns a lock.  Here's a perhaps better alternative.  It's also possible we could restructure this instead?

### DIFF
--- a/lib/perl/Genome/SoftwareResult/User.pm
+++ b/lib/perl/Genome/SoftwareResult/User.pm
@@ -159,12 +159,12 @@ sub _lock_and_get_or_create {
 
     my $resource = $class->_resolve_lock_name($params);
     my $lock = Genome::Sys->lock_resource(resource_lock => $resource, scope => 'site');
-    $class->_get_or_create($params);
+    $class->_load_or_create($params);
 
     return $lock;
 }
 
-sub _get_or_create {
+sub _load_or_create {
     my $class = shift;
     my $params = shift;
 
--- a/lib/perl/Genome/SoftwareResult/User.pm
+++ b/lib/perl/Genome/SoftwareResult/User.pm
@@ -99,14 +99,14 @@ sub _register_users {
     my $label = $newly_created ? 'created' : 'shortcut';
     $user_hash{$label} = $requestor;
 
-    my @all_params;
+    my @param_sets;
     while(my ($label, $object) = each %user_hash) {
         my %params = (
             label           => $label,
             user            => $object,
             software_result => $software_result,
         );
-        push @all_params, \%params;
+        push @param_sets, \%params;
     }
 
     my $observer;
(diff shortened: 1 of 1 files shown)


nnutter:
Maybe...

- `_get_or_create` should be `_load_or_create`.
- `@all_params` should be `@param_sets`.

For `_get_or_create_with_lock` we could do one of the following:
 
- `_get_or_create_with_lock` could return `($user, $lock)`.
- `_get_or_create_with_lock` could add the `commit` observer instead of letting the caller do that.  Then you wouldn't have to return the lock object.
- We could give User objects `lock()` and `unlock()` methods.  Leave it called `_get_or_create_with_lock` and return a locked User object.  The `commit` callback can then just call unlock on the User object.

I think I would lean towards the latter.

tmooney:
I've renamed `&_get_or_create` and `@all_params` as suggested.  The new names do sound clearer!

(Still need to address the `_get_or_create_with_lock` part.)

tabbott:
I'm down with `_load_or_create`.

I wonder if we should even be writing "get" in `_lock_and_get_or_create`. To me it's more like `_create_if_not_exists`, `_atomic_create`, `_ensure_exists`, `_commit_unique`, or maybe just `_commit`.

tabbott:
I think the reasoning of not registering the commit-time observers in the function that does the creation was to avoid having a bunch of them. Put differently, we thought it might be better to build up a list of all the stuff to do and have one precommit hook and one commit hook. Maybe that was a silly idea?

nnutter:
> I'm down with `_load_or_create`.

> I wonder if we should even be writing "get" in `_lock_and_get_or_create`. To me it's more like `_create_if_not_exists`, `_create_if_needed`, `_atomic_create`, `_ensure_exists`, `_commit_unique`, or maybe just `_commit`.

I'd prefer something like that as well but I think keeping consistency with SR is something to be considered.

nnutter:
> I think the reasoning for not registering the commit-time observers in the function that does the creation was to avoid having a bunch of them. Put differently, we thought it might be better to build up a list of all the stuff to do and have one precommit hook and one commit hook. Maybe that was a silly idea?

Are you responding to me?  If so, I don't think that was a silly idea.  I think 2) would break that but I'm still in favor of 3) which would not.

tabbott:
@nnutter Yeah, having lock/unlock methods and returning user objects also makes "get" a more reasonable part of the method name :)

tmooney:
I don't like the asymmetry that, under (3), `lock` cannot be an object method, but `unlock` would be one.  Also since avoiding deadlocks requires avoiding locking in some circumstances, we would be returning possibly-locked but possibly-not-locked User objects to a location that only needs locks and ignores the User objects.

My interpretation of the various comments is that it was a mistake to `return` the user objects from what is now `_load_or_create`, as they aren't needed anywhere except the list of things to be committed.  This only served to confuse readers about what was happening.

tmooney:
I've rebased this branch to address the merge conflict and changed another name and a return value (that's not checked, anyway).

Is this better than where it started?  If so, are there other changes, either previously suggested or not, that should be included before this is merged?

tmooney:
In the absence of a response, the current contents of this PR seem harmless and might be better than what's there now, so I guess I'll merge it!

PR #476, opened 2015-02-13
Vep standalone process

apregier:
The new requirement for a requestor user broke the ability to run `genome db ensembl vep` and `genome db ensembl run vep` as standalone commands.  
To get around this, I created a process that would simply run the vep command and can be recorded as the requestor.  This is similar to what was done in a42c5e9ea9c4d8b230d6065a27438885a24347da
Up for debate could be if we should
-make some kind of generic process that could be used in any stand-alone command that involves software results
-add more info to the process (right now it basically is just a default process with no extra inputs)

Note that these commands are used quite often and their loss is quite acutely felt by analysts, so it would be good to get a fix out quickly.


tmooney:
+1 :snowboarder:  I like that running `Process`es inline is now more official!

I lean towards more specific `Process`es than generic ones.  In this case I'm not sure how much benefit there would be to adding more info--it would be mostly duplication of information already on the result.

acoffman:
+1

I'm a little worried that Process continues to ever more closely resemble Build. We should probably address that soon! Also, I think that eventually we want to largely eliminate people running stuff by hand - do we need to make pipeline improvements so that analysts are getting the vep results they need in a more automated fashion?

apregier:
@acoffman Yes, we probably do need to get vep running in the pipeline!  Although I suspect there will always be some degree of running stand-alone.
@tmooney  The one thing about this process is that no result is actually created (usually).  The software results in question are "reference" results - there should be one result per version of ensembl/organism pair.  So most of the parameters will be "lost" unless we store them on the process.  Although they will be recorded in the workflow inputs hash, which is written to the file system in the Process base class.

PR #477, opened 2015-02-13
Start filtering instrument data by config rules on mass assignment

acoffman:
This isn't passing tests yet, but I wanted to get feedback on the idea first. @tmooney especially should probably take a look at this.

This actually grew out of these JIRA issues: 
 * https://jira.gsc.wustl.edu/browse/BIO-1326
 * https://jira.gsc.wustl.edu/browse/BIO-1315

Essentially, the rules in the configuration specified only bisulfite data, but `auto_assign_inst_data` was turned on. The `Genome::Model::Command::InstrumentData::Assign` commands don't know anything about config, and thus just do their own checks (do the subjects match? trsn? etc) and then blindly assign everything that's compatible.

We discussed the following potential fixes:
 * Add a condition to `compatible_instrument_data` in appropriate models that checks to make sure that the library preps are the same

* Add a condition to the `G::M::C::ID::Assign` commands that checks for similar library preps

And decided the we didn't like the idea of specifying this information in both the rules and somewhere in code. This is the general idea we came up with. Every piece of instrument data that we AutoAssign to a model must itself pass the ruleset for that config item. Thoughts?

### DIFF
--- a/lib/perl/Genome/Config/Command/ConfigureQueuedInstrumentData.pm
+++ b/lib/perl/Genome/Config/Command/ConfigureQueuedInstrumentData.pm
@@ -98,22 +98,23 @@ sub _process_models {
                 $model->id, ($created_new ? 'created' : 'found'), $instrument_data->id ));
 
         $self->_assign_model_to_analysis_project($analysis_project, $model, $config_profile_item, $created_new);
-        $self->_assign_instrument_data_to_model($model, $instrument_data, $created_new);
+        $self->_assign_instrument_data_to_model($model, $instrument_data, $config_profile_item, $created_new);
         $self->_update_model($model);
         $self->_request_build_if_necessary($model, $created_new);
     }
 }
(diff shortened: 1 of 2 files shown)


tmooney:
Since we store the configuration on the AnalysisProject-Model bridge, could we revise `compatible_instrument_data` to consult the configuration?  At that point we could start leaving `auto_assign_inst_data` on for pretty much everything (or remove it as a property?), since it should never assign outside the configuration.

acoffman:
Hrm, that's definitely a possibility. We had talked about that a little bit (and even thought about calling that method from CQID and passing in the config) but we kind of thought we didn't want to start introducing AnP-ish logic into models. Right now, aside from those accessors, they operate blissfully unaware of config/analysis projects/etc. Maybe that's not actually a good thing though?

acoffman:
Superseded by #482

PR #479, opened 2015-02-16
Don't try to add users to deleted results.

tmooney:
If the result is deleted for any reason, we don't still want to try to add users.  (We know it's not coming back unless another pre-commit hook resurrects it...)

For example, `Genome::InstrumentData::WithIntermediateResults` gets the intermediate results (thus adding a callback) immediately before deleting them.

### DIFF
--- a/lib/perl/Genome/SoftwareResult/User.pm
+++ b/lib/perl/Genome/SoftwareResult/User.pm
@@ -121,6 +121,7 @@ sub _register_users {
             }
             my @locks;
             for my $params (@all_params) {
+                next if $params->{user}->isa('UR::DeletedRef');
                 push @locks, $class->_get_or_create_with_lock($params);
             }
             my $unlocker;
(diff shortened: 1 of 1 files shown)


brummett:
It should skip creating the commit observer if ```@locks``` is empty

tmooney:
Agreed--check added!

brummett:
+1

acoffman:
+1

tmooney:
Merging to hopefully fix model tests.

nnutter:
I feel like an alternate implementation would be to register a delete observer on the user to remove the precommit observer; same goes for the software result.

tmooney:
I considered another observer, but I thought it was becoming overly complicated once I considered a subsequent rollback observer to restore the original observer when the deletion is not kept, as well as keeping straight nested transactions (since we only make the users once the process-level context is about to commit).

nnutter:
This whole topic is starting to frustrate me, I'm checking out.

PR #481, opened 2015-02-16
Sample Import: remove _created_objects method

ebelter:


tmooney:
+1 :department_store:

apregier:
+1

PR #482, opened 2015-02-16
Respect Config when assigning all "Compatible Instrument Data"

tmooney:
It's still a bit weird in that the model has to do the work itself to find out about matches, but it works in more situations--if someone runs the `genome instrument-data assign all-compatible` command by hand, it should still respect the Analysis Project.

I'd like to delete the `else` case in `compatible_instrument_data` entirely (and replace the `warning_message` with an `error_message`), but I'm not sure if anyone's still using it who would be upset at its demise.

This supersedes #477.


acoffman:
Yeah, as you've noted, I don't like that models are reaching into APConfig land now. I liked it when the knowledge was only unidirectional (excepting that single accessor). However, this seems like a reasonable stopgap until we can revisit this.

+1 :8ball:

nnutter:
Is there test coverage for this?  I don't see anything in `AllCompatible.t` like was removed with `AnalysisProject.t` but maybe `ConfigureQueuedInstrumentData.t` covers it?

lib/perl/Genome/ModelDeprecated.pm
```diff
@@ -631,38 +631,53 @@ sub _resolve_type_name_for_class {
...
-                   sample_id => \@sample_ids,
-               );
+    if($self->analysis_project) {
+        my $config_item = $self->analysis_project_bridge->config_profile_item;
+        unless($config_item) {
+            die $self->error_message('Configuration missing from analysis project association.  Cannot determine compatible instrument data for model %s and analysis project %s.', $self->__display_name__, $self->analysis_project->__display_name__);
+        }
+
+        my $rmm = Genome::Config::Translator->get_rule_model_map_from_config($config_item);
+        my @project_instrument_data = $self->analysis_project->instrument_data;
```
nnutter: Should this use an iterator since some projects will have thousands of instrument data associated with them?

tmooney: It's an interesting trade-off.  The most likely scenario is that we'll be processing many instrument-data sequentially from the same project, and so we'd end up loading them repeatedly.  But since it's likely we will have several large projects at once I think you're right that we shouldn't blithely load them all like this.

Would it be worthwhile to try to turn the rules into a BoolExpr and limit the amount of instrument data we're loading in the first place?

tmooney:
It is now being called within the CQID test, but it would probably be better to incorporate the deleted test as a case for the altered AllCompatible command.  I'll add that!

tmooney:
Test updated.  It's almost the same as the old one, except I added some additional instrument data that should not match.  They are in the AnalysisProject, but do not match the configuration.

lib/perl/Genome/ModelDeprecated.pm
nnutter: I don't think converting the rule into the BoolExpr wouldn't solve the trade-off problem you've brought up but it would obviously be more performant than doing the matching client side.

CQID could be using `AutoUnloadPool` around `_process_models` or something.  Do we not have memory problems with CQID now?  It looks like the old code would load everything as well?

nnutter:
:+1: Any memory issue there may be doesn't look like a regression but may still be an issue.

tmooney:
Sounds good--I'll merge this since many are eager for this feature to be available.  @acoffman and I were discussing a few different options just now for making this better; we can follow up with another PR.

PR #483, opened 2015-02-17
Relax requirements for scheduling reference alignment reports.

tmooney:
It has been requested that we still run reports (particularly MapCheck) even if no variant detection is performed.


ebelter:
+1

acoffman:
+1 :feelsgood:

PR #484, opened 2015-02-17
Make vep wrapper more robust

apregier:
A couple tweaks that fix problems I encountered while trying to run a Trio process.

### DIFF
--- a/lib/perl/Genome/Db/Ensembl/Command/Run/Base.pm
+++ b/lib/perl/Genome/Db/Ensembl/Command/Run/Base.pm
@@ -382,7 +382,6 @@ sub api {
     my $self = shift;
     return Genome::Db::Ensembl::Api->get_or_create(
         version => $self->ensembl_version,
-        test_name => $ENV{GENOME_SOFTWARE_RESULT_TEST_NAME},
         users => $self->_result_users,
     );
 }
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Db/Ensembl/Command/Run/Base.pm
```diff
@@ -382,7 +382,6 @@ sub api {
     my $self = shift;
     return Genome::Db::Ensembl::Api->get_or_create(
         version => $self->ensembl_version,
-        test_name => $ENV{GENOME_SOFTWARE_RESULT_TEST_NAME},
```
tmooney: Would the `GENOME_ALIGNER_INDEX_TEST_NAME` variable be useful here?  That's stretching the name of it a bit, but that's the variable for foundational results that depend on only things like the reference.

lib/perl/Genome/Db/Ensembl/Command/Run/Base.pm
```diff
@@ -756,8 +754,7 @@ sub cache {
             $cache_result_params{sift} = 0;
         }
         $cache_result_params{users} = $self->_result_users;
-        eval {$cache_result = Genome::Db::Ensembl::VepCache->get_or_create(%cache_result_params);
-        };
+        $cache_result = Genome::Db::Ensembl::VepCache->get_or_create(%cache_result_params);
```
tmooney: I'm quite glad this `eval` is gone! :smile:

lib/perl/Genome/Db/Ensembl/Command/Run/Base.pm
apregier: That sounds good to me!   Amendment pushed.

tmooney:
+1 :musical_note:

ebelter:
+1 :ghost:

PR #485, opened 2015-02-17
Sample Importer Attribute Cleanup

ebelter:
The PR mostly addresses the elimination of the passing around of entity attributes. Also removed is the updating of attributes of existing samples. WIP - We are still passing the library extension around, but that will be it's own PR.

### DIFF
--- a/lib/perl/Genome/Sample/Command/Import/Base.pm
+++ b/lib/perl/Genome/Sample/Command/Import/Base.pm
@@ -76,9 +76,7 @@ sub execute {
     $resolve_attrs_ok = $self->_resolve_library_attributes;
     return if not $resolve_attrs_ok;
 
-    my $import = $self->_import(
-        taxon => $self->taxon_name,
-    );
+    my $import = $self->_import;
     return if not $import;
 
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Sample/Command/Import/Base.pm
```diff
@@ -44,10 +44,13 @@ class Genome::Sample::Command::Import::Base {
         # source
         _individual => { is => 'Genome::Individual', is_optional => 1, },
         _individual_name => { is => 'Text', },
+        _individual_attributes => { is => 'Hash', },
```
tmooney: I think we more commonly use `HASH`, since that matches `ref ($cmd->_individual_attributes)`.

ebelter:
UR should accept both. Updated.

tmooney:
I agree; I believe it does currently work either way.  (It gets converted to `UR::Value::HASH` internally.)

+1 :barber:

acoffman:
+1 :palm_tree:

ebelter:
Thanks! Random emoji: :nail_care:

PR #486, opened 2015-02-17
Limit MergeReports to only two files

susannasiebert:
Previously MergeReports and its software result MergedReport took a list of reports to merge. One had to specify a main report via use_header_from to designate which report's header to use for the merged report. Now the interface designates a base_report, whose header will be used, and a supplemental_report that will be appended to the base_report. See https://jira.gsc.wustl.edu/browse/AT-678


lib/perl/Genome/VariantReporting/Command/CreateMergedReports.pm
```diff
@@ -150,46 +150,33 @@ sub connect_merge_operations {
...
+                    destination_property => 'base_report',
+                );
+
+                $dag->create_link(
+                    source => $snvs_dag,
+                    source_property => $output_name,
+                    destination => $merge_op,
+                    destination_property => 'supplemental_report',
                 );
             } else {
```
tmooney: The redundant link declarations here could be eliminated by determining a `$base_report_source` and `$supplemental_report_source` and passing them to the two links (or, alternatively, by a variable controlling the `destination_property` for each case).

tmooney:
+1 :mans_shoe:

lib/perl/Genome/VariantReporting/Framework/Component/Report/MergedReport.pm
```diff
@@ -277,16 +276,14 @@ sub can_be_merged {
 
 sub merge_parameters {
     my $self = shift;
-    my @report_results = $self->report_results;
-    my $result_class = $report_results[0]->class;
+    my $result_class = $self->base_report->class;
     return $result_class->merge_parameters;
```
apregier: I probably wrote this originally, but I don't remember - is there a reason why we have to call merge_parameters on the class and don't just cal $self->base_report->merge_parameters?

apregier:
+1

susannasiebert:
@tmooney @apregier I pushed a few commits to address your comments. I also made another passthrough on MergedReport to eliminate the need for reports_with_size

lib/perl/Genome/VariantReporting/Framework/Component/Report/MergeCompatible.pm
```diff
@@ -23,4 +23,10 @@ sub report_path {
         $self->class);
 }
 
+sub has_size {
+    my $self = shift;
+
+    return -s $self->report_path;
+}
```
tmooney: Is it worth memoizing this to avoid stat-ing the file repeatedly?

susannasiebert: Possibly. Although I though we should try to avoid memoizing with/on an object. I might just makes this a transient attribute.

susannasiebert: I pushed a commit to make has_size a transient attribute. I'm not 100% sure that the implementation caches successfully when the file doesn't have size since I believe that -s returns undef in this case. Suggestions are definitely welcome.

tmooney: You could make it `-s $self->report_path || 0` so that it is `defined`.  (Or, if you prefer, something like `-s $self->report_path? 1 : 0` to normalize all values.)

Another form of memoization without using `Memoize` that has been used here is [this](https://github.com/genome/genome/blob/6d979af90a3558c93447477f46e950122aa21bc0/lib/perl/Genome/FeatureList/Command/Import.pm#L129-L137), where a particular hash key on `$self` is checked and set explicitly.

PR #487, opened 2015-02-17
Use a Test Factory Taxon in View Tests

tmooney:
Although "mouse" was a less poor choice than "human", this test will run much more quickly with only five individuals.


ebelter:
+1 :pouting_cat:   I would prefer smelly cat, but null cat will do.

acoffman:
+1 :neckbeard:

PR #488, opened 2015-02-17
update gmt joinx vcf-report to check versions properly

tabbott:
it was using floating point comparisons for the 'minimum version' check.
the base class already had a "check_minimum_version" method that does
the right thing. added a test too.


tabbott:
Relevant RT ticket: 104174

tmooney:
+1 :station:

gsanders:
+1

PR #489, opened 2015-02-17
Sample Import: Library Handling Update

ebelter:
WIP. Update the library handling in the sample importer including making the library extension a command property.

### DIFF
--- a/lib/perl/Genome/Sample/Command/Import/Base.pm
+++ b/lib/perl/Genome/Sample/Command/Import/Base.pm
@@ -37,6 +37,12 @@ class Genome::Sample::Command::Import::Base {
             is_many => 1,
             doc => 'Additional attributes to add to the library. Give as key value pairs. Separate key and value with an equals (=) and pairs with a comma (,). Ex: attr1=val1,attr2=val2',
         },
+        library_extension => {
+            is => 'Text',
+            default_value => 'extlibs',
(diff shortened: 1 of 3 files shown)


apregier:
Do you intend to keep the library creation in the sample importer?  It seems to me like it might make more sense to have a library importer that can create individuals and samples as needed.  That way if multiple libraries are needed for the same sample they can all be defined.

ebelter:
I would prefer this to be a library importer, and just have 'sample create'. I am up for moving it to that space. This is a WIP and I am trying to space the commits so they can be reviewed.

lib/perl/Genome/Sample/Command/Import/Base.pm
```diff
@@ -37,6 +37,12 @@ class Genome::Sample::Command::Import::Base {
             is_many => 1,
             doc => 'Additional attributes to add to the library. Give as key value pairs. Separate key and value with an equals (=) and pairs with a comma (,). Ex: attr1=val1,attr2=val2',
         },
+        library_extension => {
+            is => 'Text',
+            default_value => 'extlibs',
+            valid_values => [qw/ extlibs microarraylib /],
+            doc => 'The extension to add to the sample name to create teh library name.',
```
tmooney: The `e` and `h` in `the` are transposed.

tmooney:
+1 :computer:

PR #491, opened 2015-02-18
Save the End of the Log when htseq-count Fails

tmooney:
The HTSeq::Count result saves the stderr of the command to a file within the result's directory.  If the command fails the result is not saved and so the log is lost.  To help debug failures, report the last 100 lines of that file as a debug message.

### DIFF
--- a/lib/perl/Genome/Model/Tools/Htseq/Count/Result.pm
+++ b/lib/perl/Genome/Model/Tools/Htseq/Count/Result.pm
@@ -298,13 +298,13 @@ sub _run_htseq_count {
             . " --idattr ${type}_id"
             . " -"
             . " '$gtf_file'"
-            . " 1>'$output_dir/${type}-counts.tsv'"
-            . " 2>'$err_file'";
+            . " 1>'$output_dir/${type}-counts.tsv'";
 
         try {
             Genome::Sys->shellcmd(
                 cmd => $cmd,
                 input_files => [$sorted_bam, $gtf_file],
+                redirect_stderr => $err_file,
             );
         } catch {
             my $error = $_;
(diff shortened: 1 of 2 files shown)


nnutter:
:+1:

This could be implemented in `Genome::Sys->shellcmd()`.  That thing's already a beast though so I'm not sure I would.

brummett:
What do you think about saving the last N bytes instead of the last M lines?  It'd be more efficient to just seek to N bytes from the end and read from there.

I can't suggest a good N without seeing the file, though.

tmooney:
@nnutter That could be useful.  `redirect_stderr` is only used in two places, so it wouldn't affect that much for now.  It does seem like this should use that instead of `2>` regardless!

@brummett I've symlinked an example file in my home directory called `example_htseq.err`.  I could go with 14000 bytes and get approximately the same result for that particular file.  If this gets moved into `Genome::Sys` it would become more important to set a byte threshold in case something out there produces very long lines.

brummett:
+1

I don't think using a byte offset is necessary to get this merged in, just a suggestion.

nnutter:
If we just stopped redirecting stderr would Workflow/Flow/Ptero end up redirect it to a "reasonable" log file?

tmooney:
Yes--that would eliminate the need for any of these shenanigans!

I think it was originally redirected because some invocations write a warning to the log for each read.  (My example log file was about 140 MB.)

nnutter:
Can we just filter those out with a pipe on the command?

nnutter:
If we pipe to filter then that feels like the same "shenanigans" just in a different form and I don't think it's necessarily better.  You could keep your Perl-based version but filter out the known-problematic message instead of trying to solve a more general problem.  You could also do that in a `finally` block instead of a `catch` so the logs are in the Workflow/Flow/Ptero logs on both success and failure.

Any changes are up to you, I still :+1: as is.

jasonwalker80:
Aside from this PR, we could update the version of HTseq and the tool would no longer produce 140MB log files.

iferguson90:
:+1: :shaved_ice: 

As far as last N lines versus last M bytes, if we decide to use the last M bytes it should either seek forward to the first newline within the last M bytes, or indicate that it's likely starting in the middle of a line.

PR #492, opened 2015-02-18
Don't try to add users to deleted results.

tmooney:
Hot on the heels of #479 (which should have been called "Don't try to add deleted users to results"), here's a commit that does the other thing.

[I've run align-reads from reference alignment with this commit, and it succeeds at committing this time.]


brummett:
+1 when tests pass

mkiwala:
+1

PR #493, opened 2015-02-19
Sample Import

ebelter:
The last bit of clean up before removing the nomenclature commands and moving the command to the library name space.

### DIFF
--- a/lib/perl/Genome/Sample/Command/Import/Base.pm
+++ b/lib/perl/Genome/Sample/Command/Import/Base.pm
@@ -91,7 +91,7 @@ sub execute {
     my $import = $self->_import;
     return if not $import;
 
-    $library = $self->_get_or_create_library;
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Sample/Command/Import/Base.pm
```diff
@@ -258,9 +248,11 @@ sub _create_individual {
     return $self->_individual($individual);
 }
 
-sub _create_sample {
+sub _create_sample_if_needed {
     my $self = shift;
 
+    return 1 if $self->_sample;
```
nnutter: Seems like the initial setting of `_sample`, currently in `_does_library_already_exist`, could be moved here?  If so then `_does_library_already_exist` is basically `_get_library` so `_does_library_already_exist` could be removed.

PR #495, opened 2015-02-19
Prepare for process integration tests

apregier:
The next step will be to create a jenkins script that runs the commands in the IntegrationTests yaml file, then compares the process outputs to a blessed process in the blessed yaml file using `genome process diff-blessed`.


susannasiebert:
+1 once tests pass. 

I'm a bit concerned about the outputs of compare_output of builds vs processes. I tried to make the process output match the build output but there might be some tweaking needed to make the diff output or processes helpful. It's beyond the scope of this PR, of course, but it's something we should have an eye on once this goes live.

apregier:
I'm hoping that eventually builds will "have" a process and most of the compare_output method for build will go away in favor of just comparing the processes.

lib/perl/Genome/Interfaces/Comparable/Command/DiffBlessed.pm
```diff
@@ -0,0 +1,57 @@
...
+
+use YAML;
+
+class Genome::Interfaces::Comparable::Command::DiffBlessed {
+    is => 'Genome::Interfaces::Comparable::Command::Diff',
+    is_abstract => 1,
+};
+
+sub source_file {
+    die "Must implement method source_file";
```
tmooney: Would it be sufficient to use `$self->__meta__->module_path` here?

apregier:
Thanks @tmooney , that is much better!

tmooney:
+1 :mountain_railway:

It might be nice to have tests for some of the new commands, if only to ensure they run to completion.  (`Process.t` covers the different possibilities for actual diffs pretty well.)

PR #497, opened 2015-02-19
Reserve tmp since our output is written there before being allocated.

tmooney:
This process was filling /tmp on one of our blades but other jobs would still be scheduled there.


ebelter:
+1 That's a lot of tmp!

tmooney:
It is!  These files can be large, since they write information about every position in the ROI.

nnutter:
:+1:

iferguson90:
:+1: :boar:

PR #498, opened 2015-02-21
Remove Sample Importer

ebelter:
This PR removes the sample importer. Coming will be a PR with the 'new' library importer.


malachig:
Hi @ebelter, wasn't there a `genome sample create` before that @mkiwala  (@mkiwala-g) created for use in the sGMS?  Does the current code supersede that?  To get custom data into the sGMS we follow this tutorial:

https://github.com/genome/gms/wiki/Beginner%27s-Guide-to-Data-Import

ebelter:
These import commands are for specific namespaces/source: TCGA, METAHIT etc. The idea with these commands was to assert naming conventions and other attributes. As it turned out, these are difficult maintain as there are always new sources coming in.

We are moving the essence of this importer to the library name space.  There it will still create all the entities needed, but be in a simpler format. This PR will not be mergd until the library importer is ready and approved.

The 'genome sample create' command does not exist in TGI GMS. I had considered turning it 'on' so it could be used. I can do that if you'd like.

apregier:
Maybe it would make sense to just require all of the entities to be created separately (like in the sGMS tutorial)?

ebelter:
@apregier that makes perfect sense. The reason these exist is for convenience. The goal here is to allow others to do this in an easy way. If you think that is with completing a command for each entity, then I am for it. There would need to be come checks added here and there, but it would be more compartmentalized. Each subject type would then have control over  it's creating. We can talk tomorrow.

ebelter:
A new importer helper has been implemented with documentation that uses a csv file to create commands to create the individuals, samples and libraries needed to run an import.

PR #500, opened 2015-02-23
A Command for Reprocessing an Analysis Project

tmooney:
If a user forgets to use the `--reprocess-existing` option on one of the other commands (or they'd prefer to wait before reprocessing), this command can be used to the same effect.


acoffman:
+1, looks good. I like adding the `reschedule` method as it limits the number of things that need to know about bridge statuses.

jweible:
:+1:

PR #501, opened 2015-02-23
fix for when GetTerminalSize is not supported

nnutter:
This fixes an issue I noticed when build tests time out or fail.  We get
these warnings:

```
00:34:47 Unable to get Terminal Size. The TIOCGWINSZ ioctl didn't work. The COLUMNS and LINES environment variables didn't work. The resize program didn't work. at /usr/lib/perl5/Term/ReadKey.pm line 362.
00:34:47 Use of uninitialized value $screen_width in subtraction (-) at /gscmnt/gc7008/info/jenkins/workspace/Genome-Long-Model-Tests/lib/perl/Genome/Command/WorkflowMixin.pm line 526.
00:34:47 Use of uninitialized value $screen_width in subtraction (-) at /gscmnt/gc7008/info/jenkins/workspace/Genome-Long-Model-Tests/lib/perl/Genome/Command/WorkflowMixin.pm line 527.
```

### DIFF
--- a/lib/perl/Genome/Command/Viewer.pm
+++ b/lib/perl/Genome/Command/Viewer.pm
@@ -18,11 +18,9 @@ sub write_report {
 }
 
 sub get_terminal_width {
-    my $screen_width = 80;
-
-    # this can fail in cases where no terminal is queriable
-    eval {($screen_width) = GetTerminalSize();};
-
+    # GetTerminalSize() returns an empty array if unsupported, e.g. in apipe-ci logs.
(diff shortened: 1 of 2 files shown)


tmooney:
+1 :aerial_tramway: Those warnings in the logs were annoying!

brummett:
+1

The docs for GetTerminalSize() imply it won't die, so the ```eval``` was unnecessary.  A quick look over its code confirms it.

PR #502, opened 2015-02-23
Picard refactoring

tabbott:
I wanted the ability to ask picard tools "what is the actual command you are going to run?". This PR is what happened.

I noticed that most of these tools were simply mapping their UR class input parameters to picard parameters in an ad hoc way, so I added a new param attribute 'picard_param_name' to automate this. A new base class (gmt::picard::base) examines these attributes and builds the necessary part of the command line. This actually fixes at least one bug that led to command lines like:

`java ... foo.jar/class ... PARAMS > output_file MORE_PARAMS`

I also cleaned up some of the version comparison logic. It was using floating point comparisons which is bad enough, but was made even worse by the fact that some of the revisions are svn-like (e.g., r436). A unit test was added to enforce the comment that versions in the version list maintained in the code are kept in decreasing order of recency.

Not all commands have been updated to use the new features (only the ones I really need to use right now for the /real/ project I'm working on). With the exception of the GC bias metrics tool, the interface to the modified commands has not changed. The GC bias tool had options to sort the bam and run the CleanSam tool to clip reads that straddle a sequence boundary. The only reference to these options was in BamQC which explicitly disabled them. These options have been removed.


lib/perl/Genome/Model/Tools/Picard/Base.pm
```diff
@@ -0,0 +1,169 @@
...
+# there was a bug that led to some commands being built like:
+#
+#   java ... foo.jar class ARGS > output_file MORE_ARGS
+#
+# that's bad. this sort of conflicts with the idea of log_file, but commands
+# like "BamIndexStats" that have no native OUTPUT option (i.e., can only spew
+# to stdout) support the existence of this function.
+sub _redirects {
+    my $self = shift;
+    return sprintf(">> %s", $self->log_file) if $self->log_file;
```
tmooney: This could also be accomplished by specifying the `redirect_stdout` option in `_shellcmd_extra_params`.

lib/perl/Genome/Model/Tools/Picard.pm
```diff
@@ -104,12 +131,12 @@ sub help_brief {
...
-gmt picard ...    
+gmt picard ...
 EOS
 }
 
-sub help_detail {                           
-    return <<EOS 
+sub help_detail {
+    return <<EOS
 More information about the Picard suite of tools can be found at http://picard.sourceforge.net/.
```
tmooney: FWIW <http://broadinstitute.github.io/picard/> is the current address.

lib/perl/Genome/Model/Tools/Picard/Base.pm
tabbott: This might be a naming issue. I am not a huge fan of the redirection code in gsys::shellcmd (for example, the business about adding a trailing newline will absolutely break some binary formats). What I really meant by _redirects is "stuff that absolutely comes after the actual command (pipes or redirects [<, >] to potentially arbitrary file descriptors etc)". I wonder what a better name would be...

lib/perl/Genome/Model/Tools/Picard.pm
tabbott: Oh my. I didn't mess with any of the docs. Let me see if it is easy to transform what exists into the actual urls...

lib/perl/Genome/Model/Tools/Picard/Base.pm
tmooney: Yeah, that extra newline is a bummer.  It stopped me from using `redirect_stdout` recently.  The name `_redirects` doesn't seem so bad, since that's what it's currently used for.

tmooney:
+1 :moneybag:

Hopefully it won't be too long before the rest of the tools are migrated to the new base!

lib/perl/Genome/Model/Tools/Picard.pm
tmooney: We could always follow it up with another PR later.

tabbott: Actually, the heuristic works for all but 4 of the existing commands, I'll take a crack at it now.

tabbott: just pushed an update to all http:// strings in the directory. they should all be 'better' now.

lib/perl/Genome/Model/Tools/Picard.t
```diff
@@ -61,4 +62,52 @@ for (0..6) {
...
+        ok($pkg->version_compare($v->[0], $v->[1]) < 0, "$v->[0] < $v->[1]");
+        ok($pkg->version_compare($v->[1], $v->[0]) > 0, "$v->[1] > $v->[0]");
+        ok($pkg->version_compare($v->[0], $v->[0]) == 0, "$v->[0] == $v->[0]");
+        ok($pkg->version_compare($v->[1], $v->[1]) == 0, "$v->[1] == $v->[1]");
+    }
+};
+
+subtest "Enforce minimum version" => sub {
+    # These are ordered from greatest to least
+    my @versions = $pkg->_versions_serial;
```
nnutter: I feel like this could be inlined since it's only used here and elsewhere one should probably use `available_picard_versions`.

tabbott: By inlining do you mean spelling out the test cases explicitly?

nnutter: I mean [Inline Method][1].

[1]: http://refactoring.com/catalog/inlineMethod.html

tabbott: Inlining means a specific thing to us old C/C++ people :)

I'm still not 100% clear on what you're after? Do you mean inlining the body of the for loop?

nnutter: Not the compiler optimization... :-p

I mean Inline Method on `_versions_serial` because it looks a lot like `available_picard_versions`.

tabbott: Now I'm even more confused, can you send a PR to my PR?

tabbott: To clarify, the only reason _versions_serial exists is to get at the even indexed elements of the PICARD_VERSIONS array for this specific test case (to make sure they are in fact in the order we want).

nnutter: I forgot about having to export `@PICARD_VERSIONS`.  I think leaving it as is is better than exporting that.

apregier:
I created AT-757 to track converting the remaining picard commands to the new base.

tabbott:
@nnutter any more thoughts on this?

nnutter:
:+1: from me.

nnutter:
@tabbott, you should comment when you push new commits because people aren't notified otherwise.

dufeiyu:
+1   This picard refactoring makes code look much better and cleaner. Down the road, we can further consolidate some G::M::T::Sam::XXXX modules whose only function is to call G::M::T::Picard::XXXX.

PR #503, opened 2015-02-23
InstData Solexa: fix param read trimmer param parsing

ebelter:


ebelter:
CI, test this. [github timed out]

lib/perl/Genome/InstrumentData/Solexa.pm
```diff
@@ -745,14 +745,15 @@ sub _convert_trimmer_to_sx_commands {
...
-                #for compatibility with pre-SX far processing profiles
-                if($trimmer_name eq 'far') {
-                    next if $key eq 'format';
-                }
-
-                push @params, $key => $value;
+            my @tokens = split(/\s\-\-/, $trimmer_params);
+            $tokens[0] =~ s/^\-\-//;
+            for my $token ( @tokens ) {
+                my ($key, $value) = split(/\s/, $token, 2);
```
nnutter: You might be able to use [Text::ParseWords::shellwords][1] or similar instead of `split`.  For that matter, you might be able to use [Getopt::Long][2] to parse the whole thing.  If you want to use the OO interface you can use `shellwords` first; that's all `GetOptionsFromString` does.

[1]: http://perldoc.perl.org/Text/ParseWords.html
[2]: http://perldoc.perl.org/Getopt/Long.html#Object-oriented-interface

lib/perl/Genome/InstrumentData/Solexa.pm
```diff
@@ -745,14 +745,15 @@ sub _convert_trimmer_to_sx_commands {
                 return;
             }
         } elsif ( $trimmer_params =~ /--\S+?[= ]\S+/ ) {
-            while($trimmer_params =~ /--(\S+)?[= ](\S+)/g) {
```
nnutter: Is it okay that the old version supported `--foo bar` and `--foo=bar` but the new version only supports `--foo bar`?

apregier:
+1

tmooney:
+1 :rice_scene:

lib/perl/Genome/InstrumentData/Solexa.pm
```diff
@@ -764,9 +756,16 @@ sub _convert_trimmer_to_sx_commands {
...
+    my $i = 0;
+    while ( $i <= $#params ) {
         my $property_name = $params[$i];
+        $property_name =~ s/^\-+//;
         $property_name =~ s/\-/_/g;
+        # The far format param is not valid anymore.
+        if ( $property_name eq 'format' and $trimmer_name eq 'far' ) {
+            $i += 2;
+            next;
+        }
```
brummett: Did you really mean to silently skip it, or do you need to complain here?

ebelter: Silently skip. This is for the legacy far interface which needs a format passed to it. The format is determined in the command itself. Ideally, we could backfill processing profiles to remove the format param, and then remove it here.

ebelter: If we would agree to backfill, then I would love to remove this. It would not impact the output of the far command.

PR #504, opened 2015-02-24
Pass users to the "dependencies".

tmooney:
AlignerIndices do a strange thing where they do work even in the `get()` case (where the usual `WithNestedResults` hash is not provided (since it's not "nested"!)).  We thus need to pass along the users explicitly in this case.

(Alternatively `WithNestedResults` could fill in the hash in all cases, but external callers should still not rely on its presence.)

### DIFF
--- a/lib/perl/Genome/Model/Build/ReferenceSequence/AlignerIndex.pm
+++ b/lib/perl/Genome/Model/Build/ReferenceSequence/AlignerIndex.pm
@@ -108,10 +108,9 @@ sub _supports_multiple_reference {
 
 sub get_or_create {
     my $class = shift;
-
-    my @objects = $class->SUPER::get_or_create(@_);
-
     my %params = @_;
+
+    my @objects = $class->SUPER::get_or_create(%params);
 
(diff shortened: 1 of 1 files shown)


ebelter:
+1

lib/perl/Genome/Model/Build/ReferenceSequence/AlignerIndex.pm
```diff
@@ -111,9 +111,11 @@ sub get_or_create {
 
     my @objects = $class->SUPER::get_or_create(@_);
 
+    my %params = @_;
```
nnutter: It's bordering on nitpicking but is this here because the above is changing `@_`?  If not how about put this line as second line of function and pass `%params` to the `get_or_create`?

tmooney: Nope, it isn't changing `@_`.  I will switch it!

nnutter:
:+1:

P.S. I re-queued `apipe-ci/bats` because it timed out.

PR #505, opened 2015-02-24
Restrict AnalysisMenu Item Create

jweible:
AnalysisMenu::Item->create now requires a pull request for the menu item repo to be merged and the 'Update Analysis Menu Items' jenkins job to have run before this command can be used to create a menu item.  If the file being used for menu item creation has not be added this way, this command will fail.

### DIFF
--- a/lib/perl/Genome/Config/AnalysisMenu/Command/Item/Create.pm
+++ b/lib/perl/Genome/Config/AnalysisMenu/Command/Item/Create.pm
@@ -35,8 +35,7 @@ sub execute {
     unless($self->_validate_file($file_path)){
         die $self->error_message("$file_path is invalid");
     }
-    my $allocated_file_path = $self->_copy_file_to_allocation($file_path);
-    my $item = Genome::Config::AnalysisMenu::Item->create(name => $name, file_path => $allocated_file_path, description => $self->description);
+    my $item = Genome::Config::AnalysisMenu::Item->create(name => $name, file_path => $file_path, description => $self->description);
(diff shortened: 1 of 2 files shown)


acoffman:
Just so I'm clear on what's going on here:

* We're not actually enforcing a PR through any mechanism other than policy right? There's nothing stopping someone from just committing.

* The mechanism by which this 'enforces' that we go through the git repo is by checking that the file lives in the path specified by `GENOME_ANALYSIS_PROJECT_DEFAULTS` which points to the git repo on network disk, and has restricted write permissions. 

* This git repo on disk is updated automatically by a Jenkins task. 

* We've just removed the "copy to an allocation" functionality as its not longer needed.

Is that all correct? 
If so, is it acceptable that anyone can add a menu item from anywhere on network disk if they just set their own `GENOME_ANALYSIS_PROJECT_DEFAULTS` value? Maybe that's okay as we usually don't design these things to stand up against malicious users, but if the objective is to lock this down, it might bear consideration.

tmooney:
A malicious user could also edit the create command to allow their arbitrary file path (or run the object `create()` directly in a script...or open a database handle and insert a row) so this doesn't begin to be secure by that measure!

It might be nice if this was a system-level property that couldn't be overridden by users.  That seems like a much larger issue than this PR, though.

acoffman:
Yeah, that was my general thought as well. Cool.

+1 :cake:

iferguson90:
+1 :boar:

PR #507, opened 2015-02-24
New module upgrade, Genome::Model::Tools::Analysis::Coverage::CoveragePlot

gschang:
New GMT analysis module for coverage analysis developed and added,
in Genome::Model::Tools::Analysis::Coverage::CoveragePlot


chrisamiller:
We're going to close this for now while Gue Su works to remove some of the dependencies to his own custom libraries.

PR #510, opened 2015-02-25
Created methylation_ratio

ygindin:
This pull request adds a replacement for mehRatio.py -- a script that calculates DNA methylation levels from bisulfite-treated, BSMAP-aligned data. The new script, methylation_ratio.py, avoids the exuberant memory requirements associated with the old script. Also new is the ability to multi-thread the wok.

### DIFF
--- a/lib/perl/Genome/Model/Tools/Bsmap/MethylationRatio.pm
+++ b/lib/perl/Genome/Model/Tools/Bsmap/MethylationRatio.pm
@@ -29,7 +29,7 @@ class Genome::Model::Tools::Bsmap::MethylationRatio {
         },
         threads => {
             is => 'Text',
-            doc => 'Numer of threads (2 minimum: one for reading and one for writing; additional threads beyond 2 will be allocated for reading; recommended number is 5)',
+            doc => 'Number of threads (2 minimum: one for reading and one for writing; additional threads beyond 2 will be allocated for reading; recommended number is 5)',
             is_output => 0,
             is_input => 1,
             default_value => '2',
--- a/lib/perl/Genome/Model/Tools/Bsmap/MethylationRatio.pm
+++ b/lib/perl/Genome/Model/Tools/Bsmap/MethylationRatio.pm
@@ -9,75 +9,91 @@ use Genome::Utility::Text qw(sanitize_string_for_filesystem);
 use File::Spec;
 use File::Basename qw();
 
-my (undef, $dir) = File::Basename::fileparse(__FILE__);
-my $METHRATIO_COMMAND = File::Spec->join($dir, 'methylation_ratio.py');
-
+my ( undef, $dir ) = File::Basename::fileparse(__FILE__);
+my $METHRATIO_COMMAND = File::Spec->join( $dir, 'methylation_ratio.py' );
 
 class Genome::Model::Tools::Bsmap::MethylationRatio {
-    is => 'Command',
-    has => [
-        bam_file => {
-            is => 'Text',
-            doc => 'An indexed bam file',
-            is_input => 1,
-        },
-        output_file => {
-            is => 'Text',
-            doc => 'File name for methyl counts (a .bgz extension will be added automatically)',
-            is_output => 1,
-            is_input => 1,
-            default_value => 'methylation',
-        },
-        threads => {
-            is => 'Text',
-            doc => 'Number of threads (2 minimum: one for reading and one for writing; additional threads beyond 2 will be allocated for reading; recommended number is 5)',
-            is_output => 0,
-            is_input => 1,
-            default_value => '2',
-        },
-        output_directory => {
-            is => 'Text',
-            doc => 'Where to output the methyl counts',
-            is_output => 1,
-            is_input => 1,
-        },
+	is  => 'Command',
+	has => [
+		bam_file => {
+			is       => 'Text',
+			doc      => 'An indexed bam file',
+			is_input => 1,
+		},
+		output_file => {
+			is => 'Text',
+			doc =>
+'File name for methyl counts (a .bgz extension will be added automatically)',
+			is_output     => 1,
+			is_input      => 1,
+			default_value => 'methylation',
+		},
+		threads => {
+			is => 'Text',
+			doc =>
+'Number of threads (2 minimum: one for reading and one for writing; additional threads beyond 2 will be allocated for reading; recommended number is 5)',
+			is_output     => 0,
+			is_input      => 1,
+			default_value => '2',
+		},
+		output_directory => {
+			is        => 'Text',
+			doc       => 'Where to output the methyl counts',
+			is_output => 1,
+			is_input  => 1,
+		},
 		reference => {
-		  is => 'Genome::Model::Build::ReferenceSequence',
-		  is_input => 1,
-		  doc => 'Specify reference build',
-		};
-    ],
+			is       => 'Text',
+			is_input => 1,
+			doc =>
+'Provide a path to the genome fasta file. Alternatively, specify 36 or 37 for human build 36 and 37, respecitively',
+		},
+	],
 };
 
 sub _reference_fasta {
-  my ($self) = @_;
-  return $self->reference->cached_full_consensus_path('fa');
+
+	my ($self) = @_;
+
+	my %mapping = (
+		36 => 'NCBI-human-build36',
+		37 => 'GRCh37-lite-build37',
+	);
+
+	my $reference = $self->reference;
+
+	if ( exists $mapping{$reference} ) {
+		my $reference_build =
+		  Genome::Model::Build::ReferenceSequence->get(
+			name => $mapping{$reference} );
+		$reference = $reference_build->cached_full_consensus_path('fa');
+	}
+
+	return $reference;
 }
 
 sub _generate_command_line {
-    my ($self) = @_;
+	my ($self) = @_;
 
-    my @cmd = (
-        'python', $METHRATIO_COMMAND,
-        '-o', File::Spec->join($self->output_directory, $self->output_file),
-        '-d', $self->_reference_fasta,
-        '-t', $self->threads
-    );
-    
+	my @cmd = (
+		'python', $METHRATIO_COMMAND, '-o',
+		File::Spec->join( $self->output_directory, $self->output_file ),
+		'-d', $self->_reference_fasta, '-t', $self->threads
+	);
 
-    push @cmd, $self->bam_file;
+	push @cmd, $self->bam_file;
 
-    return join ' ', @cmd;
+	return join ' ', @cmd;
 }
 
 sub execute {
-    my ($self) = @_;
-    Genome::Sys->shellcmd(
-        cmd => $self->_generate_command_line,
-        input_files => [$self->bam_file, $self->_reference_fasta],
-    );
+	my ($self) = @_;
+	Genome::Sys->shellcmd(
+		cmd         => $self->_generate_command_line,
+		input_files => [ $self->bam_file, $self->_reference_fasta ],
+	);
 
-    return 1;
+	return 1;
 }
 
 1;
--- a/lib/perl/Genome/Model/Tools/Bsmap/MethylationRatio.pm
+++ b/lib/perl/Genome/Model/Tools/Bsmap/MethylationRatio.pm
@@ -7,8 +7,10 @@ use Genome;
 
 use Genome::Utility::Text qw(sanitize_string_for_filesystem);
 use File::Spec;
+use File::Basename qw();
 
-my $METHRATIO_COMMAND = File::Spec->join(getcwd, 'methylation_ratio.py');
+my (undef, $dir) = File::Basename::fileparse(__FILE__);
+my $METHRATIO_COMMAND = File::Spec->join($dir, 'methylation_ratio.py');
 
 class Genome::Model::Tools::Bsmap::MethylationRatio {
     is => 'Command',
--- a/lib/perl/Genome/Model/Tools/Bsmap/MethylationRatio.pm
+++ b/lib/perl/Genome/Model/Tools/Bsmap/MethylationRatio.pm
@@ -12,11 +12,6 @@ use File::Basename qw();
 my (undef, $dir) = File::Basename::fileparse(__FILE__);
 my $METHRATIO_COMMAND = File::Spec->join($dir, 'methylation_ratio.py');
 
-reference => {
-  is => 'Genome::Model::Build::ReferenceSequence',
-  is_input => 1,
-  doc => ...,
-};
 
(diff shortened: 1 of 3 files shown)


lib/perl/Genome/Model/Tools/Bsmap/MethylationRatio.pm
```diff
@@ -0,0 +1,99 @@
...
+
+use strict;
+use warnings;
+use Cwd;
+use Genome;
+
+use Genome::Utility::Text qw(sanitize_string_for_filesystem);
+use File::Spec;
+
+my $METHRATIO_COMMAND = File::Spec->join(getcwd, 'methylation_ratio.py');
```
tmooney: This path needs to be relative to the module's directory rather than the current working directory.  Something like 
```perl
use File::Basename qw();
my (undef, $dir) = File::Basename::fileparse(__FILE__);
my $METHRATIO_COMMAND = File::Spec->join($dir, 'methylation_ratio.py');
```

lib/perl/Genome/Model/Tools/Bsmap/MethylationRatio.pm
```diff
@@ -0,0 +1,99 @@
...
+        output_file => {
+            is => 'Text',
+            doc => 'File name for methyl counts (a .bgz extension will be added automatically)',
+            is_output => 1,
+            is_input => 1,
+            default_value => 'methylation',
+        },
+        threads => {
+            is => 'Text',
+            doc => 'Numer of threads (2 minimum: one for reading and one for writing; additional threads beyond 2 will be allocated for reading; recommended number is 5)',
```
tmooney: Typo: `Number` is missing its `b`.

lib/perl/Genome/Model/Tools/Bsmap/MethylationRatio.t
```diff
@@ -0,0 +1,88 @@
...
+use File::Slurp;
+use File::Compare;
+use List::AllUtils qw(max);
+use Test::More tests => 3;
+
+my $class = 'Genome::Model::Tools::Bsmap::MethylationRatio';
+use_ok($class);
+
+my $test_root = File::Spec->join($ENV{GENOME_TEST_INPUTS},
+    qw(Genome-Model-Tools-Bsmap-MethylationRatio v1));
```
tmooney: If you prefer, this could also be:
```perl
use Genome::Utility::Test;
my $test_root = Genome::Utility::Test->data_dir($class, 'v1');
```

lib/perl/Genome/Model/Tools/Bsmap/MethylationRatio.pm
```diff
@@ -0,0 +1,99 @@
...
+
+    my $reference = $self->reference;
+
+    if (exists $mapping{$reference}) {
+        my $reference_build = Genome::Model::Build::ReferenceSequence->get(
+            name => $mapping{$reference}
+        );
+        $reference = $reference_build->cached_full_consensus_path('fa');
+    }
+
```
ebelter: The reference should be specified in the class definition as a Genome::Model::Build::ReferenceSequence. This also has built in erro checking, ensuring that  the id/name given is valid.

reference => {
  is => 'Genome::Model::Build::ReferenceSequence',
 is_input => 1,
 doc => ...,
}

Then getting reference fasta is easier:

sub _reference_fasta {
  my ($self) = @_;
  return $self->reference->cached_full_consensus_path('fa');
}

lib/perl/Genome/Model/Tools/Bsmap/MethylationRatio.pm
```diff
@@ -0,0 +1,99 @@
...
+sub execute {
+    my ($self) = @_;
+    my $return = Genome::Sys->shellcmd(
+        cmd => $self->_generate_command_line,
+        input_files => [$self->bam_file, $self->_reference_fasta],
+    );
+
+    unless ($return) {
+        die $self->error_message('Failed to execute: returned %s', $return);
+    }
```
ebelter: Genome::Sys->shellcmd dies on failure. You will need to use Tiny::Try's try/catch to capture the error, or remove this block altogether.

lib/perl/Genome/Model/Tools/Bsmap/MethylationRatio.t
nnutter: This would probably be good to do rather than just developer preference.

lib/perl/Genome/Model/Tools/Bsmap/methylation_ratio.py
```diff
@@ -0,0 +1,181 @@
...
+import pdb
+from multiprocessing import Process, Queue
+
+def write_data(result_queue, out_file, work_units):
+    start = time.clock()
+    fout = open(out_file, 'w')
+    fout.write('chr\tpos\tratio\ttotal_C\tmethy_C\tunmeth_C\n')
+    tab = '\t'
+    new_line = '\n'
+    print 'writer opens shop. waiting for ', work_units, ' work units'
```
nnutter: Should the three `print` calls in `write_data` be `logging.info` instead?

lib/perl/Genome/Model/Tools/Bsmap/methylation_ratio.py
```diff
@@ -0,0 +1,181 @@
...
+    logging.basicConfig(level=logging.DEBUG,
+                        format='(%(threadName)-10s) %(message)s')
+
+    genome = pysam.FastaFile(fasta_path)
+    nuc_A = 'A'
+    nuc_T = 'T'
+    nuc_C = 'C'
+    nuc_G = 'G'
+    cg = ['C', 'G']
+    while True:
```
nnutter: Where does this terminate?  I don't see a `break`.  Is there another way to escape (I don't know Python well)?

lib/perl/Genome/Model/Tools/Bsmap/methylation_ratio.py
nnutter: I guess this applies throughout the script.  I guess there's no harm in printing to stdout since you are writing to a file.  However, it is often the case that the default output (when no file specified) would be stdout in which case these `print` calls would interfere.

lib/perl/Genome/Model/Tools/Bsmap/methylation_ratio.py
```diff
@@ -0,0 +1,181 @@
...
+    def get_context(self):
+        return self.context
+
+if __name__ == '__main__':
+     
+    do_debug = False
+    usage = "usage: %prog [options] BSMAP_MAPPING_FILES (sorted BAM files)"
+    parser = optparse.OptionParser(usage=usage)
+    parser.add_option("-o", "--out", dest="outfile", metavar="FILE", help="output file name. (required)", default="")
+    parser.add_option("-d", "--ref", dest="reffile", metavar="FILE", help="reference genome fasta file. (required)", default="")
```
nnutter: I'm personally a fan of required options, I liken then to named arguments, but it's worth noting,

> Lots of people want their programs to have “required options”. Think about it. If it’s required, then it’s not optional! If there is a piece of information that your program absolutely requires in order to run successfully, that’s what positional arguments are for.

Source: [optparse](https://docs.python.org/2/library/optparse.html)

That also gets back to my comment earlier of the default output usually being stdout.

lib/perl/Genome/Model/Tools/Bsmap/methylation_ratio.py
```diff
@@ -0,0 +1,181 @@
...
+        worker = Process(target=process_region_pileup, args=(i, infile[0], options.reffile, work_queue, write_queue,))
+        worker.daemon = True
+        worker.start()
+    
+    raw_file = tempfile.NamedTemporaryFile()
+    writer = Process(target=write_data, args=(write_queue, raw_file.name, work_units,))
+    writer.start()
+    writer.join()
+    print "Sorting and compressing reads..."
+    command = ''.join(['tail -n +2 ', raw_file.name, ' | sort -k1V -k2n | bgzip > ', options.outfile, '.bgz'])
```
nnutter: I don't think this or the command a few lines down will support an `outfile` with spaces in it.

lib/perl/Genome/Model/Tools/Bsmap/MethylationRatio.pm
nnutter: This would fix the test failure I think.

nnutter:
Thanks for making a pull request @ygindin!

This is a welcome message for to share that the purpose for code review is primarily to be a learning tool.  It’s a chance for developers of varying experiences to collaborate and offer constructive feedback to improve the correctness, clarity, and maintainability of code.  Anyone that has provided feedback is willing to help you understand why the feedback was suggested and to aid you in implementing it.  If you find yourself stuck trying to implement some suggested change please seek their assistance (email, chat, or face-to-face).

Please also be aware [documentation][1] is available to help get people up to speed so they can participate in others' pull requests and manage their own pull requests.  While GitHub does make it easier than ever to get involved, some working knowledge of `git` and GitHub can help make the tech disappear so you can focus on the people and their code.  We also have a [Coding Standards and Best Practices][2] page that is worth reviewing periodically.

[1]: https://github.com/genome/docs/wiki
[2]: https://github.com/genome/docs/wiki/Coding-Standards-and-Best-Practices

lib/perl/Genome/Model/Tools/Bsmap/MethylationRatio.pm
ygindin: Good catch!

lib/perl/Genome/Model/Tools/Bsmap/MethylationRatio.t
ygindin: Done.

lib/perl/Genome/Model/Tools/Bsmap/MethylationRatio.pm
ygindin: Done

lib/perl/Genome/Model/Tools/Bsmap/methylation_ratio.py
ygindin: I completely agree. Print statements are now removed.

lib/perl/Genome/Model/Tools/Bsmap/methylation_ratio.py
ygindin: This is launched as a daemon thread and exits when the main program terminates. In a previous version of the code, this thread would exit when the work queue becomes exhausted. However, queue.size() is not implemented for some operating systems.

lib/perl/Genome/Model/Tools/Bsmap/methylation_ratio.py
ygindin: This is a valid concern. However, the out_file option is specified as part of the command line options. An argument with a space in it will not be parsed correctly, unless enclosed with quotes or some such measure. This is true for almost any program that accepts command line options.

lib/perl/Genome/Model/Tools/Bsmap/methylation_ratio.py
ygindin: My preference for the command line _options_ is that they can be specified in whatever order the user prefers. It's also clearer to read back: command --infile=mooo --outfile=neigh versus command mooo neigh. At the same time, I see your point and it is not something that I have considered before. In the future, I'll try and lean more on positional arguments, leaving command line options to tune the program behaivor.

lib/perl/Genome/Model/Tools/Bsmap/methylation_ratio.py
nnutter: Thanks for the explanation.

lib/perl/Genome/Model/Tools/Bsmap/methylation_ratio.py
nnutter: But the shell strips the quotes so when you pass it on to a new shell command here you would need to re-add them or rather there is probably an API that allows you to pass a list/array so you don't have to worry about escaping.  In Perl, this would be [IPC::Run](https://metacpan.org/pod/IPC::Run).

What I would propose here is either support spaces or check if `out_file` has spaces early in the program and error if it does.

nnutter: (`Genome` is horrible about dealing with spaces so this could be considered a double standard.)

nnutter:
@ygindin, have you seen the latest test failure?  If you click on Details beside the `apipe-ci/quick-tests` status you'll be taken to our internal `apipe-ci` server where you can see the output from the failed test.

ygindin:
@nnutter, the latest test failures have to do with the old version of pysam installed. @mkiwala is working on remedying that soon.

mkiwala:
@ygindin pysam 0.7.8 should get installed next week. I understand that `methylation_ratio.py` was originally designed to work with pysam 0.8.1, so I expect that you'll push those changes before the tests can pass again.

ygindin:
Thank you @mkiwala!

ygindin:
I've implemented the changes that were suggested, with one notable exception: changing the type of reference sequence input parameter from `text` to `Genome::Model::Build::ReferenceSequence`. Doing so would likely prevent a user from using a custom path to a reference file.

lib/perl/Genome/Model/Tools/Bsmap/MethylationRatio.pm
```diff
@@ -0,0 +1,99 @@
...
+use File::Basename qw();
+
+my ( undef, $dir ) = File::Basename::fileparse(__FILE__);
+my $METHRATIO_COMMAND = File::Spec->join( $dir, 'methylation_ratio.py' );
+
+class Genome::Model::Tools::Bsmap::MethylationRatio {
+	is  => 'Command',
+	has => [
+		bam_file => {
+			is       => 'Text',
```
nnutter: Instead of `Text` you can use `FilePath` just as a form of metadata that these are intended to be actual paths.

lib/perl/Genome/Model/Tools/Bsmap/MethylationRatio.pm
```diff
@@ -0,0 +1,99 @@
...
+		output_file => {
+			is => 'Text',
+			doc =>
+'File name for methyl counts (a .bgz extension will be added automatically)',
+			is_output     => 1,
+			is_input      => 1,
+			default_value => 'methylation',
+		},
+		threads => {
+			is => 'Text',
```
nnutter: Here you could also use `Number` instead of `Text`; again, mostly it's just metadata.

nnutter:
:+1: though I think making the types on the command parameters as accurate as possible would be an improvement.

PR #511, opened 2015-02-25
improvements to genome-env-tests

nnutter:
First two commits are to reduce the number of ephemeral databases that get
created when running these tests.  The remaining commits are refactors to
deduplicate the testexecs.


tmooney:
+1 :rugby_football:

PR #512, opened 2015-02-25
Remove adaptors from the variant reporting framework

susannasiebert:
Originally the adaptor was implemented to take the inputs from the plan file and translate them, if necessary. This functionality can be implemented by the expert command so adaptors aren't needed. This PR removes adaptors and moves their functionality to Expert::Command as well as the expert's Run child-modules that inherit from Expert::Command.

Bam readcount still needs a very lightweight 'adaptor' to allow for the usage of parallel_by on an is_many input. There is not adaptor base class anymore so this adaptor is standalone.

See https://jira.gsc.wustl.edu/browse/AT-666 for more details.


susannasiebert:
@apregier It would be good if you can have a look at the show-docs changes. It seems like a lot of arguments are now showing up for the experts that didn't previously and I'm not sure if that is correct or not. We might need to tweak the show-docs framework a bit to get the correct output for this.

lib/perl/Genome/VariantReporting/Framework/Component/Expert/Command.pm
```diff
@@ -15,33 +16,114 @@ use Genome::VariantReporting::Framework::FileLookup qw(
...
     is_abstract => 1,
-    is => ['Genome::Command::DelegatesToResult', 'Genome::VariantReporting::Framework::Component::Base'],
-    has_input => [
+    is => ['Genome::Command::DelegatesToResult', 'Genome::VariantReporting::Framework::Component::WithTranslatedInputs'],
+    attributes_have => {
+        is_planned => {
+            is => "Boolean",
+            default => 0,
+        },
+        is_argument => {
```
apregier: Let's add a doc to this to explain that these are properties that get passed along to the software result

lib/perl/Genome/VariantReporting/Framework/Component/Expert/Command.pm
```diff
@@ -63,7 +145,7 @@ sub is_many_input_names {
     my $self = shift;
 
     my @properties = $self->__meta__->properties(
-        is_many => 1, is_input => 1);
+        is_many => 1, is_argument => 1);
```
apregier: maybe we could eliminate the need for the is_argument specification by looking at the properties on the result class instead and only getting those?

lib/perl/Genome/VariantReporting/Framework/Component/Expert/Command.pm
```diff
@@ -101,7 +185,7 @@ sub input_hash {
 
     $hash{test_name} = $ENV{GENOME_SOFTWARE_RESULT_TEST_NAME};
 
-    for my $key (qw(requestor user sponsor process_id label)) {
+    for my $key (qw(requestor user sponsor process_id label plan_json)) {
```
apregier: Do we still need to delete these?  It seems like since we are limiting to is_argument maybe they won't be getting put in the hash?

lib/perl/Genome/VariantReporting/Framework/Plan/ExpertPlan.pm
```diff
@@ -21,21 +21,22 @@ sub category {
 
 sub needed_translations {
     my $self = shift;
-    return Set::Scalar->new(map {$self->adaptor_params->{$_}}
-        $self->get_class->adaptor_class->required_translated_input_names);
+    my $needed = Set::Scalar->new(map {$self->run_params->{$_}} $self->get_class->run_class->required_translated_input_names);
+    $needed->insert(map {@{$self->run_params->{$_}}} $self->get_class->run_class->required_translated_is_many_input_names);
```
apregier: Since we had to fix this both here and in the plan base, what if we would fix required_translated_input_names to include both is_many and non-is_many properties?

lib/perl/Genome/VariantReporting/Framework/Plan/ExpertPlan.pm
```diff
@@ -50,7 +51,7 @@ sub __class_errors__ {
     my @errors = $self->SUPER::__class_errors__;
     return @errors if @errors; # Can't know anything else in this case
 
-    for my $accessor ('run_class', 'adaptor_class') {
+    for my $accessor ('run_class') {
```
apregier: Since we only have one item now, we can get rid of the for-loop :-)

lib/perl/Genome/VariantReporting/Suite/BamReadcount/Annotate.pm
```diff
@@ -5,16 +5,52 @@ use warnings FATAL => 'all';
 
 class Genome::VariantReporting::Suite::BamReadcount::Annotate {
     is => 'Genome::VariantReporting::Framework::Component::Expert::Command',
+    attributes_have => {
+        is_argument => {
```
apregier: Since this is now an Expert::Command we shouldn't need to define this attributes_have

lib/perl/Genome/VariantReporting/Suite/BamReadcount/Expert.pm
```diff
@@ -80,6 +77,31 @@ sub annotate_op {
...
+        command => 'Genome::VariantReporting::Suite::BamReadcount::Adaptor',
+    );
+}
+
+sub connected_adaptor_operation {
+    my ($self, $dag) = validate_pos(@_, 1, 1);
 
+    my $adaptor_operation = $self->adaptor_operation;
+    $dag->add_operation($adaptor_operation);
+    for my $name qw(plan_json) {
```
apregier: don't need the for loop for one item!

lib/perl/Genome/VariantReporting/Framework/Plan/ExpertPlan.pm
susannasiebert: We would loose a bit of parallelism with the other subroutines in WithTranslatedInputs. I might just add another subroutine in WithTranslatedInputs that returns both.

susannasiebert: I just realized that combining these doesn't make much sense because we need to handle them differently in terms of dereferencing the run_params call.

susannasiebert:
@apregier : I pushed a few commits to address your comments.

apregier:
+1 This looks great!  Thanks for cleaning up those things that we missed the first time around

lib/perl/Genome/VariantReporting/Framework/Component/Base.pm
```diff
@@ -50,7 +50,16 @@ sub properties_in_plan {
             implied_by => undef,
             is_structural => undef)) {
         next if $property->{is_param} && $property->property_name =~ m/_md5$/;
-        next if $property->{is_param} && $property->property_name =~ m/_count$/;
+        if ($property->{is_param} && $property->property_name =~ m/^(.*)_count$/) {
```
tmooney: Would it make sense for the `_md5` property to to also check for its related `_count` property?

susannasiebert: We thought it was unlikely that an expert would have a property that ends in _md5 so we didn't think it was necessary. I can add it in for completeness if you'd like.

davidlmorton:
+1 I was scared when I saw how many lines were added but it looks like it is mostly new tests, so that's grrrreat!

tmooney:
+1 :koala:

PR #513, opened 2015-02-25
updated to template with 5.10-reference-alignment-lane-qc data

nnutter:


tmooney:
```
$ test-db template list | grep bdeae22 | wc -l
1
```

:shipit:

PR #514, opened 2015-02-25
More SR User Locking Improvements

tmooney:
* Use the new `once => 1` for observers.
* Don't let lock names get too long by including the whole label.  (Another approach might be to `substr` the label instead.  That would preserve backwards compatibility with most existing locks, but it could lead to collisions that cause deadlocks.)

### DIFF
--- a/lib/perl/Genome/SoftwareResult/User.pm
+++ b/lib/perl/Genome/SoftwareResult/User.pm
@@ -167,14 +167,14 @@ sub _resolve_lock_name {
     my $class = shift;
     my $params = shift;
 
-    my $label = $params->{label};
-    if(length($label) >= 32) {
-        $label = Genome::Sys->md5sum_data($label);
+    my $lock_name = Genome::Utility::Text::sanitize_string_for_filesystem(
+        join('_', $params->{label}, $params->{user}->id, $params->{software_result}->id)
+    );
+    if (length($lock_name) > 255) {
+        $lock_name = Genome::Sys->md5sum_data($lock_name);
     }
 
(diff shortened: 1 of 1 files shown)


brummett:
You could use a combination of the md5 of the whole string, plus a substr of it.  That way, the source of the lock can still be tracked down if necessary.

For example:
```
my $shortened_label = substr($params->{label}, 0, 100);
join('_', Genome::Sys->md5sum_data($params->{label}), $params->{user}->id, $params->{software_result}->id);
```
I don't know if the beginning of the string is most significant, or the end, or if 100 chars is distinguishable enough.

tmooney:
I don't think it's important that the lock name be human-readable, so long as it is one-to-one with the objects it's protecting.  I believe SoftwareResults lock using their lookup hash, for example.

nnutter:
I feel like it would minimize collateral damage if you also lock the old way if it's less than 255 characters (or whatever the limit is).  Once the bad code is deprecated you could then remove the old way.

tmooney:
I've pushed a commit to only hash if the label was at least 32 characters.  That way we get the old lock names in the vast majority of cases (Contention for the locks are most likely for `uses`, `sponsor`, `shortcut`, and `created`.), but also have an upper bound on the lock length to rein in those labels that have ROI names in them.

apregier:
I kind of like Nathan's idea of doing both so that we can clean it up in the (nearish) future - but I would like to get this working soon!

tmooney:
I've added a commit that uses the old lock name if it fit within the length limit of 255 and uses the hashed version otherwise.  This should work with all existing locks!

apregier:
+1

nnutter:
:+1: This isn't what I would do but I think it's reasonable.

PR #515, opened 2015-02-26
Split entry_sources of MergeReports into base_report_source and supplemental_report_source

susannasiebert:
Now that MergeReports only support two report inputs and report sources has been split into base_report and supplemental_report we don't need entry_sources to be is_many anymore. Instead this PR splits it into a base_report_source and a supplemental_report_source. 

This is to finish up https://jira.gsc.wustl.edu/browse/AT-678.

### DIFF
--- a/lib/perl/Genome/VariantReporting/Framework/Component/Report/MergedReport.pm
+++ b/lib/perl/Genome/VariantReporting/Framework/Component/Report/MergedReport.pm
@@ -38,8 +38,11 @@ class Genome::VariantReporting::Framework::Component::Report::MergedReport {
         separator => {
             is => 'Text',
         },
-        entry_sources => {
-            is_many => 'Text',
+        base_report_source => {
+            is => 'Text',
+            is_optional => 1,
+        },
+        supplemental_report_source => {
             is => 'Text',
             is_optional => 1,
         },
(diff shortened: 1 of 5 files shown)


nnutter:
:eyes: Looks clean.  I can't easily judge implementation change.

lib/perl/Genome/VariantReporting/Framework/Component/Report/MergedReport.pm
```diff
@@ -365,8 +382,7 @@ sub get_header {
 
 sub has_entry_sources {
     my $self = shift;
-    my @entry_sources = $self->entry_sources;
-    return scalar(@entry_sources);
+    return ($self->base_report_source && $self->supplemental_report_source);
```
apregier: might want to say "defined" in case they are "0"

apregier:
+1

lib/perl/Genome/VariantReporting/Framework/Component/Report/MergedReport.pm
```diff
@@ -196,7 +200,15 @@ sub merge_files {
         } else {
             $file_to_merge = $report->report_path;
         }
-        my $with_source = $self->add_source($report->report_path, $file_to_merge);
+        my $report_source_accessor = $report_name . '_source';
+        my $report_source = $self->$report_source_accessor,
```
tmooney: I would end this line with a `;` instead of a `,`.

tmooney:
+1 :chart_with_upwards_trend:

PR #516, opened 2015-02-26
Process needs to be accessible by workflow step

apregier:
The process hasn't been committed yet, so shelling out for the
workflow steps will fail.  By setting NO_LSF, the workflow will
be run in the same context and the process will be accessible.


tmooney:
To have Perl restore the original value at the end of the scope: `local $ENV{NO_LSF} = 1;`.

tmooney:
+1 :green_apple: either way!

apregier:
Thanks, @tmooney I always forget about that!

acoffman:
+1 :pineapple:

PR #517, opened 2015-02-26
Disable auto_assign_inst_data as a feature for now.

tmooney:
The current implementation, while it does a better job of restricting data than before, is too slow for large projects like FinnSeq and may not always do the correct thing (since it filters only on the "rules" section of the configuration and not also the "instrument_data_properties" section).


nnutter:
:+1:

ebelter:
+1 :rabbit2:

tmooney:
I'm going to merge this early so we can turn CQID back on sooner.

PR #519, opened 2015-02-27
Bam readcount vcf encoding

davidlmorton:


tmooney:
+1 :custard:

ernfrid:
I'm in a hurry and merging. This is a simple change.

apregier:
Do we need to set test names on old results?

davidlmorton:
No, this should be a backwards compatible change since there will be no '|' characters to translate back into commas.  If those results are to be used with Joinx though they will need to be recomputed so in that sense it might be warranted.

PR #524, opened 2015-02-28
override default timeout when creating an ephemeral database

nnutter:
We've had these errors every once in awhile,

> Got error response 500: SSL read timeout:  at
> /usr/bin/test-db-database-create line 15

Previously we added the `--timeout` option but we didn't think it should be
necessary.  Now I am thinking it may be that the template locks while creating
the new database so contention is what is causing the timeouts. For now I am
just increasing the timeout as a workaround.  We'll need to see if this is
actually what's going on.

### DIFF
--- a/bin/genome-env
+++ b/bin/genome-env
@@ -399,7 +399,7 @@ function setup_test_db {
     fi
 
     echo "=> Creating test database..."
-    eval "$($TESTDB database create --bash --owner $TESTDBSERVER_DB_USER --based-on $DB_SNAPSHOT_NAME)"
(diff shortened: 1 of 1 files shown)


nnutter:
When [2-Genome-Model-Tests-2-Run-Models #1482](https://apipe-ci.gsc.wustl.edu/job/2-Genome-Model-Tests-2-Run-Models/1482/) started only one of the ~12 jobs didn't timeout.

nnutter:
Merging early because I'm pretty sure this is required to get model tests working with the new `genome-env`.

PR #525, opened 2015-03-01
New Genome::Model::Tools::Analysis::Coverage::CoveragePlot added

gschang:
I have updated the whole source codes according to Chris's comments. Module dependency is solved by re-writing subroutines for external module replacement.

### DIFF
--- a/lib/perl/Genome/Model/Tools/Analysis/Coverage/CoveragePlot.R
+++ b/lib/perl/Genome/Model/Tools/Analysis/Coverage/CoveragePlot.R
@@ -36,11 +36,6 @@ cs_scale <- function(data, breaks=5, range=NULL, fraction=0.0, order=FALSE)
     {
         return(p);
     }
-    
-    # updated on 2010-08-26, by using the "pretty" function
-    # computes the histgram of the given data
-    #h <- hist(data, breaks=breaks, plot=F);
-    #return(h$breaks);
 }
 
 
@@ -55,10 +50,6 @@ cs_scale_scope <- function(scale, factor=0.05)
     
     if (s[3]) return(c(r[2], r[1]))
     else return(c(r[1], r[2]));
-
-    # updated on 2010/09/22
-    #r <- range(scale);
-    #return (range(scale) + c(-factor, factor) * diff(range(scale)));
 }
 
 
@@ -77,7 +68,7 @@ cs_range <- function(x)
 
 
 
-cs_range_limit2 <- function(x, y, xlim=NULL, ylim=NULL)
+cs_range_limit_xy <- function(x, y, xlim=NULL, ylim=NULL)
 {
     if (is.null(xlim)) stop("xlim not defined");
     
@@ -87,10 +78,6 @@ cs_range_limit2 <- function(x, y, xlim=NULL, ylim=NULL)
     miny = min(y);
     maxy = max(y);
     
-    #print(c(minx, maxx, miny, maxy));   # for debug
-    #print(xlim);  # for debug
-    #print(ylim);  # for debug
-        
     # if all x and y data points are within the given range
     if (xlim[1] <= minx & maxx <= xlim[2])
     {
@@ -136,8 +123,6 @@ cs_range_limit2 <- function(x, y, xlim=NULL, ylim=NULL)
             }
         }
     }
-    
-    #print(data);    # for debug
     
     return(data);
 }
@@ -187,10 +172,6 @@ cs_vmargin <- function(x, visible=T, rot=0, label=FALSE, default=unit(1.5, "line
     
     # determines the margin where the X label is placed
     if (label) margin <- margin + unit(3, "strheight", "O");
-    
-    # for debug
-    #print(max);
-    #print(convertUnit(margin, "mm"));
     
     return (margin);
 }
@@ -242,10 +223,6 @@ cs_hmargin <- function(y, visible=T, rot=0, label=FALSE, default=unit(1.5, "line
     # determines the margin where the Y label is placed
     if (label) margin <- margin + unit(3, "strheight", "O");
     
-    # for debug
-    #print(max);
-    #print(convertUnit(margin, "mm"));
-    
     return (margin);
 }
 
@@ -270,11 +247,6 @@ cs_viewport2 <- function(left, bottom, xscale, xfactor=0.05, yscale, yfactor=0.0
 
         name = name
     );
-    
-    # for debug
-    #str(vp);
-    #print(paste("x=", vp$x));
-    #print(paste("y=", vp$y));
     
     return(vp);
 }
@@ -319,17 +291,6 @@ cs_base <- function(title=NULL, lines=2, xlab=NULL, xmargin=2, ylab=NULL, ymargi
     {
         # uses the given viewport
         cv <- vp;
-        
-        # the margins obtained from the given viewport
-        # The following margins do not work in an accurate way
-        #xmar = vp$x;
-        #ymar = vp$y;
-        
-        # for debug
-        #print(xmar);
-        #print(vp$x);
-        #print(ymar);
-        #print(vp$y);
     }
 
 
@@ -381,12 +342,6 @@ cs_base <- function(title=NULL, lines=2, xlab=NULL, xmargin=2, ylab=NULL, ymargi
          gp = gp,
          vp = parent
         );    # cs_base
-    
-    # for debug
-    #grid.draw(g);
-        
-    # for debug, prints out the summary information of the gTree
-    #sumg(g);
 
     return(g);
 }
@@ -491,9 +446,6 @@ grid.cs_vruler <- function(scale=NULL, breaks=NULL, type=c("top", "bottom"), inn
         if (type == "bottom") bottom = TRUE;
     }
     
-    #print(top);
-    #print(bottom);
-    
     
     # TODO: data custimization (optional)
     
@@ -539,8 +491,6 @@ grid.cs_vruler <- function(scale=NULL, breaks=NULL, type=c("top", "bottom"), inn
     i = 0;
     while(cur <= max(scale))
     {
-        #print(cur); # for debug
-        
         # create a major and minor mark
         if (i == 0)
         {
@@ -659,9 +609,6 @@ grid.cs_hruler <- function(scale=NULL, breaks=NULL, type=c("left", "right"), inn
         if (type == "left") left = TRUE;
     }
     
-    #print(right);
-    #print(left);
-    
     
     # TODO: data custimization (optional)
     
@@ -707,8 +654,6 @@ grid.cs_hruler <- function(scale=NULL, breaks=NULL, type=c("left", "right"), inn
     i = 0;
     while(cur <= max(scale))
     {
-        #print(cur); # for debug
-        
         # create a major and minor mark
         if (i == 0)
         {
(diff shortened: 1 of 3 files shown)


lib/perl/Genome/Model/Tools/Analysis/Coverage/CoveragePlot.pm
```diff
@@ -0,0 +1,1439 @@
...
+	    is => 'String',
+	    is_optional => 0,
+	    doc => 'Directory to save output files such as PDF file and tables for plotting, and ref-cov stats files',
+	},
+	
+	bam_files => {
+	    is => 'String',
+	    is_optional => 1,
+	    doc => 'Comma-separated list of bam files to generate coverage plots from, Must specify either --bam_files or --refalign_models',
+	},
```
susannasiebert: If you specify this (and other comma-separated properties) as `is_many => 1` you don't have to do any splitting up into an array during execution. UR will do that for you when you call `$self->bam_files`.

lib/perl/Genome/Model/Tools/Analysis/Coverage/CoveragePlot.pm
```diff
@@ -0,0 +1,1439 @@
...
+	
+	refalign_models => {
+	    is => 'Number',
+	    is_optional => 1,
+	    doc => 'Comma-separated list of refalign model IDs to generate coverage plots from, bam_files will override refalign_models if given',
+	},
```
susannasiebert: You should make this a `is => 'Genome::Model::ReferenceAlignment'`. This will automatically get the model object for each id for you and it will do some error checking that the id is indeed a reference alignment model. You don't need to do the retrieval and check on lines 188 and 189. 

This should also be a `is_many => 1`

lib/perl/Genome/Model/Tools/Analysis/Coverage/CoveragePlot.pm
```diff
@@ -0,0 +1,1439 @@
...
+    my $min_depth_filters = $self->min_depth_filters;
+    my $wingspan = $self->wingspan;
+    
+    my $bw = $self->bw;
+    $bw = "nrd0" if ($bw =~ /^\s*$/);
+    
+    
+    # as default
+    die "roi-file-path required" unless -e $roi_file_path;
+    print STDERR sprintf("%d region(s) of interest from %s\n", `wc -l $roi_file_path | cut -d " " -f 1`, $roi_file_path);
```
brummett: All these prints to STDERR should instead use the methods ```debug_message()```, ```status_message()```, ```warning_message()``` or ```error_message()``` depending on the severity of the message.  For example, this one could be:
```
$self->status_message("%d region(s) of interest from %s\n",
                      `wc -l $roi_file_path | cut -d " " -f 1`,
                      $roi_file_path);
```

brummett:
The ```create_hash()``` is actually a constructor for some new class and really belongs in its own package.  As it is, it is returning objects blessed into the ```Genome::Model::Tools::Analysis::Coverage::CoveragePlot``` package when they don't belong there.

Here are a couple of links that describes how to make a simple hash-based class in Perl:
http://stackoverflow.com/questions/1037312/how-do-i-create-a-perl-class
http://www.zentut.com/perl-tutorial/perl-oop/
All the methods that operate on those objects (like ```write()``` and ```from_file()```) need to be moved into the class' package as well.  It's also good practice to put each kind of thing in its own class.  For example, it looks like things you'd call ```from_file()``` on are a different kind of thing that those you'd call ```gradient_2colors()``` on

lib/perl/Genome/Model/Tools/Analysis/Coverage/CoveragePlot.pm
```diff
@@ -0,0 +1,1439 @@
...
+    
+    my $bam_files = $self->bam_files;
+    my $refalign_models = $self->refalign_models;
+    
+    my $labels = $self->labels;
+    my $min_depth_filters = $self->min_depth_filters;
+    my $wingspan = $self->wingspan;
+    
+    my $bw = $self->bw;
+    $bw = "nrd0" if ($bw =~ /^\s*$/);
```
dufeiyu: This "nrd0" should be set as default_value when making "bw" parameter (line 64). Also to me "band_width" seems to be a better parameter name than "bw".

gschang:
Thank you for all your nice comments. I will address and get back to you again next week.
--Gue Su

lib/perl/Genome/Model/Tools/Analysis/Coverage/CoveragePlot.R
```diff
@@ -0,0 +1,1549 @@
...
+    }
+    
+    #print(data);    # for debug
+    
+    return(data);
+}
+
+
+
+cs_range_limit3 <- function(x, y, z, xlim=NULL, ylim=NULL, zlim=NULL)
```
chrisamiller: Many of these function names could be made more descriptive. The difference between cs_range_limit2 and cs_range_limit3 is not at all apparent from the function name.

lib/perl/Genome/Model/Tools/Analysis/Coverage/CoveragePlot.R
```diff
@@ -0,0 +1,1549 @@
...
+        # the margins obtained from the given viewport
+        # The following margins do not work in an accurate way
+        #xmar = vp$x;
+        #ymar = vp$y;
+        
+        # for debug
+        #print(xmar);
+        #print(vp$x);
+        #print(ymar);
+        #print(vp$y);
```
chrisamiller: Lots of comments that could be removed. Debugging statements should be taken out. If the above chunk that's labeled "The following margins do not work in an accurate way" is old code that will not be used, just remove it.

lib/perl/Genome/Model/Tools/Analysis/Coverage/CoveragePlot.R
```diff
@@ -0,0 +1,1549 @@
...
+                if(inner) m <- linesGrob(x=unit(c(0, ssize), "npc"), y=unit(c(cur, cur), "native"), name=paste("b", cur, sep=""), gp=gp, vp=vp)
+                else m <- linesGrob(x=unit(c(-ssize, 0), "npc"), y=unit(c(cur, cur), "native"), name=paste("b", cur, sep=""), gp=gp, vp=vp);
+                
+                gcs <- addGrob(gcs, m);
+            }
+        }
+        
+        # creates a gridline
+        if (! is.null(lty))
+        {
```
chrisamiller: In general, the preferred TGI style is to do:

    if(statement){
        code
    }

with the open parentheses on the same line as the statement (makes for more compact code).  I'm not going to ask you to clean all of these up at this point, but it's something to note next time you write a script.

lib/perl/Genome/Model/Tools/Analysis/Coverage/CoveragePlot.R
```diff
@@ -0,0 +1,1549 @@
...
+    
+    return(g);
+}
+
+
+
+
+# --------------------------------------
+# gets the command line arguments when run by Rscript
+args <- commandArgs(trailingOnly=TRUE);
```
chrisamiller: Wouldn't be a bad idea to verify that you get the expected number of arguments here.

lib/perl/Genome/Model/Tools/Analysis/Coverage/CoveragePlot.R
gschang: Those comments were already cleaned up. I also realized some functions 
are unnecessary (never used) in the CoveragePlot code. So, I had updated 
the codes accordingly, but I don't know how to push them up onto the 
previous commitment (which you are now seeing). Let me know how to do by 
commands.


On 3/4/2015 11:57 AM, Chris Miller wrote:
>
> In lib/perl/Genome/Model/Tools/Analysis/Coverage/CoveragePlot.R 
> <https://github.com/genome/genome/pull/525#discussion_r25796352>:
>
> > +    }
> > +    else
> > +    {
> > +        # uses the given viewport
> > +        cv <- vp;
> > +
> > +        # the margins obtained from the given viewport
> > +        # The following margins do not work in an accurate way
> > +        #xmar = vp$x;
> > +        #ymar = vp$y;
> > +
> > +        # for debug
> > +        #print(xmar);
> > +        #print(vp$x);
> > +        #print(ymar);
> > +        #print(vp$y);
>
> Lots of comments that could be removed. Debugging statements should be 
> taken out. If the above chunk that's labeled "The following margins do 
> not work in an accurate way" is old code that will not be used, just 
> remove it.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/genome/genome/pull/525/files#r25796352>.
>



____

gschang:
Hi, all. I have updated the source code of CoveragePlot with your comments. Some didn't work and I am going to add individual messages here. Looking forward to your next comments. Thank you for your comments again.

lib/perl/Genome/Model/Tools/Analysis/Coverage/CoveragePlot.pm
gschang: Actually, this option does not work. It showed "ERROR: Unknown option: refalign-models". I also commented about this error message in the source code.

lib/perl/Genome/Model/Tools/Analysis/Coverage/CoveragePlot.pm
gschang: As you suggested in the comment of "create_hash()" below, apipe does not want to create a class namespace arbitrarily and does not allow individual developer to push their own class into Genome Model library. So, I extracted from my class library and pasted only necessary subroutines in this tool. You are right, but I took it as a practical solution that is not for object-oriented programming here.

lib/perl/Genome/Model/Tools/Analysis/Coverage/CoveragePlot.R
gschang: I have cleaned up the source codes. In fact, many unnecessary functions were pasted from my R grid library, which also makes a confusion in the function name. Now, it looks clearer.

lib/perl/Genome/Model/Tools/Analysis/Coverage/CoveragePlot.pm
brummett: ```debug_message()``` and the others are methods of the base class, UR::Object.  They take the same arguments as ```printf()``` and are the standard way to communicate short messages to the user.

```debug_message``` is intended to be information to the developer.  Their messages are normally suppressed.
```status_message``` is what we normally use to print general information to the user.
```warning_message``` is for when something unusual or unexpected has happened.
```error_message``` is for when something dire has happened.  The program will usually die or exit after sending the message.

File: lib/perl/Genome/Model/Tools/Analysis/Coverage/CoveragePlot.pm [hunk 9]
```diff
@@ -0,0 +1,1375 @@
...
+    # gets the file name
+    my $filepath = shift;
+
+    # On Unix returns ("baz", "/foo/bar", ".txt") from "/foo/bar/baz.txt"
+    my ($name, $path, $ext) = fileparse($filepath, qr/\.[^.]*/);
+    return $name . $ext;
+}
+
+
+sub is_numeric {
```
brummett: The function ```looks_like_number``` in the module Scalar::Util does this same job and is more thorough.

lib/perl/Genome/Model/Tools/Analysis/Coverage/CoveragePlot.pm
brummett: create_hash() and its related methods should definitely be in another class.  Returning entities blessed into this Command module's class will be confusing to the next person to work on this, and is bound to cause trouble in the future.

If you don't want to create another .pm file for fear of cluttering up the filesystem, you could just declare another ```package``` at the bottom of this same file, and move the relevant methods down there.  Let us know if you need help making the changes.

lib/perl/Genome/Model/Tools/Analysis/Coverage/CoveragePlot.pm
ebelter: @brummett  It that preferred to Regexp::Common?

brummett: I guess in this case, you could use either.  Scalar::Util::looks_like_number() uses the same low-level, perl internal mechanism used when it converts a string to a number.  Any string that passes looks_like_number() won't generate warnings or exceptions if you try to use it in a numerical expression.

I'd try looks_like_number() first, and only use something more complicated like a regex if it didn't work.  A regex will always be slower.

gschang:
I appreciate your helpful comments. I addressed most comments and updated my source codes. Regarding your create_hash() comment, I solved it both by creating a new class (in fact, this class is a part of my original class CSLib::Tabular, and I pasted only necessary methods) and by converting some to normal subroutines, as you suggested. I want to see how you think.

brummett:
This looks much improved and removes my concerns.
+1

chrisamiller:
+1 let's merge this

PR #526, opened 2015-03-01
temporarily disable test

nnutter:
See AR-953.


nnutter:
Merging early to get tests passing.

PR #527, opened 2015-03-02
configure a site module for the SGMS.

gatoravi:
add SGMS.pm which configures backends for scopes ("unknown", "site") which are required for builds to run in the SGMS. "tgisan" seems to be TGI specific and can be ignored. This addresses https://github.com/genome/gms/issues/162

### DIFF
--- a/lib/perl/Genome/Sys/Lock.pm
+++ b/lib/perl/Genome/Sys/Lock.pm
@@ -208,11 +208,13 @@ sub all_backends {
 }
 
 sub scopes {
-    if ($ENV{GENOME_SYS_ID} && $ENV{GENOME_SYS_ID} ne 'GMS1') {
-        return ('site', 'unknown');
-    } else {
-        return ('site', 'tgisan', 'unknown');
-    }
+   if ($ENV{GENOME_LOCK_SCOPES}) {
+      return split(":", $ENV{GENOME_LOCK_SCOPES});
+   } else {
+      # Ultimately move this into the TGI config and
+      # not have it in code at all.
+      return ('site', 'tgisan', 'unknown');
+   }
 }
 
 sub clear_backends {
--- a/lib/perl/Genome/Sys/Lock.pm
+++ b/lib/perl/Genome/Sys/Lock.pm
@@ -208,7 +208,11 @@ sub all_backends {
 }
 
 sub scopes {
-    return ('site', 'tgisan', 'unknown');
+    if ($ENV{GENOME_SYS_ID} && $ENV{GENOME_SYS_ID} ne 'GMS1') {
+        return ('site', 'unknown');
+    } else {
+        return ('site', 'tgisan', 'unknown');
+    }
 }
 
 sub clear_backends {
--- a/lib/perl/Genome/Env/GENOME_LOCK_SCOPES.pm
+++ b/lib/perl/Genome/Env/GENOME_LOCK_SCOPES.pm
@@ -1,0 +1,19 @@
+package Genome::Env::GENOME_LOCK_SCOPES;
+
+=pod
+
+=head1 NAME
+
+GENOME_LOCK_SCOPES
+
+=head1 DESCRIPTION
+
+The GENOME_LOCK_SCOPES defines the scopes for which locking backends are set and released.
+
+=head1 DEFAULT VALUE
+
+ Not set
+
+=cut
+
+1;
--- a/lib/perl/Genome/Site/SGMS.pm
+++ b/lib/perl/Genome/Site/SGMS.pm
@@ -1,0 +1,20 @@
+package Genome::Site::SGMS;
+use strict;
+use warnings;
+
+BEGIN {
+    unless ($INC{'Genome/Sys/Lock.pm'}) {
+        use Genome::Sys::Lock;
+    }
+};
+
+require Genome::Sys::Lock::FileBackend;
+Genome::Sys::Lock->add_backend('site',
+    Genome::Sys::Lock::FileBackend->new(is_mandatory => 1,
+        parent_dir => $ENV{GENOME_LOCK_DIR}));
+
+Genome::Sys::Lock->add_backend('unknown',
+    Genome::Sys::Lock::FileBackend->new(is_mandatory => 1,
+        parent_dir => '/'));
+
+1;
--- a/lib/perl/Genome/Site/SGMS.t
+++ b/lib/perl/Genome/Site/SGMS.t
@@ -1,0 +1,13 @@
+#!/usr/bin/env genome-perl
+
+use strict;
+use warnings;
+
+use above 'Genome';
+use Test::More tests => 1;
+
+# This test was auto-generated because './Site/TGI.pm'
+# had no '.t' file beside it.  Please remove this test if you believe it was
+# created unnecessarily.  This is a bare minimum test that just compiles Perl
+# and the UR class.
+use_ok('Genome::Site::SGMS');


malachig:
@brummett .  I'm a bit lost here, but it also seems non-generic to have this list `('site', 'tgisan', 'unknown');` in the code.  

Since 'tgisan' is decidedly site specific, might this make sense as an environment variable, rather than hardcoded in `Genome/Sys/Lock.pm`?  Then we could configure the scopes that we need and we could customize the list of names to validate against.  And inside TGI we could do the same.

Probably I am just not understanding though.  I think I understand what you are saying about the inappropriateness of throwing an exception when a scope has no backends and no locks.  Not totally clear to us how specifically we would modify that though...

malachig:
@gatoravi is there a reason why this PR is into `gms-pub-2015.02.22` instead of `master`.

lib/perl/Genome/Sys/Lock.pm
```diff
@@ -208,7 +208,11 @@ sub all_backends {
 }
 
 sub scopes {
-    return ('site', 'tgisan', 'unknown');
+    if ($ENV{GENOME_SYS_ID} && $ENV{GENOME_SYS_ID} ne 'GMS1') {
+        return ('site', 'unknown');
+    } else {
+        return ('site', 'tgisan', 'unknown');
+    }
 }
```
sakoht: Malachi suggested we let the scopes be in the environment entirely for the standalone version.

This would let the internal version switch to env too instead of having hard-coded values:
```
sub scopes {
   if ($ENV{GENOME_LOCK_SCOPES}) {
      return split(":", $ENV{GENOME_LOCK_SCOPES})
   }
   else {
      # Ultimately we should move this into the TGI config and not have it in code at all.
      return ('site', 'tgisan', 'unknown');
   }
}
```

gatoravi:
I think the original goal was to get it into gms-pub-2015.02.22 to get on with the SGMS issues. But I don't see why this change couldn't  go into master as well. I was hoping to get some consensus on a fix before suggesting that.

brummett:
I like the environment variable idea.  We can get it into our environment here at TGI, then remove the ```else``` case in a later commit.  It'll need a matching module under Genome/Env.

gatoravi:
ok, that sounds good. I've just pushed a commit which adds this change.

brummett:
You'll also need a module Genome/Env/GENOME_LOCK_SCOPES.pm.  You can base it on GENOME_BUILD_ID.pm so it's not required and has some documentation with it.

gatoravi:
thanks for the headsup @brummett I've added this module.

malachig:
Oh yeah, thanks @brummett for reminding us of this requirement for `GENOME_*` env variables.

brummett:
+1 looks good to me

gatoravi:
I'm merging this into master as well unless someone has an objection.

nnutter:
The scopes were hard-coded because they weren't intended to be configured.  The current scopes were a half-step.  There shouldn't be an `unknown` scope and `tgisan` would be migrated to a new `host` scope.  So you would be left with `site` and `host`.  In a single-host system you would configure an appropriate backend for `site` and `host` and they would probably be the same backend.  In a multi-host system these would be different backends.

This PR probably broke locking in the GMS config.  For example, what happens if you try to call `cached_full_consensus_path` on a `Genome::Model::Build::ReferenceSequence` which locks [`tgisan`][1]?

[1]: https://github.com/genome/genome/blob/0579ea9b9a39f4f3ad755c68fab696326b329316/lib/perl/Genome/Model/Build/ReferenceSequence.pm#L776

malachig:
I don't quite follow.  In the GMS, the same should happen that has always happened in TGI because $GENOME_LOCK_SCOPES is not set there.  Not being set, Lock.pm will return `return ('site', 'tgisan', 'unknown');` just as it did before.   Maybe I am not understanding the code change quite right?  It is the intent of the PR to not change anything regarding configuration at TGI.  Only to conditionally allow a different list of scopes if you actively set them (as in the sGMS).

nnutter:
It is correct that if `GENOME_LOCK_SCOPES` is not set then no effective change has happened.  What I am saying is that what you are trying to allow to happen when you set `GENOME_LOCK_SCOPES` won't "just work".  Whatever code calls `lock_resource` has to specify a scope so you can't just configure new scopes; they are part of the source code.  What you can configure is the backend implementation for a given scope.  If you are imagining a scope other than 'site' or 'host' an example may be useful to help clarify.

malachig:
Basically, we are just trying to work around an exception thrown if these scope are not defined.  As described here: https://github.com/genome/gms/issues/162.  Basically, the implementation as it currently stands blocks installation of the the sGMS.

We are not trying to have a scope other than 'site'.  We are not sure what the implications of having or not having 'unknown' would be.  We are simply trying to limit to a subset of those already defined at TGI and load that subset from SGMS.pm.  If there was a HOST.pm, that would also make sense I think, but it doesn't exist yet right?

Anyway, we are simply trying to get around being blocked by the current half-step implementation of the scopes.  The code seems to require that those scopes listed in Lock.pm be defined.  It seemed weird to define 'tgisan' in the sGMS because it is TGI specific, so we wanted to at least limit the list to those that might make sense.

If there is some other code change that would make more sense right now, we are open to suggestions.  If you want to make the other 'half step' of implementing scopes part of this PR, I have no objection to that, but we are probably not in a position to do it, lacking a lot of context as to what is going on with these backends/scopes.

Another option, is to wait on that, and for now, leave this PR on the sgms specific branch where it is now.  I'm not really in favor of that approach but we have a pretty hard deadline for this work.  So if there are strong objections to this PR and no clarity on what could go into master we may just have to move on.

Another option is to meet and discuss further offline if you want.

tmooney:
It would be sufficient to have the sGMS define a backend for the `tgisan` scope.  Despite the name, it is not TGI-specific.  (The TGI-specific value for the `GENOME_FS_LOCAL_NETWORK_CACHE` environment variable happens to include the string `tgisan`, but the scope is really related to that variable.)

Without defining the `tgisan` scope, it seems likely that no pipelines would successfully run because that scope is currently used [in reference sequences](https://github.com/genome/genome/blob/7e4b64d5d8c8d11f543d458109cb28544bced902/lib/perl/Genome/Model/Build/ReferenceSequence.pm#L776).

I believe @nnutter is trying to say that using an environment variable for this doesn't solve the problem, because the pipelines will crash as soon as they encounter a place in the code where the scope that was left out happened to be used.

nnutter:
I'm sorry I'm making the issue confusing.  Tom is correctly interpreting my concerns.

malachig:
When we started all of this we had a conversation with @mkiwala and @brummett .  At that point we were working under the assumption that `tgisan` was related to certain file system mounts that were accessible on a subset of systems at the TGI.  And that `unknown` was somehow related to reference sequences maybe.  It sounds like maybe we have gone down a garden path based on that assumption that doesn't make sense.

Based on this info, maybe we should just define all three scopes in SGMS.pm.  Then we do not need the new environment variable at all.  And we do not need the conditional logic in Lock.pm.  

We still need to load the SGMS.pm to define these scopes.  That can be done by defining $ENV{GENOME_CONFIG}.

malachig:
Since this was already merged into the branch `genome:gms-pub-2015.02.22`, I think we we will revert and take another stab at fixing this issue that takes the above discussion into account.  That will be a new PR but we'll reference back to this PR (https://github.com/genome/genome/pull/527) and the related issue (https://github.com/genome/gms/issues/162).  Apologies for all the confusion guys.  And thanks to @brummett, @mkiwala, @nnutter and @tmooney for helping us understand and work towards a solution that makes sense.

nnutter:
That sounds good to me.

PR #528, opened 2015-03-02
Adding reminder that AnPs are created in a Pending state

chrisamiller:


chrisamiller:
I'm hoping that this will prevent me from forgetting to release AnPs, which happens like once a month...

acoffman:
+1

PR #529, opened 2015-03-02
Use parens around list

tmooney:
(Perl 5.14+ doesn't like a bare qw().)


susannasiebert:
Should we change/fix this everywhere in our code base?

tmooney:
Yes!  I'm surprised it's as common a pattern as it is.

brummett:
[Here is the explanation](http://docs.activestate.com/activeperl/5.14/lib/pods/perl5140delta.html#use_of_qw______as_parentheses) for the change.  The parens where always supposed to be there, but ```qw()``` tricked a bug in the parser.

brummett:
+1

acoffman:
+1

iferguson90:
+1 :cow:

PR #530, opened 2015-03-03
Print the display name for blessed objects.

tmooney:
Updating the reference would produce a message like:
```
setting reference to Genome::Model::Build::ImportedReferenceSequence=HASH(0x6e2ef40)
```
Now it will be like:
```
setting reference to human-build38 (a123b4567cd89e01f234a5b67cd89ef0)
```

### DIFF
--- a/lib/perl/Genome/FeatureList/Command/Update.pm
+++ b/lib/perl/Genome/FeatureList/Command/Update.pm
@@ -4,6 +4,7 @@ use strict;
 use warnings;
 
 use Genome;
+use Scalar::Util qw();
 
 class Genome::FeatureList::Command::Update{
     is => 'Command::V2',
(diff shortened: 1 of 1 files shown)


iferguson90:
Seems reasonable! :sheep:
+1

ebelter:
Maybe display_name_for_value in Genome::Command::Crud should be pulled out to a more useful place.

tmooney:
@ebelter Perhaps! I've sometimes thought we should make `UR::Object`s use `__display_name__` for their default stringification.  Then this code goes back to the way it was.

ebelter:
@tmooney That would be great. Then you do not get useless information when stringifying. I am not against pushing this out. I am just advocating a more centralized solution. If interested, we could pair and knock it out :)

brummett:
The problem with overloading stringification is that the overloading will also be used whenever you test it in a boolean way, like in
```
if ($obj) { ... }
```
Which slows things down.  At one point, UR::BoolExpr's overloaded stringification sub was the top-most subroutine which profiling.  We had to find all the places UR was using a BoolExpr as a boolean and replace it with something like
```
if (defined $bx) { ... }
```
So, it's not as great an idea as it may immediately seem.

tmooney:
Yeah, that would be bad!  Could we add `bool => sub { 1 }` to our overloads to get around that?  (Is a `UR::Object` ever false?)

brummett:
That would make it call the 'bool' overload function instead of the stringification function, but there's still the overhead of invoking the overload machinery.  I'd want to quantify how much slower it is before adding stringification overloading to UR::Object by default.

An alternative is to have a Genome class/mixin/role to do the stringification and only apply it in certain places.

ebelter:
@brummett That sounds like a good solution.

PR #531, opened 2015-03-03
somatic-filter workflow

gatoravi:
This PR adds a module to filter Varscan SNPs using a workflow. The current way to do this is GMT Varscan SomaticParallelFilter.pm uses manual bjobs to do this. A workflow makes it easier to run this tool as part of a build.


lib/perl/Genome/Model/Tools/Capture/FormatSnvs.pm
```diff
@@ -60,10 +63,19 @@ sub execute {                               # replace with real execution logic.
...
-	my $output_file = $self->output_file;
+	my $output_file;
+	if($self->output_file) {
+		$output_file = $self->output_file;
+	} elsif($self->outdir) {
+		$output_file = File::Spec->join(
+		$self->outdir,
+		basename($variants_file) . ".formatted");
+	} else {
+		die $self->error_message("Specify variants file or outdir.");
```
tmooney: Should this error message refer to the output_file rather than the variants file?

lib/perl/Genome/Model/Tools/Capture/FormatSnvs.t
```diff
@@ -4,10 +4,56 @@ use strict;
...
-# created unnecessarily.  This is a bare minimum test that just compiles Perl
-# and the UR class.
 use_ok('Genome::Model::Tools::Capture::FormatSnvs');
+
+BEGIN {
+  $ENV{UR_DBI_NO_COMMIT} = 1;
+  $ENV{UR_USE_DUMMY_AUTOGENERATED_IDS} = 1;
+}
+
+my $expected_output_dir = $ENV{"GENOME_TEST_INPUTS"} . "Genome-Model-Tools-Capture-FormatSnvs/2015-02-12/";
```
tmooney: The fancy new way of doing this is:
```perl
use Genome::Utility::Test; #with the other "use" statements
Genome::Utility::Test->data_dir_ok('Genome::Model::Tools::Capture::FormatSnvs','2015-02-12');
```

It will assemble the directory path from the module name and version, and add a single ok() test that it exists.

lib/perl/Genome/Model/Tools/Capture/FormatSnvs.t
```diff
@@ -4,10 +4,56 @@ use strict;
...
+ok(-e $expected_output_dir, "Found test dir: $expected_output_dir") or die;
+
+my $variants_file = File::Spec->join(
+  $expected_output_dir,
+  "varscan.snps",
+);
+ok(-e $variants_file, "Found variant file: $variants_file") or die;
+
+#my $temp_dir = Genome::Sys->create_temp_directory();
+my $temp_dir = ".";
```
tmooney: This should be swapped back to a real temporary directory.

lib/perl/Genome/Model/Tools/Capture/FormatSnvs.t
```diff
@@ -4,10 +4,56 @@ use strict;
...
+$formatter->execute();
+
+my $expected_outfile = File::Spec->join(
+  $expected_output_dir,
+  "varscan.snps.filtered"
+);
+ok(-e $expected_outfile, "Found expected outfile: $expected_outfile") or die;
+my $outfile = File::Spec->join(
+  $temp_dir,
+  "varscan.snps.filtered",
```
tmooney: I believe the code appends `.formatted` rather than `.filtered` to the filename.

lib/perl/Genome/Model/Tools/Capture/FormatSnvs.t
```diff
@@ -4,10 +4,56 @@ use strict;
...
+);
+ok(-e $outfile, "Found outfile: $outfile") or die;
+
+#Perform a diff between the stored results and those generated by this test
+my @diff = `diff -r $outfile $expected_outfile`;
```
tmooney: `Genome::Utility::Test` also provides a test that might be handy here.
```perl
use Genome::Utility::Test qw(compare_ok); #as before but exporting compare_ok
compare_ok($expected_outfile, $outfile);
```
If there is a diff, the output is something like:
```
# First diff:
# --- /expected/file.out line 1
# +++ /actual/file.out line 1
# -[expected]
# +[actual]
not ok 2
```

lib/perl/Genome/Model/Tools/Capture/FormatSnvs.t
```diff
@@ -4,10 +4,56 @@ use strict;
...
+ok(@diff == 0, "Found only expected number of differences between expected results and test results")
+or do {
+  diag("expected: $expected_output_dir\nactual: $temp_dir\n");
+  diag("differences are:");
+  diag(@diff);
+  my $diff_line_count = scalar(@diff);
+  print "\n\nFound $diff_line_count differing lines\n\n";
```
tmooney: It would be better if this were a `diag` or at least printed to `STDERR` (if it's not replaced with `compare_ok` entirely).

lib/perl/Genome/Model/Tools/Somatic/FilterFalsePositives.pm
```diff
@@ -220,9 +239,12 @@ sub execute {
         die;
     }
 
+    my $output_file =
+      $self->output_file?$self->output_file:File::Spec->join($self->outdir, basename($self->variant_file) . ".filtered");
```
tmooney: You could simplify this to:
```perl
my $output_file = $self->output_file // File::Spec->join($self->outdir, basename($self->variant_file) . ".filtered");
```

lib/perl/Genome/Model/Tools/Somatic/FilterFalsePositives.pm
```diff
@@ -636,6 +665,10 @@ sub fails_homopolymer_check {
 sub readcount_program {
     my $self = shift;
     my $reference = $self->reference;
+    if($self->bam_readcount_version) {
+      my $bam_rc = "bam-readcount" . $self->bam_readcount_version;
```
tmooney: This is better than a hardcoded version! :smile:  A future improvement could be defer to `Genome::Model::Tools::Sam::Readcount` to get the path and/or run this.

lib/perl/Genome/Model/Tools/Varscan/SomaticFilterWorkflow.pm
```diff
@@ -0,0 +1,231 @@
...
+
+use strict;
+use warnings;
+
+use Genome;
+use Workflow::Simple;
+use File::Basename;
+
+BEGIN {
+    $ENV{WF_USE_FLOW} = 1;
```
tmooney: I would set this at the top of `execute` with `local $ENV{WF_USE_FLOW} = 1;`.  That way it is only changed when the command is run.

lib/perl/Genome/Model/Tools/Varscan/SomaticFilterWorkflow.t
```diff
@@ -0,0 +1,60 @@
...
+    $ENV{UR_DBI_NO_COMMIT} = 1;
+    $ENV{UR_USE_DUMMY_AUTOGENERATED_IDS} = 1;
+    $ENV{NO_LSF} = 1;
+};
+
+use above "Genome";
+use Test::More;
+
+use_ok('Genome::Model::Tools::Varscan::SomaticFilterWorkflow') or die;
+my $expected_output_dir = $ENV{"GENOME_TEST_INPUTS"} .
```
tmooney: Similarly to the other test, this and the `diff` later could use `Genome::Utility::Test`.

lib/perl/Genome/Model/Tools/Capture/FormatSnvs.t
```diff
@@ -4,10 +4,48 @@ use strict;
 use warnings;
 
 use above 'Genome';
-use Test::More tests => 1;
+use Test::More;
+use Genome::Utility::Test;
```
tmooney: The `use Genome::Utility::Test qw(compare_ok);` should be sufficient to cover both uses of the module, and this one can be removed.

lib/perl/Genome/Model/Tools/Varscan/SomaticFilterWorkflow.t
gatoravi: As I understand compare_ok under Genome::Utility::Test compares two files, is there a recommended way of comparing corresponding files in two directories ?

gatoravi:
CI, test this.

nnutter:
I suspect this needs input metadata added to the test database; see [Test Database Template](https://confluence.gsc.wustl.edu/display/APIPE/Test+Database+Template) for that process.  Please get some help from APIPE if needed.

lib/perl/Genome/Model/Tools/Somatic/FilterFalsePositives.pm
gatoravi: it seemed like getting the bam-readcount path from that module could be a little complicated since `sub readcount_path` uses `$self->use_version` instead of a parameter to that routine.

gatoravi:
I think I've made all of the changes except two which I'm not sure how to incorporate, is this ok to merge in ?

lib/perl/Genome/Model/Tools/Somatic/FilterFalsePositives.pm
tmooney: Yeah, you're right--that one isn't set up like most of them with a `path_for_bam_readcount_version`-like subroutine.  I woudln't worry about it!

tmooney:
+1 :earth_americas:

gatoravi:
merging so that we can start using this module :tokyo_tower:

PR #532, opened 2015-03-03
Sys: Do not close the fh in write_file if writing to STDOUT

ebelter:
Writing to STDOUT more than once via Sys's write_file causes a crash. This is a PR has tests showing the failure. Should this be fixed?

@brummett please take a look.

### DIFF
--- a/lib/perl/Genome/Sys.t
+++ b/lib/perl/Genome/Sys.t
@@ -2,7 +2,6 @@
 use strict;
 use warnings;
 use above 'Genome';
-use Test::Exception;
 use Test::More;
 
 use Genome::Sys;
--- a/lib/perl/Genome/Sys.t
+++ b/lib/perl/Genome/Sys.t
@@ -2,6 +2,7 @@
 use strict;
 use warnings;
 use above 'Genome';
+use Test::Exception;
 use Test::More;
 
 use Genome::Sys;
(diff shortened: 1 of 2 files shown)


ebelter:
The subtest I added is misnamed. I realized there were no tests for write_file and read_file and started to add them. But they failed because the file was not being flushed. So I removed them, and focused on the STDOUT issue.

brummett:
Looks like ```write_file()``` needs to skip calling close if $fname is "-"

ebelter:
I did a check test to see if opening STDOUT multiple times causes a problem and it does not. I will add the suggested skip if the fname is '-'.

ebelter:
@brummett @tmooney any thoughts on this PR?

tmooney:
Seems fine!  I see `open_file_for_writing` specially returns `STDOUT` when passed `-` instead of opening a new filehandle.

lib/perl/Genome/Sys.t
```diff
@@ -2,6 +2,7 @@
 use strict;
 use warnings;
 use above 'Genome';
+use Test::Exception;
```
tmooney: Is this still being used?

tmooney:
(Test failure is because of issue addressed by #536.)

brummett:
+1

ebelter:
CI, test this.

PR #534, opened 2015-03-03
New Command to Define an AnalysisProject Template.

tmooney:
Once upon a time we added support for Analysis Projects that were "templates".  They would let someone store a set of configurations in one place for easy copying using `genome analysis-project copy-config`.  However, there was no way to create one... until now!


acoffman:
+1

jasonwalker80:
+1

PR #535, opened 2015-03-03
Retain build-status in metadata if defined

gatoravi:
Annotation builds with status "Dummy" prevent builds being launched on the SGMS. This PR retains the build's current status.

### DIFF
--- a/lib/perl/Genome/Model/Command/Export/Metadata.pm
+++ b/lib/perl/Genome/Model/Command/Export/Metadata.pm
@@ -341,7 +341,7 @@ sub add_to_dump_queue {
     }
 
     if ($obj->isa("Genome::Model::Build")) {
-        $obj->{status} = "Dummy";
+        $obj->{status} = $obj->status || "Dummy";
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Model/Command/Export/Metadata.pm
```diff
@@ -341,7 +341,7 @@ sub add_to_dump_queue {
     }
 
     if ($obj->isa("Genome::Model::Build")) {
-        $obj->{status} = "Dummy";
+        $obj->{status} = $obj->status || "Dummy";
```
tmooney: `$obj->{status}` should already hold the current value of `$obj->status`, so it should be sufficient to remove this line and the immediately following one that changes `{db_committed}`.

gatoravi:
it seems that a check for a valid status would be appropriate here. A commit to do that has been added.

tmooney:
+1 :ballot_box_with_check:

gatoravi:
merging early since this is relatively minor and would help with SGMS issues.

PR #536, opened 2015-03-04
Use the test build to test the ClinSeq file accessors.

tmooney:
The Full Perl tests are now failing because this test depends on the current blessed ClinSeq build, which is not included in the new testing database.  Update this test to not having a moving target of database dependency.


gatoravi:
This looks ok to me but someone more familiar with the new testing database than me should probably have a look at it.

tmooney:
I selfishly would like the tests to pass, so I'm going to merge this given that >30 hours have elapsed!

PR #539, opened 2015-03-05
No longer hide build queue in docs.

tmooney:
Now that "queueing" a build will run it as the run_as user, users can use this to queue their own builds.


acoffman:
+1

apregier:
+1

PR #540, opened 2015-03-05
New genome-env option

apregier:
I'm not sure if this is a good idea or not, but it seemed useful to us, for example for debugging model tests

### DIFF
--- a/bin/genome-env
+++ b/bin/genome-env
@@ -45,6 +45,8 @@ function help {
     -i db_id:db_host:db_port
           The connection information of an existing database to use.  Format is name:host:port
 
+    -p    Persist the test-db outside of this shell.
+
     -U    Disable configuration of UR.
 
     -u ur_repo
@@ -130,7 +132,7 @@ function main {
         EXIT=$?
     fi
 
-    if test -z "$DISABLE_DB" -a -z "$DB_ID"
+    if test -z "$DISABLE_DB" -a -z "$PERSIST_DB"
     then
         test-db delete database $TESTDBSERVER_DB_NAME > /dev/null
     fi
@@ -145,10 +147,10 @@ function validate_bash {
 }
 
 function parse_opts {
-    DISABLE_DB="" DISABLE_MIGRATIONS="" DISABLE_UR="" DISABLE_WF="" DB_SNAPSHOT_NAME="" DB_ID="" DB_HOST="" DB_PORT="" SQITCH_REPO="" UR_REPO="" WF_REPO=""
+    DISABLE_DB="" DISABLE_MIGRATIONS="" DISABLE_UR="" DISABLE_WF="" DB_SNAPSHOT_NAME="" DB_ID="" DB_HOST="" DB_PORT="" SQITCH_REPO="" UR_REPO="" WF_REPO="" PERSIST_DB=""
 
     local opts
-    while getopts :hMm:Dd:i:Uu:Ww: opts "$@"
+    while getopts :hMm:Dd:i:pUu:Ww: opts "$@"
     do
         case $opts in
             h)
@@ -173,6 +175,9 @@ function parse_opts {
                 DB_ID=${args[0]}
                 DB_HOST=${args[1]}
                 DB_PORT=${args[2]}
+                PERSIST_DB="TRUE"
+                ;;
+            p)  PERSIST_DB="TRUE"
                 ;;
             U)
                 DISABLE_UR="TRUE"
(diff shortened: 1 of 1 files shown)


acoffman:
I may not be the best person to review the shell script itself, but this definitely does seem like a useful feature, and it looks to be implemented correctly!

+1

bin/genome-env
```diff
@@ -89,7 +102,14 @@ function main {
...
-        log.info "Test DB Snapshot: $DB_SNAPSHOT_NAME"
+        if test -n "$DB_ID"
+        then
+            log.info "Using existing test DB"
+            log.info "Test DB Host: ${TESTDBSERVER_DB_HOST}"
+            log.info "Test DB Port: ${TESTDBSERVER_DB_PORT}"
+        else
+            log.info "Test DB Snapshot: $DB_SNAPSHOT_NAME"
+        fi
         log.info "Test DB Name: $TESTDBSERVER_DB_NAME"
```
ebelter: Why not merge this if/else block with the one above?

apregier: I'm not sure I know what you mean, but if you mean the block starting at line 80, I guess the reason is to maintain the order of the log messages.

ebelter: That is the block I was siting.

nnutter:
This either may not be needed or will be insufficient as is.  How do you intend on debugging model tests?  If you mean we get a failure and you want to be able to enter that ephemeral database with this then you'll also need an option to disable removing the ephemeral database when the `genome-env` process ends.  If you just want to be able to manually run a model tests and debug then this isn't necessary as you can just drop into a shell with `genome-env`, run the model test, and then debug as needed.

apregier:
Yes, I was thinking of looking at failed model tests from Jenkins, or even just checking on ones that are currently running.  I did include a check before deleting the db to see if it was loaded with the new option (line 133).

nnutter:
I meant that the model tests are going to delete their ephemeral database when they complete.  (Though I imagine there's probably going to be some resulting inconsistency if someone's connected to the database when they try to do that.)

apregier:
Oh ok - yes, I was thinking we need that option too!  Is it true that the dbs expire after a month anyway?  It seems like it would be useful to have them for at least a couple days in the case of failed builds, and in the case of succeeded builds we could explicitly delete them.

nnutter:
The current default is 7 days.  If that needs to be changed or, more likely, configurable we'll need to add that to `test-db`.

apregier:
7 days seems reasonable to me - or do you think it will be too much of a resource hog to keep the dbs for model tests around that long?  I pushed a commit that adds the option of persisting the db beyond the life of the shell.

bin/genome-env
```diff
@@ -541,6 +585,8 @@ default_value TESTDBSERVER_DB_USER "genome"
 default_value TESTDBSERVER_DB_PASS "mypassword"
 default_value TESTDBSERVER_URL "https://apipe-test-db.gsc.wustl.edu"
 default_value GE_ENV_NAME "genome-env"
+default_value TESTDBSERVER_DB_HOST "apipe-test-db.gsc.wustl.edu"
+default_value TESTDBSERVER_DB_PORT "5432"
```
nnutter: `TESTDBSERVER_DB_HOST` and `TESTDBSERVER_DB_PORT` defaults shouldn't be in here I don't think.  They get set by `$TESTDB database create --bash` and is information that should either come from the developer, for `-i`, or from `test-db`.

bin/genome-env
```diff
@@ -89,7 +104,14 @@ function main {
 
     if test -z "$DISABLE_DB"
     then
-        log.info "Test DB Snapshot: $DB_SNAPSHOT_NAME"
+        if test -n "$DB_ID"
+        then
+            log.info "Using existing test DB"
+            log.info "Test DB Host: ${TESTDBSERVER_DB_HOST}"
+            log.info "Test DB Port: ${TESTDBSERVER_DB_PORT}"
```
nnutter: I think it would be fine to always show the host and port, i.e. move them down with "Test DB Name".

nnutter:
> 7 days seems reasonable to me - or do you think it will be too much of a resource hog to keep the dbs for model tests around that long?

We'll just have to monitor disk usage on `apipe-test-db` to see.

bin/genome-env
apregier: I agree - those were from before when we were trying to just pass in the db name with -i; I should have deleted them!

bin/genome-env
apregier: Ok, done!

nnutter:
:+1:

PR #541, opened 2015-03-05
Park beta move

davidlmorton:


tmooney:
This seems like a step in the right direction!  Is the intention to eventually add this to genome.sh, or to keep using the value specified in code here?

+1 :boat:

gsanders:
+1! :100: :100: :100: (that makes 300, this is sparta!)

iferguson90:
:dolphin: :+1:

PR #542, opened 2015-03-05
Add max_haplotypes as option for ccsvcf for indels

davidlmorton:
This option is already supported by Park, this just allows the gmt to set it.


gsanders:
I see no obvious mistakes, +1.

I will state the obvious and say:
1)  it seems dangerous to not have tests around this, such that if you and I miss a typo, stuff is broken.
2) I was able to do as you suggest and follow along from GMT to park land, so I think that speaks positively of things. However, I think this would be difficult for a park-newbie.

davidlmorton:
To be clear, there are unit-tests on this (the Genome) side, that go right up to what is passed to the Park system, and there is good test-coverage of this stuff in Park itself.  I agree that some integration tests would be nice though.

PR #543, opened 2015-03-05
Variant reporting vaf interpreter bugfix

susannasiebert:
This PR fixes a bug in the per-library-vaf interpreter where flatten_hash would stomp on any previously calculated values so only that last sample's vaf values were getting populated. A test case was added to cover this case.

This PR also takes care of some minor cleanup that I noticed while working on this code.


lib/perl/Genome/VariantReporting/Suite/BamReadcount/PerLibraryVafInterpreter.pm
```diff
@@ -39,19 +38,20 @@ sub _interpret_entry {
...
+        $return_values{$alt_allele} = {map {$_ => $self->interpretation_null_character} $self->available_fields};
+    }
 
     for my $sample_name ($self->sample_names) {
         my $readcount_entries = $self->get_readcount_entries($entry, $sample_name);
         unless (defined($readcount_entries)) {
-            for my $alt_allele (@$passed_alt_alleles) {
-                $return_values{$alt_allele} = {map {$_ => $self->interpretation_null_character} per_sample_field_descriptions($sample_name)};
-            }
+            return %return_values;
```
tmooney: Should this be a `next` instead of `return`ing out of the loop?

susannasiebert: You're right. Fixing.

apregier:
+1

acoffman:
+1

PR #544, opened 2015-03-05
SUPER::create() may return undef--pass it along.

tmooney:
`get_or_create()` will then try to `get()` one last time (if `create` detected
that the object was made already).  Crashing prevents this from working.


ebelter:
+1

apregier:
+1

PR #546, opened 2015-03-06
Compare vcf in process

apregier:
There is probably a much cleaner way to do this, maybe using roles instead of inheritance, once roles are ready?  I am open to other suggestions as well.  It would be nice if it didn't become quite as entangled as the file comparison logic in Build.pm and I am afraid this is heading down the same path by using inheritance.


tmooney:
It seems like the logic for both deciding how to do the comparisons and also performing them doesn't really need to be in either `Build` or `Process`. Instead it could be in some common location that cares about filetypes and diffing, and both can delegate their work to that.  (Bonus points if it also supports `Genome::Utility::Test::compare_ok`!)

Such a comparator might end up in `Genome::Utility` or even `Genome::File`?  Naming things is hard!

apregier:
I pulled at least some of the file comparison logic out into a base class.  I didn't go for the bonus points of supporting compare_ok!  Another possibility would be to also pull out the directory comparison logic, but it is quite different between Process and Build.  Build has a bunch of additional stuff to match up files that don't have exactly the same filename.

apregier:
CI, test this.

apregier:
CI, test this.

apregier:
This is ready for reviews again - I also ran the model-tests to make sure the build differs still work as expected.

lib/perl/Genome/Model/Build/GenePrediction/Bacterial.pm
```diff
@@ -172,18 +172,19 @@ sub regex_files_for_diff {
...
-sub regex_for_custom_diff {
+sub special_compare_functions {
     return (
-        ace         => '\.ace$',
-        ks_bz2      => '\.ks\.bz2$',
-        sorted_bz2  => '\.sorted\.bz2$',
-        fasta       => '\.fasta$',
-        sqlite      => 'sqlite-[^-]+-\d{4}_\d{2}_\d{2}.dat$',
-    ), $_[0]->SUPER::regex_for_custom_diff(@_);
+        qr(\.ace$) => sub {my ($a, $b) = @_; !Genome::Model::Build::GenePrediction::Bacterial::diff_ace($a, $b);},
```
tmooney: It might be simpler if each of these passed `@_` along instead of declaring `$a` and `$b` (which are also magic variables for sorting).

apregier: Great suggestion - implemented!

apregier:
CI, test this.

PR #547, opened 2015-03-06
clinseq: update defaults in update-analysis

gatoravi:
Update the default annotation-build, dbsnp version, cancer-annotation db and ref-align processing-profile used by ClinSeq/UpdateAnalysis.

PR #548, opened 2015-03-06
ClinSeq: update analysis now specifies cancer, misc, cosmic DBs

malachig:
ClinSeq: update analysis now specifies cancer, misc, cosmic DBs in clin-seq model define commands when a non-default version is specified by the user.

`genome model clin-seq update-analysis` was correctly examining which version of the cosmic, cancer and misc databases were being used in candidate models.  If these parameters did not match those specified by the user (or default if the user did not say), then such clin-seq models were considered non matching.  When no clin-seq model matching all desired criteria was found, it advised the user to create one.  

The problem is, at this point it did not present the correct command to define a model with the correct cosmic, cancer, and misc databases when the user specified a non-default option.

This pull request modifies update-analysis so that the correct `genome model define clin-seq` command is produced when non-default DBs are requested by the user.

I tested this on a use case provided by @kkrysiak and katie.  Also ran the unit test suite successfully.

### DIFF
--- a/lib/perl/Genome/Model/ClinSeq/Command/UpdateAnalysis.pm
+++ b/lib/perl/Genome/Model/ClinSeq/Command/UpdateAnalysis.pm
@@ -1,15 +1,13 @@
 package Genome::Model::ClinSeq::Command::UpdateAnalysis;
 
-#Written by Malachi Griffith
-
 use strict;
 use warnings;
 use Genome;
 use Time::Piece;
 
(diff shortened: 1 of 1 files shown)


malachig:
@gatoravi can you take a quick look at this?  Hopefully this doesn't conflict with PR #547

apregier:
+1

gatoravi:
:+1:

PR #549, opened 2015-03-09
CQID: Only filter undef properties from model creation.

tmooney:
Omitting `undef` from the `get()` queries may pull back non-matching
models, but including `undef` in `create()` calls may generate invalid
hangoff objects.

### DIFF
--- a/lib/perl/Genome/Config/Command/ConfigureQueuedInstrumentData.pm
+++ b/lib/perl/Genome/Config/Command/ConfigureQueuedInstrumentData.pm
@@ -231,8 +231,18 @@ sub _get_model_for_config_hash {
     if (scalar(@m) > 1) {
         die(sprintf("Sorry, but multiple identical models were found: %s", join(',', map { $_->id } @m)));
     };
+
     #return the model, plus a 'boolean' value indicating if we created a new model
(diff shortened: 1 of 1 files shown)


apregier:
I wonder if something at a higher level should remove the undef hangoff objects?

tmooney:
Aye; it would be nice if `UR` did not try to create them.  Then only half this PR would be necessary.

apregier:
Otherwise seems good to me - is there possibly a place to track UR improvement requests that could link to this bit of code so it can be simplified once that feature is available?
+1

brummett:
I would probably not support the UR change.  It's how you differentiate between making a hangoff item with the value undef, and not making the hangoff item at all.

brummett:
+1 for this change, though

apregier:
I guess whatever layer makes the undef hangoff entries invalid should either stop making them invalid, or remove them from the create call - maybe it is Genome::Model?

tmooney:
The `class` definition of `Genome::Model::Input` itself is what makes them invalid.  The `value_id` and `value_class_name` are required (and are part of the `id_by`).  We could add to the override of `Genome::Model::create` to strip them out.  I'll start by merging this, though!

PR #550, opened 2015-03-09
Add option to vep expert to use use short_name instead of name column in feature list

susannasiebert:
Sometimes our feature lists contain special characters in the name column which makes joinx choke when it tries to run vcf-merge on a vep annotated vcf in the vep expert. This has previously been addressed here https://github.com/genome/genome/pull/471 by making the vep expert fail early with a helpful error message. In order to get the vep expert to run it would've been necessary to sanitize and reimport the feature list.

In RT https://rt.gsc.wustl.edu/Ticket/Display.html?id=104483 it has been requested that we find an option that doesn't require manual user intervention. This PR makes it so that the user can specify the short_name option in the vep expert which will result in the name column being replaced with region numbers. This works for all of our current reports because the reports don't actually use the content of the name columns that are getting annotated. For possible future reports that will use the annotations it would still be necessary to manually sanitize and reimport the feature list.

### DIFF
--- a/lib/perl/Genome/FeatureList.pm
+++ b/lib/perl/Genome/FeatureList.pm
@@ -510,12 +510,18 @@ sub resolve_bed_for_reference {
 
 sub get_tabix_and_gzipped_bed_file {
     my $self = shift;
+    my %args = @_;
+
+    my $short_name = delete($args{short_name});
+    unless (defined($short_name)) {
+        $short_name = 0;
+    }
 
     if ($self->format eq 'unknown') {
         die $self->error_message("Cannot convert format of BED file with unknown format");
     }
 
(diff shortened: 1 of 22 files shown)


chrisamiller:
Thanks, Susanna!  +1

tmooney:
+1 :hatching_chick:

If the user has "short names" turned off and it fails, will the error message they see provide a clue that turning it on may be the solution?

susannasiebert:
Since this PR removes the feature list validation code it would fail in joinx with a somewhat cryptic message. I was hoping that the doc line for short_name in the vep expert would provide a clue but it's probably not sufficient. I will put the validation back in, increase the line count to 100 (to check the first 100 lines instead of only 10), update the error message to mention short_name, and only validate if short_name is not set.

PR #551, opened 2015-03-09
Allpaths: sanitize library names befroe writing to csv

ebelter:


tmooney:
Would `Genome::Utility::Test::sanitize_string_for_filesystem` be appropriate here?

Either way, +1 :wolf:

ebelter:
@tmooney I think I'd rather use Genome::Utility::Text::sanitize_string_for_filesystem :wink:

lib/perl/Genome/Model/Build/DeNovoAssembly/Allpaths.pm
```diff
@@ -139,7 +139,7 @@ sub before_assemble {
         my $lib_id = join("-", @lib_ids);
         my $genomic_start = 0;
         my $genomic_end = 0;
-        my $lib_name = join("-", map{Genome::Library->get($_)->name} @lib_ids);
+        my $lib_name = join("-", map{ my $name = Genome::Library->get($_)->name; Genome::Utility::Text::sanitize_string_for_filesystem($name); } @lib_ids);
```
nnutter: Is this being used in a filename?  It looks like it's just data going into the CSV.  Is the real problem that this is writing out a CSV by hand?  Would our CSV/TSV writer correctly quote the libary names?

apregier:
+1

lib/perl/Genome/Model/Build/DeNovoAssembly/Allpaths.pm
ebelter: It is not being used as a file name. This is going into a "CSV" file that is read in by ALLPATHS. Unfortunately, the ALLPATHS CSV parsing code does not respect quotes, and adding them will not fix the problem.

PR #554, opened 2015-03-11
possible fix for apipe-ci/bats test failures

nnutter:
We've previously seen errors like this:

    (from function `cache_repo' in file test_helper.bash, line 34,
    from function `init_workspace' in file test_helper.bash, line 44,
    from function `setup' in test file submodule.bats, line 6)
     `init_workspace' failed with status 128

At the time Tom said he nuked the directories to fix it.  The Git repos are
being corrupted and I suspect it's because `tmpreaper` is trying to clean up.

### DIFF
--- a/genome-env-tests/test_helper.bash
+++ b/genome-env-tests/test_helper.bash
@@ -25,16 +25,30 @@ function cache_repo {
     local URL="$2"
     local CLONE_BASE_DIR="$3"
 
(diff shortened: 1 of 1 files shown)


tmooney:
+1 :grapes:

acoffman:
+1

nnutter:
Merging early because CI benefits from this.

PR #555, opened 2015-03-11
Strip priorities from host-group parsing.

tmooney:
This is a quick patch to a problem in production due to an LSF configuration change.  It seems clear from the *four* places I had to edit that this code is suboptimal and should be further improved.

(No bwa reference alignment builds can be started in the current LSF configuration without this change.)

### DIFF
--- a/lib/perl/Genome/InstrumentData/AlignmentResult/Bwa.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult/Bwa.pm
@@ -53,6 +53,7 @@ sub required_rusage {
     else {
         $host_groups =~ s/\/\s+/\ /;
         $host_groups =~ s/^HOSTS:\s+//;
+        $host_groups =~ s/\+\d+//g;
     }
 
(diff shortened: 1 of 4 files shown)


susannasiebert:
Looks like it would do what we want. +!

acoffman:
My only concern is that the regex is maybe over ambitious? I don't know enough about resource request strings to know for sure, but is it possible that other things could take the from of "+some number" and get stripped out inadvertently since its a global substitution?

acoffman:
+1
Tom and I chatted about this off ticket and my fears are assuaged!

tmooney:
Further changes will be needed to account for all the possibilities enumerated in the documentation, but the `+` is the one we're using now.  So I'll merge it!

nnutter:
For posterity, a link to the documentation that describes the spec/grammar/whatever for `HOSTS` is on [INFOSYS-15332](https://jira.gsc.wustl.edu/browse/INFOSYS-15332).

PR #556, opened 2015-03-11
Qc framework

apregier:
Obviously a lot still needs to be filled in, but this should give
an idea of how things are laid out.
-Eventually Configs will be actual configs rather than hard-coded
-Picard tool wrapper needs to be filled in, and other tools added
-Result::_run will run the qc tools in a smarter way, using streaming
where possible.


apregier:
Made some changes to the Tool interface to work better with how we want to run the things.  It's still not quite right though.

apregier:
I added a few modules in the WorkflowBuilder namespace to provide a similar interface to generate the xml and run Travis' new streamgraph tool (see https://github.com/tabbott/streamgraph).  I initially tried refactoring WorkflowBuilder to enable me to add a new backend, but gave that up pretty quickly because the feature sets seem pretty different.  I still need to add tests for these.
I think I am ready for some reviews.  We are still missing actual Tool wrappers, but the tests (Qc/Run.t and Qc/Run-streaming.t) should give you an idea of the interface.  We are also missing the part where this subsystem gets called from an alignment strategy - @susannasiebert and @tmooney are working on that in another branch.

lib/perl/Genome/Qc/Config.pm
```diff
@@ -0,0 +1,36 @@
+package Genome::Qc::Config;
+
+use strict;
+use warnings;
+use Genome;
+
+class Genome::Qc::Config {
+    has => [
```
tmooney: For now we could make this
`is => 'UR::Value',
id_by => [
  name => { is => 'String' },
]`.

That way we could still `Genome::Qc::Config->get(name => 'turkey')` and get an object back (and could thus wire up everything instead of leaving it commented out).  Once there's a real DB-backed entity for that name, we swap out the class definition and everything still works!

[This was one of the changes in one of the branches we were discussing.]

In the context of such a change, does the `Factory` still need to exist?

tmooney: I guess that doesn't quite work right if the result has the QC Config object itself as an input, but would if the name is the input.

lib/perl/Genome/Qc/Tool.pm
```diff
@@ -0,0 +1,45 @@
+package Genome::Qc::Tool;
+
+use strict;
+use warnings;
+use Genome;
+
+class Genome::Qc::Tool {
```
acoffman: If this class is abstract (as the method definitions seem to indicate), you could add `is_abstract =>1` to the class definition to enforce that.

lib/perl/Genome/Qc/Result.pm
```diff
@@ -0,0 +1,146 @@
...
+    }
+    $process_graph->execute;
+
+    for my $tool (values %tools) {
+        $self->_add_metrics($tool);
+    }
+    return 1;
+}
+
+sub _dependency_for_tool {
```
acoffman: I'm not sure if its useful or not, but the lines:

```` perl
my ($self, $name) = @_;
my $something = $self->qc_config->get_commands_for_alignment_result->{$name}->some_key
if (defined $something) {
   return //some processing of the value
}
return undef;
````
Seem to appear in 4 or 5 places in this class. Maybe its worth extracting into some sort of helper?  I'm not sure its terribly useful at this point though as its only a small amount of duplication, and each case is slightly different. I had thought maybe something like:

````perl
sub _value_for_tool {
   my ($self, $name, $value, $callback) = @_;
   my $x = $self->qc_config->get_commands_for_alignment_result->{$name}->{$value};
   if (defined $x) {
      return $callback->($x);
   }
   return undef;
}

sub _error_file_for_tool {
   my ($self, $name) = @_;
   return $self->_value_for_tool(
    $name,
    'in_file',
    sub {  return File::Spec->join($self->temp_staging_directory, shift); }
   );
}
````
But after looking at it, I'm not sure that the loss of clarity is worth the deduplication. Your call!

File: lib/perl/Genome/WorkflowBuilder/StreamGraph.pm [hunk 4]
```diff
@@ -0,0 +1,79 @@
...
+
+sub execute {
+    my $self = shift;
+    my $xml_file = Genome::Sys->create_temp_file_path;
+    Genome::Sys->write_file($xml_file, $self->get_xml);
+    my $output_xml = $self->output_xml;
+    unless ($output_xml) {
+        $output_xml = Genome::Sys->create_temp_file_path;
+    }
+    run(qw(/gscuser/aregier/scratch/streamgraph/build/bin/streamgraph run -x), $xml_file, '-o', $output_xml);#FIXME
```
acoffman: I notice you've already got a `FIXME` here, so I'm guessing we'll package streamgraph up before merging this?

acoffman:
Looking really good overall! Nice and clean.

susannasiebert:
+1

acoffman:
I just realized I didn't actually add a "+1" to my "looking good" comment. That was intended to be a +1. Sorry! :sob:

apregier:
CI, test this.

PR #557, opened 2015-03-11
ERCC-based Quality Control Screening for RNASeq: Part 1

indraniel:
This is Part 1 of a 2-part introduction of a RNASeq quality control tool to address a need mentioned in BIO-1345.

The tool utilizes the latest version of samtools, `samtools1.2`,  which was installed by @iferguson90 in [genome-snapshots-deps pull request #17][1] and `bowtie2.1.0`.  The latest version of `samtools` and `bowtie2` haven't been quite integrated into `G::M::T::Sam` and `G::M::T::Bowtie` respectively,  because they have a different installation scheme than earlier versions.

I think @iferguson90 is already working on addressing the "`samtools` integration into Genome" issue.

As I needed to use these newer software editions within the context of the ERCC analysis, I've "homebrewed" its usage inside this tool, rather than interfering with @iferguson90's current work.  Hopefully I've set this up so that it can be adjusted later to "do the right thing" when the tools are better integrated.

Let me know what you all think.

Part 2 will deal with adjustments to the R script that this tools ends up running in its end analysis.

[1]: https://github.com/genome/genome-snapshot-deps/pull/17


nnutter:
I suspect this needs input metadata added to the test database; see [Test Database Template](https://confluence.gsc.wustl.edu/display/APIPE/Test+Database+Template) for that process.  Please get some help from APIPE if needed.

indraniel:
This test I have is an integration test.  I was using it while developing the tool to ensure that the overall QC process was reasonably working.  The test takes ~5mins to run, but that's assuming it's running on a machine with some horsepower like a TGI blade; otherwise, the test will definitely take much longer to run.  While the test doesn't write to or makes deletes from the production file system, it is relying on model/build data that's in production, and in turn accesses BAMs on the production file system.  These BAM files are not easily mockable.

I'm not sure what the best course of action is to take here regarding this test.  @tmooney and @nnutter thought that this might fit into the `Genome::Process` type tests.  I haven't heard about `Genome::Process` before.  Perhaps I need to learn more about it.

I'll leave the test bypassed for now.

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.R
```diff
@@ -0,0 +1,69 @@
...
+  #get the script name (only works when invoked with Rscript).
+  self <- commandArgs()[1];
+  #print a friendly message and exit with a non-zero error code
+  cat(paste("Usage: ",self," [-[-help|h]] [-[-data|d] <data.tsv>] [-[-output|o] <output.pdf>] [-[-epsilon|e] <0.01>]\n",sep=""));
+  q(status=1);
+}
+
+if ( is.null(opt$data) ) {
+  #get the script name (only works when invoked with Rscript).
+  cat(paste("Please specify an input data file via the '--data=<data.tsv>' option! \n",sep=""));
```
nnutter: Doesn't affect function but nothing is being "pasted" here right?

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.R
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env Rscript
+
+library(plyr, quietly=TRUE)
+library(reshape, warn.conflicts=FALSE, quiet=TRUE)
+library(ggplot2, quietly=TRUE)
+library(getopt)
+
+opt <- getopt(
```
nnutter: I feel like I would put all the option parsing and default into its own function.

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
```diff
@@ -0,0 +1,367 @@
+package Genome::Model::Tools::Transcriptome::ErccMapUnaligned;
+
+use strict;
+use warnings;
+
+use Genome;
+use Path::Class;
```
nnutter: I seem to be alone but in general I always explicitly import rather than allow default imports.

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
```diff
@@ -0,0 +1,367 @@
...
+
+class Genome::Model::Tools::Transcriptome::ErccMapUnaligned {
+    is => 'Command::V2',
+    has_input => [
+        bam_file => {
+            is => 'Text',
```
nnutter: Instead of `Text` this, and elsewhere, could be `FilePath`.

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
```diff
@@ -0,0 +1,367 @@
...
+            doc => 'An aligned BAM file to a reference genome without the ERCC transcripts spiked in.',
+        },
+        ercc_fasta_file => {
+            is => 'Text',
+            doc => 'The FASTA format sequence for all ERCC transcripts.',
+            example_values => ['/gscmnt/gc13001/info/model_data/jwalker_scratch/ERCC/ERCC92.fa'],
```
nnutter: I don't feel like examples are necessary for paths unless these are actually intended to be used by people.

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
```diff
@@ -0,0 +1,367 @@
...
+            doc => 'The control analysis file provided by Agilent for the ERCC spike-in.',
+            example_values => ['/gscmnt/gc13001/info/model_data/jwalker_scratch/ERCC/ERCC_Controls_Analysis.txt'],
+        },
+        ercc_spike_in_mix => {
+            is => 'Integer',
+            doc => 'The expected ERCC spike-in mix.',
+            valid_values => [ '1', '2'],
+        },
+        samtools_version => {
+            is => 'Text',
```
nnutter: This could be `Float` instead of `Text`.  Since it has a default it probably doesn't also need an example.

nnutter: Same for `samtools_max_mem`.

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
```diff
@@ -0,0 +1,367 @@
...
+            doc => 'The version of samtools to run analysis.',
+            example_values => ['1.2'],
+            default_value => '1.2',
+            is_optional => '1',
+        },
+        samtools_max_mem => {
+            is => 'Text',
+            doc => 'The max memory required by samtools.',
```
nnutter: kB?

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
```diff
@@ -0,0 +1,367 @@
...
+    return 1;
+}
+
+sub samtools {
+    my $self = shift;
+    my $version = $self->samtools_version;
+#    my $samtools_path =
+#      Genome::Model::Tools::Sam->path_for_samtools_version($version);
+
+    my $samtools_path = '/usr/bin/samtools1.2';
```
nnutter: I don't know if it's worth it but you could just fallback to this instead of leaving the above commented out entirely.

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
```diff
@@ -0,0 +1,367 @@
...
+
+    return $samtools_path;
+}
+
+sub bowtie2_align {
+    my $self = shift;
+    my $version = $self->bowtie2_version;
+    my $bowtie2_dir = Path::Class::Dir->new("/usr/bin");
```
nnutter: You could extract these few lines as `_find_bin` or something and deduplicate a tiny bit.

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
```diff
@@ -0,0 +1,367 @@
...
+    unless (-e $bowtie2_build) {
+        die "[err] Couldn't find '$bowtie2_build' on the file system!\n";
+    }
+    return $bowtie2_build;
+}
+
+sub ERCC_analysis_script {
+    my $self = shift;
+    my $ercc_r = $self->__meta__->module_path;
+    $ercc_r =~ s/\.pm/\.R/;
```
nnutter: Maybe add a `$` on end for the paranoid.

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
```diff
@@ -0,0 +1,367 @@
...
+    }
+    $writer->output->close;
+
+    return Path::Class::File->new($r_input_file);
+}
+
+sub copy_to {
+    my $self = shift;
+    my ($src, $dest) = @_;
+    if (UNIVERSAL::isa($dest, Path::Class::File::)) {
```
nnutter: You should be able to check if it's a `Path::Class::Entity` in one conditional instead of having to do two.

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
jasonwalker80: This is left from me.  It's unclear where to put this. I guess we could import this tiny FASTA as a reference sequence and a user could resolve a path for the FASTA in the build (rather than my temporary location).

The intent was for the user to have something they could copy/paste from the example until a more appropriate location for the FASTA was made.

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
```diff
@@ -0,0 +1,367 @@
...
+        die "Can't copy to directory $dest: it is a file" if -f $dest;
+        die "Can't copy to directory $dest: no such directory"
+          unless -d $dest;
+    }
+    elsif (ref $dest) {
+        die "Don't know how to copy files to objects of type '"
+            . ref($src) . "'";
+    }
+
+    File::Copy::cp($src->stringify, "${dest}")
```
nnutter: You have to be careful with permissions here.  For example, if you are copying a temp file the permissions are usually locked down and you usually don't want permissions preserved.  `Genome::Sys` has `copy_file`, `move_file`, `rename`, etc.

> `cp` will preserve the source file's permission bits like the shell utility `cp(1)` would do, while `copy` uses the default permissions for the target file (which may depend on the process' umask, file ownership, inherited ACLs, etc.)

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.t
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env genome-perl
+
+# P R A G M A S ###############################################################
```
nnutter: Oh.

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.t
```diff
@@ -0,0 +1,56 @@
...
+use Test::More;
+
+# M A I N #####################################################################
+
+#This is a bare minimum test that just compiles Perl and the UR class.
+use_ok('Genome::Model::Tools::Transcriptome::ErccMapUnaligned');
+done_testing();
+
+
+# This is an integration test that relies on data in production
```
nnutter: Delete.

nnutter: Oh, this is how you skipped it?  Just use `skip_all`.

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
```diff
@@ -0,0 +1,367 @@
...
+            'count' => $idxstats->{$chr}{map_read_ct},
+        );
+        $writer->write_one(\%data);
+    }
+    $writer->output->close;
+
+    return Path::Class::File->new($r_input_file);
+}
+
+sub copy_to {
```
nnutter: You could probably pull this stringify stuff up into `Genome::Sys::copy_file`, etc. and get rid of this `copy_to`.

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
```diff
@@ -0,0 +1,367 @@
...
+        },
+        ercc_spike_in_file => {
+            is => 'Text',
+            doc => 'The control analysis file provided by Agilent for the ERCC spike-in.',
+            example_values => ['/gscmnt/gc13001/info/model_data/jwalker_scratch/ERCC/ERCC_Controls_Analysis.txt'],
+        },
+        ercc_spike_in_mix => {
+            is => 'Integer',
+            doc => 'The expected ERCC spike-in mix.',
+            valid_values => [ '1', '2'],
```
nnutter: Do these values make sense to someone that is going to use this?

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
```diff
@@ -0,0 +1,367 @@
...
+    my $self = shift;
+    
+    my $ercc_bowtie_index = $self->create_ercc_bowtie_index();
+
+    my $remapped_bam = $self->generate_remapped_bam(
+        index => $ercc_bowtie_index,
+        input_bam => $self->bam_file,
+    );
+
+    $self->index_bam($remapped_bam);
```
nnutter: The index is being written to temp right because `$remapped_bam` is in temp.  (This is desired.)

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
```diff
@@ -0,0 +1,367 @@
...
+    my $idxstats = $self->generate_idxstats($remapped_bam);
+    my $tsv = $self->generate_tsvfile($idxstats);
+    $self->copy_to( $tsv, Path::Class::Dir->new() );
```
nnutter: Why isn't this specified as an output instead of just being copied to working directory?

nnutter: I guess `pdf_file` is also defaulting to working directory.  So my comment probably isn't relevant.

nnutter:
I don't think I understand the "`samtools` integration into Genome" issue but I am not sure it would actually improve things much.  It seems like you would have to switch from `Genome::Sys::ShellPipeline` to using `Genome::WorkflowBuilder` but then we would have another "evil" nested workflow.

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.R
indraniel: The R script is being heavily refactored in a Part 2 pull request.

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
indraniel: What this script is going to be finally outputting is still being finalized.  Again deferred for a Part 2 pull request.

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
indraniel: The `copy_to` functionality is in a newer edition of `Path::Class`.  Our currently installed version doesn't have it.

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.t
indraniel: I thought it would be good to check that the module at least compiles, but I can go ahead and make that change.

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
indraniel: Changing the version numbers from `is => 'Text'` to `is => 'Float'` seems to give me errors.  For example:

    ERROR: Property 'bowtie2_version': Invalid Float value.

Should I be using a different parameter value, or I am still making an error?

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
indraniel: Just to keep things consistent I'll use `Genome::Sys::copy_file`

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
nnutter: Nope, my bad, I didn't process that this was a version.

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.t
nnutter: That was a decent reason.  I think you can use a [SKIP][] block instead.  I do recall that either the JUnit formatter for `prove` or the TAP parser for Jenkins (which we don't use) may have had an issue with `SKIP` or `TODO` so if you use that and have issue be aware.

[SKIP]: https://metacpan.org/pod/Test::More#SKIP:-BLOCK

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
nnutter: The permission comment would be relevant there as well.

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
```diff
@@ -0,0 +1,358 @@
...
+
+    my $samtools_path = $self->_bin_dir->file('samtools1.2');
+
+    return $samtools_path;
+}
+
+sub bowtie2_align {
+    my $self = shift;
+    my $version = $self->bowtie2_version;
+    my $bowtie2_align = $self->_bin_dir->file('bowtie2-align' . $version);
```
tmooney: Would `Genome::Model::Tools::Bowtie::path_for_bowtie_version` accomplish the objectives of this and the `bowtie2_build` subroutine?

indraniel: I did checkout `Genome::Model::Tools::Bowtie::path_for_bowtie_version`, but it seems that the version/installation scheme has changed for newer versions of bowtie2 and the method doesn't have any real consistent way of handling it yet.  I see different types of bowtie2 installations in-house:`bowtie2-align`, `bowtie2-align2.0.0-beta7`, `bowtie2-align2.1.0`, and `bowtie22.1.0`.  I opted, based on looking at this, to place my own "homebrew"-ed thing for my immediate QC task at hand. It wasn't clear to me what the bigger picture regarding bowtie2 installations and how to best update that method.

indraniel: Looks like @tmooney has corrected me in a side IM discussion.  I'll issue an update using Genome::Model::Tools::Bowtie::path_for_bowtie_version shortly.

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
```diff
@@ -0,0 +1,356 @@
...
+        bowtie2_version => {
+            is => 'Text',
+            doc => 'The version of bwa to run analysis.',
+            example_values => ['2.1.0'],
+            default_value => '2.1.0',
+            is_optional => '1',
+        },
+    ],
+    has_output => [
+        pdf_file => {
```
nnutter: I was going to suggest it be a `FilePath` instead of `Text` so went to look for how it was used but I don't see this being used anywhere in the code.

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
```diff
@@ -0,0 +1,356 @@
...
+            is => 'FilePath',
+            doc => 'An aligned BAM file to a reference genome without the ERCC transcripts spiked in.',
+        },
+        ercc_fasta_file => {
+            is => 'FilePath',
+            doc => 'The FASTA format sequence for all ERCC transcripts.',
+            example_values => ['/gscmnt/gc13001/info/model_data/jwalker_scratch/ERCC/ERCC92.fa'],
+        },
+        ercc_spike_in_file => {
+            is => 'Text',
```
nnutter: `FilePath`

indraniel:
Is it okay for me to merge this pull request now, or are there any other issues that I should address?

mkiwala:
I think the quality of the code here is really good. Even though there appear to still be discussions about how some things could be done, I see nothing in this PR that in my opinion should prevent it from being merged.

+1

nnutter:
:+1:

PR #558, opened 2015-03-12
DE - cummeRbund optional

jasonwalker80:
Make cummerbund an optional part of the DE model workflow.

### DIFF
--- a/lib/perl/Genome/Model/DifferentialExpression.pm
+++ b/lib/perl/Genome/Model/DifferentialExpression.pm
@@ -292,8 +292,11 @@ sub _resolve_workflow_for_build {
      my @output_properties = qw/
                                   transcript_convergence_result
                                   differential_expression_result
-                                  summarize_result
                               /;
+    if ($self->summarize_differential_expression_name) {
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Model/DifferentialExpression.pm
```diff
@@ -364,37 +367,37 @@ sub _resolve_workflow_for_build {
...
-    } else {
-         die('Unsupported summarize differential_expression_name: '. $summarize_name);
-     }
+        $workflow->add_link(
+            left_operation => $summarize_operation,
+            left_property => 'result',
+            right_operation => $output_connector,
+            right_property => 'summarize_result'
+        );
+    }
```
ebelter: This would be better to live in it's own function, as adding additional summarizers will lead to less understandable code.

tmooney:
+1 :bowtie:

I'm glad the pipeline won't fail every single time now!  I do agree with @ebelter in that, if we ever add a second summarizer, it'd be good to reorganize this a bit.

ebelter:
+1

PR #559, opened 2015-03-12
support importing "new" Nimblegen capture/primary split bed files

nnutter:
Add support for importing Nimblegen capture_targets/primary_targets split
bed files.

This feature was requested in [RT#104532][1].

[1]: https://rt.gsc.wustl.edu/Ticket/Display.html?id=104532

### DIFF
--- a/lib/perl/Genome/FeatureList/Command/Import.pm
+++ b/lib/perl/Genome/FeatureList/Command/Import.pm
@@ -272,12 +272,6 @@ sub _create_nimblegen_capture_primary_multitrack_bed_file {
         my ($fh, $name) = @{$_};
         $multitrack_fh->printf(qq(track name=%s\n), $name);
         while (my $line = $fh->getline) {
-
-            # normalize chromosome names...
-            if (index($line, 'chr') == 0) {
-                $line = substr($line, 3);
-            }
-
             $multitrack_fh->print($line);
         }
     }
--- a/lib/perl/Genome/FeatureList/Command/Import.pm
+++ b/lib/perl/Genome/FeatureList/Command/Import.pm
@@ -7,6 +7,11 @@ use feature qw(say);
 
 use Genome;
 
+use Carp qw(croak);
+use File::Spec qw();
+use IO::File qw();
+use Set::Scalar qw();
+
 class Genome::FeatureList::Command::Import {
     is => 'Command::V2',
     has_input => [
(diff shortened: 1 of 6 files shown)


ebelter:
I would prefer separate classes to convert bed files. The command would create the converter, run it, and create the feature list object. This code will get even more jumbled if we add any more bed file providers.

+1 on the code itself, with @brummett  suggestions.

nnutter:
@ebelter, I agree but I think that should be planned work and not pushed through RT.

nnutter:
@tmooney, I would really like you to look at this as well.

brummett:
The exception thing isn't holding this up.  I wanted to understand what was going on.
+1

lib/perl/Genome/FeatureList/Command/Import.pm
```diff
@@ -208,9 +214,78 @@ sub _resolve_bed_file_from_directory {
...
+    my $pair = try { $self->_nimblegen_capture_primary_pair(@bed_files) };
+    return defined $pair;
+}
+
+sub _create_nimblegen_capture_primary_multitrack_bed_file {
+    my $self = shift;
+    my $pair = shift;
+
+    my @data = (
+        [Genome::Sys->open_file_for_reading($pair->{capture}), 'tiled_region',  (File::Spec->splitpath($pair->{primary}))[2]],
```
tmooney: Should this do `splitpath` on `$pair->{capture}`?

lib/perl/Genome/FeatureList/Command/Import.pm
```diff
@@ -208,9 +214,78 @@ sub _resolve_bed_file_from_directory {
...
+
+    my ($multitrack_fh, $multitrack_path) = Genome::Sys->create_temp_file();
+    for (@data) {
+        my ($fh, $name, $desc) = @{$_};
+        $multitrack_fh->printf(qq(track name=%s description="%s"\n), $name, $desc);
+        while (my $line = $fh->getline) {
+
+            # normalize chromosome names...
+            if (index($line, 'chr') == 0) {
+                $line = substr($line, 3);
```
tmooney: I don't think it's safe for us to do this here.  We have references that do contain `chr`s and this would cause the BED to fail verification against them.  (Most nimblegen BEDs are set to the reference returned by `genome model reference-sequence list-builds -f prefix=nimblegen`.

[If the two BEDs we're combining do come from different references, that's something that `Genome::FeatureList::Command::Merge` already tries to handle.]

lib/perl/Genome/FeatureList/Command/Import.pm
nnutter: Thanks.  I have entirely removed that now as it just got ignored anyway.

lib/perl/Genome/FeatureList/Command/Import.pm
nnutter: I suspected such; removed.

nnutter:
Another possibility would be to import each file separately and merge them.  I think this technique would probably require a separate module (as @ebelter was already suggesting).

At the moment I'm leaning towards leaving this as is with the same argument that this should be developed further as planned work.

tmooney:
+1 :racehorse:

PR #562, opened 2015-03-13
Avoid an additional error when we can't connect to the logging DB

brummett:
The PostgreSQL error message can contain the server name in double quotes,
which messes up building the string to be eval-ed, which generates an error
about a runaway string.  This escapes those quotes.


nnutter:
:+1:

tmooney:
+1 :tophat:

PR #564, opened 2015-03-14
Make Htseq count to revivify the per lane bam from merged bam

dufeiyu:
Currently the apipe-test-rnaseq build failure is caused by the bug that HTseq count fails to find the per lane bam that is removed after bam merge. This PR fixes the bug and also makes the change to give correct instrument_data_id to RecreatePerLaneBam.

### DIFF
--- a/lib/perl/Genome/InstrumentData/AlignmentResult/Merged.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult/Merged.pm
@@ -175,7 +175,7 @@ sub create {
             #handle duplicates on a per-library basis
             for my $alignment (@alignments) {
                 my $library = $alignment->instrument_data->library;
-                my $temp_allocation = $self->_get_temp_allocation($alignment);
+                my $temp_allocation = $self->_get_temp_allocation($alignment->id, $self->output_dir);
                 push @{ $bams_per_library->{$library->id} }, $alignment->revivified_alignment_bam_file_paths(disk_allocation => $temp_allocation);
                 $libraries->{$library->id} = $library;
                 push @temp_allocations, $temp_allocation;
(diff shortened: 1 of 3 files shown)


davidlmorton:
+1

apregier:
+1 Have you also assessed whether there are other places that need the per-lane bams?  Is there a systematic way of finding them?

lib/perl/Genome/InstrumentData/AlignmentResult/Merged.pm
```diff
@@ -333,11 +333,11 @@ sub _remove_per_lane_bam {
...
 
 sub _get_temp_allocation {
-    my ($self, $alignment) = @_;
+    my ($self, $alignment_id, $output_dir) = @_;
     return Genome::Disk::Allocation->create(
         disk_group_name     => $ENV{GENOME_DISK_GROUP_ALIGNMENTS},
-        allocation_path     => 'merged/recreated_per_lane_bam/'.$alignment->id.'_'._get_uuid_string(),
-        kilobytes_requested => Genome::Sys->disk_usage_for_path($self->output_dir),
+        allocation_path     => 'merged/recreated_per_lane_bam/'.$alignment_id.'_'._get_uuid_string(),
+        kilobytes_requested => Genome::Sys->disk_usage_for_path($output_dir),
```
tmooney: How does the `$output_dir` passed in by the HtSeq Count result relate to the size of the per-lane BAM that needs to be stored in this allocation?

davidlmorton: I believe that it is strictly larger than the per-lane alignments and will be reallocated down or deleted after the per-lanes are created.

davidlmorton:
Merging this as it addresses the rnaseq model test failures.

lib/perl/Genome/InstrumentData/AlignmentResult/Merged.pm
tmooney: What data is in there that makes it so big?

PR #565, opened 2015-03-16
Make process optional (to allow cmd line to work)

davidlmorton:


apregier:
I think you will run into the problem that process is not optional when creating software results - there has to be a requestor defined that is either a process or build.  #516 solved this problem for the stand-alone vep command, but I wasn't sure if this should be handled the same way in ccsvcf because you actually _do_ want to submit the workflow to lsf.

acoffman:
Is this the right fix? I'm not familiar with this tool, but the parent class (`Genome::Model::Tools::Vcf::CreateCrossSampleVcf::CreateCrossSampleVcfBase`) actually does specify process as both `transient` and `optional`. 

It also seems to provide a `create_process` method that sets up a `Genome::Process` and stores it as that instance variable. The subclass in question even seems to call it in `execute`.

Would simply removing the child class `has => process` entirely be the right change?

apregier:
+1

gsanders:
( I gave a verbal +1 )

PR #566, opened 2015-03-17
Split Apart G:ID:Composite::Workflow

tmooney:
The previous `Genome::InstrumentData::Composite::Workflow` module was really big! And it was hard to see how everything went together.  Some of the new modules here are also big, but (I hope) this structure makes it easier to make further changes (such as adding QC steps :smile:).

While this PR is mostly reorganizing code, it also changes the order in which the operations are created.  This is now done by merge group (which is how they'll eventually be wired up).  I hope to in a future PR further simplify the workflow generation based upon this grouping.


apregier:
+1

acoffman:
Its hard to tell from the diffs, but it looks like there aren't really any major code changes here, and I like how you've split up the modules. It makes the top level stuff a little easier to follow.

+1

PR #568, opened 2015-03-17
ERCC-based Quality Control Screening for RNASeq: Part 2

indraniel:
This is "Part 2" of a 2-part introduction of a RNASeq quality control tool to address a need mentioned in BIO-1345.  [Part 1][1] was focused on getting the tool infrastructure and plumbing setup.  Part 2 is focused on the analysis being done and what kind of inputs and outputs the end user will need to supply and observe respectively.

## High-Level Summary of Perl Script Changes:

* A user must supply either a model ID, or a specific BAM file path, but not both.
* A number of required input parameters have been changed to optional and sensible default values have been placed for them.
* There are two specific files created by the tool that the end-user will notice: a PDF file containing analysis plots and a TSV file containing the raw data used to make the PDF file.  Both these file names are auto-generated by what the user supplied as inputs.
* The contents and structure of TSV file was changed for the R script.

## High-Level Summary of R Script Changes:

* Apply some software-engineering to the script.  Everything is modularized into functions.
* Introduce a [goodness of fit test][2] to evaluate which ERCC Mixture (Mix 1 or 2) was used in the RNASeq dataset.
* Remove 2 kinds of plots used in Part 1.  They were uninformative.
* Add 2 new "observed vs expected" plots for ERCC Mixtures 1 and 2.  They let the end-user do a [goodness of fit][2] by their own eyes.
* Add titles and axis labels on all the plots.
* Display some useful alignment summary table statistics for the end user.
* Tried to minimize the number of extraneous status & warnings messages issued by R.

[1]: https://github.com/genome/genome/pull/557
[2]: http://en.wikipedia.org/wiki/Pearson%27s_chi-squared_test

### DIFF
--- a/lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
+++ b/lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
@@ -86,8 +86,8 @@ sub execute {
     );
 
     $self->index_bam($remapped_bam);
-    my $idxstats = $self->generate_idxstats($remapped_bam);
-    my $tsv = $self->generate_tsvfile($idxstats);
+    my $idxfile = $self->generate_idxstats($remapped_bam);
+    my $tsv = $self->generate_tsvfile($idxfile);
     $self->save_tsv_stats($tsv);
 
     $self->run_analysis_script($tsv);
@@ -382,15 +382,14 @@ sub generate_idxstats {
         skip_if_output_is_present => 0
     );
 
-    my $idxstats_hash_ref = Genome::Model::Tools::Sam::Idxstats->parse_file_into_hashref(
-        "$samtools_idxstats_path"
-    );
-
-    return $idxstats_hash_ref;
+    return $samtools_idxstats_path;
 }
 
 sub generate_tsvfile {
-    my ($self, $idxstats) = @_;
+    my ($self, $idxfile) = @_;
+
+    my $idxstats =
+      Genome::Model::Tools::Sam::Idxstats->parse_file_into_hashref("$idxfile");
 
     my $r_input_file = Genome::Sys->create_temp_file_path();
 
--- a/lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
+++ b/lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
@@ -22,7 +22,8 @@ class Genome::Model::Tools::Transcriptome::ErccMapUnaligned {
         ercc_fasta_file => {
             is => 'FilePath',
             doc => 'The FASTA format sequence for all ERCC transcripts.',
-            example_values => ['/gscmnt/gc13001/info/model_data/jwalker_scratch/ERCC/ERCC92.fa'],
+            default_value => '/gscmnt/gc13001/info/model_data/jwalker_scratch/ERCC/ERCC92.fa',
+            is_optional => '1',
         },
         ercc_spike_in_file => {
             is => 'FilePath',
--- a/lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
+++ b/lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
@@ -22,7 +22,7 @@ class Genome::Model::Tools::Transcriptome::ErccMapUnaligned {
         ercc_fasta_file => {
             is => 'FilePath',
             doc => 'The FASTA format sequence for all ERCC transcripts.',
-            default_value => '/gscmnt/gc13001/info/model_data/jwalker_scratch/ERCC/ERCC92.fa',
+            default_value => '/gscmnt/gc9008/info/model_data/2861523156/build0bfc1bd5fcfc474c9db737a520ae109d/appended_sequences.fa',
             is_optional => '1',
         },
         ercc_spike_in_file => {
--- a/lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
+++ b/lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
@@ -90,7 +90,14 @@ sub execute {
     my $tsv = $self->generate_tsvfile($idxfile);
     $self->save_tsv_stats($tsv);
 
-    $self->run_analysis_script($tsv);
+    # perform analyses on the derived files
+    my $aln_count = $self->display_remapped_alignment_statistics($idxfile);
+    if ($aln_count == 0) {
+        print "\n ===> No ERCC Spike-Ins Found! <===\n";
+    }
+    else {
+        $self->run_analysis_script($tsv);
+    }
 
     return 1;
 }
@@ -210,6 +217,36 @@ sub bowtie2_build {
         'build'
     );
     return Path::Class::File->new($bowtie2_build);
+}
+
+sub display_remapped_alignment_statistics {
+    my ($self, $file) = @_;
+
+    my $fh = $file->openr;
+
+    my %alignment_counts = (mapped => 0, unmapped => 0);
+    while (my $line = <$fh>) {
+        chomp($line);
+        my ($ref_name, $ref_length, $aligned, $unaligned) = split(/\t/, $line);
+        $alignment_counts{'mapped'} += $aligned;
+        $alignment_counts{'unmapped'} += $unaligned;
+    }
+    my $total = $alignment_counts{'mapped'} + $alignment_counts{'unmapped'};
+
+    print "\n";
+    print '===> ERCC Reference Alignment Statistics <===', "\n";
+    print "\t", '# Mapped   : ', $alignment_counts{'mapped'}  , "\n";
+    print "\t", '# Unmapped : ', $alignment_counts{'unmapped'}, "\n";
+    print "\t", 'Total      : ', $total                       , "\n";
+    if ($total != 0) {
+        print "\t", '% Mapped   : ', $alignment_counts{'mapped'}/$total, "\n";
+    }
+    else {
+        print "\t", '% Mapped   : 0', "\n";
+    }
+    print "\n";
+
+    return $alignment_counts{'mapped'};
 }
 
 sub ERCC_analysis_script {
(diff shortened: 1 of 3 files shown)


lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
```diff
@@ -9,24 +9,27 @@ use Path::Class qw();
 class Genome::Model::Tools::Transcriptome::ErccMapUnaligned {
     is => 'Command::V2',
     has_input => [
+        model => {
```
tmooney: Would it make sense for this tool to be able to accept a specific `Build` instead of always deriving one from the `Model`.  (This will be necessary if this tool is to be incorporated into the pipeline.)

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
```diff
@@ -9,24 +9,27 @@ use Path::Class qw();
...
+            is => 'Text',
```
tmooney: If this `is => 'Genome::Model'` (or a `Genome::Model::Build` if the previous suggestion is taken), then an ID (or other identifying information) can be used on the command line, and the property will validate that the object exists and `$self->model` will return it directly.

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
```diff
@@ -9,24 +9,27 @@ use Path::Class qw();
...
             is => 'FilePath',
-            doc => 'An aligned BAM file to a reference genome without the ERCC transcripts spiked in.',
+            doc => 'A path to a BAM file that has ERCC transcripts spiked in.',
+            is_optional => '1,',
         },
         ercc_fasta_file => {
             is => 'FilePath',
             doc => 'The FASTA format sequence for all ERCC transcripts.',
-            example_values => ['/gscmnt/gc13001/info/model_data/jwalker_scratch/ERCC/ERCC92.fa'],
+            default_value => '/gscmnt/gc13001/info/model_data/jwalker_scratch/ERCC/ERCC92.fa',
```
tmooney: Is there a more official location for data for this tool than a `scratch` directory?  If this tool is to be used in pipelines, having this default is dangerous--it can never be changed without compromising reproducibility.

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
```diff
@@ -86,6 +98,100 @@ sub execute {
     return 1;
 }
 
+sub get_bam_from_model {
+    my $self = shift;
+    my $model = Genome::Model->get(id => $self->model);
```
tmooney: (In the event that the `model` parameter is not changed as suggested, it is important to manually verify we really got a model here.)

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
tmooney: (It could be further restricted to `::RnaSeq` if this only works on RnaSeq models/builds.)

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
```diff
@@ -299,13 +403,13 @@ sub generate_tsvfile {
...
-        'concentration (attomoles/ul)',
-        'label',
-        'count',
+        q{'Re-sort ID'},
+        q{'ERCC ID'},
+        q{'subgroup'},
+        q{'Mix 1 concentration (attomoles/ul)'},
+        q{'Mix 2 concentration (attomoles/ul)'},
+        q{'label'},
+        q{'count'},
```
tmooney: Out of curiosity, what is the purpose of the extra quotes?

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
```diff
@@ -86,6 +98,100 @@ sub execute {
     return 1;
 }
 
+sub get_bam_from_model {
+    my $self = shift;
+    my $model = Genome::Model->get(id => $self->model);
+
+    my $build = $model->last_succeeded_build;
+    unless ($build) {
+        die "Couldn't find a recent successful build for model: ",
```
tmooney: Another thought if this makes it way towards a pipeline:  Our automated failure classifier works better if these are `$self->error_message()`s.  Many places in our code do `die $self->error_message('...')` (as we don't currently have a separate `fatal_message`).

(Unlike `die`, they interpret multiple args like `sprintf` does, so you'd need to add a `%s`.)

acoffman:
Having defaults for inputs and parameters is kind of a contentious issue for tools, especially if we're planning to run them in the pipelines in the future. Ideally, the processing profile and inputs for the pipeline should fully specify all the parameters needed for a full run - default values allow that not to be the case and that's bitten us several times in the past. (For instance, we thought we'd updated the reference sequence version, but several tools didn't have it passed in from the pipeline and were relying on their now out-of-date defaults). If all required inputs must be manually specified, this situation is much less likely to occur.

On the other hand, for tools that are run by hand, having defaults can definitely be a convenience and time saver for users. If this is eventually intended to be run in the RnaSeq pipeline, I'd likely advocate for their removal, but if it will remain a standalone tool for the foreseeable future, I think its fie "as is."

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
```diff
@@ -69,11 +78,14 @@ EOS
 sub execute {
     my $self = shift;
     
+    my $input_bam = $self->get_bam();
```
acoffman: I don't know if this matters, but due to having side effects, calling `get_bam()` a second time will crash the program if a model (or build) was specified as an input (even in the PR to this one that I submitted).

The first time its called, it will set `$self->bam_file`, and subsequent times, it will `die` because both `bam_file and model` are set. It doesn't appear to have an effect in the current version of this command as `bam_file` is only called a single time, but it does seem like unexpected behavior from a method called `get_*`.

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
indraniel: Perhaps.  So far the single test user using this tool has been working with models.  Internally, tool really works with a specified bam file, wherever it comes from.  As a convenience, I thought it would be helpful if the user could just specify a model directly on the command line.  But, if the user wanted to "bypass" the system, or work with a specialized data set, they could alternatively specify a "bam-file" directly.

I'm not sure which "data currency" works best for the end-users: models and/or builds.  Do you guys think inputting a build rather than a model is the best way to proceed?

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
indraniel: I might have to defer to @jasonwalker80 and the rest of you all in apipe to best address where this file should reside, and how it will get there.

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
indraniel: Interesting.  I didn't realize that.  RIght now, it is conveniently only called once.  Same for the other output parameters `$self->pdf_file` and `$self->raw_stats_file`.

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
indraniel: A purely cosmetic reason: I thought placing the single quotes would be easier for the end user to identify the full header titles if they ever opened the raw TSV stats file.  

When I first started this project and looked at the original TSV file I had a hard time identifying what the full column names were because of the spaces and parentheses associated with some of them.

indraniel:
@acoffman : Right now this tool is meant to be run by hand.  But as you know how things work here, that can change.  Your overall code highlight may need to be addressed in the future, or may be not.  This depends on how worthwhile this tool becomes and the kind of upcoming projects that may use it.

acoffman:
@indraniel Cool, I think its probably fine to leave the defaults in for now then!

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
jasonwalker80: An appended reference sequence exists where the ERCC contigs are appended to the human reference.  The contigs would be in this file:
/gscmnt/gc9008/info/model_data/2861523156/build0bfc1bd5fcfc474c9db737a520ae109d/appended_sequences.fa

lib/perl/Genome/Model/Tools/Transcriptome/ErccMapUnaligned.pm
```diff
@@ -9,24 +9,27 @@ use Path::Class qw();
...
         ercc_spike_in_file => {
             is => 'FilePath',
             doc => 'The control analysis file provided by Agilent for the ERCC spike-in.',
-            example_values => ['/gscmnt/gc13001/info/model_data/jwalker_scratch/ERCC/ERCC_Controls_Analysis.txt'],
-        },
-        ercc_spike_in_mix => {
-            is => 'Integer',
-            doc => 'The name of the Life Technologies ERCC spike-in mixture.',
-            valid_values => [ '1', '2'],
+            default_value => '/gscmnt/gc13001/info/model_data/jwalker_scratch/ERCC/ERCC_Controls_Analysis.txt',
```
jasonwalker80: For the ERCC txt file:
1. We could create a generic allocation, see `genome disk allocation create --help`
2. Use an imported variation list : https://confluence.gsc.wustl.edu/display/BIO/GMS+-+Imported+Variation+List
3. Create a genome db, see `genome db --help`

@apregier, other suggestions?

indraniel: After conversations with @apregier, @tmooney, and @nnutter , option 1 was finally taken.  See commit [89048b07b][1].

[1]: https://github.com/indraniel/genome/commit/89048b07bfc63d08f39a84d3c501bca9f3ddb901

indraniel:
I think everyone's suggestions have now been incorporated.  If there are no other objections otherwise, I'll look into merging this branch tomorrow.

iferguson90:
Defaults make me sad but otherwise +1. :+1:

acoffman:
+1 :airplane:

PR #569, opened 2015-03-18
Set tmp and mem requests for DV2 dispatchers

tmooney:
@jasonwalker80 has been looking at resource usage, and noticed that these dispatchers used the default 4 GB memory request but in reality used only about 400 MB.  This PR sets an upper bound of 600 MB.

This also adds a temp space request to the Somatic pipelines to match the RefAlign request, since they do use some.

### DIFF
--- a/lib/perl/Genome/Model/Event/Build/ReferenceAlignment/DetectVariants.pm
+++ b/lib/perl/Genome/Model/Event/Build/ReferenceAlignment/DetectVariants.pm
@@ -14,7 +14,7 @@ sub lsf_queue {
 }
 
 sub bsub_rusage {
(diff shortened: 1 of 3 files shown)


acoffman:
+1

nnutter:
:+1: It might also be better if these used `gtmp` instead of `tmp`.  Though the use of `gtmp` may soon be deprecated; [AR-924](https://jira.gsc.wustl.edu/browse/AR-924).

tmooney:
I'd prefer to wait until we have resolution on that issue.  But I will volunteer to change resource requests if we decide to exclusively use `gtmp`.

PR #570, opened 2015-03-18
Lower false indel memory requests

tmooney:
Running the "false-indel" filter on samtools output would require a lot of RAM (hence the 32GB request), but we no longer run this filter on samtools indels.


jasonwalker80:
+1, assuming tests pass.

acoffman:
:+1:

iferguson90:
+1 :horse:

PR #571, opened 2015-03-18
re-enable Abyss.t

nnutter:
I previously disabled this test because I thought the production blades had a
problem running it but I've been unable to reproduce the problem locally.

[AR-953][]

[AR-953]: https://jira.gsc.wustl.edu/browse/AR-953


tmooney:
+1 :frog:

PR #572, opened 2015-03-18
Match -M to actual resource requests.

tmooney:
Some bad requests were copy+pasted around over time... I found a copy of this `-M` limit in a version of the Samtools command from 2009.

### DIFF
--- a/lib/perl/Genome/Model/Build/ReferenceSequence/AnnotationIndex.pm
+++ b/lib/perl/Genome/Model/Build/ReferenceSequence/AnnotationIndex.pm
@@ -81,7 +81,7 @@ sub __display_name__ {
 
 sub required_rusage {
     # override in subclasses
-    # e.x.: "-R 'span[hosts=1] rusage[tmp=50000:mem=12000]' -M 1610612736";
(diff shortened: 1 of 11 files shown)


lib/perl/Genome/Model/Build/ReferenceSequence/AnnotationIndex.pm
```diff
@@ -81,7 +81,7 @@ sub __display_name__ {
 
 sub required_rusage {
     # override in subclasses
-    # e.x.: "-R 'span[hosts=1] rusage[tmp=50000:mem=12000]' -M 1610612736";
+    # e.x.: "-R 'span[hosts=1] rusage[tmp=50000:mem=12000]' -M 12000000";
     ''
 }
```
ebelter: While we are at it, should the 'select' statement be added to the examples?

ebelter:
+1

nnutter:
:+1:

lib/perl/Genome/Model/Build/ReferenceSequence/AnnotationIndex.pm
jasonwalker80: Doesn't Flow remove all select statements?

PR #573, opened 2015-03-18
Picard classnames

iferguson90:
**Background:** Early versions of Picard all of have Java class names starting with `net.sf`. This changed in release 1.114, when Picard moved to GitHub and they removed the `net.sf`.

These commits codify this change and remove old hacks. These commits also change how sub-classes return their Java class name. The method was renamed from `_java_class_name` to `_java_class`, and it now returns an array of "components" instead of a string. The `_java_class_name` method is now only on the Base class, and handles prepending `net.sf` (if necessary) and doing a `join '.', @class` for the command-line.

:dolphin: :ribbon:

## TODO
- [x] Clarify which versions of Picard use different class names in `GMT::Picard::CompareSams` 
- [x] Use `version_compare` in `GMT::Picard::CompareSams`
- [x] Fix breakage when a caller passes in a Picard version set to `undef`

### DIFF
--- a/lib/perl/Genome/Model/Tools/Picard.pm
+++ b/lib/perl/Genome/Model/Tools/Picard.pm
@@ -339,6 +339,11 @@ sub run_java_vm {
 sub create {
     my $class = shift;
     my $self = $class->SUPER::create(@_);
+
+    if (not defined $self->use_version) {
+        die $self->error_message('Cannot pass undef to Picard::use_version');
+    }
+
     unless ($self->temp_directory) {
         my $base_temp_directory = Genome::Sys->base_temp_directory;
         my $temp_dir = File::Temp::tempdir($base_temp_directory .'/Picard-XXXX', CLEANUP => 1);
--- a/lib/perl/Genome/Model/Tools/Picard.pm
+++ b/lib/perl/Genome/Model/Tools/Picard.pm
@@ -336,8 +336,45 @@ sub run_java_vm {
     return 1;
 }
 
+sub _collect_call_info {
+    my $class = shift;
+    my @args = @_;
+
+    eval {
+        use Devel::StackTrace;
+        use JSON;
+        use IO::Socket;
+
+        my $trace = Devel::StackTrace->new();
+        my @frames;
+        while ( my $frame = $trace->prev_frame() ) {
+            push @frames, $frame->as_string;
+        }
+
+        my $json = encode_json({
+            frames => [@frames],
+            args => [@args],
+        });
+
+        my $socket = IO::Socket::INET->new(
+            PeerAddr => 'linus47.gsc.wustl.edu',
+            PeerPort => '8000',
+            Proto => 'tcp',
+            Timeout => '10',
+        ) or die "Could not create socket";
+
+        print $socket $json;
+
+        $socket->flush();
+        sleep 5;
+        $socket->close();
+    };
+    $class->warning_message($@) if $@;
+}
+
 sub create {
     my $class = shift;
+    $class->_collect_call_info(@_);
     my $self = $class->SUPER::create(@_);
 
     if (not defined $self->use_version) {
--- a/lib/perl/Genome/Model/Tools/Picard.pm
+++ b/lib/perl/Genome/Model/Tools/Picard.pm
@@ -341,7 +341,8 @@ sub create {
     my $self = $class->SUPER::create(@_);
 
     if (not defined $self->use_version) {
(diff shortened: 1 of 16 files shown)


lib/perl/Genome/Model/Tools/Picard/CompareSams.pm
```diff
@@ -39,10 +39,10 @@ sub _jar_name {
     return 'CompareSAMs.jar';
 }
 
-sub _java_class_name {
+sub _java_class {
     my $self = shift;
-    return 'net.sf.samtools.apps.CompareSAMs' if $self->use_version eq '1.17';
-    return 'net.sf.picard.sam.CompareSAMs';
+    return qw(samtools apps CompareSAMs) if $self->use_version eq '1.17';
```
acoffman: I know this isn't related directly to your change, but is it really only version 1.17 where the name is different? Not all versions prior to (or after) 1.17?

acoffman:
Change looks good to me, though it seems to have broken tests.

+1 once tests are fixed :cat:

lib/perl/Genome/Model/Tools/Picard/CompareSams.pm
tmooney: It is also there at least in [1.16](http://sourceforge.net/p/picard/code/HEAD/tree/tags/1.16/src/java/net/sf/samtools/apps/) (which we call `r116` for some reason).

susannasiebert: It would probably be better to use Picard::version_compare instead of eq.

susannasiebert:
+1 once tests are fixed

lib/perl/Genome/Model/Tools/Picard/CompareSams.pm
acoffman: I agree with @susannasiebert on using `version_compare` as well :tada:

lib/perl/Genome/Model/Tools/Picard.pm
```diff
@@ -318,6 +339,11 @@ sub run_java_vm {
 sub create {
     my $class = shift;
     my $self = $class->SUPER::create(@_);
+
+    if (not defined $self->use_version) {
+        die $self->error_message('Cannot pass undef to Picard::use_version');
```
iferguson90: I'm worried that this will actually cause crashes in all sorts of places that aren't picked up by the test suite.

The test failures for earlier commits in this PR were due to the Bwamem test not setting any value for `picard_version` on the test `AlignmentResult` object. Then, `AlignmentResult::Bwamem` would blindly pass the undefined `$self->picard_version` to `Picard::use_version`. In other words, `use_version` would be set to `undef`, not the default value.

However, until now, `use_version` was only used like so:
```
sub picard_path {
    my $self = shift;
    return $self->path_for_picard_version($self->use_version); # in this case, undef
}

sub path_for_picard_version {
    my ($class, $version) = @_;
    $version ||= $PICARD_DEFAULT; # the default gets filled in
    ...
}
```

I think the reasoning behind this was to allow someone to call `GMT::Picard->path_for_picard_version()` and get the path to the "default" Picard. However this had the side effect of masking bad behavior from callers passing in an undefined `use_version`. Until now...

I'm hesitant to push code that could break poorly written callers but I also don't want to further legitimize the use of default version (by filling in the default if someone incorrectly tries to use `undef` as a version). In my opinion there should be no default version. Either pick one or always use the latest version if none is specified.

apregier: This comment is orthogonal to your comment above: Rather than putting this check in create, you can make it non-optional, right?

apregier: For now, the thing to do might be to make a wrapper to use_version:
```perl
sub _use_version {
    my $self = shift;
    return $self->use_version ||= $PICARD_DEFAULT;
}
```

lib/perl/Genome/Model/Tools/Picard.pm
```diff
@@ -315,9 +336,52 @@ sub run_java_vm {
     return 1;
 }
 
+sub _collect_call_info {
```
iferguson90: This sub is temporary. We're hoping to collect metrics about how Picard is called in practice before I start messing with default version handling.

I'm hoping that wrapping in an `eval` and passing a timeout to `IO::Socket::INET->new()` is sufficient protection against this sub taking down a whole build or getting stuck waiting on a socket.

PR #574, opened 2015-03-18
Refactor per-lane locking out of Merged.pm and into AlignmentResult.pm

dufeiyu:
This PR refactors per-lane bam locking code out of G::I::AR::Merged and into G::I::AlignmentResult. It also removes an extra reallocation.

### DIFF
--- a/lib/perl/Genome/InstrumentData/AlignmentResult.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult.pm
@@ -1919,5 +1919,48 @@ sub add_to_temporary_input_files_queue {
     return 1;
 }
 
+sub lock_bam_file_access {
+    my $self = shift;
+    unless ($self->get_merged_alignment_results) {
+        my @bams = $self->alignment_bam_file_paths;
+        unless (@bams) {
+            die $self->error_message("Alignment result with class (%s) and id (%s) has neither ".
+                "merged results nor valid bam paths. This likely means that this alignment result ".
+                "needs to be removed and realigned because data has been lost. ".
+                "Please create an apipe-support ticket for this.", $self->class, $self->id);
+            #There is no way to recreate per lane bam if merged bam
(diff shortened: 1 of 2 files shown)


acoffman:
+1 :pizza:

lib/perl/Genome/InstrumentData/AlignmentResult.pm
```diff
@@ -1919,5 +1919,48 @@ sub add_to_temporary_input_files_queue {
...
+                "needs to be removed and realigned because data has been lost. ".
+                "Please create an apipe-support ticket for this.", $self->class, $self->id);
+            #There is no way to recreate per lane bam if merged bam
+            #does not exist and per lane bam is removed. Software
+            #result of this per lane alignment needs to be removed and
+            #this per lane instrument data needs to be realigned
+            #with that aligner.
+        }
+
+        my $lock_var = File::Spec->join('genome', __PACKAGE__, 'lock-per-lane-alignment-'.$self->id);
```
apregier: During the transition to this code, there may be two lock names, since __PACKAGE__ used to be the merged alignment result and is now alignment result.  We may need to take steps to mitigate this?

dufeiyu: Gabe and I get a solution to handle both old and new locks during transition. Code is pushed as below. This code need be reverted later.

nnutter: You could also use `LockMigrationProxy` once I merge genome/genome#592.

nnutter: I just merged that in so if you want to use it you can merge `master` into your branch.

dufeiyu:
A new PR 648 replaces this. Close this one.

dufeiyu:
Closed

PR #575, opened 2015-03-20
Initial ptero integration

mkiwala:
This PR introduces code which translates legacy workflow xml into Ptero workflow json and tests that the json is how we expect it to look.  Submitting the workflow to Ptero is disabled in this PR.  The intent of this PR is to focus only on the generation of correct Ptero workflow json from legacy workflow xml.


lib/perl/Genome/WorkflowBuilder/Command.pm
```diff
@@ -40,6 +41,65 @@ sub create {
...
+    return Ptero::Builder::ShellCommand->new(
+        name => 'shortcut',
+        parameters => {
+            commandLine => [
+                'genome', 'ptero', 'wrapper', 'command',
+                '--command-class', $self->command,
+                '--method', 'shortcut',
+            ],
+            environment => \%ENV,
+            user => $ENV{USER},
```
nnutter: I don't know if it matters or if this is an exception but I think we usually use `Genome::Sys->username` instead of `$USER`.

lib/perl/Genome/WorkflowBuilder/DAG.pm
```diff
@@ -61,6 +61,93 @@ sub add_link {
...
+sub execute {
+    my $self = shift;
+
+    my %p = Params::Validate::validate(@_, {
+        inputs => {type => HASHREF},
+    });
+
+    my $inputs = {%{$self->constant_values}, %{$p{inputs}}};
+
+    my $backend = $ENV{GENOME_WORKFLOW_BUILDER_BACKEND} || 'workflow';
```
nnutter: This is a duplication of the default value since it's also set in the `Genome::Env` module.

nnutter:
A couple possible improvements but I don't see any blockers, :+1:.

mkiwala:
@nnutter thanks for catching those mistakes!

PR #576, opened 2015-03-23
Make AlignmentResult an AlignedBamResult

susannasiebert:
We need this for the QC framework so that they have a common interface with MergedAlignmentResult.

### DIFF
--- a/lib/perl/Genome/InstrumentData/AlignmentResult_regenerate.t
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult_regenerate.t
@@ -246,10 +246,14 @@ subtest 'test per lane bam removal and recreation - AlignedBamResult accessors' 
 
     # The revivified bam will be in a different location
     my $new_path = $ar2->revivified_alignment_bam_file_paths(disk_allocation => $temp_allocation);
-    # This test currently fails because the bam_path points to the original
-    # output_dir. This will change once we have a universal method to get to
-    # the bam's path (which shouldn't be called bam_path)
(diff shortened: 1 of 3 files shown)


iferguson90:
This seems reasonable, and tests have passed. +1

ebelter:
+1

lib/perl/Genome/InstrumentData/AlignmentResult_regenerate.t
```diff
@@ -235,4 +235,23 @@ sub get_test_alignment_results {
...
+        mount_path => $temp_allocation_dir,
+        owner => $owner,
+    );
+
+    # The revivified bam will be in a different location
+    my $new_path = $ar2->revivified_alignment_bam_file_paths(disk_allocation => $temp_allocation);
+    # This test currently fails because the bam_path points to the original
+    # output_dir. This will change once we have a universal method to get to
+    # the bam's path (which shouldn't be called bam_path)
+    # is($ar2->bam_path, File::Spec->join($temp_allocation_dir, $per_lane_bam), 'bam_path correct after revivification');
```
tmooney: It might be better to have this single test [skip](http://search.cpan.org/~exodist/Test-Simple/lib/Test/More.pm#Conditional_tests) instead of being commented out.

tmooney:
+1 :diamond_shape_with_a_dot_inside:

lib/perl/Genome/InstrumentData/AlignmentResult_regenerate.t
susannasiebert: Good point. I put this test in a skip block.

PR #577, opened 2015-03-23
Don't Warn about Each Missing ".bases" File if We Didn't Make Any

tmooney:
Previously, the choice of FASTA in the `_list_bases_files` was the original input which may or may not have been indexed.  Now it will be called on the build's FASTA which should be indexed during the build process.  (Additionally, cache the result of whether or not to make bases files, so the choice of FASTA is less relevant.)

I don't believe that the `--skip-bases-files` flag to `genome model define imported-reference-sequence` is currently working as intended... but this change shouldn't affect that.


ebelter:
+1

apregier:
+1

PR #580, opened 2015-03-24
Prefer the existing per-lane BAM if we have one.

tmooney:
It does not follow from the existence of a merged result that we can
necessarily revivify this BAM.  The merge may have happened on an older
snapshot and the revivification will in such a case fail for lack of the
.header file.


tmooney:
@dufeiyu indicates that his backfill script is currently running, which would render this issue moot.  It isn't safe to commit this change while it is running, so we'll have to contend with the failures another way for now.

PR #581, opened 2015-03-24
Optionally add tag(s) when associating menu item(s) to the AnP.

tmooney:
This feature already existed when creating a configuration from a file directly.  This makes the two commands more similar.  (If more features are added to these, they would probably benefit from a common base.)


nnutter:
:+1:

apregier:
+1

PR #582, opened 2015-03-24
Create qc tool wrappers

susannasiebert:


lib/perl/Genome/Model/Tools/Picard/CollectWgsMetrics.pm
```diff
@@ -0,0 +1,82 @@
...
+        reference_sequence => {
+            is => 'String',
+            doc => 'The reference sequence fasta aligned to.',
+            picard_param_name => 'REFERENCE_SEQUENCE',
+        },
+        minimum_mapping_quality => {
+            is => 'Integer',
+            is_optional => 1,
+            doc => "Minimum mapping quality for a read to contribute coverage. Default value: 20. This option can be set to 'null' to clear the default value.",
+            picard_param_name => 'MINIMUN_MAPPING_QUALITY',
```
nnutter: I assume `MINIMUN` is a typo?

lib/perl/Genome/Model/Tools/Picard/Base.pm
```diff
@@ -128,18 +128,26 @@ sub _redirects {
...
-    my $java_vm_cmd = sprintf '%s -Xmx%dm -XX:MaxPermSize=%dm %s -cp /usr/share/java/ant.jar:%s %s %s',
+    return (
         $self->java_interpreter,
-        int(1024 * $self->maximum_memory),
-        $self->maximum_permgen_memory,
-        $jvm_options,
-        $self->_jar_path,
+        sprintf('-Xmx%dm', int(1024 * $self->maximum_memory)),
+        sprintf('-XX:MaxPermSize=%dm', $self->maximum_permgen_memory),
+        split(' ', $jvm_options),
```
nnutter: Might be safer to use [Test::ParseWords::shellwords](http://perldoc.perl.org/Text/ParseWords.html).

lib/perl/Genome/Model/Tools/Picard/MarkDuplicates.pm
```diff
@@ -121,15 +140,11 @@ sub execute {
...
-    $self->run_java_vm(
-        cmd => $dedup_cmd,
-        input_files => [$self->input_file],
-        output_files => [$self->output_file, $self->metrics_file],
-        skip_if_output_is_present => 0,
-    );
-    return 1;
+sub build_cmdline_list {
+    my $self = shift;
+    return split(' ', $self->_java_cmdline($self->build_cmdline_string));
```
nnutter: Same here, might be safer to use [Test::ParseWords::shellwords](http://perldoc.perl.org/Text/ParseWords.html).

lib/perl/Genome/Model/Tools/Picard/Base.pm
nnutter: Actually, `shellwords` might be more harmful due to lack of symmetry with the `join`.

lib/perl/Genome/Model/Tools/Picard/CollectWgsMetrics.pm
susannasiebert: Fixed.

lib/perl/Genome/Qc/Tool.pm
```diff
@@ -25,7 +25,15 @@ class Genome::Qc::Tool {
 
 sub cmd_line {
     my $self = shift;
-    die $self->error_message("Abstract method run must be overriden by subclass");
+    my $cmd = $self->gmt_class->create($self->gmt_params);
+    return $cmd->build_cmdline_list;
```
apregier: I think this method might belong in Genome::Qc::Tool::Picard - since gmt_class and build_cmdline_list aren't necessarily common interfaces for all tools, just for picard.

apregier: Same thing with output_file and output_file_accessor - they are not needed outside of Picard, right?

lib/perl/Genome/Qc/Tool/Picard/CollectMultipleMetrics.t
```diff
@@ -0,0 +1,60 @@
...
+    '/usr/share/java/ant.jar:/gscmnt/sata132/techd/solexa/jwalker/lib/picard-tools-1.123/CollectMultipleMetrics.jar',
+    'net.sf.picard.analysis.CollectMultipleMetrics',
+    'ASSUME_SORTED=true',
+    sprintf('INPUT=%s', __FILE__),
+    'MAX_RECORDS_IN_RAM=500000',
+    sprintf('OUTPUT=%s', $output_file),
+    sprintf('TMP_DIR=%s', __FILE__),
+    'VALIDATION_STRINGENCY=SILENT',
+);
+$DB::single=1;
```
apregier: retained breakpoint

lib/perl/Genome/Qc/Tool/Picard/MarkDuplicates.t
```diff
@@ -0,0 +1,71 @@
...
+
+my $data_dir = __FILE__.".d";
+
+my $metrics_file = File::Spec->join($data_dir, 'output_file.txt');
+
+my $tool = $pkg->create(
+    gmt_params => {
+        input_file => __FILE__,
+        metrics_file => $metrics_file,
+        output_file => __FILE__,
```
apregier: This makes me a little bit nervous - I know we're not executing the command, but is there a chance that someone would add that to the test and then it could stomp over the source file?

lib/perl/Genome/Qc/Tool.pm
susannasiebert: I agree with you about the code in cmd_line and will move it to Picard.pm

I was hoping that output_file_accessor could become a more global solution via the logic in Result.pm for any tool that needs to parse an output_file for metrics. Right now we really only have Picard tools that fit that criteria.

lib/perl/Genome/Qc/Tool/Picard/MarkDuplicates.t
susannasiebert: Yeah, I will change all of those to temp_file_path

susannasiebert:
@apregier I pushed a few commits to address your comments.

lib/perl/Genome/Model/Tools/Picard/MarkDuplicates.pm
```diff
@@ -90,8 +90,27 @@ sub get_max_filehandles_param {
...
+    );
+    return 1;
+}
+
+sub build_cmdline_string {
+    my ($self, $cmd) = @_;
+
+    my $dedup_cmd = $self->picard_path .'/MarkDuplicates.jar';
+    if ($self->version_compare($self->use_version, '1.122') >= 0) {
+        $dedup_cmd .= ' net.sf.picard.sam.markduplicates.MarkDuplicates';
```
tmooney: The `net.sf.` part was removed from Picard beginning with version 1.114.  Should it also be removed here?

tmooney:
This and #573 will have to be reconciled to each other.

susannasiebert:
MarkDuplicates is currently not a Picard::Base so they shouldn't conflict. Once #573 makes it into master I was planning on making MarkDuplicates inherit from Picard::Base and not overriding these subroutines.

lib/perl/Genome/Model/Tools/Picard/MarkDuplicates.pm
susannasiebert: I think you are right. I'm kinda hoping that #573 gets merged soon so I can just get rid of this altogether.

apregier:
+1

tmooney:
The new `CollectWgsMetrics` herein is a `Picard::Base`, though.

lib/perl/Genome/Model/Tools/Picard/MarkDuplicates.pm
susannasiebert: I pushed a fix for the MarkDuplicates package name.

susannasiebert:
I have pulled in #573 and reconciled things. I also made MarkDuplicates a G::M::T::Picard::Base.

susannasiebert:
CI, test this.

lib/perl/Genome/Model/Tools/Picard/Base.pm
```diff
@@ -49,6 +49,7 @@ sub _picard_param_metas {
 sub _format_picard_arg {
     my ($type, $name, $value) = @_;
     $value = $value ? 'true' : 'false' if $type eq 'Boolean';
+    $value = "\"$value\"" if $type eq 'Text';
```
tmooney: I like using `qq` when needing to use quotes within a string, e.g.:
```perl
$value = qq{"$value"} if $type eq 'Text';
```

lib/perl/Genome/Model/Tools/Picard/MarkDuplicates-filehandles.t
```diff
@@ -15,17 +15,14 @@ my $cmd = $class->create(
...
     metrics_file => '/tmp/baz',
     temp_directory => '/tmp',
-    use_version => "1.34"
+    use_version => "1.36"
     );
 
-my $fh = $cmd->get_max_filehandles_param;
-like($fh, qr/^MAX_FILE_HANDLES=[0-9]+$/, "Version 1.34 sets MAX_FILE_HANDLES");
+my $fh = $cmd->build_cmdline_string;
+like($fh, qr/MAX_FILE_HANDLES=[0-9]+$/, "Version > 1.34 sets MAX_FILE_HANDLES");
```
tmooney: Should this be a `>=`?  Reading the new version of the test I wasn't sure what `1.34` should do.

lib/perl/Genome/Model/Tools/Picard/MarkDuplicates.pm
```diff
@@ -7,46 +7,44 @@ use Genome;
...
             doc => 'Assume the input data is sorted.  default_value='. $DEFAULT_ASSUME_SORTED,
             default_value => $DEFAULT_ASSUME_SORTED,
             is_optional => 1,
+            picard_param_name => 'ASSUME_SORTED',
         },
         remove_duplicates => {
-            is => 'Integer',
-            valid_values => [1, 0],
+            is => 'Boolean',
             doc => 'Merge the seqeunce dictionaries. default_value='. $DEFAULT_REMOVE_DUPLICATES,
```
tmooney: It's not something you changed, but these doc lines don't need to include the defaults since the help text will already show them.   Additionally this doc line doesn't match the parameter it is attached to!

lib/perl/Genome/Qc/Tool/Picard.pm
```diff
@@ -0,0 +1,53 @@
...
+    return $cmd->build_cmdline_list;
+}
+
+sub get_metrics {
+    my $self = shift;
+
+    my $file = $self->output_file;
+    my $gmt_class = $self->gmt_class;
+    my %metrics = $self->metrics;
+    my %metric_results = %{$gmt_class->parse_file_into_metrics_hashref($file)};
```
tmooney: Is this being de-referenced and re-referenced in order to make a copy?  Otherwise it could be passed along.

lib/perl/Genome/Qc/Tool/Picard.pm
```diff
@@ -0,0 +1,53 @@
...
+}
+
+sub _get_metrics {
+    my ($self, $metrics, $metric_results) = @_;
+
+    my %desired_metric_results;
+    while (my ($metric, $metric_details) = each %{$metrics}) {
+        my $metric_key = $metric_details->{metric_key};
+        unless (defined($metric_key)) {
+            ($metric_key) = keys %{$metric_results};
```
tmooney: What does it mean when we end up grabbing a random key from `%$metric_results` to use as the `$metric_key`?

susannasiebert: If the Picard gmt class doesn't have _metric_header_as_key defined than the metric_key during creation of the hash will be set to the first column's header and value. As far as I can tell this is the case when the metrics are output on only one line and aren't subcategorized. When _metric_header_as_key is defined this subroutine defines the column that is used for subcategorizing (e.g., read1, read2, both) where each key then contains values that are applicable to that category (e.g., % reads aligned for the read1 category). In the rest of the cases we don't have categories so the metrics_key is just a useless middle layer.

susannasiebert: It might be a good idea to check that there is really only one hash key and die if there are more. It probably means that you forgot to define the metric_key in the metrics hash.

susannasiebert:
@tmooney I pushed a few commits to address your comments.

tmooney:
+1 :spaghetti:

PR #583, opened 2015-03-24
Add decorators and a "QC" decorator to the Alignment Dispatcher

tmooney:
This is the means of hooking the QC Framework in #556 into the pipelines.

### DIFF
--- a/lib/perl/Genome/Command/DelegatesToResult.pm
+++ b/lib/perl/Genome/Command/DelegatesToResult.pm
@@ -30,7 +30,12 @@ class Genome::Command::DelegatesToResult {
         label => {
             is => 'Text',
             doc => 'Label for the additional user (if any) of the result',
-        }
+        },
+        result_users => {
+            is => 'HASH',
+            doc => 'for programmatically passing all users',
+            is_transient => 1,
+        },
     ],
 };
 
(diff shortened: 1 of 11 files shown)


tmooney:
I should probably add more tests before merging this.

lib/perl/Genome/Command/DelegatesToResult.pm
```diff
@@ -63,12 +68,15 @@ sub input_hash {
...
 
-    my $result_users = {
-        sponsor   => $self->sponsor // Genome::Sys->current_user,
-        requestor => $self->requestor,
-    };
-    if($self->label and $self->user) {
-        $result_users->{$self->label} = $self->user;
+    my $result_users = $self->result_users;
+    unless ($result_users) {
+        my $result_users = {
```
apregier: probably don't want the 'my' here since you want to use the value outside the scope of this unless block

lib/perl/Genome/Command/DelegatesToResult.pm
```diff
@@ -97,7 +105,7 @@ sub __errors__ {
         }
     }
 
-    unless(defined $self->requestor) {
+    unless(defined $self->requestor or defined $self->result_users) {
```
apregier: Do you want to check specifically that result_users has an entry for requestor?

lib/perl/Genome/Command/DelegatesToResult.pm
tmooney: Indeed!  This is what's making all the tests fail right now :smile:

lib/perl/Genome/Command/DelegatesToResult.pm
apregier: and sponsor for that matter?

tmooney: I was reluctant to go too far down the road of duplicating the validation that already occurs inside `Genome::SoftwareResult::User`.  If the caller gave us a `result_users` value, it seems like they're at least trying to comply (and we know they didn't use the CLI), so it might be less important to catch it early here to display an error.

Do you think it'd be better to be more thorough here?

apregier: Ok, I see what you are saying - that's fine

apregier:
+1 - I don't disagree about adding more tests - it will make our lives easier in the long run.

acoffman:
+1

PR #584, opened 2015-03-24
InstData Import: pass description on to created instdasta

ebelter:


lib/perl/Genome/InstrumentData/Command/Import/Basic_bam.t
```diff
@@ -51,6 +52,7 @@ my $instrument_data = $instrument_data[0];
 is($instrument_data->original_data_path, $source_bam, 'original_data_path correctly set');
 is($instrument_data->import_format, 'bam', 'import_format is bam');
 is($instrument_data->sequencing_platform, 'solexa', 'sequencing_platform correctly set');
+is($instrument_data->description, 'We took some DNA and sheared it and sequenced it.', 'description cprrectly set');
```
tmooney: The `o` in `correctly` has gained a tail!

tmooney:
+1 :pig2:

apregier:
+1 - another option would be not to expose description as a special parameter to the command and just let people set it via instrument_data_properties

ebelter:
@apregier I would prefer that, too. Chip was very adamant about having this be a separate option as some of his scripts depend on it.

PR #585, opened 2015-03-24
Don't log rollbacks on build requested

tmooney:
Issues this PR is attempting to address:
* Processes will sometimes acquire the lock and hold onto it for a long time even though they haven't attempted to modify the `build_requested` value.
* Processes spam the model notes with calls from `_reverse_all_changes`.

This short script demonstrates the existing behaviour:
```
$ genome-perl -M5.10.0 -MGenome -e 'my $m = Genome::Model->get(name => "apipe-test-reference-alignment"); say scalar(@{[$m->notes]}); UR::Context->rollback(); say scalar(@{[$m->notes]});'
13035
13036
Removing remaining resource lock: 'build_requested/2880887847' at /.../Genome/Sys/Lock/FileBackend.pm line 244.
```

Here's the behaviour with the changes:
```
$ genome-perl -M5.10.0 -e 'use above "Genome"; my $m = Genome::Model->get(name => "apipe-test-reference-alignment"); say scalar(@{[$m->notes]}); UR::Context->rollback(); say scalar(@{[$m->notes]});'
Using libraries at /.../git/genome-fast/lib/perl
13035
13035
```

While this seems to be effective, I do wonder if there might be a better way to accomplish the goal than inspecting `caller`.

### DIFF
--- a/lib/perl/Genome/Model-build_requested.t
+++ b/lib/perl/Genome/Model-build_requested.t
@@ -22,6 +22,8 @@ my $sample = create_test_sample('test_sample');
 my $pp = create_test_pp('test_pp');
 my $model = create_test_model($sample, $pp, 'test_model');
 
+my $tx = UR::Context::Transaction->begin();
+
 $model->build_requested(0);
 is($model->build_requested, 0, 'unset build requested');
 {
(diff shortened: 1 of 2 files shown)


nnutter:
It seems like maybe UR shouldn't be using `build_requested` when rolling back but should instead be using `__build_requested` since that's probably the inverse of the change it's tracking.  However, if that is intended then couldn't we just stop overloading `build_requested` by changing the property/column to `_build_requested`?  Then we would lose the ability to filter on `build_requested`?  Maybe the API should be `request_build` instead of setting `build_requested`?

tmooney:
Based on your suggestion I removed the `caller` checks and instead made the column back a `_build_requested` property.  Using UR's alias feature we can still query `build_requested`, but it will rollback the underlying property instead.

nnutter:
:+1:

brummett:
+1

PR #587, opened 2015-03-25
Filter out G::ID::AlignmentResult

susannasiebert:
This is to fix failing model tests.

Since per-lane alignments are now G::ID::AlignedBamResult they need to be explicitly filtered out.


tmooney:
+1 :honey_pot:

apregier:
+1

PR #588, opened 2015-03-25
Handle symlink comparison more explicitly when comparing process output

susannasiebert:
Right now we only print the symlink paths which isn't helpful when the symlinks are temporary, e.g. during Jenkins process tests. This will make it so that the symlink paths as well as the target paths are printed.


susannasiebert:
@nnutter You had concerns in PR https://github.com/genome/genome/pull/391 about the usage of file vs target in the last else block in the file diff error message.

Can you please have a look at this PR which changes the handling of symlinks in the comparison?

lib/perl/Genome/Process.pm
```diff
@@ -710,12 +710,22 @@ sub _compare_output_directories {
...
+                if (-l $file || -l $other_file) {
+                    if (-d $target && -d $other_target) {
+                        my %additional_diffs = $self->_compare_output_directories($target, $other_target, $other_process);
+                        %diffs = (%diffs, %additional_diffs);
+                    }
+                    if (-f $target && -f $other_target) {
+                        if (compare($target, $other_target)) {
+                            $diffs{File::Spec->abs2rel($file, $output_dir)} = sprintf(
+                                'symlinks are not the same (diff -u %s %s) (diff -u %s %s)',
+                                $file, $other_file, $target, $other_target
```
apregier: In terms of making the output useful, we may not want to include the symlinks at all, since they will always be temp files

susannasiebert: Not necessarily. I guess if we assume that it will always be called by _compare_output_files, they will be. I originally was only printing the targets but I changed that after @nnutter's comment on PR #391.

lib/perl/Genome/Process.pm
```diff
@@ -710,19 +710,28 @@ sub _compare_output_directories {
...
+                            );
+                        }
+                        else {
+                            #Symlink targets are in fact the same content - do nothing
+                        }
+                    }
+                    return;
                 }
+                $diffs{File::Spec->abs2rel($file, $output_dir)} = sprintf(
```
nnutter: I believe the recording of the diff here is meant as an else block of the preceding conditional.  I find the mix of if/else and if/return confusing.  I personally prefer if/return style but I am only referring to the mixing of the two styles here.

susannasiebert: Argh - the return on line 729 is erroneous. This diff should apply here as well as for the "else-case" for the inner if-else-block that checks for target type. I was wanting to avoid code duplication. 

Under this circumstance do you still prefer explicit else clauses with duplicated code?

lib/perl/Genome/Process.pm
nnutter: I think I might change my opinion at this point.  To me there's a conflict caused by the fact that things are being symlinked in to compare them with whether a symlink could exist in the result.  I don't know which side I fall on anymore.

lib/perl/Genome/Process.pm
nnutter: I would try it both ways and maybe also look for another solution.  However, you also don't have to agree with me. ;-)

lib/perl/Genome/Process.pm
```diff
@@ -710,19 +710,28 @@ sub _compare_output_directories {
...
-                }
-                elsif (-f $target && -f $other_target && !compare($target, $other_target)) {
-                    #Files are in fact the same - do nothing
-                }
-                else {
-                    $diffs{File::Spec->abs2rel($file, $output_dir)} = sprintf(
-                        'files are not the same (diff -u %s %s)',
-                        $file, $other_file
-                    );
+                if (-l $file || -l $other_file) {
```
nnutter: Are both not being symlinks possible?  Does that imply they are different?

lib/perl/Genome/Process.pm
susannasiebert: I prefer explicit else's because it's easy to mess up things otherwise, as I just demonstrated. Just wish there was a way to avoid the code duplication. Oh well.

lib/perl/Genome/Process.pm
nnutter: Maybe concern about whether they are symlinks should be limited to how the `sprintf` message is generated instead of as part of the comparison logic?

susannasiebert: It would be very unusual for only one of them to be symlinks and it would imply that something is really wrong. For the most part both of them would be symlinks. In that case we still need to differentiate between directories and files.

nnutter: Is that based on an assumption that this will always be called by `_compare_output_files`?

susannasiebert: Yes.

susannasiebert: Or at least that the input directories belong to comparable processes.

susannasiebert:
@apregier I pulled in the latest origin/master which contained your file comparer changes. This caused a conflict in Process.pm. I think I reconciled the conflict correctly but it would be great if you could have a look as well.

susannasiebert:
I made some more commits to this PR and it's ready for another round of code review.

lib/perl/Genome/Process.pm
```diff
@@ -709,19 +709,38 @@ sub _compare_output_directories {
...
+                            );
+                        }
+                        else {
+                            #Symlink targets are in fact the same content - do nothing
+                        }
+                    }
+                    else {
+                        $diffs{File::Spec->abs2rel($file, $output_dir)} = sprintf(
+                            'files are not the same (diff -u %s %s)',
+                            $file, $other_file
```
apregier: Should this give the "symlinks are not the same" message?

susannasiebert: Maybe we need something more generic. Only one of them could be a symlink and the other could be something else. "Path" maybe?

apregier: Sure, the main thing is that we will probably want to see the $target versions of the path in addition to the $file

susannasiebert: Most of our diff cases will be caught with the case on line 718 which outputs both the file paths as well as the target paths. This specific case is probably unlikely to ever occur but I agree that it wouldn't hurt to output both the file paths and the target paths here as well.

susannasiebert: @apregier I pushed a commit for this change.

apregier:
+1 when tests pass (I believe the latest test failures are load related)

tmooney:
+1 with passing tests :baggage_claim:

PR #589, opened 2015-03-25
Framework for process tests

susannasiebert:


davidlmorton:
I really like the yaml file for configuring the process tests!

apregier:
process_yaml is used by the jenkins script: https://github.com/genome/wugc-hudson/pull/17/files

susannasiebert:
@davidlmorton I pushed a commit to address your comment as well as a corresponding commit to https://github.com/genome/wugc-hudson/pull/17

davidlmorton:
+1 :baby_chick:

lib/perl/Genome/Model/SomaticValidation/Command/AlignmentStatsSummary.pm
```diff
@@ -143,7 +143,13 @@ sub _alignment_metrics_from_result {
     for my $lane (@per_lane_alignments) {
         $total_bases += $lane->total_base_count;
         $total_mapped_bases += $lane->total_aligned_base_count;
-        $mismatches += $lane->total_base_count * $lane->instrument_data->filt_error_rate_avg / 100;
+        my $filt_error_rate_avg;
+        if ($mismatches ne "NA" and $lane->instrument_data->can("filt_error_rate_avg")) {
```
acoffman: I believe this is intentional, but I just wanted to make sure I understood. If any one lane "cannot" `filt_error_rate_avg` then the value of both `mismatches` and `error_rate` for the whole report will be `NA` correct?

acoffman:
+1 :chart_with_upwards_trend: 

Not entirely related to this PR, but I might pull the error_rate, total_bases, total_mapped_bases, and mismatches calculations into their own sub. It seems like a pretty self contained calculation with a small number of inputs and outputs. Don't think that should hold up a merge though!

lib/perl/Genome/Model/SomaticValidation/Command/AlignmentStatsSummary.pm
susannasiebert: I think that is correct although @apregier made this commit so she should be able to confirm but she's on vacation this week.

apregier: Yes, because if any of the lanes doesn't have an error rate, then we don't want to report it at all.

PR #590, opened 2015-03-25
InstData Import: give default value to resolved instdata properties

ebelter:


tmooney:
+1 :bridge_at_night:

apregier:
+1

PR #592, opened 2015-03-25
LockProxy and LockMigrationProxy

nnutter:
The primary goal is to cleanup/finish the scoping of locks that was started
previously with the addition of the `scope` option and the definition of the
`site`, `tgisan`, and `unknown` scopes.  I believe the goal being to have just
`site` and `host` scopes at this time.

This will require migrating locks from one scope to another at dozens of
locations.  We have also had occasions where we need to migrate from one
resource (name) to another.  So I am trying to make a helper class,
`Genome::Sys::LockMigrationProxy`, to make these types of migrations almost trivial.

As part of that work I needed to lift the restriction on not allowing
locks at multiple scopes.  In trying to do that I discovered that
`unlock_resource` was looking up the scope instead of requiring it as an
argument.  Really a lock is defined by the resource (name) and the scope so I
created an object to encapsulate those, `Genome::Sys::LockProxy`.

Related PRs:

- genome/genome#386
- genome/genome#527
- genome/genome#514 and genome/genome#537

### DIFF
--- a/lib/perl/Genome/Sys/LockMigrationProxy.pm
+++ b/lib/perl/Genome/Sys/LockMigrationProxy.pm
@@ -26,6 +26,9 @@ A hash reference containing the constructor parameters for the desired/new lock.
 
 sub new {
     my $class = shift;
+    if (ref $class) {
+        croak 'new() should be called as a class method';
+    }
     my %params = validate(@_, {
         old => { type => HASHREF },
         new => { type => HASHREF },
--- a/lib/perl/Genome/Sys/LockMigrationProxy.pm
+++ b/lib/perl/Genome/Sys/LockMigrationProxy.pm
@@ -4,6 +4,7 @@ use strict;
 use warnings;
 
 use Carp qw(croak);
+use Genome::Sys::Lock qw();
 use Genome::Sys::LockProxy qw();
 use Params::Validate qw(validate validate_with HASHREF);
 
@@ -53,11 +54,7 @@ sub lock {
     my $self = shift;
     my %params = validate_with(
         params => \@_,
-        spec => {
-            block_sleep => 0,
-            max_try => 0,
-            wait_announce_interval => 0,
-        },
+        spec => Genome::Sys::Lock::PROXY_LOCK_PARAMS_SPEC(),
     );
 
     unless ($self->{old}->lock(%params)) {
--- a/lib/perl/Genome/Sys/LockMigrationProxy.pm
+++ b/lib/perl/Genome/Sys/LockMigrationProxy.pm
@@ -1,0 +1,100 @@
+package Genome::Sys::LockMigrationProxy;
+
+use strict;
+use warnings;
+
+use Carp qw(croak);
+use Genome::Sys::LockProxy qw();
+use Params::Validate qw(validate validate_with HASHREF);
+
+=item new()
+
+Keyword Arguments:
+
+=over
+
+=item old
+
+A hash reference containing the constructor parameters for the existing/old lock.
+
+=item new
+
+A hash reference containing the constructor parameters for the desired/new lock.
+
+=cut
+
+sub new {
+    my $class = shift;
+    my %params = validate(@_, {
+        old => { type => HASHREF },
+        new => { type => HASHREF },
+    });
+
+    my $primoridal_ooze = {
+        old => Genome::Sys::LockProxy->new(%{$params{old}}),
+        new => Genome::Sys::LockProxy->new(%{$params{new}}),
+    };
+
+    return bless $primoridal_ooze, $class;
+}
+
+=item lock()
+
+C<lock()> is an instance method with optional parameters C<block_sleep>,
+C<max_try>, or C<wait_announce_interval>.  It behaves similar to
+C<Genome::Sys::Lock::lock_resource()> but on sucess C<lock()> returns an
+instance of C<Genome::Sys::LockMigrationProxy> that can later be unlocked with
+C<unlock()>.  Both the C<old> lock and the C<new> lock will be acquired, in
+that order, or neither will.
+
+=cut
+
+sub lock {
+    my $self = shift;
+    my %params = validate_with(
+        params => \@_,
+        spec => {
+            block_sleep => 0,
+            max_try => 0,
+            wait_announce_interval => 0,
+        },
+    );
+
+    unless ($self->{old}->lock(%params)) {
+        return;
+    }
+
+    unless ($self->{new}->lock(%params)) {
+        $self->{old}->unlock();
+        return;
+    }
+
+    return $self;
+}
+
+=item unlock()
+
+C<unlock()> is an instance method that will unlock both the C<new> lock and the
+C<old> lock, in that order, and should throw an exception if it fails to
+release either lock.
+
+=cut
+
+sub unlock {
+    my $self = shift;
+
+    unless ($self->{new}->unlock()) {
+        return;
+    }
+
+    # If old fails to unlock we can't relock new due to lack of context so I
+    # would like to die but I am hoping that unlock already does that as
+    # `lock_resource` is documented as doing.
+    unless ($self->{old}->unlock()) {
+        return;
+    }
+
+    return $self;
+}
+
+1;
--- a/lib/perl/Genome/Sys/LockMigrationProxy.t
+++ b/lib/perl/Genome/Sys/LockMigrationProxy.t
@@ -1,0 +1,170 @@
+#!/usr/bin/env genome-perl
+
+use strict;
+use warnings;
+
+use Genome;
+use Test::More tests => 3;
+
+use Genome::Sys::Lock qw();
+use Genome::Sys::Lock::MockBackend qw();
+use Genome::Sys::LockMigrationProxy qw();
+use List::Util qw(shuffle);
+use Scope::Guard qw();
+use Sub::Override qw();
+
+subtest 'old fails' => sub {
+    plan tests => 4;
+
+    my %backends = (
+        test_old => [ Genome::Sys::Lock::MockBackend->new() ],
+        test_new => [ Genome::Sys::Lock::MockBackend->new() ],
+    );
+    my $backend_guard = backend_guard();
+    my $scopes_guard = Sub::Override->new('Genome::Sys::Lock::scopes', sub { keys %backends });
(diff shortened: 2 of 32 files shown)


nnutter:
I need to review the commits where I've converted to `LockObect` and be more careful because some places use the lock (e.g. previously `$resource_lock`) in error messages, etc.

nnutter:
Oh, and I expect tests will fail because of 0131324dcdf66127b6fe848e5f98c2e792cfb63d.

davidlmorton:
I like the `LockObject` class.  I think `ResourceProxy` might be a good name for it with attributes named `name` and `scope`.  An `is_locked` method would be useful in the case that these are being passed around to consumers that don't know the state lock state.

I think something to ease the burden of multiple locks for migration is good too.  Perhaps a way of adding two ResourceProxy objects to get a CombinedResourceProxy that could be locked makes sense.

nnutter:
> I like the `LockObject` class. I think `ResourceProxy` might be a good name for it with attributes named `name` and `scope`.

Does `ResourceProxy` mean something or are you primarily using the word proxy here and resource just comes from `resource`/`resource_lock`?

> An `is_locked` method would be useful in the case that these are being passed around to consumers that don't know the state lock state.

I would be interested in concrete examples.

> I think something to ease the burden of multiple locks for migration is good too. Perhaps a way of adding two `ResourceProxy` objects to get a `CombinedResourceProxy` that could be locked makes sense.

That sounds better than just another method on `Genome::Sys::Lock`.

nnutter:
>> I like the `LockObject` class. I think `ResourceProxy` might be a good name for it with attributes named `name` and `scope`.
> 
> Does `ResourceProxy` mean something or are you primarily using the word proxy here and resource just comes from `resource`/`resource_lock`?

[Proxy pattern](http://en.wikipedia.org/wiki/Proxy_pattern)?

davidlmorton:
I guess I just meant that the object isn't actually the resource but is a proxy for the resource, albeit in this case a proxy with very limited utility (it can only lock and unlock the resource).  Another name choice I like is `LockableResource`

nnutter:
This PR should be ready for general review.  I believe it will pass tests but is currently failed due to GitHub issues.

tmooney:
Should [`Genome::Disk::Volume::get_lock`](https://github.com/nnutter/genome/blob/9188c79feb0b8724bd8a68dfc917d0190ffced85/lib/perl/Genome/Disk/Volume.pm#L194) also be updated to use this new lock object?

(If you're accepting late naming suggestions, it seems like the thing one uses to lock or unlock something is called a `Key`!)

nnutter:
> Should `Genome::Disk::Volume::get_lock` also be updated to use this new lock object?

I have only gone after places that called `unlock_resource`.  It looks like maybe this didn't show up because it's not even be used.  I'll look into it.

nnutter:
> (If you're accepting late naming suggestions, it seems like the thing one uses to lock or unlock something is called a Key!)

I definitely don't have any issue renaming things.  I'll think about that one.

lib/perl/Genome/Sys/LockMigrationProxy.pm
```diff
@@ -0,0 +1,100 @@
...
+
+sub lock {
+    my $self = shift;
+    my %params = validate_with(
+        params => \@_,
+        spec => {
+            block_sleep => 0,
+            max_try => 0,
+            wait_announce_interval => 0,
+        },
```
tmooney: Could this one also benefit from `Genome::Sys::Lock->LOCK_RESOURCE_PARAMS_SPEC`?

lib/perl/Genome/Sys/LockMigrationProxy.t
```diff
@@ -0,0 +1,170 @@
...
+    my $new_lock = Genome::Sys::Lock->lock_resource(
+        resource_lock => $resource,
+        scope => 'test_new',
+        %lock_params,
+    );
+    is($new_lock, undef, q(failed to lock on 'test_new' after LockMigrationProxy locked));
+};
+
+sub random_string {
+    my @chars = map { (shuffle 'a'..'z', 'A'..'Z', 0..9)[0] } 1..10;
```
tmooney: It's not random, but if we're trying for uniqueness would `Data::UUID->create_str` serve the purpose?

Alternatively, some version of:
```perl
my @valid_chars = ('a'..'z', 'A'..'Z',0..9);
my @chars = map { $valid_chars[rand @valid_chars] } 1..10;
return join('', @chars);
```
should have the same effect without needing to shuffle the entire list each time.

tmooney:
+1 :closed_lock_with_key:

lib/perl/Genome/Sys/LockProxy.pm
```diff
@@ -0,0 +1,106 @@
...
+
+sub new {
+    my $class = shift;
+    my $spec = Genome::Sys::Lock::LOCK_RESOURCE_PARAMS_SPEC();
+    $spec->{resource} = delete $spec->{resource_lock};
+    my %params = validate_with(
+        params => \@_,
+        spec => $spec,
+    );
+    return bless \%params, $class;
```
tmooney: This will crash if `new()` is called on an existing object.  Should we also validate that `$class` is not a reference?

lib/perl/Genome/Sys/LockProxy.pm
```diff
@@ -0,0 +1,106 @@
...
+C<max_try>, or C<wait_announce_interval>.  It behaves similar to
+C<Genome::Sys::Lock::lock_resource()> but on sucess C<lock()> returns an
+instance of C<Genome::Sys::LockProxy> that can later be unlocked with
+C<unlock()>.
+
+=cut
+
+sub lock {
+    my $self = shift;
+    my %params = validate_with(
```
acoffman: Part of me wonders if this `validate_with` should be in the Genome::Sys::Lock::lock_resource` method that this is ultimately delegating to.

acoffman:
I generally like the name `LockProxy` over `Resource*` with the interface as it currently is. Proxy implies (at least to me) some sort of exposure of the interface of an underlying object (through delegation, decoration, whatever). In this case, the underlying interface that is "exposed" by the proxy is that of `Lock` and not of the resource being locked.

It seems like in almost all cases, we invoke the constructor and immediately chain a call to `lock()` on it. `my $lock = LockProxy->new()->lock()`. Would it make sense to expose some sort of `init_and_lock` class method that encapsulates that pattern while still leaving `new` and `lock` in tact for those who may want to use them independently?

Either way, this looks good to me, I'd be fine merging it as is.

+1 :key:

lib/perl/Genome/Sys/LockMigrationProxy.t
nnutter: I was curious how each would perform.  It seems like my list shuffling would be pretty bad.  It turns out it is bad but UUID is worse.  I'll switch this to the `rand()` version.

```
           Rate    uuid shuffle    rand
uuid     9625/s      --    -39%    -83%
shuffle 15798/s     64%      --    -72%
rand    56497/s    487%    258%      --
```

lib/perl/Genome/Sys/LockProxy.pm
nnutter: I'll see what the crash looks like and decide.

lib/perl/Genome/Sys/LockProxy.pm
nnutter: This is just a lightweight check for parameter names.  The real validation is still in [Genome::Sys::Lock::lock_resource][1].  The lightweight check is duplication though.  I think I'll split `Genome::Sys::Lock::LOCK_RESOURCE_PARAMS_SPEC` up so I can use it here.

[1]: https://github.com/nnutter/genome/blob/988981633940f9cd10844450659399325e107601/lib/perl/Genome/Sys/Lock.pm#L73

lib/perl/Genome/Sys/LockMigrationProxy.pm
nnutter: Same response as to [Adam's](https://github.com/genome/genome/pull/592/files#r27582981).

nnutter:
> It seems like in almost all cases, we invoke the constructor and immediately chain a call to `lock()` on it. `my $lock = LockProxy->new()->lock()`. Would it make sense to expose some sort of `init_and_lock` class method that encapsulates that pattern while still leaving `new` and `lock` in tact for those who may want to use them independently?

I almost made `new()` do double duty and support construction or the `init_and_lock` style.  I decided to keep `new()` simple instead but I think that double duty is somewhat common in Perl.  Ultimately I felt like `LockProxy->new()->lock()` wasn't really different than `LockProxy->init_and_lock()` except that we don't use chaining very often (and I don't know why).

nnutter:
> > (If you're accepting late naming suggestions, it seems like the thing one uses to lock or unlock something is called a Key!)
> 
> I definitely don't have any issue renaming things. I'll think about that one.

I think that `LockProxy` is more like an address than a key.  I don't think our locks have keys. :-p

acoffman:
Cool, I actually like chained methods in general and use them quite often in my ruby projects. I just haven't seen us use them much in `Genome` land before, hence the comment. I'm good with keeping them chained.

tmooney:
>> > (If you're accepting late naming suggestions, it seems like the thing one uses to lock or unlock something is called a Key!)
>> 
>> I definitely don't have any issue renaming things. I'll think about that one.
>
>I think that `LockProxy` is more like an address than a key.  I don't think our locks have keys. :-p

Fair enough!  Perhaps they are `Combination` locks! :smile:

lib/perl/Genome/Sys/LockProxy.pm
nnutter: It seems like it would be silly if someone called `new()` on an object.  Is that a thing?  I am checking that it's not a ref so instead of,

```
Attempt to bless into a reference at /gscuser/nnutter/genome/lock-pair/lib/perl/Genome/Sys/LockProxy.pm line 32.
```

you'll get,

```
new() should be called as a class method at -e line 1
```

but maybe I should just make it bless into the class of the object?

nnutter:
Thanks @acoffman, @davidlmorton, and @tmooney for your review.  I'm going to merge this in later today if it passes tests so if you have anything else to add or if anyone else does let me know.

davidlmorton:
+1 Nothing else to add

PR #593, opened 2015-03-27
Add shortcut to adaptor class

apregier:
I noticed this was being submitted every time as a separate lsf job.  There is no reason why this should ever need to run with lots of resources.


nnutter:
I can see arguments for either the current form or @davidlmorton's suggestion.  In either form I :+1:.

apregier:
I initially was thinking that calling execute might have side effects since it implements an abstract Command method - but on further reflection that is not true.

PR #594, opened 2015-03-27
Filter out AlignmentResults since they're now AlignedBamResults

susannasiebert:
This is needed to make model tests pass.

### DIFF
--- a/lib/perl/Genome/Model/SomaticValidation/Command/IdentifyDnp.pm
+++ b/lib/perl/Genome/Model/SomaticValidation/Command/IdentifyDnp.pm
@@ -55,7 +55,7 @@ sub params_for_result {
     my $build = $self->build;
     my @result_users = Genome::SoftwareResult::User->get(user => $build);
     my @results = map($_->software_result, @result_users);
-    my @merged_results = grep { $_->isa('Genome::InstrumentData::AlignedBamResult') } @results;
+    my @merged_results = grep { ! $_->isa('Genome::InstrumentData::AlignmentResult') } grep { $_->isa('Genome::InstrumentData::AlignedBamResult') } @results;
     my ($tumor_alignment_result) = grep( ($_->instrument_data)[0]->sample eq $build->tumor_sample, @merged_results);
 
     my $tumor_aligment_result_id = $tumor_alignment_result->id;


iferguson90:
If I understand correctly, tests failed because this module started to (incorrectly) pick up AlignmentResults? This will certainly fix it. Any other places that this needs to be fixed?

Otherwise +1 :boar:

susannasiebert:
I really hope this is the last place but it's hard to check for it. It is certainly the last place that does a refinement like this.

susannasiebert:
And yes to "this module started to (incorrectly) pick up AlignmentResults?"

lib/perl/Genome/Model/SomaticValidation/Command/IdentifyDnp.pm
```diff
@@ -55,7 +55,7 @@ sub params_for_result {
     my $build = $self->build;
     my @result_users = Genome::SoftwareResult::User->get(user => $build);
     my @results = map($_->software_result, @result_users);
-    my @merged_results = grep { $_->isa('Genome::InstrumentData::AlignedBamResult') } @results;
+    my @merged_results = grep { ! $_->isa('Genome::InstrumentData::AlignmentResult') } grep { $_->isa('Genome::InstrumentData::AlignedBamResult') } @results;
```
nnutter: Why can't this be looking for `Genome::InstrumentData::AlignmentResult::Merged`?  Filtering out select subclasses does not feel like the right solution.

nnutter: I see `Genome::InstrumentData::BamUtil::ClipOverlapResult` and `Genome::InstrumentData::Gatk::Base` are also subclasses of `Genome::InstrumentData::AlignedBamResult`.

susannasiebert: I believe it's because we also want refinement results which are not AlignmentResult::Merged. The `@merged_results` is a bit of a misnomer. @apregier @tmooney can you confirm?

susannasiebert: Right so we want AlignmentResult::Merged and BamUtil::ClipOverlapResult and Gatk::Base results.

nnutter: Maybe there should be a class that those things inherit from instead of `Genome::InstrumentData::AlignedBamResult`?  I don't know enough about the bigger picture.  The fact that you have had to make the same change in more than one place hints there is something wrong.  I just don't know what the right solution is yet.  Don't consider anything I'm saying as blocking the PR if there's a bug to be fixed.

susannasiebert: @nnutter I implemented your suggestion in https://github.com/genome/genome/pull/598. Closing this PR.

PR #595, opened 2015-03-27
Fix test failure when running ./WorkflowBuilder/StreamGraph.t

iferguson90:
The test would work when running like so:
```
$ genome-perl WorkflowBuilder/StreamGraph.t
```
But not like so:
```
$ genome-perl ./WorkflowBuilder/StreamGraph.t
```
(`WorkflowBuilder` versus `./WorkflowBuilder`)

Not sure if using `File::Spec->canonpath` is the best fix, but I think this should fix Perl tests, which appear to be calling this test with the `./`:
```
#   Failed test 'xml produced correctly'
#   at ./WorkflowBuilder/StreamGraph.t line 59.
```

### DIFF
--- a/lib/perl/Genome/WorkflowBuilder/StreamGraph.t
+++ b/lib/perl/Genome/WorkflowBuilder/StreamGraph.t
@@ -9,13 +9,14 @@ use strict;
 use warnings;
 
 use above "Genome";
+use File::Spec;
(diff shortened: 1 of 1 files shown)


tmooney:
This was duplicated by #596 which has been merged.

PR #596, opened 2015-03-27
use canonpath to normalize $test_dir

nnutter:
Use canonpath to normalize path as something in StreamGraph seemed to do the
same.  For example, `./WorkflowBuilder` becomes just
`WorkflowBuilder` which was breaking the full test suite.


nnutter:
Merging early to fix tests.

PR #597, opened 2015-03-30
Call Correct Accessor in ADSP QC tools

tmooney:
There's a lot of duplication between them... but with this change at least they should work!


ebelter:
+1

nnutter:
:+1:

PR #598, opened 2015-03-30
Make a class for merged AlignedBamResults and use it where appropriate

susannasiebert:
This allows us to differentiate between per-lane and merged aligned bam results.

This replaces https://github.com/genome/genome/pull/594 and will fix model tests.


susannasiebert:
I ran the somatic validation model test locally on this change and it succeeded.

nnutter:
:+1:

tmooney:
+1

PR #599, opened 2015-03-30
Narrow down choices of convertible references by preferring an input.

tmooney:
If there are many candidates for conversion but exactly one is in our original query set, then use it rather than force the user to choose.

### DIFF
--- a/lib/perl/Genome/FeatureList/Command/Merge.pm
+++ b/lib/perl/Genome/FeatureList/Command/Merge.pm
@@ -6,6 +6,7 @@ use warnings;
 use feature qw(say);
 
 use List::MoreUtils qw(any uniq all);
+use UR::Util;
 use Genome;
 
 class Genome::FeatureList::Command::Merge {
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/FeatureList/Command/Merge.pm
```diff
@@ -188,14 +188,20 @@ sub _find_convertible_reference {
...
-            'The references of the input feature-lists can be converted to multiple references:',
-            @convertible_references,
-        );
+    if (scalar(@convertible_references) < 2) {
+        return $convertible_references[0];
     }
 
-    return $convertible_references[0];
+    #prefer a reference from our original query set
+    my @preferred_references = grep { my $target = $_; any { $_ eq $target } @_ } @convertible_references;
```
nnutter: `@_` could be `@references`?  If you're worried about `@references` changing I would probably save `@_` again at beginning of function rather than use `@_` anywhere besides first couple lines of function.

nnutter: `Set::Scalar` might make it clearer that you're intersecting the two lists but what you have is not complicated.

nnutter:
:+1: but would prefer to see the use of `@_` changed.

lib/perl/Genome/FeatureList/Command/Merge.pm
tmooney: Yes, that should definitely be `@references`--updated!

acoffman:
+1. I second @nnutter's thought that using set intersection might be marginally clearer but I'm also fine to merge "as is".

tmooney:
I've switched this to use `UR::Util::intersect_lists` instead of a homespun intersection.

PR #600, opened 2015-03-30
handle use case where reference sequence is missing

nnutter:
This addresses the unclear error noticed in [RT#104836][],

> Could not start build, reason: Can't call method "__display_name__" on an undefined value at /gsc/scripts/opt/genome/snapshots/genome-3587/lib/perl/Genome/Model/Build.pm line 1221.

[RT#104836]: https://rt.gsc.wustl.edu/Ticket/Display.html?id=104836


ebelter:
+1

PR #601, opened 2015-03-30
Lock all SR User construction to enforce uniqueness.

tmooney:
This makes sure the user relationship is not created more than once.

I find the `linked alignment ... to the build` debug message to still be useful, but maybe we don't need to separately verify the user was created here?

### DIFF
--- a/lib/perl/Genome/Model/Event/Build/ReferenceAlignment/AlignReads.pm
+++ b/lib/perl/Genome/Model/Event/Build/ReferenceAlignment/AlignReads.pm
@@ -459,6 +459,7 @@ sub _process_and_link_alignments_to_build {
     my @errors;
 
     my $result_users = Genome::SoftwareResult::User->user_hash_for_build($build);
+    $result_users->{uses} = $build;
 
     my %segment_info;
     if (defined $self->instrument_data_segment_id) {
@@ -488,7 +489,7 @@ sub _process_and_link_alignments_to_build {
     }
 
     for my $alignment (@alignments) {
-        my $link = $alignment->add_user(user => $build, label => 'uses');
+        my $link = $alignment->user(user => $build, label => 'uses');
         if ($link) {
             $self->debug_message("Linked alignment " . $alignment->id . " to the build");
         }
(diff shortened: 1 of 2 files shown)


lib/perl/Genome/Model/Event/Build/ReferenceAlignment/AlignReads.pm
```diff
@@ -488,7 +489,7 @@ sub _process_and_link_alignments_to_build {
     }
 
     for my $alignment (@alignments) {
-        my $link = $alignment->add_user(user => $build, label => 'uses');
+        my $link = $alignment->user(user => $build, label => 'uses');
```
nnutter: Would it be safer to `$alignment->user(...) || $alignment->add_user(...)`?

tmooney: If we get here and the user does not exist, there's a bug in the code handling the  `$result_users`.  I am more inclined to remove the existence check here altogether and just print the `Linked alignment ...` debug messages for all alignments.

nnutter: Where is the duplicate `add_user`?  I've gone across 3-4 classes and haven't found it...

tmooney: The problem is that `add_user` can be called from both `shortcut` and `execute`.  In this case `shortcut` failed in the flow wrapper after `commit` added the user.  Thereafter `execute` was run which blindly added the user again.

The other `add_user` calls are also susceptible to this problem.  I should change them all.  This is the only one that has come up as a problem in practice (since I did the massive cleanup) so it was the first to be changed.

tmooney:
I added a more general "fix" for this problem.  We'll need a follow-up commit to remove the old locks... It'll simply the code in here a good bit, too.

acoffman:
+1

davidlmorton:
+1 its better than the SR::User specific locking, and until we have a clear path to avoid the locking I don't see a better alternative.

tmooney:
I will wait to merge this until after the next snapshot promotion.  No need to combine too many wide-ranging changes at once!

PR #602, opened 2015-03-30
update parent-dir for locks

gatoravi:
Trying to create a lock on "/" causes a step to fail in https://github.com/genome/gms/issues/181

### DIFF
--- a/lib/perl/Genome/Site/SGMS.pm
+++ b/lib/perl/Genome/Site/SGMS.pm
@@ -1,6 +1,7 @@
 package Genome::Site::SGMS;
 use strict;
 use warnings;
+use File::Spec;
 
 BEGIN {
     unless ($INC{'Genome/Sys/Lock.pm'}) {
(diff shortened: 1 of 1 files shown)


nnutter:
This is probably okay because on sGMS there is really only one level of locking since it runs specifically on one host.  However, since the `scope` is like a namespace it's possible that you would run into accidental collisions.

gatoravi:
Would something like `parent_dir => $ENV{GENOME_LOCK_DIR}/unknown` and  `parent_dir => $ENV{GENOME_LOCK_DIR}/tgisan` be better?

nnutter:
> Would something like `parent_dir => $ENV{GENOME_LOCK_DIR}/unknown` and `parent_dir => $ENV{GENOME_LOCK_DIR}/tgisan` i.e subdirectories within the main locking directory be better?

Probably.  You might also want to have the `site` scope go to a `site` subdirectory.

gatoravi:
I've added a commit for the scope specific subdirectories.
PS - this PR is aimed at the sgms branch but we'd like to incorporate the same change into master to minimize divergence.

nnutter:
:+1:

gatoravi:
I think the test-failures are because this is on the SGMS branch. Merging this in so that we can test this.

PR #603, opened 2015-03-30
change default version of G::M::T::Sam::Readcount to use bam-readcount v...

ernfrid:
...0.7


nnutter:
:+1: It passed tests.  Changing a default is usually more painful...

tmooney:
+1 :card_index:

PR #606, opened 2015-03-31
Fix gmt picard typos; remove duplicate method

tmooney:


susannasiebert:
I haven't totally confirmed this but parse_file_into_metrics_hashref and parse_file_into_histogram_hashref seem to be almost identical. Maybe we should attempt some deduplication.

tmooney:
I've removed the second one altogether.  It wasn't called anywhere and--as noted--was almost identical to the first one.

susannasiebert:
+1

nnutter:
:+1:

PR #605, opened 2015-03-31
Fix class name

susannasiebert:

PR #607, opened 2015-03-31
RNA Seq: fix capitalization

ebelter:


susannasiebert:
There is already a PR for this that has been reviewed and is just waiting on github to stabilize so that tests can pass: https://github.com/genome/genome/pull/605

PR #608, opened 2015-03-31
Make new_feature_list output available from merge.

tmooney:
This makes merge more similar to import, both of which are being used by LIMS for similar purposes.


nnutter:
:+1:

ebelter:
+1 :mahjong:

PR #609, opened 2015-04-01
InstData Generate Cmds: check if file is empty

ebelter:


lib/perl/Genome/InstrumentData/Command/Import/GenerateBase.pm
```diff
@@ -81,9 +81,10 @@ sub _open_file_parser {
         });
     die $self->error_message('Failed to create Text::CSV parser!') if not $parser;
 
+    die $self->error_message('File (%s) is empty!', $file) if not -s $file;
     my $fh = Genome::Sys->open_file_for_reading($file);
     my $headers = $parser->getline($fh);
-    die $self->error_message('File (%s) does not have any headers! Is it empty?', $file) if not $headers;
+    die $self->error_message('File (%s) does not have any headers! Is is empty?', $file) if not $headers;
```
tmooney: The `Is it empty?` sentence could be removed completely--we now know it's not.

acoffman: Either way, it currently says `is is empty` so that's probably not correct!

acoffman:
+1 with typo fix!

ebelter:
I removed the check all together. The headers are checked later on anyway.

tmooney:
+1 :monkey:

iferguson90:
+1 :boar:

PR #610, opened 2015-04-01
Remove disused variables.

tmooney:
Not only are they not used, they duplicate two other variables declared just above them!


acoffman:
+1

nnutter:
:+1:

iferguson90:
+1 Torch them! :fire:

PR #611, opened 2015-04-01
Revert "Remove incomplete disused "parse_file_into_histogram_hashref"."

tmooney:
This reverts commit fdbcd444d7ea17d24014c8d6af86525b749d6eae.

(For the second time in a week I have failed to use grep correctly, it would seem!)

This subroutine is still mostly duplicative, which is sad.

### DIFF
--- a/lib/perl/Genome/Model/Tools/Picard.pm
+++ b/lib/perl/Genome/Model/Tools/Picard.pm
@@ -507,5 +507,60 @@ sub parse_file_into_metrics_hashref {
     return \%data;
 }
 
+sub parse_metrics_file_into_histogram_hashref {
+    my ($class,$metrics_file,$metric_header_as_key) = @_;
+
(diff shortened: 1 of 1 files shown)


susannasiebert:
+2

brummett:
You left a ```$DB::single=1``` in there

nnutter:
For reference, this is reverting genome/genome#606.

nnutter:
Sorry, I missed that too.  It's definitely [used elsewhere](https://github.com/genome/genome/search?utf8=%E2%9C%93&q=parse_metrics_file_into_histogram_hashref).

tmooney:
I will follow-up with another PR that does something better with this subroutine (and removes the breakpoint!).  But merging this should fix failing model tests.

iferguson90:
Sad that this had to be reverted, but +1 when tests pass.

PR #612, opened 2015-04-01
GMT CG Hub: metadata can be created from XML and file

ebelter:
GMT CG Hub: metadata can be created from XML and file.


tmooney:
Is the `xml` ever needed after the object is created?  If not, it might be better to have a `create_from_xml` to go along with `create_from_file`, both of which create an object that only stores the `HASH` version as a property.  As it is here, two copies of the data will be stored for the life of the object.

PR #614, opened 2015-04-01
Picard needs to know the name in the VCF.

tmooney:
(This is different from the name for TCGA samples, for example.)


ebelter:
+1

acoffman:
+1 :name_badge:

PR #617, opened 2015-04-01
When Tiering in SomVar only Consider Results we "Use"

tmooney:
The specific scenario motivating this PR is that the `copy-number-with-bam-window` detector runs two mini DV2s inside of itself.  The results generated by those detectors were previously not linked to the build, but the user-tracking system means that the build is now their `creator`.  Since all the official DV2 results should [have the `uses` label applied](https://github.com/genome/genome/blob/166fc8310843c2f73f34e887731b9fc2bbdd9a07/lib/perl/Genome/Model/SomaticVariation/Command/DetectVariants.pm#L87-L90), limit our search here to those results.


acoffman:
+1

apregier:
+1

PR #618, opened 2015-04-02
Add readcount checks

ernfrid:
After talking to @malachig, we're convinced that filtering on VAF and depth shouldn't be allowed when feeding this tool multiple BAMs. I've added code to check this and updated the docs ever so slightly. 

@malachig,@obigriffith,@kkrysiak  - let me know if you think the doc updates look good.


malachig:
+1 looks good to me.

tmooney:
+1 :wrench:

PR #619, opened 2015-04-02
only build workflows should go in build workflow queue

nnutter:
This doesn't really solve the problem pointed out in RT#104979 but instead
probably just moves it to another queue.  A better solution should be possible
after the LSF queues are redesigned in a couple weeks.

### DIFF
--- a/lib/perl/Genome/Process/Command/Run.pm
+++ b/lib/perl/Genome/Process/Command/Run.pm
@@ -148,7 +148,7 @@ sub _bsub_in_pend_state {
         log_file => $self->workflow_process_out_log,
         hold_job => 1,
         project => $self->process->workflow_name,
-        queue => $ENV{GENOME_LSF_QUEUE_BUILD_WORKFLOW},
+        queue => $ENV{GENOME_LSF_QUEUE_DV2_WORKFLOW},
         send_job_report => 1,
         never_rerunnable => 1,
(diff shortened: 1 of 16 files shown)


lib/perl/Genome/Process/Command/Run.pm
```diff
@@ -148,7 +148,7 @@ sub _bsub_in_pend_state {
         log_file => $self->workflow_process_out_log,
         hold_job => 1,
         project => $self->process->workflow_name,
-        queue => $ENV{GENOME_LSF_QUEUE_BUILD_WORKFLOW},
+        queue => $ENV{GENOME_LSF_QUEUE_DV2_WORKFLOW},
```
tmooney: I would make `Genome::Process::Command::Run` an exception to this change.  `Process`es and `Build`s are nearly identical in terms of running workflows (and I still hope they are eventually deduplicated!).

nnutter: Don't Builds run Processes?

tmooney: I believe that a `Build` that runs a `Process` should `execute` it rather than `schedule` it.

nnutter: Alright, I'll revert this instance.

PR #620, opened 2015-04-02
Deduplicate histogram and metrics processing subroutines.

tmooney:
[As promised in #611,] This PR is a second attempt at cleaning up the duplicate subroutines for metrics processing.

One thing of note: this no longer silently skips rows where the number of headers and number of values don't match in the "histogram" case.


ebelter:
+1 :tanabata_tree:

brummett:
+1
I don't want to hold this up because of the default value thing.

PR #621, opened 2015-04-02
Figure out bedtools path better

gatoravi:
This PR modifies the way this GMT figures out the bedtools path.

### DIFF
--- a/lib/perl/Genome/Model/Tools/CopyNumber/Cnmops.pm
+++ b/lib/perl/Genome/Model/Tools/CopyNumber/Cnmops.pm
@@ -50,6 +50,11 @@ class Genome::Model::Tools::CopyNumber::Cnmops {
     default => 'tgi/cancer-annotation/human/build37-20130401.1',
     is_optional => 1,
   },
+  bedtools_version => {
+      is => "String",
+      doc => "Bedtools version to use",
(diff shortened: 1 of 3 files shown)


lib/perl/Genome/Model/Tools/CopyNumber/Cnmops.pm
```diff
@@ -199,12 +204,22 @@ sub annotate_cnvs {
   }
 }
 
+sub _bedtools_path {
```
nnutter: This logic already exists in [Genome::Model::Tools::Bwa::RunMem](https://github.com/genome/genome/blob/fcb7058fe1aa98d03997d5600cb694d94ea7e95f/lib/perl/Genome/Model/Tools/Bwa/RunMem.pm#L205) which means this is duplication and instead of doing that you should centralize it.  So it seems like this should be implemented in `Genome::Model::Tools::BedTools`.

gatoravi:
I agree, I've moved this to BedTools.pm

nnutter:
:+1: Looks good. =)

malachig:
Thanks @nnutter. Looks good to me @gatoravi.   +1

gatoravi:
merging it in to test on SGMS.

PR #626, opened 2015-04-03
minimize use of Genome::Sys::Lock API

nnutter:
I went looking for `lock_resource()` usage to migrate to `LockProxy` but there
wasn't much left after previously converting locations that used
`unlock_resource()`.


ebelter:
+1 w/ tests passing

acoffman:
+1

PR #628, opened 2015-04-03
Module to identify LOH regions.

gatoravi:
This module helps identify LOH regions in the tumor compared to the normal. It takes a clin-seq or somatic-variation build as input. Thanks to @chrisamiller for working out the details.


lib/perl/Genome/Model/ClinSeq/Command/IdentifyLoh.pm
```diff
@@ -0,0 +1,280 @@
...
+    $self->status_message("varscan snvs file found: " .
+        join("\t", @varscan_snvs));
+    $self->split_snvs($varscan_snvs[0], $snv_prefix);
+}
+
+#Get tumor/normal BAMs from somatic-variation
+sub get_tumor_normal_bam {
+    my $self = shift;
+    my $somvar_build = shift;
+    my $tumor_bam = $somvar_build->tumor_model->last_succeeded_build->
```
tmooney: In general, when starting from a build, it is bad practice to jump to a model and thence to its last succeeded build.  There is a `tumor_build` on the `$somvar_build` that corresponds to the build actually used by the SomaticVariation build (which may be quite different from some later build).

lib/perl/Genome/Model/ClinSeq/Command/IdentifyLoh.pm
```diff
@@ -0,0 +1,280 @@
...
+    my $self = shift;
+    my $snv_prefix = shift;
+    my $snv_combined = $snv_prefix . ".combined";
+    my $snv_combined_sorted = $snv_prefix . ".combined.sorted";
+    my $germline = $snv_prefix . "*.formatted.LOH.hc.filtered";
+    my $loh = $snv_prefix . "*.formatted.Germline.hc.filtered";
+    unless(scalar glob("$germline $loh")) {
+        die $self->error_message("Unable to find $germline or $loh");
+    }
+    my $cat_cmd = "cat $germline $loh > $snv_combined";
```
tmooney: You may be interested in `Genome::Sys->concatenate_files` or `Genome::Sys->cat` (from `Genome::Site::TGI::Extension::Sys`).  The former was created to handle potentially troublesome filenames; the latter long lists that may exceed the allowed length of a command.

lib/perl/Genome/Model/ClinSeq/Util.pm
```diff
@@ -1446,5 +1446,21 @@ sub _is_copycat_somvar {
     }
 }
 
+sub get_best_somvar_build {
+    my $self = shift;
+    my $clinseq_build = shift;
+    my $somvar_build =
+        $clinseq_build->model->wgs_model->last_succeeded_build;
```
tmooney: A similar caution against getting the build via `last_succeeded_build` here.  `$clinseq_build->wgs_build` would be more direct.

lib/perl/Genome/Model/ClinSeq/Command/IdentifyLoh.pm
```diff
@@ -0,0 +1,280 @@
...
+in the normal sample and compare with stretches of homozygosity in the tumor.
+
+If a clinseq build is supplied as input, the tool attempts to use the results
+from the WGS-somatic-variation build input, if these are unavailable the tool
+looks for WEx-somatic-variation build.
+
+Alternatively, instead of a clin-seq build you can supply a
+somatic-variation build as an input.
+
+Thanks to Chris Miller for working out the workflow.
```
chrisamiller: I just wrote the bash script around the workflow that Nathan Dees and Dan Koboldt developed back in the day. Feel free to drop attribution here. As with many projects, this was a team effort.

lib/perl/Genome/Model/ClinSeq/Command/IdentifyLoh.pm
```diff
@@ -0,0 +1,280 @@
...
+        next unless $build;
+        my $m = $build->model;
+        if ($m->can("reference_sequence_build")){
+            return $m->reference_sequence_build;
+        }
+    }
+}
+
+#Filter out putative false-positive SNVs
+sub filter_snvs {
```
chrisamiller: Is this function necessary? If Varscan has been run in somatic-variation, should this filter have been run already? If so, then you could just grab the results from the som-var build.

tmooney:
A general question... What is different here from the existing LOH steps in the `SomaticVariation` and `SomaticValidation` pipelines?  And if those differences are important, should they be incorporated into the pipelines instead of a new tool?

chrisamiller:
The LOH steps in the somatic pipelines do (best I can tell), absolutely nothing useful. They're legacy and we should get rid of them at some point. 

That said, you raise a good point. Why is this in clinseq and not in somatic variation? In my mind, the distinction is that basic variant calling goes in the somatic model, and more integrative stuff goes in clinseq. (I have the same question about exome CN calling, FWIW...)

gatoravi:
Good question @tmooney, the LOH step in the somatic-variation pipeline seems to be a quick way to remove LOH SNPs from the final somatic variation list that is produced(the 'hq' and 'lq' files.) This module attempts to identify regions or stretches with a loss of heterozygosity in the tumor, you could argue that the LOH SNPs filtered out in the somatic-variation pipeline could be used for this purpose but its very likely that the calls from VarScan which we use here offer better sensitivity due to the statistical approach(though they probably are very similar in some cases.) There is an additional segmenting of the LOH SNPs to identify stretches which we currently don't do in somatic-variation I don't think. This module will be evaluated with some samples and if useful can be plugged into a pipeline, if not I'm all for removing it :)
@chrisamiller, I think there is definitely a number of steps in ClinSeq which could be argued to be in somatic-variation e.g exome-cnv, clonality analysis, snv-indel-report, microarray-cnv etc. My understanding is that most of these steps run in less than an hour, hence quick to regenerate and are well suited for clin-seq. I suspect this is up for debate though, I'd like to see what @malachig thinks.

lib/perl/Genome/Model/ClinSeq/Command/IdentifyLoh.pm
gatoravi: I'm having a bit of trouble trying to use wild-cards(*) in either of those two, is that expected?
In the first method, the quotes result in the file not being found. In the second method, the -e operator is used, this also results in an 'input files not found' error. Is there a suggested alternative? I'm not able to clearly predict the names of my inputs here because of the variable per-chromosome files that might be generated.

jasonwalker80:
@ernfrid proposed removing the existing LOH step back in 6/2014.  It's my understanding it was added to support SomaticSniper.  Dave, please correct me where I'm wrong, but I also believe it could be removed or replaced.

As for addition of LOH to existing somatic pipelines, this JIRA issue provides some historical context:
https://jira.gsc.wustl.edu/browse/AT-234

In general, I believe we should migrate the "somatic-only" analysis, ie. does not rely on another data type like germline, RNA, bisulfite, etc. back to the somatic models.  The LOH Epic above is a good example.  It's been near the top of the TODO list for a long time, but has never hit a critical mass or urgency to prioritize it above other requests.  

Summary : ClinSeq seems like as good a place as any for the time being until development priorities change.

lib/perl/Genome/Model/ClinSeq/Command/IdentifyLoh.pm
tmooney: Yes, I would not expect wildcards to work directly in those.  You would need to `glob` the list of needed files in advance.

malachig:
I believe we have determined that the "LOH analysis" done in somatic-variation is as not really genome-wide LOH analysis but rather a way of eliminating certain SNPs that get called as somatic mutations because they are within an LOH region.  If we have the broader goal of defining all LOH regions (the segments of the genome that exhibit loss of heterozygosity) then it is best to step back and identify all SNP sites in the *normal sample* that are heterozygous, characterize their VAFs in the tumor, and identify the likely segments of LOH.  This last step is conceptually similar to the copy number segment identification done in CopyCat, but the segmentation algorithm operates on VAF derived values for a series of SNP positions rather than copy number estimates (read counts) for a series of fixed size windows.

I agree that both the Exome CopyNumber and LOH analysis steps ultimately belong in somatic-variation.  That's where they make the most sense.  We are simply testing out these implementations in clin-seq to stimulate such discussion and would be glad for help to migrate into somatic variation if it becomes sufficiently high priority?  As much as possible we want to keep all heavy lifting and large data files out of clin-seq.  Except where they specifically relate to clinical interpretation, manual review, final report generation, and cross data type integration (Array vs. WGS vs. Exome vs. RNA-seq).

For example, In ClinSeq it could still make sense to have analysis related to both CNV and LOH where we are comparing across data types (e.g. WGS CNV vs Exome CNV vs Array CNV), where we are integrating different event types (e.g. correlating LOH regions with allelic specific expression in RNA-seq, identifying somatic mutations that are also within regions of LOH, etc.).

Its been a while since we talked about cancer pipeline development priorities...  Maybe we should touch base on that front again?

lib/perl/Genome/Model/ClinSeq/Command/IdentifyLoh.pm
gatoravi: This is a really good point since this filter takes up quite a bit of run-time. From what I can tell, the original filtering module "GMT::VarScan::SomaticParallelFilter.pm" is not being run as part of any build. There are probably other modules attempting similar filtering but not as stringent on the Gemline/LOH calls.
I might be looking at the wrong place. I looked at the test build [here](https://gscweb.gsc.wustl.edu/gscmnt/gc9006/info/model_data/2888676250/build9dc0385a1b634c9bb85eb2017b3a0c73/variants/snv/varscan-somatic-2.2.6-d41d8cd98f00b204e9800998ecf8427e/varscan-high-confidence-v1-d41d8cd98f00b204e9800998ecf8427e/) for example, files like `snvs.Germline.hc` and `snvs.LOH.hc` haven't been filtered with the same filters. 
Testing the module on this build, after the filter I'd expect to see about ~234,000 calls in chr1 for this sample, but we see ~280,000 which is closer to the pre-filtered number.

ernfrid: You guys seem to have it right as far as my knowledge goes. The LOH 
filtering step in the pipeline is completely obsolete as of Sniper1.0.4 
as you can achieve the same affect with -G and -L. I had thought this 
had been removed, but perhaps it was just expected to be turned off via PP.



____

chrisamiller: > From what I can tell, the original filtering module "GMT::VarScan::SomaticParallelFilter.pm" is not being run as part of any build. 

Hmm - interesting.  I wonder if the results from the existing somatic filters would be adequate, especially if they'd save us some time. Can you elaborate a bit on the differences?

chrisamiller:
@malachig - yeah, we should discuss this elsewhere. At the end of the day, I don't care where it's being run, but anything we can to do to keep the different pipelines conceptually coherent will probably pay off in the long run. Let's shoot for a meeting among the key folks in the next few weeks.

I don't have any objections to this pull, as long as we're thinking about where this should live long-term, and thinking about whether the filtering can be done in one step to save some time.  Nice work, Avi!  +1

lib/perl/Genome/Model/ClinSeq/Command/IdentifyLoh.pm
```diff
@@ -0,0 +1,281 @@
...
+}
+
+#Combine germline, LOH calls and sort them
+sub combine_sort_snvs {
+    my $self = shift;
+    my $snv_prefix = shift;
+    my $snv_combined = $snv_prefix . ".combined";
+    my $snv_combined_sorted = $snv_prefix . ".combined.sorted";
+    my $germline = glob $snv_prefix . "*.formatted.LOH.hc.filtered";
+    my $loh = glob $snv_prefix . "*.formatted.Germline.hc.filtered";
```
tmooney: Just to check--are you expecting one file to match each of these or possibly several?  (If it's the latter, then these will need to be `@germline` and `@loh` instead.)

lib/perl/Genome/Model/ClinSeq/Command/IdentifyLoh.pm
gatoravi: I'm trying to compare the results from the two sets of filtered SNVs. Whats a good place to look at how the "snvs.Germline.hc" etc are generated in a somatic-variation build?

lib/perl/Genome/Model/ClinSeq/Command/IdentifyLoh.pm
gatoravi: thanks for pointing this out Tom. we are expecting more than one file here, I've fixed this.

malachig:
+1

After reviewing this again with Avi it seems like the existing results in SomaticVariation - both the old "LOH" results, and the filtered results associated with VarScan are not what we need for LOH analysis.  The existing VarScan results are filtered, but only with respect to somatic variants.

We can get the existing Germline + LOH variants from the VarScan run in SomaticVariation, but we then need to do false positive filtering on all of these before moving on to segmentation, summary and visualization.

So we are left with the strategy that Avi has implemented here.  We are going to run all past clin-seq cases through this approach and summarize the results.  At that point we will have a better sense of how compelling this analysis is.  We should definitely meet and talk about how to get this migrated from clin-seq back to somatic though.

gatoravi:
@tmooney I think I've addressed all of your comments on this PR?

lib/perl/Genome/Model/ClinSeq/Command/IdentifyLoh.pm
```diff
@@ -0,0 +1,289 @@
...
+
+#Get rid of intermediate files
+sub cleanup {
+    my $self = shift;
+    my $unwanted;
+    foreach my $suffix ("*.Somatic*", "*.readcounts", "*.hc", "*.LOH",
+        "*.lc", "*.Germline", "*.removed", "*.formatted",
+        "*.hc.err", "*.other") {
+        my $unwanted = File::Spec->join($self->outdir, $suffix);
+        if(-e $unwanted) {
```
tmooney: This will only match if the filenames literally contain `*`.  You'd need something like:
```perl
for my $file (glob($unwanted))  {
    unlink $file;
}
```

tmooney:
+1 save for my one new comment in `cleanup`. :sparkles:

gatoravi:
thanks Tom, I've addressed this.

PR #630, opened 2015-04-03
De Novo Assembly: add number of threads to lsf resource when processing instrument data.

ebelter:


nnutter:
I'm skeptical that this works.  Have you tried running a build and seen that LSF gets the thread count?

ebelter:
@tmooney indicated something of the like would work - maybe I implemented it incorrectly. I meant to run a build, but forgot. I will send one off and see what happens!.

ebelter:
@nnutter @tmooney The inputs passed in have a build, but it is a 'Workflow::Link::Instance'. When calling 'value' on it, this error is encountered:
 ERROR: Can't call method "output" on an undefined value at /gsc/scripts/opt/genome/snapshots/genome-3588/lib/perl/Workflow/Link/Instance.pm line 49.

So it seems there is no operation instance for the link, even though it has an id.

tmooney:
@ebelter This is a peculiarity of `Workflow`'s multi-process behaviour... The IDs referenced only refer to objects in the other process.  I think maybe [`$type->resource_for_instance(...)`](https://github.com/genome/tgi-workflow/blob/43813893069460826e8db143081219f33d07b8ba/lib/Workflow/Server/Hub.pm#L512) should use `$payload->{input}` instead of consulting `$instance->input`.

(I suppose it shouldn't be too surprising there are bugs in this feature, since we don't really use it... and it only [theoretically] works in the Workflow system we're moving away from.)

ebelter:
@tmooney Perhaps a hard coded class selection based on threading when building  the workflow:

if threads
 Class::RunWithThreads
else
 Class::Run?

tmooney:
That would work!

If you only ever need one resource requirement for a given OperationType, it might also work to override the resource requirements when constructing the workflow so it gets into the XML with the desired thread count.  (I know some past pipelines worked that way; I'm not sure if that still works.)

ebelter:
@tmooney I took the approach you suggested. Thanks!

tmooney:
+1 It worked when tested with `Workflow`! :dash:

PR #632, opened 2015-04-04
fix typo during LockProxy conversion

nnutter:


tmooney:
+1 :dog: Sorry for missing that before!

PR #633, opened 2015-04-06
Fix BQM's ability to decide what column to use for "build_requested".

tmooney:
Ironically the pointer we're using is changing more frequently than the value it points to!


acoffman:
+1

ebelter:
+1

PR #635, opened 2015-04-06
File-Backed Genome::Config Implementation

nnutter:
I still don't consider this ready for merge but I think this is ready for a stringent code review.  After any issues with this pass of code review are addressed I'll start looking at deploying specs and configs and converting the code base to using them.

There are three main classes:

- `Genome::Config`
- `Genome::ConfigSpec`
- `Genome::Configurable`

Each has POD describing them so I won't repeat that here but `Genome::Config` is sort of the base-level API to lookup values and `Genome::Configurable` is a higher-level role to integrate `Genome::Config` into a class definition, typically a Command.

To continue to try and keep the PR reviewable I'm separating out the conversion of the codebase as a separate PR, see nnutter/genome#2.


tmooney:
This reminds me somewhat of [genome/configuration-manager](https://github.com/genome/configuration-manager).

davidlmorton:
I have a couple of concerns, which you've probably already thought about but I just wanted to make sure they were discussed:

1.  These values are only loaded from files at runtime, so in the case of a Genome::Model::Build, the values could be different for different steps if the configuration changes while a build is running.  This is not ideal, especially with one of the configuration files living in the user's HOME directory.  A user could accidentally crash all their builds by editing that file, and HOME directory slowness would start equating to production slowness.

2.  These values are not persisted in any way, so we lose reproducibility and the ability to debug after the fact.  I know that this is already the case for all of the environment variables in our system currently, so this isn't a regression.  One solution would be to have the Genome::Model::Build framework and maybe the Genome::Process framework concretize the configuration into env variables when a Build or Process is launched, and persist the environment either in a file or in the DB.

acoffman:
Is this intended to be general purpose or specific to Genome/UR? If its the later, I'm a little unsure about the idea of the validators. It seems like properties should already have type annotations and classes should validate themselves. Adding the validations to the config seems like we're just duplicating that information in another place.

lib/perl/Genome/ConfigSpec.pm
```diff
@@ -0,0 +1,54 @@
...
+
+has 'env'           => (is => 'ro', isa => 'Str', predicate => 'has_env');
+has 'default_value' => (is => 'ro', isa => 'Str', predicate => 'has_default_value');
+has 'validators'    => (is => 'ro', isa => 'ArrayRef[CodeRef]', default => sub { [] });
+
+sub new_from_file {
+    my $class = shift;
+    my $file = Path::Class::File->new(shift);
+
+    (my $key = $file->basename) =~ s/\.yaml$//i;
```
acoffman: My gut instinct was that surely `Path::Class::File` had methods for working with file extensions and this manual munging wouldn't be needed, but after looking at the docs, that doesn't seem to be the case. Seems like an odd omission.

lib/perl/Genome/ConfigSpec.pm
```diff
@@ -0,0 +1,54 @@
...
+    my $data = YAML::Syck::LoadFile($file->stringify);
+
+    my @validators;
+    for my $v (@{$data->{validators}}) {
+        my $module_name = join('::', 'Genome', 'ConfigValidator', ucfirst($v));
```
acoffman: Is it a convention that validator will need to have single word names? (Its fine if so!) This will break if something needed to be camel cased.

lib/perl/Genome/ConfigSpec.pm
brummett: File::Basename::fileparse() will do it, though not case-insensitively.

lib/perl/Genome/Config.pm
```diff
@@ -0,0 +1,90 @@
...
+    if (!$spec->has_default_value && !defined($value)) {
+        croakf('required config key missing: %s', $key);
+    }
+
+    if (!defined($value)) {
+        return $spec->default_value;
+    }
+
+    for my $v (@{$spec->validators}) {
+        if (my $error = $v->($value, $spec)) {
```
acoffman: It seems odd that a call to validator's validate method (well, coderef) returns a truthy value on failure and a falsey value on success. I do like how you're using it to sort of auto-generate error messages though.

I wonder if part of the interface of a validator should be an `error_message` class method or something? I also wonder if its worth running all validators rather than dying after the first failure? That way you don't fix one problem only to stumble into the next one when you try again. Obviously this is just pseudocode but maybe something like: 

````perl
my @errors = map { $_->error_message($value) } 
        grep { ! $_->is_valid($value) }
        $spec->validators;

croakf(join(@error_messages, "\n")) if @error_messages;
````

Unrelated, but writing that made me wish perl had fold/reduce in core instead of an external module.

File: lib/perl/Genome/ConfigSpec.pm (hunk 1 above)
Comment (acoffman) [inline lib/perl/Genome/ConfigSpec.pm]: Ah, good to know, thanks for the pointer!

Comment (ebelter) [inline lib/perl/Genome/ConfigSpec.pm]: You can pass fileparse a suffix list. Presumably you are only looking for '.YAML' and '.yaml'.

Comment (nnutter) [inline lib/perl/Genome/ConfigSpec.pm]: I switched to using `fileparse`.

File: lib/perl/Genome/ConfigSpec.pm (hunk 2 above)
Comment (nnutter) [inline lib/perl/Genome/ConfigSpec.pm]: We can change it as needed.

File: lib/perl/Genome/Config.pm (hunk 3 above)
Comment (nnutter) [inline lib/perl/Genome/Config.pm]: I don't think I've quite addressed what you're suggesting but I did make it so the spec class creates the full error message,

```
my @errors = $spec->validate($value);
if (@errors) {
  my $msg = $spec->validation_error(@errors);
  croakf($msg);
}
```

**Comment by nnutter** [comment]
I still don't consider this ready for merge but I think this is ready for a stringent code review.  After any issues with this pass of code review are addressed I'll start looking at deploying specs and configs and converting the code base to using them.

**Comment by nnutter** [comment]
@davidlmorton,

>     These values are only loaded from files at runtime, so in the case of a Genome::Model::Build, the values could be different for different steps if the configuration changes while a build is running. This is not ideal, especially with one of the configuration files living in the user's HOME directory. A user could accidentally crash all their builds by editing that file, and HOME directory slowness would start equating to production slowness.

I'm not too concerned about this.  Either jobs should run in controlled environments, e.g. `apipe-builder`, or the user should be responsible for changes they make.  I did add the `sticky` option to "preserve" the value in the (shell) environment but I think most of the current configuration values would not use this as they are not configuration that affect build results.

>     These values are not persisted in any way, so we lose reproducibility and the ability to debug after the fact. I know that this is already the case for all of the environment variables in our system currently, so this isn't a regression. One solution would be to have the Genome::Model::Build framework and maybe the Genome::Process framework concretize the configuration into env variables when a Build or Process is launched, and persist the environment either in a file or in the DB.

I think this is orthogonal to how config values are resolved so long as there is a way to get all config values which there is.

File: lib/perl/Genome/ConfigSpec.t [hunk 4]
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/env genome-perl
+
+use strict;
+use warnings;
+
+use Genome;
```
tmooney: This one doesn't have the `qw()` that the other instances of `use Genome` in this PR do.  (It's a strange world if `Genome` starts exporting subroutines by default!)

lib/perl/Genome/Config.pm
```diff
@@ -0,0 +1,158 @@
...
+use File::Find::Rule qw();
+use Genome::Carp qw(croakf);
+use Genome::ConfigSpec qw();
+use Path::Class qw();
+use YAML::Syck qw();
+
+=item get()
+
+C<get($key)> retrieves the configuration value specified by the input key.  It
+does a hierarchival search for the value,
```
tmooney: `hierarchival` should be `hierarchical`.  This sentence might be clearer with a clause introducing the list, e.g. `...search for the value, which examines the following places:`.

lib/perl/Genome/Config.pm
```diff
@@ -0,0 +1,158 @@
...
+
+- environment variable
+- user config file
+- snapshot config file
+- global config files
```
tmooney: Is the global configuration special in that there are multiple files, but the other locations have only one?

lib/perl/Genome/Config.pm
```diff
@@ -0,0 +1,158 @@
...
+Values are specified in C<genome/config.yaml> in one of the source directories
+or via the corresponding environment variable.
+
+Configuration values are "registered" by creating a spec file, see
+L<Config::Spec>.  Spec files can go in either the snapshot config directory or
+the global config directories in the C<genome> subdirectory, e.g.
+C</etc/genome/foo.yaml>.
+
+If you want the configuration value to be optional you must specify a
+C<default_value> even if it is zero or an empty string but it must be defined.
```
tmooney: This is a run-on sentence.  One possibility is to split it in two: `...specify a C<default_value>.  The default can be zero or an empty string, but it must be defined.`

lib/perl/Genome/Config.pm
```diff
@@ -0,0 +1,158 @@
...
+sub get {
+    my $key = shift;
+    my $spec = spec($key);
+
+    my $value = _lookup_value($spec);
+    if (!$spec->has_default_value && !defined($value)) {
+        croakf('required config key missing: %s', $key);
+    }
+
+    if (!defined($value)) {
```
tmooney: Would the logic here be more clear if both this and the previous conditional were combined with a check on `$spec->has_default_value` inside?

lib/perl/Genome/Config.pm
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    my @errors = $spec->validate($value);
+    if (@errors) {
+        my $msg = $spec->validation_error(@errors);
+        croakf($msg);
+    }
+
+    if ($spec->sticky) {
+        unless ($spec->has_env) {
```
tmooney: Would it make sense for the `ConfigSpec` object to check that `env` is set when `sticky` is set?

lib/perl/Genome/ConfigSpec.t
nnutter: It's strange world when anything can export subroutines by default!

lib/perl/Genome/Config.pm
nnutter: Yes, the global directories are specified by a colon-delimited environment variable, `XGENOME_CONFIG_DIRS`, defaulting to `/etc`.

lib/perl/Genome/Config/Command/Get.pm
```diff
@@ -0,0 +1,26 @@
+package Genome::Config::Command::Get;
+
+use strict;
+use warnings;
+
+use Genome qw();
+use Genome::Config qw();
+
+class Genome::Config::Command::Get {
+    is => 'Command',
```
tmooney: Is there a reason to prefer `Command` over `Command::V2`?  (It seems like we should stop having two versions, but...)

tmooney:
Would it be useful to have a command that is sort of a cross between `genome config get` and `genome config validate` that lists all of the configuration sort of like `git config -l`?

lib/perl/Genome/Config.pm
nnutter: I'll have to think about that.

lib/perl/Genome/Config/Command/Get.pm
nnutter: For some reason I thought `Command` was the latest version not `V1`.  I'll switch it.

nnutter:
> Would it be useful to have a command that is sort of a cross between `genome config get` and `genome config validate` that lists all of the configuration sort of like `git config -l`?

I've thought about it but I'm not sure yet.  I'd probably just make `genome config get` default to all keys unless one or more is specified.

nnutter:
One thing I'm worrying about is whether requiring a defined value is practical or not.  I'd like it to be.

lib/perl/Genome/Config.pm
```diff
@@ -0,0 +1,158 @@
...
+    if (!defined($value)) {
+        if (!$spec->has_default_value) {
+            croakf('required config key missing: %s', $key);
+        }
+        $value = $spec->default_value;
+    }
+
+    my @errors = $spec->validate($value);
+    if (@errors) {
+        my $msg = $spec->validation_error(@errors);
```
tmooney: Maybe `$spec` could provide a method that's basically
```perl
sub {
  my ($self, $value) = @_;
  my @errors = $self->validate($value);
  return unless @errors;
  return $self->validation_error(@errors); 
}
```

lib/perl/Genome/Config.pm
```diff
@@ -0,0 +1,158 @@
...
+User's config files are discovered in C<XGENOME_CONFIG_HOME> which defaults to
+C<$HOME/.config>.  Global config files are discovered in C<XGENOME_CONFIG_DIRS>
+which defaults to C</etc> and can be a colon-delimited list of directories like
+C<PATH>.
+
+Values are specified in C<genome/config.yaml> in one of the source directories
+or via the corresponding environment variable.
+
+Configuration values are "registered" by creating a spec file, see
+L<Config::Spec>.  Spec files can go in either the snapshot config directory or
```
tmooney: This is currently `Genome::ConfigSpec` rather than `Config::Spec`.

lib/perl/Genome/ConfigSpec.pm
```diff
@@ -0,0 +1,123 @@
...
+
+Genome::ConfigSpec defines the properties a given configuration value should
+have.  These properties are: C<key>, C<type>, C<validators>, C<env>,
+C<default_value>, and C<sticky>.
+
+=head2 Required Properties
+
+C<key> is the an identifier for the configuration value.
+
+C<type> species the type of the value...
```
tmooney: Are the `...`s in this POD suggesting more is yet to be written?

lib/perl/Genome/Configurable.pm
```diff
@@ -0,0 +1,69 @@
...
+        set_prop_desc_with_warn($prop_desc, 'data_type', $spec->type);
+        set_prop_desc_with_warn($prop_desc, 'calculated_default', $default_method_name);
+    }
+
+    return $desc;
+}
+
+sub set_prop_desc_with_warn {
+    my ($prop_desc, $key, $value) = @_;
+    if ($prop_desc->{$key}) {
```
tmooney: A false value seems bogus, anyway, but would it better capture the intent if this were a `defined` or `exists` check instead?

tmooney:
I like the way this is turning out!  (And as a bonus `Genome::Site::TGI::LegacyConfig` might finally go away, too :smile:)

nnutter:
CI, test this.

lib/perl/Genome/Config/Command/Get.pm
```diff
@@ -0,0 +1,31 @@
...
+            is_many => 1,
+            is_optional => 1,
+            shell_args_position => 1,
+        },
+    ],
+};
+
+sub execute {
+    my $self = shift;
+    my @keys = $self->keys || Genome::Config::all_keys();
```
tmooney: Using `$self->keys` as a logical value will mean `@keys` gets the count instead of the keys themselves.

tmooney:
Should we move the existing modules under `Genome/Config/` to maybe claim the `Genome/AnalysisProject/` namespace instead?  (Or, possibly other things go under `Genome/Config/AnalysisProject/` and we leave that alone?)  They're sort of related, but sort of not.  

(For example, this *could* use `Genome::Config::Parser` to parse the YAML, but that seems kind of coincidental.)

lib/perl/Genome/Config/Command/Get.pm
nnutter: Thanks.  I also discovered that the new `calculated_default` calls in scalar context. ;-)

nnutter: Actually, I guess I'm just using it wrong.  It should probably return an arrayref.

nnutter:
> Should we move the existing modules under `Genome/Config/` to maybe claim the `Genome/AnalysisProject/` namespace instead?  (Or, possibly other things go under `Genome/Config/AnalysisProject/` and we leave that alone?)  They're sort of related, but sort of not.  

We'll probably want to do something.  In my mind that can wait until towards the end of this PR or maybe even as a follow-up PR.

nnutter:
I did another rebase (mostly to pickup a UR fix).  aab374b is the equivalent of the pre-rebase state.  I've added commits that remove a couple environment variables and converted a couple configuration variables.

nnutter:
I'm don't think the `type` means much in the `ConfigSpec` so I'm going to rethink that a bit.  I think this gets back to one of the feedback @acoffman had.

I'm also debating exporting a `config()` or `get_config()` (or similar) to use instead of `Genome::Config::get()`.

nnutter:
To continue to try and keep the PR reviewable I'm separating out the conversion of the codebase as a separate PR, see nnutter/genome#1.

nnutter:
CI, test this.

nnutter:
I screwed up nnutter/genome#1 by accident so now that work is on nnutter/genome#2.

nnutter:
CI, test this.

lib/perl/Genome/Test/Config.pm
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    my $config_file = Path::Class::File->new($genome_dir, $params{name});
+    YAML::Syck::DumpFile($config_file, $params{data});
+}
+
+sub temp_dir_helper {
+    my @temp_dirs;
+    my $new_temp_dir = sub {
+        push @temp_dirs, File::Temp->newdir();
```
tmooney: It might be good if this had the behaviour of `Genome::Sys->create_temp_directory` where it is aware of LSF's temp directory location when `bsub`bed.

lib/perl/Genome/Config.t
```diff
@@ -0,0 +1,137 @@
...
+
+subtest 'Genome does not need to be loaded for Genome::Config' => sub {
+    use_ok('Genome::Config');
+    ok(!scalar(grep { $_ eq 'Genome.pm' } keys %INC), 'Genome has not been loaded');
+};
+
+subtest 'basic lookup' => sub {
+    plan tests => 2;
+
+    my ($temp_dirs, $new_temp_dir) = Genome::Test::Config::temp_dir_helper();
```
tmooney: I believe that, as long as the caller here holds a reference to `$new_temp_dir`, the `@temp_dirs` inside will not be cleaned up even if a reference to that array is not held here.

nnutter: I think you are describing what I wanted.  I want this scope/`subtest` to hold the (only) reference to the `File::Temp` object so that it is cleaned up when the `subtest` ends.

nnutter:
One major decision I have left is whether to migrate everything to a new environment variable prefix to break ties with the legacy system.  The cost of that is missing any the places where we have override `/etc/profile.d/genome.sh`.

nnutter:
I'll need to coordinate changes to `puppet-modules` and `wugc-hudson` as part of this.

nnutter:
I've pulled out the `XGENOME_` prefix change and will do that as a separate PR.

nnutter:
I've removed the _work in progress_ label as I think all major changes are done.  I think this is close to being ready to ship.  I will be following this up PR up with a separate that changes the environment variables prefixes so we can get a clean break from the legacy system.

nnutter:
If it helps anyone wanting to help review, right now the newly added files are:

```
etc/genome/*
lib/perl/Genome/Config.t
lib/perl/Genome/Config/Command/Get.pm
lib/perl/Genome/Config/Command/Get.t
lib/perl/Genome/Config/Command/List.pm
lib/perl/Genome/Config/Command/List.t
lib/perl/Genome/Config/Command/SetEnv.pm
lib/perl/Genome/Config/Command/SetEnv.t
lib/perl/Genome/Config/Command/Validate.pm
lib/perl/Genome/Config/Command/Validate.t
lib/perl/Genome/ConfigSpec.pm
lib/perl/Genome/ConfigSpec.t
lib/perl/Genome/ConfigValidator/Defined.pm
lib/perl/Genome/ConfigValidator/LSFQueue.pm
lib/perl/Genome/ConfigValidator/Numeric.pm
lib/perl/Genome/ConfigValidator/Numeric.t
lib/perl/Genome/ConfigValidator/Port.pm
lib/perl/Genome/ConfigValidator/Positive.pm
lib/perl/Genome/ConfigValidator/UR.pm
lib/perl/Genome/ConfigValidator/URI.pm
lib/perl/Genome/ConfigValidator/URL.pm
lib/perl/Genome/ConfigValidatorBase.pm
lib/perl/Genome/Configurable.pm
lib/perl/Genome/Configurable.t
lib/perl/Genome/Test/Config.pm
```

Found by doing `git log --name-status @{u}.. | grep '^A       ' | awk '{ print $2 }' | sort -u`.

You should definitely look at `Genome::Config` to see how values are resolved.  Basically it's environment > home > snapshot > global.   For commands we'll be able to use `Genome::Configurable` which will allow the default value for an option/parameter to be looked up that way so for `Genome::Configurable` it's command line > environment > home > snapshot > global.  Shell scripts can make use of `genome config get` and `genome config set-env` to get or set values.

nnutter:
The `apipe-ci/bats` tests always run against `master` so fail.  I'll either have to ignore them or update them to run using the PR's genome/ur/workflow/sqitch code.

nnutter:
I've been waiting to merge this until model tests are passing again.  Once they are I'll merge it.

PR #636, opened 2015-04-06
Pass bedtools version to merge

tmooney:
The `BwamemStream` aligner uses a bedtools version.  This was wired up for `ReferenceAlignment` but also needs to be added to the `MergeAlignments` step used by the alignment dispatcher.

### DIFF
--- a/lib/perl/Genome/InstrumentData/Command/MergeAlignments.pm
+++ b/lib/perl/Genome/InstrumentData/Command/MergeAlignments.pm
@@ -47,6 +47,11 @@ class Genome::InstrumentData::Command::MergeAlignments {
             is => 'Text',
             doc => 'The version of Samtools to use when needed by mergers/deduplicators',
         },
+        bedtools_version => {
+            is => 'Text',
+            is_optional => 1,
+            doc => 'The version of bedtools to use when needed',
+        },
         result_users => {
             is => 'HASH',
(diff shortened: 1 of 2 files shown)


ebelter:
+1 :lollipop:

lib/perl/Genome/InstrumentData/Command/MergeAlignments.pm
```diff
@@ -108,6 +113,7 @@ sub params_for_merged_alignment {
         duplication_handler_params => $self->duplication_handler_params,
         duplication_handler_version => $self->duplication_handler_version,
         samtools_version => $self->samtools_version || undef,
+        bedtools_version => $self->bedtools_version || undef,
```
apregier: Can you remind me why we often have the "|| undef" added to these param hashes?  Wouldn't it jus be undef anyway?

apregier:
+1 I didn't see a real obvious place to add test coverage to this new feature, but it would be nice ;-)

lib/perl/Genome/InstrumentData/Command/MergeAlignments.pm
tmooney: I did it because the `samtools_version` did it. :smile: 

This normalizes `0` to `undef`, though that doesn't seem like something we really need to do.  (SoftwareResults already normalize `''` to `undef`, I believe.)  It's probably not useful here.

PR #637, opened 2015-04-06
remove skipping of transcripts with no domains.

ernfrid:
@apregier - can you look at this? Seems to do the right thing if I remove it, but I'm a little nervous it was in here for a reason that I'm not observing at the moment.


apregier:
+1
I believe Dan and Krishna made some figures by commenting this out last fall, and I didn't hear that there were problems.

PR #638, opened 2015-04-07
Allow spaces in comments again.

tmooney:
I wonder if a longer-term solution might be to add a minimum version attribute to Picard properties and have the base class handle this instead of it being a special case?


susannasiebert:
+1 once tests pass

jweible:
+1

PR #639, opened 2015-04-07
Exclude yaml from trio process diffs

apregier:
The yaml is just metadata, not a real output of the process.

### DIFF
--- a/lib/perl/Genome/VariantReporting/Process/Trio.pm
+++ b/lib/perl/Genome/VariantReporting/Process/Trio.pm
@@ -141,6 +141,7 @@ sub special_compare_functions {
     my $self = shift;
     my @functions = $self->SUPER::special_compare_functions(@_);
     push @functions, qr(\.xml$) => sub {Genome::VariantReporting::Process::Trio::compare_igv_xml(@_)};
+    push @functions, qr(\.yaml$) => sub {return 0};
     return @functions;
 }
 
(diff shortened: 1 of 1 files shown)


susannasiebert:
+1 once tests pass

apregier:
I realized my commit message is misleading.  The diff is not caused by result ids and paths, but rather by the library_name_labels being written in a nondeterministic order.  Another option would be to make that order deterministic, but since it is metadata it didn't seem worthwhile.

lib/perl/Genome/VariantReporting/Process/Trio.pm
```diff
@@ -140,7 +140,8 @@ sub get_cle_input_results {
 sub special_compare_functions {
     my $self = shift;
     my @functions = $self->SUPER::special_compare_functions(@_);
-    push @functions, qr(\.xml$) => sub {my ($a, $b) = @_; Genome::VariantReporting::Process::Trio::compare_igv_xml($a, $b)};
+    push @functions, qr(\.xml$) => sub {Genome::VariantReporting::Process::Trio::compare_igv_xml(@_)};
+    push @functions, qr(\.yaml$) => sub {return 0};
```
tmooney: If you wanted, you could have the first `push @functions` take both functions instead of repeating it for each one.

tmooney:
+1 :crossed_flags:

PR #641, opened 2015-04-07
Include QC failed reads when recreating per lane bam from merged bam

dufeiyu:
Most of our in-house bams do not include qc-failed reads, but some bams do include those, while currently recreate_bam command has include_qc_failed turned off. This bug causes unexpected bam flagstat diff and bam recreation failing. This PR is to fix the bug and recreate per lane bam to its original.

### DIFF
--- a/lib/perl/Genome/InstrumentData/AlignmentResult/Command/RecreatePerLaneBam.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult/Command/RecreatePerLaneBam.pm
@@ -41,6 +41,12 @@ class Genome::InstrumentData::AlignmentResult::Command::RecreatePerLaneBam {
             is  => 'FilePath',
             doc => 'The path of flagstat file that is used to compare with that of recreated bam',
         },
+        include_qc_failed => {
+			is          => 'Boolean',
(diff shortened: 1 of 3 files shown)


lib/perl/Genome/InstrumentData/AlignmentResult/Command/RecreatePerLaneBam.pm
```diff
@@ -41,6 +41,12 @@ class Genome::InstrumentData::AlignmentResult::Command::RecreatePerLaneBam {
             is  => 'FilePath',
             doc => 'The path of flagstat file that is used to compare with that of recreated bam',
         },
+        include_qc_failed => {
+			is          => 'Boolean',
+			doc         => 'Include reads that were marked within the bam as having failed QC',
+			is_optional => 1,
+			default     => 0,
+		},
```
ebelter: The indentation is off here.

dufeiyu: Messed up during copy/paste. Fixed.

acoffman:
+1

tmooney:
+1 :newspaper:

ebelter:
+1

PR #642, opened 2015-04-07
InstData Import: sort then sanitize; add attrs to skip for reimport

ebelter:


lib/perl/Genome/InstrumentData/Command/Import/Basic.pm
```diff
@@ -423,7 +426,10 @@ sub _add_sort_bam_op_to_workflow {
     my $sort_bam_op = $self->helpers->add_operation_to_workflow_by_name($self->_workflow, 'sort bam');
     $self->_workflow->add_link(
         left_operation => $previous_op,
-        left_property => 'output_bam_path',
+        #left_property => 'output_bam_path',
+        left_property => ( $previous_op->name eq 'verify not imported' ) # not ideal...
```
tmooney: Would it be better or worse to use something like:
```perl
$previous_op->operation_type->command_class_name->can('output_bam_path')
```

Alternatively, a silly solution might be to add a converge or block operation after `verify not imported` that is only used to transform the property name(s) as needed.

apregier:
+1

PR #644, opened 2015-04-08
fix lock directory being created with incorrect permissions

nnutter:
Partial fix for [RT#105060](https://rt.gsc.wustl.edu/Ticket/Display.html?id=105060).

### DIFF
--- a/lib/perl/Genome/Sys/Lock/FileBackend.pm
+++ b/lib/perl/Genome/Sys/Lock/FileBackend.pm
@@ -6,10 +6,11 @@ use warnings;
 use Carp qw(carp croak);
 use File::Basename;
 use File::Temp;
+use Genome::Utility::File::Mode qw(mode);
 use MIME::Lite;
+use Path::Class qw();
 use Sys::Hostname qw(hostname);
 use Time::HiRes;
-use Path::Class qw();
 
 use Genome::Logger;
 use Genome::Utility::Instrumentation;
@@ -294,8 +295,7 @@ sub tempdir_for_resource {
         Carp::croak("Failed to create temp lock directory ($tempdir)");
     }
 
-    # make this readable for everyone
-    chmod(0770, $tempdir) or Carp::croak("Can't chmod 0770 path ($tempdir): $!");
+    mode($tempdir)->add_group_rwx();
 
     return $tempdir;
 }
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Sys/Lock/FileBackend.pm
```diff
@@ -276,17 +277,11 @@ sub lock_dir_for_resource {
...
-    my $obj = Path::Class::dir($dir_path);
-    $obj->mkpath(0, 0777);
-}
-
 sub tempdir_for_resource {
     my ($self, $resource_lock) = @_;
 
     my $lock_dir = $self->lock_dir_for_resource($resource_lock);
-    $self->make_dir_path($lock_dir);
+    Genome::Sys->create_directory($lock_dir . '');
```
acoffman: My best guess is that this is here to force `$lock_dir` to stringify, but it [appears](https://github.com/nnutter/genome/blob/lock-pair/lib/perl/Genome/Sys/Lock/FileBackend.pm#L277) that we already have a string. Am I missing something? (Highly likely!). If it is to force stringfication, is there a reason to prefer this to `"$lock_dir"`?

lib/perl/Genome/Sys/Lock/FileBackend.pm
```diff
@@ -300,8 +295,7 @@ sub tempdir_for_resource {
         Carp::croak("Failed to create temp lock directory ($tempdir)");
     }
 
-    # make this readable for everyone
-    chmod(0770, $tempdir) or Carp::croak("Can't chmod 0770 path ($tempdir): $!");
+    mode($tempdir)->add_group_rwx();
```
acoffman: It probably doesn't actually matter, but do we want group execute permissions or just r/w?

nnutter: It's a directory so you want execute.

nnutter: You arguably don't need write.

lib/perl/Genome/Sys/Lock/FileBackend.pm
nnutter: You're right, I thought it was a `Path::Class::File`.  If it were you can `"$lock_dir"`, `$lock_dir . ''`, or `$lock_dir->stringify`.  TIMTOWTDI.

nnutter: Oh, and the forced stringification... it's paranoia and I probably shouldn't have bothered.

lib/perl/Genome/Sys/Lock/FileBackend.pm
acoffman: Ah yes, you're totally correct!

acoffman:
+1 :lock: :file_folder:

tmooney:
+1 :swimmer:

PR #645, opened 2015-04-08
In case the alignment result has a test name set

apregier:
This is a possible fix for rt 105030.  It's not great...


tmooney:
This might cause more problems than it solves.  Not finding the per-lane result is generally correct if it has a different test_name set.

brummett:
Agreed.  And if you've already found more than one, removing a filter isn't going to reduce the number of results.

Also...
```unless (@alignments and @alignments == 1)```
is more simply written
```unless (@alignments == 1)```

PR #646, opened 2015-04-08
New command to find references that have the same chromosome names as a query file.

tmooney:
Sometimes one may have a BED (or other tab-delimited) file but be unsure what reference sequences were involved in its creation (e.g. human plus several viruses).  This tool helps narrow down the candidates.

### DIFF
--- a/lib/perl/Genome/Model/ReferenceSequence/Command/FindPotentialMatches.pm
+++ b/lib/perl/Genome/Model/ReferenceSequence/Command/FindPotentialMatches.pm
@@ -75,10 +75,8 @@ sub execute {
     my @subsets;
     my @partials;
     while (my $reference = $reference_iterator->next) {
-        my $seqdict_path = File::Spec->join($reference->data_directory, 'seqdict', 'seqdict.sam');
-        next unless(-e $seqdict_path); #don't try to generate this now
-
-        my $reference_chromosomes = $reference->chromosome_array_ref;
+        my $reference_chromosomes = $reference->chromosome_array_ref(create_seqdict => 0);
+        next unless defined $reference_chromosomes;
 
         my ($matching, $only_query, $only_reference) = UR::Util::intersect_lists($query_chromosomes, $reference_chromosomes);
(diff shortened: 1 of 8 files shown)


lib/perl/Genome/Model/ReferenceSequence/Command/FindPotentialMatches.pm
```diff
@@ -0,0 +1,186 @@
...
+
+    my $reference_iterator = $self->_reference_iterator;
+
+    my @matches;
+    my @supersets;
+    my @subsets;
+    my @partials;
+    while (my $reference = $reference_iterator->next) {
+        my $seqdict_path = File::Spec->join($reference->data_directory, 'seqdict', 'seqdict.sam');
+        next unless(-e $seqdict_path); #don't try to generate this now
```
davidlmorton: I'm assuming that `$reference->chromosome_array_ref` relies on the presence of this seqdict file in some way?  Perhaps a method on that class would be more appropriate?  Something like `has_seqdict` or `has_chromosome_array_ref` or the like.

davidlmorton:
Yay for hand-crafted (small) test data that lives in the repo!

lib/perl/Genome/Model/ReferenceSequence/Command/FindPotentialMatches.pm
tmooney: Yes, `$reference->get_sequence_dictionary(...)` will try to create the seqdict file if it doesn't exist, and `$reference->chromosome_array_ref` calls it.  We have some oddball references for which the sequence dictionary can't be created, so those were crashing this tool.  (l'm not sure what purpose such references serve.)

That said, I agree it would be better for this to ask the reference about the existence of the file instead of doing this itself.  I'll change it!

apregier:
+1
How will this interact with the reference converter?

tmooney:
I don't believe it will interact with the reference converter at all.

acoffman:
+1

PR #647, opened 2015-04-09
Limit merged results to ones that match all params

apregier:
This is an alternate solution to rt 105030 (instead of #645 )

### DIFF
--- a/lib/perl/Genome/InstrumentData/AlignmentResult.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult.pm
@@ -1726,9 +1726,15 @@ sub get_merged_bam_to_revivify_per_lane_bam {
 sub get_merged_alignment_results {
     my $self = shift;
     # Always load from the database, since other merged results may have committed since we updated the UR cache
+    my %params;
+    for my $param ($self->__meta__->params) {
+        my $name = $param->property_name;
+        $params{$name} = $self->$name;
+    }
     my @results = Genome::InstrumentData::AlignmentResult::Merged->load(
         'inputs.value_id' => $self->instrument_data_id,
         test_name => $self->test_name,
+        %params,
     );
     my @filtered_results = $self->filter_non_database_objects(@results);
     return $self->filter_non_matching_results(@filtered_results);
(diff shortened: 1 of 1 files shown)


apregier:
This might make the query a lot slower (I'm not sure - haven't tried it empirically)

tmooney:
If this does trigger the slow query, maybe an alternative would be to `define_boolexpr` with the params inside of `filter_non_matching_results` in order to `evaluate` each one before trying to `collect_individual_alignments`.

apregier:
It's definitely not orders of magnitude slower - I have a script that runs on 8 alignment results in < 4 seconds, so I don't think I will worry about optimizing it unless other evidence arises.

apregier:
I guess one risk is that adding the params could limit the query too much and exclude too many merged alignment results, so that we wouldn't find one even if one existed.  Does anyone see a possibility of that happening?

nnutter:
Maybe you can avoid the slow query by using the `lookup_hash` instead (like `collect_individual_alignments` does?

apregier:
We can't do lookup hash because we don't have the full set of params (specifically, we are missing the merger and deduplication version, params, etc.
I also don't think the query seems to be that slow

nnutter:
One might not be slow.  Regardless, I'd rather see slow and correct than fast and broken.  We could instrument the call and if it is too slow in use then we can figure out an alternative.

lib/perl/Genome/InstrumentData/AlignmentResult.pm
```diff
@@ -1726,9 +1726,15 @@ sub get_merged_bam_to_revivify_per_lane_bam {
 sub get_merged_alignment_results {
     my $self = shift;
     # Always load from the database, since other merged results may have committed since we updated the UR cache
+    my %params;
+    for my $param ($self->__meta__->params) {
+        my $name = $param->property_name;
+        $params{$name} = $self->$name;
+    }
```
nnutter: @brummett, can't you go from a UR::Object to a UR::BoolExpr?  I thought there was something like,

```
my %params = UR::BoolExpr->resolve($self)->params_list;
```

apregier:
Oops, the script I thought was testing this was not doing at all what I thought - the query is in fact slow when I actually do it.  I will explore options.

gsanders:
How slow is it, really? Unless it is severe I would be inclined to agree that you don't need to optimize now and we can just go with this.

tmooney:
+1  :speedboat:, this version seems reasonable and avoids loading tons of individual alignment results!

gsanders:
I should have included a +1 with my previous comment. But if you find that optimization is necessary I can re-review!

apregier:
Thanks, @gsanders I did implement @tmooney's suggestion of using boolexpr to evaluate, so this isn't slow anymore.

nnutter:
:+1:

PR #648, opened 2015-04-09
Refactor per-lane bam locking out of Merged.pm and into AlignmentResult.pm

dufeiyu:
This PR refactors per-lane bam locking code out of G::I::AR::Merged and into G::I::AlignmentResult and handles both old and new lock during the migration. It also removes an extra reallocation. This PR replaces the old PR#574.

### DIFF
--- a/lib/perl/Genome/InstrumentData/AlignmentResult.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult.pm
@@ -1969,15 +1969,20 @@ sub lock_bam_file_access {
     unless ($self->get_merged_alignment_results) {
         my @bams = $self->alignment_bam_file_paths;
         unless (@bams) {
-            die $self->error_message("Alignment result with class (%s) and id (%s) has neither ".
-                "merged results nor valid bam paths. This likely means that this alignment result ".
-                "needs to be removed and realigned because data has been lost. ".
-                "Please create an apipe-support ticket for this.", $self->class, $self->id);
-            #There is no way to recreate per lane bam if merged bam
-            #does not exist and per lane bam is removed. Software
-            #result of this per lane alignment needs to be removed and
-            #this per lane instrument data needs to be realigned
-            #with that aligner.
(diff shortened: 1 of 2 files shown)


lib/perl/Genome/InstrumentData/AlignmentResult.pm
```diff
@@ -1963,5 +1963,61 @@ sub add_to_temporary_input_files_queue {
     return 1;
 }
 
+sub lock_bam_file_access {
+    my $self = shift;
+
+    unless ($self->get_merged_alignment_results) {
+        my @bams = $self->alignment_bam_file_paths;
+        unless (@bams) {
+            die $self->error_message("Alignment result with class (%s) and id (%s) has neither ".
```
tmooney: Unless we have already locked, could the merged alignment result remove the BAM between `$self->get_merged_alignment_results` and `$self->alignment_bam_file_paths` and trigger this `die`?

dufeiyu: It's unlikely considering that longer time will be taken between merge commit  and per lane bam removal. Even it happens and triggers the die, the new build will be ok since merged_alignment_results become available by then.

davidlmorton: That's even worse!  This process gets the bam file path here, and then the merged alignment step commits and releases the lock, then we get the lock.  Now we have a path to a bam file that will be deleted out from under us.

dufeiyu: alignment_bam_file_paths is used here only to check for some extreme cases where neither merged alignment results nor per lan bams exist. If the merge from other process commits, removes the per lane bam and releases the lock, this process will get the lock, but immediately unlock this lock after checking get_merged_alignment_results again that returns true this time.

dufeiyu:
The PR#663 makes this PR obsolete. Close it.

PR #649, opened 2015-04-10
InstData Import: remove support for direct interaction with cghub

ebelter:


ebelter:
Merging early. This PR removes code and I need to make more changes that are dependent upon it.

PR #651, opened 2015-04-10
Genome::Carp

nnutter:
This wraps Carp's functions to support `sprintf`, for example,

    carpf('%s died: %s', $self->__display_name__, $!);

One suggestion I've already received was to move this to Genome::Utility.
Please let me know if you think that's a better place for this.


lib/perl/Genome/Carp.t
```diff
@@ -0,0 +1,32 @@
...
+    like($exception, qr/^hello joe/, "$die_method did die");
+}
+
+my @warn_methods = qw(carpf cluckf warnf);
+for my $warn_method (@warn_methods) {
+    my $method = \&{ "Genome::Carp::$warn_method" };
+    my $warning;
+    local $SIG{__WARN__} = sub { $warning = $_[0] };
+    $method->('hello %s', 'joe');
+    like($warning, qr/^hello joe/, "$warn_method did warn");
```
ebelter: Could you use Test::Warn here?

lib/perl/Genome/Carp.pm
```diff
@@ -0,0 +1,41 @@
...
+    my $carp_method_name = substr($method_name, 0, -1);
+    my $carp_method = Carp->can($carp_method_name);
+    unless ($carp_method) {
+        die 'method not found: Carp::' . $carp_method_name;
+    }
+
+    install_sub({
+        code => sub {
+            my ($template, @args) = @_;
+            local $Carp::CarpLevel = 1;
```
ebelter: Carp doc says CarpLevel is hard to get to work correctly. It suggests @CARP_NOT should be used instead to remove the class from the call stack. Do you disagree?

lib/perl/Genome/Carp.t
nnutter: We could.  It's not a current dependency.  If I were to choose I would probably choose [Test::Warnings](https://metacpan.org/pod/Test::Warnings).

lib/perl/Genome/Carp.pm
nnutter: I don't know if I disagree but I did willfully ignore that warning.

One reason is that,

> Usually people skip too many call frames.

made me less nervous since I was setting it `1`.  Another was that the wrapper function would have to inspect `caller` which bothered me a tiny bit.

ebelter: The fact you are skipping 1 is probably not going to cause an issue.

ebelter:
+1

lib/perl/Genome/Carp.t
ebelter: I did not check the dependencies. I saw that is is 'installed'. Good either way.

lib/perl/Genome/Carp.pm
brummett: Shouldn't it be
```local $Carp::CarpLevel = $Carp::CarpLevel + 1;```
In case the caller has also set a CarpLevel?

File: lib/perl/Genome/Carp.pm [hunk 3]
```diff
@@ -0,0 +1,41 @@
...
+
+sub dief {
+    my ($template, @args) = @_;
+    return Carp::croak(sprintf($template, @args));
+}
+
+sub warnf {
+    my ($template, @args) = @_;
+    return Carp::carp(sprintf($template, @args));
+}
```
brummett: If the caller is using ```dief``` or ```warnf```, then they probably don't want Carp's location reporting behavior, and instead want die and warn's behavior.

nnutter: With `$Carp::CarpLevel` set to default, `0`, this behaves like I would expect it:

```
$ cat foo.pl 
use strict;
use warnings;
use Genome::Carp qw(dief warnf);
warnf 'this is a %s', 'warning';
dief 'this is a %s', 'die';
```

```
$ genome-test-env -D genome-perl foo.pl
...
this is a warning at foo.pl line 4
this is a die at foo.pl line 5
```

Maybe I should `local $Carp::CarpLevel = 0`?

lib/perl/Genome/Carp.pm
nnutter: An alternative would be to only set it to `1` if it was `0` (or `undef`).  At the moment I would lean towards `1` if `0` as being less surprising.

lib/perl/Genome/Carp.t
nnutter: Turns out there is no pre-existing package for `Test::Warnings` but is for `Test::Warn` so that would motivate using the latter.

lib/perl/Genome/Carp.pm
brummett: You're right - I was thinking too much about it.  The issue I mention only arises if the caller's package puts Genome::Carp in its ```@CARP_NOT``` list, which is unlikely.

lib/perl/Genome/Carp.pm
brummett: I think you always have to increment it.  In this test, Dier::do_die sets its own CarpLevel to make the message report its caller's caller (GrandparentCaller::calls_do_die), but the reported location is Caller::calls_do_die, because ```croakf``` is stomping over the existing CarpLevel.  If you swap the two ```local``` lines in croakf, then you get the location the caller wants.

```
use Carp qw();

package Dier;

sub do_die {
    local $Carp::CarpLevel = 1;
    Genome::Carp::croakf('in do_die');
}

package Caller;

sub calls_do_die {
    Dier::do_die();
}

package GrandparentCaller;

sub calls_do_die {
    Caller::calls_do_die();
}

package Genome::Carp;

sub croakf {
    my($tmpl, @params) = @_;
    local $Carp::CarpLevel = 1;  # Existing, wrong
#    local $Carp::CarpLevel =  $Carp::CarpLevel + 1;  # Right
    Carp::croak(sprintf($tmpl, @params));
}

package main;

GrandparentCaller::calls_do_die();
```

lib/perl/Genome/Carp.t
ebelter: :tongue:

lib/perl/Genome/Carp.pm
ebelter: Why not use @CARP_NOT ?

nnutter:
I've pushed new commits.

brummett:
+1

PR #652, opened 2015-04-11
InstData Import: move setting original_data_path and do not set if exists

ebelter:


ur
```diff
@@ -1 +1 @@
-Subproject commit af01a9f82e18d0b11ef2c1953b7e3df7444a66dd
+Subproject commit c80b913aeddf01bfcdab0487cf99d27749b3be5c
```
tmooney: Did you mean to include this in the PR?  This appears to revert the inclusion of genome/UR#69.

ebelter: No. Removing!!

ebelter:
I am going to close and redo the PR.

PR #653, opened 2015-04-13
reserve 8 GB memory for ClinSeq's MicroarrayCnv

nnutter:
The MRS1 were failing due to reaching the default memory limit of 4 GB. As a
quick fix I increased memory to 16 GB and the models were run but only
consumed about 6.5 GB so for this change I've limited the reservation to 8 GB.

See [RT#105133][1] for additional context.

[1]: https://rt.gsc.wustl.edu/Ticket/Display.html?id=105133


malachig:
+1.  Thank @nnutter this approach sounds very reasonable to me.

ebelter:
+1

PR #654, opened 2015-04-13
Fill a hole in the system of tracking result users

apregier:
The build uses this result relationship to find the result later and this one seems to be missing.  I am running the model test as well to make sure it finds the correct results and fixes the diff.

### DIFF
--- a/lib/perl/Genome/Model/SomaticVariation/Command/TierVariants.pm
+++ b/lib/perl/Genome/Model/SomaticVariation/Command/TierVariants.pm
@@ -236,6 +236,7 @@ sub link_result_to_build {
 
         Genome::Sys->create_symlink($f, join('/', $effects_dir, $name));
     }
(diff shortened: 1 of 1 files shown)


nnutter:
:+1:

tmooney:
[This](https://github.com/apregier/genome/blob/77a2e300d3ee2b709278d7bd0fb762b34c80a6c4/lib/perl/Genome/Model/SomaticVariation/Command/TierVariants.pm#L211) should be linking the tiered results to the build.  The hole is that results of the [target class](https://github.com/apregier/genome/blob/77a2e300d3ee2b709278d7bd0fb762b34c80a6c4/lib/perl/Genome/Model/SomaticVariation/Command/TierVariants.pm#L199) are not found when running tiering.

apregier:
This isn't the real solution.

PR #657, opened 2015-04-13
GMT CG Hub - Add command to help TCGA imports

ebelter:


apregier:
+1
It did take me a long time to understand this code, but I guess a lot of it is cleaning up the confusing data structure that you get back from CgHub

lib/perl/Genome/Model/Tools/CgHub/Metadata.pm
```diff
@@ -13,175 +13,143 @@ class Genome::Model::Tools::CgHub::Metadata {
...
-    else {
-        # unknown library strategy
-        $target_region = 'none';
-    }
+    my $result = $self->metadata->{$lookup_id};
+    return if not $result;
+    my $value = $result->{$name};
 
-    return $target_region;
+    return if not defined $value or $value eq '';
```
apregier: This line is pretty redundant with the next line, right?

ebelter: Yeah - There is some code leftover as I updating having the metadata hold multiple analysis ids.

lib/perl/Genome/Model/Tools/CgHub/Metadata.pm
```diff
@@ -13,175 +13,143 @@ class Genome::Model::Tools::CgHub::Metadata {
...
+    return $value;
 }
 
 sub _files {
-    my $self = shift;
+    my ($self, $lookup_id) = Params::Validate::validate_pos(@_, {type => HASHREF}, {type => SCALAR});
```
tmooney: Some places are validating `$self` as an `OBJECT` while others are validating it as a `HASHREF`.  They should probably all be `OBJECT` (or alternatively `{isa => __PACKAGE__}` or the like).

PR #659, opened 2015-04-13
InstData Import: move setting original_data_path and do not set if exists

ebelter:


acoffman:
+1

tmooney:
+1 :eight_spoked_asterisk:

PR #660, opened 2015-04-14
check for empty segments

gatoravi:
This PR adds a check for the empty segments case. Currently this errors out with 'replacement empty' error.


chrisamiller:
+1

acoffman:
+1

PR #661, opened 2015-04-14
disable diagnostics when using compare_ok with binary files

nnutter:
We should disable diagnostics when using `compare_ok()` with binary files.


tmooney:
+1 :minidisc:

nnutter:
I forgot `compare_ok` supported disabling diagnostics (diff) message so I rebased with an alternate fix.

acoffman:
+1

tmooney:
Still +1 :smile:

PR #662, opened 2015-04-14
fix bug exposed in newer Bash versions

nnutter:
@mcallaway discovered this bug when trying to use `genome-env` on a newer Bash
version,

    $ genome-env -D ...
    .../genome-env: line 279: SUBMODULES: unbound variable

### DIFF
--- a/bin/genome-env
+++ b/bin/genome-env
@@ -274,9 +274,9 @@ function assert_dir {
     fi
 }
 
-declare -a SUBMODULES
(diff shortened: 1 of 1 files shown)


apregier:
+1

nnutter:
_Note: I did reproduce the problem and test that this fixed it._

tmooney:
+1 :space_invader:

PR #663, opened 2015-04-14
A new method get_bam_file to handle lock and return revivified bam or copied in-place bam

dufeiyu:
A new method get_bam_file is added to G::I::AR to handle the per lane bam lock and return either revivified_alignment_bam_file_path or copied in-place bam path depending on get_merged_alignment_results.

### DIFF
--- a/lib/perl/Genome/InstrumentData/AlignmentResult.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult.pm
@@ -1795,7 +1795,7 @@ sub _get_temp_allocation {
     );
     UR::Context->process->add_observer(
         aspect   => 'pre-commit',
-        once     => 1
+        once     => 1,
         callback => sub { $temp_allocation->delete; },
     );
     return $temp_allocation;
--- a/lib/perl/Genome/InstrumentData/AlignmentResult.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult.pm
@@ -1794,7 +1794,8 @@ sub _get_temp_allocation {
         owner_id            => Genome::Sys->username,
     );
     UR::Context->process->add_observer(
-        aspect => 'pre-commit',
+        aspect   => 'pre-commit',
+        once     => 1
         callback => sub { $temp_allocation->delete; },
     );
     return $temp_allocation;
--- a/lib/perl/Genome/InstrumentData/AlignmentResult.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult.pm
@@ -1645,6 +1645,97 @@ sub alignment_bam_file_paths {
     return glob(File::Spec->join(shift->output_dir, "*.bam"));
 }
 
+
+sub get_bam_file {
+    my $self = shift;
+
+    if ($self->get_merged_alignment_results) {
+        return $self->revivified_alignment_bam_file_path;
+    }
+    else {
+        my $lock = $self->get_bam_lock;
+        if ($self->get_merged_alignment_results) {
+            $lock->unlock;
+            return $self->revivified_alignment_bam_file_path;
+        }
+        else {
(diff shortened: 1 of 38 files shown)


lib/perl/Genome/InstrumentData/AlignmentResult.pm
```diff
@@ -1642,9 +1642,101 @@ sub verify_alignment_data {
...
+        my $lock = $self->get_bam_lock;
+        if ($self->get_merged_alignment_results) {
+            $lock->unlock;
+            return $self->revivified_alignment_bam_file_path;
+        }
+        else {
+            my @bams = $self->alignment_bam_file_paths;
+            unless (@bams) {
+                $lock->unlock;    
+                if ($self->get_merged_alignment_results) {
```
tmooney: Is it possible for this to become true when it was false on line right after acquiring the lock?

dufeiyu: It is unlikely. Adding check here just in case that it happens somehow.

lib/perl/Genome/InstrumentData/AlignmentResult.pm
```diff
@@ -1699,14 +1783,21 @@ sub revivified_alignment_bam_file_path {
...
         allocation_path     => 'merged/recreated_per_lane_bam/'.$self->id.'_'._get_uuid_string(),
-        kilobytes_requested => $self->bam_size || 100_000_000, # This should always be set, but just in case we reserve a lot
+        kilobytes_requested => $bam_size_kilobytes || 100_000_000, # This should always be set, but just in case we reserve a lot
         owner_class_name    => 'Genome::Sys::User',
         owner_id            => Genome::Sys->username,
     );
+    UR::Context->process->add_observer(
+        aspect => 'pre-commit',
+        callback => sub { $temp_allocation->delete; },
+    );
```
tmooney: This observer should have the `once => 1` property set.

lib/perl/Genome/InstrumentData/AlignmentResult/Merged/BamQc.pm
```diff
@@ -160,9 +148,12 @@ sub create {
...
 
     # use per lane instrument data id as link name
-    if ($lane_flag) {
-        my $instr_data = $self->alignment_result->instrument_data;
+    my @instrument_data = $self->alignment_result->instrument_data;
+    my $lane_flag;
+    if (@instrument_data == 1) {
+        my $instr_data = $instrument_data[0];
         $bam_qc_params{bam_link_name} = $instr_data->id;
+        $lane_flag = 1;
```
tmooney: Is it okay that the `$lane_flag` is now being set on merged results with a single instrument data whereas it was not before?

lib/perl/Genome/InstrumentData/AlignmentResult.pm
dufeiyu: ok, I will add it.

lib/perl/Genome/InstrumentData/AlignmentResult/Merged/BamQc.pm
dufeiyu: Good catch. I will change to set lane_flag if alignment_result isa('Genome::InstrumentData::AlignmentResult').

ebelter:
+1 :four_leaf_clover: once tests are passing.

gsanders:
+1 once tests pass. Everything looks great! This is much better than what was in place previously. Nice work!

tmooney:
+1 with passing tests! :confetti_ball:

susannasiebert:
I think we still need to have a conversation about what bam_path and bam_file in G::ID::AlignmentResult are supposed to return. However, that can be handled in a different, QC-related PR after some additional talks with Allison and Feiyu. 

+1

PR #664, opened 2015-04-14
Upgrade in Genome::Model::Tools::Analysis::Coverage::MergeReadcounts and...

gschang:
Upgrade and bug fixation in

Genome::Model::Tools::Analysis::Coverage::MergeReadcounts
Genome::Model::Tools::Analysis::Coverage::CoveragePlot

### DIFF
--- a/lib/perl/Genome/Model/Tools/Analysis/Coverage/MergeReadcounts.pm
+++ b/lib/perl/Genome/Model/Tools/Analysis/Coverage/MergeReadcounts.pm
@@ -6,23 +6,26 @@ use warnings;
 
 
 class Genome::Model::Tools::Analysis::Coverage::MergeReadcounts{
-    is => 'Command',
+    is => 'Command::V2',
     has => [
 	bam_files => {
 	    is => 'String',
 	    is_optional => 0,
+	    is_many => 1,
 	    doc => 'comma-separated list of bam files to grab readcounts from, Output columns will be appended in this order',
 	},
 	
 	variant_files => {
 	    is => 'String',
 	    is_optional => 0,
+	    is_many => 1,
 	    doc => 'coma-separated list of text files containing snvs in annotation format (1-based, first 5-cols = [chr, st, sp, ref, var]), all variant sites will be combined to a list',
 	},
 	
 	variant_sources => {
 	    is => 'String',
-	    is_optional => 0,
+	    is_optional => 1,
+	    is_many => 1,
 	    doc => 'coma-separated list of the name for each variant_file, used in the source column (1-based index number as default such as 1, 2, 3, ...)',
 	},
 	
@@ -85,6 +88,7 @@ class Genome::Model::Tools::Analysis::Coverage::MergeReadcounts{
 	header_prefixes => { 
 	    is => 'String',
 	    is_optional => 1,
+	    is_many => 1,
 	    doc => 'Comma-separated list - if the file has a header, three column titles get added for each bam ("ref_count","var_count","VAF"). This specifies a prefix for those columns. (i.e.  "Normal" will lead to "Normal_ref_count","Normal_var_count","Normal_VAF"). As a default, the ascending integers in the same BAM order, e.g. 1, 2, 3, ...',
 	},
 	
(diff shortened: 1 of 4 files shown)


lib/perl/Genome/Model/Tools/Analysis/Coverage/MergeReadcounts.pm
```diff
@@ -210,17 +213,16 @@ sub execute {
...
     
     
     # reads the variant files and combine
     # for a hash with the keys chr, start, stop, ref, var, type
     my (%hash, @header);
-    for (my $n=0; $n<@variants; $n ++)
+    for (my $n=0; $n<@variant_files; $n ++)
     {
-	my $variant = $variants[$n];
+	my $variant = $variant_files[$n];
```
nnutter: There's nothing wrong with this at all but I wanted to share an alternative way to loop through `@variant_files` and `@sources` using [List::MoreUtils::each_array][1] in case you find it more expressive.

```
use List::MoreUtils qw(each_array);

my $ea = each_array(@variant_files, @sources);
while (my ($variant, $source) = $ea->()) {
     ...
}
```

[1]: https://metacpan.org/pod/List::MoreUtils#each_array-ARRAY1-ARRAY2

lib/perl/Genome/Model/Tools/Analysis/Coverage/MergeReadcounts.pm
```diff
@@ -237,7 +239,17 @@ sub execute {
 	else
 	{
 	    my @new = split /\t/, $line;
-	    die "Inconsistent annotation format found: " . $variant unless scalar(@header) == scalar(@new);
+	    
+	    die "Inconsistent table format found in the header line: " . $variant unless scalar(@header) == scalar(@new);
+	    
+	    # double-checks whether they share the same column header
+	    for (my $i=0; $i<@header; $i ++)
```
nnutter: Same here.  Nothing wrong at all but here's an alternative using `List::MoreUtils`' [all][1] and [pairwise][2].
```
use List::MoreUtils qw(all pairwise);

unless (all { $_ } pairwise { $a eq $b } @header, @new) {
    die ...
}
```

Both [List::Util][3] and [List::MoreUtils][4] are very handy libraries.  Do be aware our version is a bit out of date so viewing the documentation for our installed version would be better, e.g. `perldoc List::MoreUtils`.

[1]: https://metacpan.org/pod/List::MoreUtils#all-BLOCK-LIST
[2]: https://metacpan.org/pod/List::MoreUtils#pairwise-BLOCK-ARRAY1-ARRAY2
[3]: https://metacpan.org/pod/List::Util
[4]: https://metacpan.org/pod/List::MoreUtils

nnutter:
:+1: I made a couple tangential comments but they are just informational.  Another informational comment I would make is that you should be prudent when changing command interfaces, i.e. removing support for comma-delimited lists for `--bam-files`, etc., as it could break scripts people may have written using the old interface.  Usually the minimum we would do is announce the change by making a "CHANGELOG" comment in Git; APIPE can help you do that if you wish to do so.

lib/perl/Genome/Model/Tools/Analysis/Coverage/MergeReadcounts.pm
```diff
@@ -6,23 +6,26 @@ use warnings;
...
 	    is_optional => 0,
+	    is_many => 1,
 	    doc => 'comma-separated list of bam files to grab readcounts from, Output columns will be appended in this order',
 	},
 	
 	variant_files => {
 	    is => 'String',
 	    is_optional => 0,
+	    is_many => 1,
 	    doc => 'coma-separated list of text files containing snvs in annotation format (1-based, first 5-cols = [chr, st, sp, ref, var]), all variant sites will be combined to a list',
```
tmooney: This isn't something you changed, but there is a typo here and in the doc for `variant_sources`--`comma` is short an `m` in both places.

tmooney:
+1, the improvements look good!

chrisamiller:
+1

susannasiebert:
+1 

The `is_many` change to `bam_files` won't change the command line invocation of this tool but it does change how people will need to call this tool from of another perl module/script. I agree with @nnutter that a CHANGELOG entry would be helpful if you think that people are using this tool.

nnutter:
@susannasiebert, thanks for correcting that.  I was not aware that `String` `is_many` would assume it could split on commas.  (I may even be to blame.)

gschang:
Hi, all.

Thank you for many nice comments. I have made minor updates with "List::MoreUtils" module. Since I got +3, I am going to merge it soon.

PR #665, opened 2015-04-14
Disable sqlite_use_immediate_transaction to restore old behaviour.

tmooney:
As part of testing this on `trusty`, the MetaDB was sometimes deadlocking itself between processes.  This was caused by a newer `DBD::SQLite` behaviour where it [acquires a lock immediately](http://search.cpan.org/~ishigaki/DBD-SQLite-1.46/lib/DBD/SQLite.pm#Transaction_and_Database_Locking) upon the first read/write from the DB.  (This had the side-effect of causing failures if the MetaDB was read-only.)  This flag restores the old behaviour we're used to on `lucid`.

There may be other ways of avoiding this problem (such as setting `AutoCommit` unless we're intending to make changes).

### DIFF
--- a/lib/perl/Genome/DataSource/Meta.pm
+++ b/lib/perl/Genome/DataSource/Meta.pm
@@ -6,12 +6,27 @@ package Genome::DataSource::Meta;
 use strict;
 use warnings;
 
+use version;
+
 use UR;
+use DBD::SQLite;
 
 UR::Object::Type->define(
     class_name => 'Genome::DataSource::Meta',
     is => ['UR::DataSource::Meta'],
 );
 
+sub _dbi_connect_args {
+    my $self = shift;
+
+    my @connection = $self->SUPER::_dbi_connect_args(@_);
+
+    if(version->parse($DBD::SQLite::VERSION) >= version->parse('1.38_01')) {
+        my $connect_attr = $connection[3] ||= {};
+        $connect_attr->{sqlite_use_immediate_transaction} = 0;
+    }
+
+    return @connection;
+}
 
 1;


nnutter:
- I don't have enough information to understand the bug.
- It seems probable that this belongs in `UR::DataSource::Meta`.

tmooney:
* Under newer `DBD::SQLite`, an exclusive lock on the SQLite database is obtained when executing the first query and retained until commit.  This leads to two problems:
 * If the MetaDB is not writable (such as in a snapshot directory) this results in an immediate failure [`attempt to write to a read-only database`].
 * If instead it is writable, then any other processes in any build (including, for instance, subtasks spawned by the parent) will wait for the lock [they get `EAGAIN` from `fcntl`].  The parent, however, is waiting on the child and so it will never release the lock.
* Moving it into `UR` seems reasonable.  I based this on [the method in `Genome::DataSource::GMSchema`](https://github.com/genome/genome/blob/2d85829566b36f58540b460e8fa471cff0e07f2e/lib/perl/Genome/DataSource/GMSchema.pm#L40-L50), but it seems likely this change would be more broadly desirable than that one.

brummett:
It's my understanding that in normal operation, access to the MetaDB should never get deadlocked.  The MetaDB is only ever read from (except at the end of ```ur update classes-from-db``` which we don't normally do), and so there should ever be a shared lock on the DB file.

[This page](http://sqlite.org/lockingv3.html) explains SQLite locking, especially part 7.0 at the bottom.

I know in the past, we've had processed deadlocking against themselves because they've inadvertently created multiple handles to the MetaDB, but I thought I've fixed all those.

tmooney:
Newer `DBD::SQLite` issues `BEGIN IMMEDIATE` upon the first query (see also the `DBD::SQLite` documentation linked in the original PR comment), which [this page](http://www.sqlite.org/lang_transaction.html) says precludes any other processes from succeeding to `BEGIN IMMEDIATE` until the first transaction is complete.

nnutter:
> To avoid this, DBD::SQLite internally issues a `BEGIN IMMEDIATE` when you begin a transaction by `begin_work` or under the `AutoCommit` mode (since 1.38_01).

Are we using `AutoCommit` mode?  It seems like we're (previously) using [UR::DataSource::RDBMS::_db_connect_args](https://github.com/genome/UR/blob/74df78e80f7e642c878e8be5350fe197fa3191a2/lib/UR/DataSource/RDBMS.pm#L441) which has `AutoCommit => 0`.  There is a [case](https://github.com/genome/UR/blob/74df78e80f7e642c878e8be5350fe197fa3191a2/lib/UR/DataSource/SQLite.pm#L110) that uses `AutoCommit` but I don't think that's used for `*::DataSource::Meta`.

But that quote seems to contradict,

> As of version 1.38_01, this attribute is set to true by default. If you really need to use deferred transactions for some reasons, set this to false explicitly.

This also may be relevant,

> If you need to `fork()`, (re)open a database after you `fork()`. You might also want to tweak `sqlite_busy_timeout` and `sqlite_use_immediate_transaction` (see below), depending on your needs.

I think one issue is there is not test case representing the bug but creating a test case is difficult because it would requiring temporarily using a newer DBD::SQLite.  (This is similar to the `genome-env` fix.)  @tmooney, were you able to reproduce the problem and confirm this fixed it?

tmooney:
I think this part best describes what we're seeing: 
> <h5>When the AutoCommit flag is off</h5>
> You're supposed to always use the transactional mode, until you explicitly turn on the `AutoCommit` flag. You can explicitly issue a `BEGIN` statement (only when an actual transaction has not begun yet) but you're not allowed to call `begin_work` method (if you don't issue a `BEGIN`, it will be issued internally). You can commit or roll it back freely. Another transaction will automatically begins if you execute another statement.

After `chmod a-w DataSource/Meta.sqlite3`, this script is sufficient to demonstrate the read-only case:

```perl
#!/usr/bin/env genome-perl

use strict;
use warnings;

use above "Genome";

my $dbh = Genome::DataSource::Meta->get_default_handle();

my @answer = $dbh->selectall_arrayref('select count(*) from dd_table');
print Data::Dumper::Dumper \@answer;
```

This script succeeds on my workstation.  It fails on the `trusty` VM, but succeeds with the change in this PR.

brummett:
I've done a bit more reading... According to the DBD::SQLite Changelog, this change happened in 1.38_01, and suggests Tom's fix to get back the old behavior.  This should definitely go in UR::DataSource::Meta instead of Genome::DataSource::Meta.

The way UR works, I'm starting to think it really belongs in UR::DataSource::SQLite...

brummett:
> The way UR works, I'm starting to think it really belongs in UR::DataSource::SQLite...

Scratch that.  The MetaDB absolutely requires multiple readers, so it needs this fix.  Other SQLite DBs may or may not require multiple readers, so we can defer to the final, concrete DataSource and let them set the policy there.

I'd say it goes in UR::DataSource::Meta

nnutter:
@brummett, I concur.  I'm really having trouble understanding how the DBD::SQLite change isn't dumb.

nnutter:
If @tmooney were to turn on `UR_DBI_MONITOR_SQL` would we see the `BEGIN IMMEDIATE`?  I'd still like to understand whether DBD::SQLite is always doing that or only if AutoCommit is on or what.  I just don't understand getting a write lock unless you're going to write...

nnutter:
Or maybe `DBI_TRACE=2`.

brummett:
[Here](https://rt.cpan.org/Public/Bug/Display.html?id=56444) is the RT thread about the change.  The complaint is that COMMIT may fail if a reader that was connected before the first write is still connected at COMMIT.

I think it's the wrong change, but the original complaint was 5 years ago, and the change made 2 1/2 years ago, so it's probably too late to get it changed back.

tmooney:
I opened genome/UR#71 with the same change but to `UR::DataSource::Meta` instead.

tmooney:
The DBI trace is the same either way (`BEGIN TRAN`), but here's the excerpt from `dbdimp.c`:
```C
            sqlite_trace(sth, imp_sth, 3, "BEGIN TRAN");
            if (imp_dbh->use_immediate_transaction) {
                rc = sqlite_exec(sth, "BEGIN IMMEDIATE TRANSACTION");
            } else {
                rc = sqlite_exec(sth, "BEGIN TRANSACTION");
            }
```

nnutter:
For posterity, I was confused by,

> To avoid this, DBD::SQLite internally issues a `BEGIN IMMEDIATE` when you begin a transaction by `begin_work` or under the `AutoCommit` mode (since 1.38_01).

DBD::SQLite only does the `BEGIN IMMEDIATE` is you have disabled `AutoCommit` on the handle.  So enabling `AutoCommit` was a valid solution as @tmooney had suggested.  However, using `sqlite_use_immediate_transaction` is necessary since UR doen't know how the handle is going to be used.

PR #666, opened 2015-04-15
adding option to stop clustering in 1d

chrisamiller:


apregier:
+1

nnutter:
Test failure is probably due to high load.  Will re-run once things calm down.

acoffman:
+1 once tests pass. As a note, I think if you supply a default value, you don't need `is_optional`. If the user doesn't pass in a value, the default will be used.

chrisamiller:
Removed is_optional, as suggested.  Merging.

PR #667, opened 2015-04-15
remove 'unknown' and 'tgisan' lock scopes

nnutter:
Due to the maintenance window all snapshots prior to genome-3591 have been
archived which means it's safe to remove these lock scopes now.


acoffman:
+1

brummett:
+1

PR #670, opened 2015-04-16
Fix more typos.

tmooney:
These are other places in the system that spelled `quitting` and `available` similarly to #669.

(I'll resist going further down the rabbit-hole by cleaning up `gmt consed traces-to-consed`.)


nnutter:
:+1:

acoffman:
+1

PR #671, opened 2015-04-16
Create common base class and update docs for AnP config-adding commands

tmooney:
The new base class should help keep the docs and functionality in synch going forward!

### DIFF
--- a/lib/perl/Genome/Config/AnalysisProject/Command/AddConfigBase.pm
+++ b/lib/perl/Genome/Config/AnalysisProject/Command/AddConfigBase.pm
@@ -4,6 +4,7 @@ use strict;
 use warnings;
 
 use Genome;
+use Carp qw();
 
 class Genome::Config::AnalysisProject::Command::AddConfigBase {
     is_abstract => 1,
@@ -42,6 +43,12 @@ sub execute {
     }
 
     return scalar(@new_items);
+}
+
+sub _create_profile_items {
+    my $self = shift;
+    my $name = $self->command_name;
+    Carp::confess sprintf('Command `%s` is missing a way to create profile items.', $name);
 }
 
 sub _apply_tags {
--- a/lib/perl/Genome/Config/AnalysisProject/Command/AddConfigBase.pm
+++ b/lib/perl/Genome/Config/AnalysisProject/Command/AddConfigBase.pm
@@ -1,0 +1,59 @@
+package Genome::Config::AnalysisProject::Command::AddConfigBase;
+
+use strict;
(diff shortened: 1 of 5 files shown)


lib/perl/Genome/Config/AnalysisProject/Command/AddConfigBase.pm
```diff
@@ -0,0 +1,59 @@
...
+};
+
+sub valid_statuses {
+    return ("Pending", "Hold", "In Progress", "Template");
+}
+
+sub execute {
+    my $self = shift;
+
+    my @new_items = $self->_create_profile_items();
```
nnutter: Should we create a `_create_profile_items()` stub?

lib/perl/Genome/Config/AnalysisProject/Command/AddModel.t
```diff
@@ -104,7 +104,9 @@ EOFILE
         store_only => 1,
         reprocess_existing => 0,
     );
-    my $profile_item = $add_cmd->execute();
+    my $ok = $add_cmd->execute();
+    ok($ok, 'executed add command');
+    my $profile_item = Genome::Config::Profile::Item->get(analysis_project => $analysis_project, status => 'disabled');
```
nnutter: Should we verify `$profile_item` doesn't exist before the `$add_cmd->execute()`?

nnutter:
:+1:

acoffman:
+1

It does seem a little odd that the `is_many` input is called `tag` rather than `tags`. It makes the line 
````perl
for my $tag ($self->tag) {}
```` 
read a little oddly. It looks like the singular naming is left over from the previous versions of the commands, so perhaps its best not to break compatibility. I'm fine with it either way!

lib/perl/Genome/Config/AnalysisProject/Command/AddConfigBase.pm
tmooney: Sure! I've added one.

lib/perl/Genome/Config/AnalysisProject/Command/AddModel.t
tmooney: Sounds like a good idea--added!

tmooney:
@acoffman You're not the only one who thinks `--tag` should be `--tags`!  (See also AUE-427.) That's what motivated this PR originally :smile:

I was thinking that if consensus was to change it (which it sounds like it is) this would be followed up with a second PR that made that change in its new central location.

PR #672, opened 2015-04-17
Add gene annotations to the DoCM report and remove the trio.t test

susannasiebert:
The trio.t test has been superseded by the trio_main process test on jenkins which is a deployment blocking step now.


lib/perl/Genome/Model/Tools/Joinx/VcfMerge.pm
```diff
@@ -168,6 +173,9 @@ sub _resolve_flags {
     if ($self->merge_samples) {
         $flags .= " -s";
     }
+    if ($self->allow_same_file) {
```
acoffman: According to the [docs here](https://github.com/genome/genome/pull/672/files#diff-cf316049f5f32c0727e11cb88580cdbaR5), this feature flag requires joinx >= 1.11.

I see that immediately below this you're checking joinx version compatibility with a different flag (exact-pos). Should we be doing the same thing here?

susannasiebert: You're right. I'm checking for version compatibility in Genome/VariantReporting/Suite/Vep/Run.pm but not here. Will add a check.

acoffman: As I got further in this PR, it looks like we are checking the version compatibility of this feature in the context of the  VariantReporting framework in: [Genome/VariantReporting/Suite/Vep/Run.pm](https://github.com/genome/genome/pull/672/files#diff-6b6f65490e0c8fad06cc6756830f83d0R85) so it is being checked somewhere. It does maybe seem like knowledge of joinx feature compatibility should live in the joinx tools rather than in a VariantReporting/Vep module.

acoffman:
+1

lib/perl/Genome/Model/Tools/Joinx/VcfMerge.pm
susannasiebert: I generally agree with you that we should avoid duplicating these checks. However, adding this version check to __planned_errors__ allows the variant reporting framework to fail early at validation time before the process/workflow is even kicked off so I will keep it in the vep expert (and add it to joinx vcf-merge itself).

acoffman: Makes perfect sense! Thanks for the explanation.

susannasiebert: You're welcome. I just pushed a commit with the additional check.

lib/perl/Genome/VariantReporting/Suite/Vep/Run.pm
```diff
@@ -65,3 +70,27 @@ sub name {
...
+
+sub _get_joinx_version_error {
+    my ($self, $version, $allow_same_file) = @_;
+
+    my @errors;
+    if ($allow_same_file && (version->parse("v".$version) < version->parse("v1.11"))) {
+        push @errors, UR::Object::Tag->create(
+            type => 'error',
+            properties => ['joinx_version'],
+            desc => "Provided parameter joinx_version: $version not supporting allow-same--file option, use 1.11 and above",
```
tmooney: There's an extra `-` between `same` and `file`.  (Also, `does not support the` might sound better than `not supporting` here).

susannasiebert: I pushed a commit that fixes this error message.

tmooney:
+1 :roller_coaster:

PR #673, opened 2015-04-17
Fixation of the test error in Genome::Model::Tools::Analysis::Coverage::...

gschang:
I got a report for the test error, and I fixed it.

### DIFF
--- a/lib/perl/Genome/Model/Tools/Analysis/Coverage/MergeReadcounts.t
+++ b/lib/perl/Genome/Model/Tools/Analysis/Coverage/MergeReadcounts.t
@@ -5,7 +5,7 @@ use warnings;
 
 use above "Genome";
 use Genome::Utility::Test qw(compare_ok);
-use Test::More tests => 11;
+use Test::More tests => 12;
 
 eval
 {


susannasiebert:
Just an FYI for the future. When fixing failing tests we'd still like for the fix to go through code review (which usually means two +1's and waiting to merge for 24 hours). What will usually happen for a fix of failing tests is that someone will give you a +2 which would be the go-ahead to merge before the 24 hour waiting period is up. If you make a PR and immediately merge it you're foregoing all the perks of code review.

susannasiebert:
Also, you should always wait for the two apipe-ci tests to run on your PR and succeed before merging any PR. This will ensure that the code fix is not causing any new failures.

gschang:
I got it. Next time, I will keep in mind what I earned here.


On 4/17/2015 10:28 AM, Susanna Siebert wrote:
>
> Also, you should always wait for the two apipe-ci tests to run on your 
> PR and succeed before merging any PR. This will ensure that the code 
> fix is not causing any new failures.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/genome/genome/pull/673#issuecomment-94009106>.
>



____

PR #674, opened 2015-04-17
fix test failure on master due to missing use

nnutter:


dufeiyu:
+1

nnutter:
For posterity, the tests passed on the genome/genome#663 because `Genome::InstrumentData::AlignmentResult` depends on `Genome::Model::Build::ReferenceSequence` which was loading `Genome::Sys::LockMigrationProxy` prior to genome/genome#667.

dufeiyu:
My bad, missed this in PR#663

nnutter:
@dufeiyu, It's understandable; Perl doesn't make it easy to catch.

nnutter:
Merging early.

PR #676, opened 2015-04-20
SGMS: configure email addresses for build reports.

gatoravi:
The current *from* address evaluates to something like "apipe@ip-XX-XX" on AWS where ip-XX-XX is the `hostname`. Addresses created this way fail validation in Genome::Report::Email which uses Email::Valid.
This PR proposes to use an environment variable for specifying the email address.

Note - This PR is based on the SGMS branch on not master. This is an SGMS specific change.

### DIFF
--- a/lib/perl/Genome/Model/Build.pm
+++ b/lib/perl/Genome/Model/Build.pm
@@ -1868,8 +1868,8 @@ sub generate_send_and_save_report {
     my $email_confirmation = Genome::Report::Email->send_report(
         report => $report,
         to => $to->id,
-        from => 'apipe@'.Genome::Config::domain(),
-        replyto => 'noreply@'.Genome::Config::domain(),
+        from => $ENV{GENOME_EMAIL_NOREPLY},
+        replyto => $ENV{GENOME_EMAIL_NOREPLY},
         # maybe not the best/correct place for this information but....
         xsl_files => [ $generator->get_xsl_file_for_html ],
     );
--- a/lib/perl/Genome/Model/Build.pm
+++ b/lib/perl/Genome/Model/Build.pm
@@ -1868,7 +1868,7 @@ sub generate_send_and_save_report {
     my $email_confirmation = Genome::Report::Email->send_report(
         report => $report,
         to => $to->id,
-        from => $ENV{GENOME_EMAIL_NOREPLY},
+        from => $ENV{GENOME_EMAIL_PIPELINE},
         replyto => $ENV{GENOME_EMAIL_NOREPLY},
         # maybe not the best/correct place for this information but....
         xsl_files => [ $generator->get_xsl_file_for_html ],
--- a/lib/perl/Genome/Env/GENOME_DISK_GROUP_BENCHMARKING.pm
+++ b/lib/perl/Genome/Env/GENOME_DISK_GROUP_BENCHMARKING.pm
@@ -1,4 +1,0 @@
-package Genome::Env::GENOME_DISK_GROUP_BENCHMARKING;
-use base 'Genome::Env::Required';
-
-1;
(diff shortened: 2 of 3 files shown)


lib/perl/Genome/Model/Build.pm
```diff
@@ -1868,8 +1868,8 @@ sub generate_send_and_save_report {
     my $email_confirmation = Genome::Report::Email->send_report(
         report => $report,
         to => $to->id,
-        from => 'apipe@'.Genome::Config::domain(),
-        replyto => 'noreply@'.Genome::Config::domain(),
+        from => $ENV{GENOME_EMAIL_NOREPLY},
```
acoffman: Should this be `GENOME_EMAIL_FROM` rather than `NOREPLY`?

gatoravi: `GENOME_EMAIL_FROM` does sound more correct. That would have to be a new environment variable that we would have to define. I'm also not sure what value to set that variable to. `apipe@genome.wustl.edu` doesn't seem appropriate for this and hence I decided to stick with the `GENOME_EMAIL_NOREPLY`

tmooney: Would the existing `GENOME_EMAIL_PIPELINE` work here?

gatoravi: yep, that would work well here. I've updated the 'from' to this environment variable.

acoffman:
+1

nnutter:
I disabled the safety check in `genome-env` to find out why this PR is failing.  The error when loading `Genome` is:

```
Environment variable GENOME_DISK_GROUP_BENCHMARKING must be configured to use Genome at /gscuser/nnutter/genome/master/lib/perl/Genome/Env/Required.pm line 14.
```

Which looks to be related to these commits:

```
(genome-env) ~/genome/master $ git log -SGENOME_DISK_GROUP_BENCHMARKING
commit 46c276757c49c70e06c09370b80d3b6e00f3203b
Author: Obi <ogriffit@GGMS>
Date:   Sat Mar 1 23:50:57 2014 -0600

    add new environment variables GENOME_DISK_GROUP_BENCHMARKING GENOME_DISK_GROUP_READS

commit 96784ef225a12a9788bedccca44fec52ff347b06
Author: Obi <ogriffit@GGMS>
Date:   Sat Mar 1 23:37:10 2014 -0600

    update disk group list to include reads and benchmarking
```

If they really are required then this PR will have to be tested manually.

nnutter:
I think this change should be on `master` as well.

nnutter:
:+1:

gatoravi:
Yes, 'GENOME_DISK_GROUP_BENCHMARKING' is SGMS-specific and we do use it, it is not used/doesn't exist on the TGI side and thats why we see this error. The only way to test any changes to the SGMS branch with current github would be to add these variables to master/TGI?

I think if I create a new PR based off of master with the same change we won't see the same test-failure.

nnutter:
What are those environment variables used for?

```
$ git checkout origin/pull/676/merge
...
HEAD is now at 3003de3... Merge 6dd40d2801c82a4445f36f99e1e8407fcad1d084 into afd25ebe1a154021d6893a90d2cf68db1c7d407b

$ git grep GENOME_DISK_GROUP_BENCHMARKING
lib/perl/Genome/Env/GENOME_DISK_GROUP_BENCHMARKING.pm:package Genome::Env::GENOME_DISK_GROUP_BENCHMARKING;

$ git grep GENOME_DISK_GROUP_READS
lib/perl/Genome/Env/GENOME_DISK_GROUP_READS.pm:package Genome::Env::GENOME_DISK_GROUP_READS;
```

gatoravi:
Ok, I was wrong. I got confused with the environment variable we use for specifying low resource usage.
These look like environment variables that were present/used in 'master' at some point and were thus added to the SGMS. It might be that they have now been removed from 'master' and not removed yet from the SGMS.
The right thing to do would be then to delete these two environment variable modules and remove them from genome.conf(SGMS configuration file) as well - https://github.com/genome/gms/blob/c055959f19ca48b9f43a3fdd423b38cb6f8a6d0c/setup/etc/genome.conf#L50-51

gatoravi:
ok, merging this to test on the SGMS. I'll open a separate PR for master.

PR #677, opened 2015-04-20
AnP Config Add: Pluralize tag-adding option.

tmooney:
It has been requested that this be `--tags` instead of `--tag`.


acoffman:
Looks good overall - I may add a CHANGELOG entry since we're breaking backwards compatibility, but I suspect in reality that there are only a couple people using this command and we can notify them manually.

+1 either way!

tmooney:
@acoffman I will add a `CHANGELOG` entry to the merge commit when the time comes.

nnutter:
:+1:

PR #678, opened 2015-04-21
Don't fail optional expert parameters as required during validation

susannasiebert:
This fixes a bug where any optional expert parameter would fail validation if it wasn't in the plan file.

See https://rt.gsc.wustl.edu/Ticket/Display.html?id=105348

### DIFF
--- a/lib/perl/Genome/VariantReporting/Command/ShowDocs.t.expected-output/experts/__test__
+++ b/lib/perl/Genome/VariantReporting/Command/ShowDocs.t.expected-output/experts/__test__
@@ -6,4 +6,6 @@
     (undocumented)
   [1m__planned__[0m
     (undocumented)
+  [1m__planned_optional__[0m
+    (undocumented)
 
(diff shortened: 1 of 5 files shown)


nnutter:
Would it be hard to add a regression test for this?

susannasiebert:
Hm, not sure. There is actual a TODO about the lack of test coverage for this part of the code base. I will look into it.

acoffman:
If, in both places that `planned_required_names` is called, we immediately convert the list into a `Set::Scalar`, does it make sense to just have the method return a set instead?

+1 either way though!

susannasiebert:
`planned_names` and `planned_required_names` are each just called once in this module. I'll have to check if these subroutines are used anywhere else but if not I don't see a reason why we can't convert these two subroutines as you suggested.

susannasiebert:
@nnutter, I added a test for the `__planned_errors__` subroutine.

@acoffman, I made the change you suggested.

acoffman:
+1 :ship:

nnutter:
:+1:

PR #679, opened 2015-04-21
GMT Integration of VCF Evaluation (part 2)

indraniel:
This is part 2 of a roll out to integrate the [VCF Evaluation scripts][1] within the genome codebase. The aim of this pull request is to get [evaluate_vcfs.pl][2] setup as a GMT. 

[1]: https://github.com/genome/vcf-evaluation
[2]: https://github.com/genome/vcf-evaluation/blob/master/evaluate_vcfs.pl

### DIFF
--- a/lib/perl/Genome/Model/Tools/Vcf/EvaluateVcf.pm
+++ b/lib/perl/Genome/Model/Tools/Vcf/EvaluateVcf.pm
@@ -165,7 +165,7 @@ sub execute {
         );
 
     my $tn_bed_size = $self->true_negative_size
-      || $self->bed_size(${tn_bed});
+      || $self->bed_size($final_tn_file);
 
     my $false_positives_in_roi = $self->number_within_roi(
         $final_input_file,
--- a/lib/perl/Genome/Model/Tools/Vcf/EvaluateVcf.pm
+++ b/lib/perl/Genome/Model/Tools/Vcf/EvaluateVcf.pm
@@ -185,7 +185,7 @@ sub execute {
 
     $self->rawstats(\%results);
 
-    $self->display_all_stats();
+#    $self->display_all_stats();
 
     return 1;
 }
--- a/lib/perl/Genome/Model/Tools/Vcf/EvaluateVcf.pm
+++ b/lib/perl/Genome/Model/Tools/Vcf/EvaluateVcf.pm
@@ -422,7 +422,7 @@ sub bed_size {
     my $self = shift;
     my $bed = shift;
     my $count = 0;
-    my $fh = IO::File->new("zcat $bed |") or die "Unable to open $bed to calculate size\n";
+    my $fh = IO::File->new($bed) or die "Unable to open $bed to calculate size\n";
     while(my $line  = $fh->getline) {
         chomp $line;
         my ($chr, $start, $stop) = split "\t", $line;
--- a/lib/perl/Genome/Model/Tools/Vcf/EvaluateVcf.pm
+++ b/lib/perl/Genome/Model/Tools/Vcf/EvaluateVcf.pm
@@ -165,7 +165,7 @@ sub execute {
         );
 
     my $tn_bed_size = $self->true_negative_size
-      || $self->bed_size("${tn_bed}.roi.bed.gz");
+      || $self->bed_size(${tn_bed});
 
     my $false_positives_in_roi = $self->number_within_roi(
         $final_input_file,
--- a/lib/perl/Genome/Model/Tools/Vcf/EvaluateVcf.pm
+++ b/lib/perl/Genome/Model/Tools/Vcf/EvaluateVcf.pm
@@ -185,8 +185,6 @@ sub execute {
 
     $self->rawstats(\%results);
 
-#    $self->display_all_stats();
-
     return 1;
 }
 
--- a/lib/perl/Genome/Model/Tools/Vcf/EvaluateVcfs.pm
+++ b/lib/perl/Genome/Model/Tools/Vcf/EvaluateVcfs.pm
@@ -102,6 +102,12 @@ sub execute {
 sub display_summary_stats {
     my ($self, $configs) = @_;
 
+    print "\n";
+    print "==================", "\n";
+    print "Valdiation Summary", "\n";
+    print "==================", "\n";
+    print "\n";
+
     my @stat_names = $self->stat_types;
     my @headers = ('Name', 'ID', 'VarType', @stat_names);
     print join("\t", @headers), "\n";
--- a/lib/perl/Genome/Model/Tools/Vcf/EvaluateVcfs.pm
+++ b/lib/perl/Genome/Model/Tools/Vcf/EvaluateVcfs.pm
@@ -12,19 +12,19 @@ class Genome::Model::Tools::Vcf::EvaluateVcfs {
         bedtools_version => {
             is => "Text",
             doc => "Bedtools version to use",
-            default_value => "2.17.0",
+            default_value => Genome::Model::Tools::Vcf::EvaluateVcf->__meta__->property(property_name => 'bedtools_version')->default_value,
         },
 
         joinx_version => {
             is => "Text",
             doc => "Joinx version to use",
-            default_value => "1.11",
+            default_value => Genome::Model::Tools::Vcf::EvaluateVcf->__meta__->property(property_name => 'joinx_version')->default_value,
         },
 
         vcflib_version => {
             is => "Text",
             doc => "vcflib toolset version to use",
-            default_value => "1.0",
+            default_value => Genome::Model::Tools::Vcf::EvaluateVcf->__meta__->property(property_name => 'vcflib_version')->default_value,
         },
 
         config_file => {
--- a/lib/perl/Genome/Model/Tools/Vcf/EvaluateVcfs.pm
+++ b/lib/perl/Genome/Model/Tools/Vcf/EvaluateVcfs.pm
@@ -239,34 +239,10 @@ sub _collect_evaluation_stats {
 
     my %stats;
 
-    $stats{'true_positive_found_exact'} =
-      $cmd->stat_true_positive_found_exact();
-
-    $stats{'total_true_positive_exact'} =
-      $cmd->stat_total_true_positive_exact();
-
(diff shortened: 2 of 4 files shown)


lib/perl/Genome/Model/Tools/Vcf/EvaluateVcf.pm
```diff
@@ -185,7 +185,7 @@ sub execute {
 
     $self->rawstats(\%results);
 
-    $self->display_all_stats();
+#    $self->display_all_stats();
```
tmooney: This should probably be either uncommented or removed entirely.

lib/perl/Genome/Model/Tools/Vcf/EvaluateVcfs.pm
```diff
@@ -0,0 +1,369 @@
...
+    my ($self, $inputs) = @_;
+
+    my ($i, $total) = (1, scalar @{$inputs});
+    $self->status_message(
+        "Performing Evaluations & Collecting Statistics (%d vcfs)",
+        $total
+    );
+
+    for my $set (@{$inputs}) {
+        $self->status_message(
```
tmooney: Depending how noisy this will be, some of these might make more sense as `debug_message`s.

lib/perl/Genome/Model/Tools/Vcf/EvaluateVcfs.pm
```diff
@@ -0,0 +1,369 @@
...
+        chomp $line;
+        my ($name, $id, $path, $variant_type, $sample) = split("\t", $line);
+        $self->validate_type($variant_type);
+        my $gold_vcf =
+          $variant_type eq "snvs" ? $self->gold_snv_vcf : $self->gold_indel_vcf;
+
+        my $clean_indels = $variant_type eq 'indels' ? 1 : 0;
+
+        my $model = Genome::Model->get($id);
+        my $build = $model->last_succeeded_build;
```
tmooney: Would it make sense for this tool to support being passed builds instead of models?  (It could try both by doing something like:
```perl
my $build = Genome::Model::Build->get($id) || Genome::Model->get($id)->last_succeeded_build;
```

Regardless it might be good to make sure we got a `$model` or a `$build` rather than dying with a `Can't call method ... on an undefined value` error.

lib/perl/Genome/Model/Tools/Vcf/EvaluateVcfs.pm
indraniel: Those shouldn't be too noisy.  I placed them there to help gauge the progress of the tool while its running.

lib/perl/Genome/Model/Tools/Vcf/EvaluateVcfs.pm
indraniel: It might make more sense being passed builds, but right now the overall work flow is pretty centered around this configuration file which uses the model id. For consistencies sake, I'll leave it as is for now.

I'll add the check if we didn't a build though.

lib/perl/Genome/Model/Tools/Vcf/EvaluateVcfs.pm
```diff
@@ -0,0 +1,375 @@
...
+    my $doc = q{
+      This tool is a wrapper process to Genome::Model::Tools::Vcf::EvaluateVcf.
+      It allows one to compare multiple SNV and Indel VCFs against a Gold Standard
+      set.
+
+      A configuration file is used to specify the list of VCFs to compare.  The
+      configuration file is a TSV file with each row containing the following ordered
+      columns: name, model id, vcf file system path, variant type, and sample name.
+
+      This tools parses each line from the configruation file, appropriately
```
tmooney: The `r` and `u` in `configuration` are transposed.

lib/perl/Genome/Model/Tools/Vcf/EvaluateVcfs.pm
```diff
@@ -0,0 +1,375 @@
...
+    return $dir;
+}
+
+sub help_brief {
+    return "Validate multiple VCF files to a specified Gold Standard Set";
+}
+
+sub help_detail {
+    my $doc = q{
+      This tool is a wrapper process to Genome::Model::Tools::Vcf::EvaluateVcf.
```
tmooney: It might make more sense for the documentation to refer to `gmt vcf evaluate-vcf` than its Perl class.

lib/perl/Genome/Model/Tools/Vcf/EvaluateVcfs.t
```diff
@@ -0,0 +1,126 @@
...
+
+# M A I N #####################################################################
+
+#This is a bare minimum test that just compiles Perl and the UR class.
+use_ok('Genome::Model::Tools::Vcf::EvaluateVcfs');
+
+# This directory contains the test data set (based from BIO-1176)
+my $basedir = Path::Class::Dir->new(
+    '/gscmnt/gc2801/analytics/idas',
+    'jira/BIO-1387/cle-data'
```
tmooney: Usually test data that can't be in the repository goes under the location pointed to by `$ENV{GENOME_TEST_INPUTS}`.  `Genome::Utility::Test` provides a `data_dir_ok` test that generates a path based on the module under test and checks that it exists.

Theoretically someone could then run this test externally if they were willing to supply their own dataset.

lib/perl/Genome/Model/Tools/Vcf/EvaluateVcfs.t
```diff
@@ -0,0 +1,126 @@
...
+);
+
+SKIP: {
+    skip "this is a long-running test & must be run within TGI", 75;
```
tmooney: Would it be possible to test some parts of this command, even if we have to skip running the `EvaluateVcf` command itself?  `long-running test` is another way of saying "test that will never be run"!

lib/perl/Genome/Model/Tools/Vcf/EvaluateVcfs.pm
```diff
@@ -0,0 +1,374 @@
...
+sub parse_config_file {
+    my $self = shift;
+    my $config = $self->fetch_config_file;
+
+    $self->status_message("Parsing configuration file : $config");
+
+    my $fh = $config->openr;
+
+    my @inputs;
+    while (my $line = <$fh>) {
```
acoffman: If you wanted, you could use TSV reader that's already in `Genome::Utility`

Something along the lines of

```perl
my $reader = Genome::Utility::IO::SeparatedValueReader->create(
  headers => [qw(name id path variant_type sample)],
  input => $self->config_file(),
  separator => "\t"
);

while (my $line_data = $reader->next()) {
   $self->validate_type($line_data->{variant_type});
   ....
}
````
Would eliminate the need for the `fetch_config_file` method as well as the manual `chomp`, `split`, and file handle handling. It would also validate that the file has the correct number of columns for you.

File: lib/perl/Genome/Model/Tools/Vcf/EvaluateVcfs.pm [hunk 9]
```diff
@@ -0,0 +1,374 @@
...
+        },
+
+        vcflib_version => {
+            is => "Text",
+            doc => "vcflib toolset version to use",
+            default_value => "1.0",
+        },
+
+        config_file => {
+            is => 'Text',
```
acoffman: Should this be a `Path` rather than `Text`?

lib/perl/Genome/Model/Tools/Vcf/EvaluateVcfs.pm
```diff
@@ -0,0 +1,374 @@
...
+use strict;
+use warnings;
+
+use Cwd;
+use Path::Class;
+
+class Genome::Model::Tools::Vcf::EvaluateVcfs {
+    is => "Command::V2",
+    has_input => [
+        bedtools_version => {
```
acoffman: Unless I'm missing something (very possible!), neither this nor `joinx_version` nor `vcflib_version` are used in this command.

indraniel: Good call!  Lemme add those in :-)

lib/perl/Genome/Model/Tools/Vcf/EvaluateVcfs.pm
indraniel: Thanks!  I didn't know about this Genome feature and it's good to know about.  However, [the initial code that's this is based off][1] of didn't use that feature.  For consistencies sake, I'll keep this section as is.  Perhaps after some more vetting of the logic I'll convert it over.

[1]: https://github.com/genome/vcf-evaluation/blob/master/evaluate_vcfs.pl

lib/perl/Genome/Model/Tools/Vcf/EvaluateVcfs.pm
```diff
@@ -0,0 +1,377 @@
...
+
+    $cmd->execute or die $self->error_message(
+        "Trouble running Genome::Model::Tools::Vcf::EvaluateVcf!"
+    );
+
+    my $stats = $self->_collect_evaluation_stats($cmd);
+    return $stats;
+}
+
+sub _collect_evaluation_stats {
```
tmooney: It seems like this subroutine could be distilled down to:

```perl
sub _collect_evaluation_stats {
  my %stats;
  for my $stat ($self->stat_types) {
    my $stat_accessor = join('_', 'stat', $stat);
    $stats{$stat} = $cmd->$stat_accessor;
  }

  return \%stats;
}
```

lib/perl/Genome/Model/Tools/Vcf/EvaluateVcfs.pm
```diff
@@ -0,0 +1,377 @@
...
+    return $config;
+}
+
+sub fetch_output_directory {
+    my $self = shift;
+    my $dir = Path::Class::Dir->new($self->output_directory());
+    unless (-d $dir) {
+        $self->status_message(
+            "'$dir' doesn't exist on file system."
+            . "Making path"
```
tmooney: There's no space between these two sentences.

acoffman:
I like both of @tmooney's new suggestions.

+1 with both of them in place :shaved_ice:

acoffman:
CI, test this.

PR #680, opened 2015-04-21
use ENV variables to configure build email

gatoravi:
This is identical to #676 but this PR is based off of master.
Note - this will change the reply-to to `donotreply@genome.wustl.edu` instead of `noreply@genome.wustl.edu`


nnutter:
:+1:

tmooney:
+1 :scroll:

PR #682, opened 2015-04-22
Support Converge Operations when using Ptero

davidlmorton:
PTero itself now has support for both Block and Converge operations, but I've just implemented support for Converge operations here in Genome.  WorkflowBuilder has no concept of Block operations, and maybe never will, so I didn't try to fix that now.

### DIFF
--- a/lib/perl/Genome/WorkflowBuilder/Converge.pm
+++ b/lib/perl/Genome/WorkflowBuilder/Converge.pm
@@ -97,12 +97,12 @@ sub _get_ptero_converge_method {
     require Ptero::Builder::Detail::Workflow::Converge;
 
     my $self = shift;
-    my @output_names = $self->output_properties;
+    my ($output_name) = $self->output_properties;
     return Ptero::Builder::Detail::Workflow::Converge->new(
         name => 'converge',
         parameters => {
             input_names => [$self->input_properties],
-            output_name => shift @output_names,
+            output_name => $output_name,
         },
     );
 }
--- a/lib/perl/Genome/WorkflowBuilder/Converge.pm
+++ b/lib/perl/Genome/WorkflowBuilder/Converge.pm
@@ -32,6 +32,7 @@ sub from_xml_element {
         name => $element->getAttribute('name'),
         input_properties => \@input_properties,
         output_properties => \@output_properties,
+        parallel_by => $element->getAttribute('parallelBy'),
     );
 }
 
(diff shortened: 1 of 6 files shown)


lib/perl/Genome/WorkflowBuilder/Converge.pm
```diff
@@ -73,4 +74,37 @@ sub notify_output_link {
     return;
 }
 
+sub get_ptero_builder_task {
+    require Ptero::Builder::Detail::Workflow::Task;
```
acoffman: It seems odd that you'd have to read through every subroutine implementation in a module to discover its potential dependencies. Maybe all the `requires` should be at the top of the module? Or, if its fine to read through every implementation to find dependencies, why not lean on the autoloader and omit the `require` entirely? Are we trying to avoid a compile time dependency on `Ptero::Builder::Detail::Workflow::Task`?

lib/perl/Genome/WorkflowBuilder/Converge.pm
```diff
@@ -73,4 +74,37 @@ sub notify_output_link {
...
+sub _get_ptero_converge_method {
+    require Ptero::Builder::Detail::Workflow::Converge;
+
+    my $self = shift;
+    my @output_names = $self->output_properties;
+    return Ptero::Builder::Detail::Workflow::Converge->new(
+        name => 'converge',
+        parameters => {
+            input_names => [$self->input_properties],
+            output_name => shift @output_names,
```
acoffman: Will there only ever be one of these? If that's the case, does it make more sense to do something like `my ($output_name) = $self->output_properties()` instead of putting it in an array and immediately shifting it off?

lib/perl/Genome/WorkflowBuilder/Converge.pm
davidlmorton: We were trying to play nice with loading unneeded libraries.  Since you have to set an environment variable to use PTero currently, the Ptero::* modules don't need to be loaded until then.  I'm not sure the tiny savings is worth it though, what do you think?  It was always intended that these imports would be moved to to top of the module once PTero became the default.

lib/perl/Genome/WorkflowBuilder/Converge.pm
davidlmorton: This is for compatibility with the legacy Converge weirdness, even if multiple outputs are specified for a Converge step, only the first one is set.  I will update to use the syntax you suggested though.

lib/perl/Genome/WorkflowBuilder/Converge.pm
acoffman: Cool, that makes sense to me - I'm fine leaving it as is for now, and moving it to the top when we migrate fully!

acoffman:
+1 One step closer to PTero! :dragon: (They didn't have a dinosaur emoji, so we'll have to settle for a dragon.)

PR #683, opened 2015-04-22
Delete the output files, not whatever is in the current directory.

tmooney:
This implements my last comment in #628.

The test data is updated to expect the files to be removed as well :smile:

### DIFF
--- a/lib/perl/Genome/Model/ClinSeq/Command/IdentifyLoh.pm
+++ b/lib/perl/Genome/Model/ClinSeq/Command/IdentifyLoh.pm
@@ -255,12 +255,14 @@ sub filter_loh {
 #Get rid of intermediate files
 sub cleanup {
     my $self = shift;
-    my $unwanted = "*.Somatic* *.readcounts *.hc *.LOH" .
-        " *.lc *.Germline *.removed *.formatted" .
-        " *.hc.err *.other";
-    for my $file (glob $unwanted) {
-        $self->status_message("unlinking $file");
-        unlink $file;
+    my @patterns = qw(*.Somatic* *.readcounts *.hc *.LOH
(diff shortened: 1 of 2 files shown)


lib/perl/Genome/Model/ClinSeq/Command/IdentifyLoh.pm
```diff
@@ -255,12 +255,14 @@ sub filter_loh {
...
-        " *.lc *.Germline *.removed *.formatted" .
-        " *.hc.err *.other";
-    for my $file (glob $unwanted) {
-        $self->status_message("unlinking $file");
-        unlink $file;
+    my @patterns = qw(*.Somatic* *.readcounts *.hc *.LOH
+        *.lc *.Germline *.removed *.formatted
+        *.hc.err *.other);
+    for my $pattern (@patterns) {
+        for my $file (glob File::Spec->join($self->outdir, $pattern)) {
```
acoffman: Super nitpicky and maybe not worth implementing, but a `map` would eliminate the nested loops here.

````perl
my @files = map { glob File::Spec->join($self->outdir, $_) } qw(*.Somatic* *.readcounts *.hc *.LOH *.lc *.Germline *.removed *.formatted *.hc.err *.other);
$self->status_message("unlinking %s", join(',' @files));
unlink @files
````
You'd lose the ability to tell which ones failed or succeeded, but we aren't currently checking that anyways.

acoffman:
+1 with or without my loop flattening suggestion

nnutter:
:+1:

acoffman:
I'm going to merge this in @tmooney's absence since he won't be back until Monday and this fixes a potentailly destructive bug.

PR #684, opened 2015-04-23
some config changes that can be isolated from Genome::Config changes

nnutter:


nnutter:
Forgot about variable name validation on `master`.

PR #687, opened 2015-04-23
Use Scope::Guard to ensure unlocking

dufeiyu:
According to Nathan's suggestion, Scope::Guard is used to make sure unlocking when it goes out of scope upon return or exception.


nnutter:
:+1:

brummett:
+1

PR #688, opened 2015-04-23
SpeedSeq Alignment Tools

jasonwalker80:
This is an initial PR for the incorporation of the SpeedSeq alignment commands as GMTs.  The version of SpeedSeq is not an official release.  The PR is to get feedback about the test, commands and gain some familiarity with the SpeedSeq tool suite.  This PR will not be merged as-is, but feedback and code review would be greatly appreciated.


lib/perl/Genome/Model/Tools/SpeedSeq/Align.pm
```diff
@@ -0,0 +1,78 @@
...
+use warnings;
+
+use Genome;
+
+class Genome::Model::Tools::SpeedSeq::Align {
+    is => 'Genome::Model::Tools::SpeedSeq::AlignBase',
+    has_input => [
+        fastq => {
+            is => 'Text',
+            doc => 'paired-end fastq file. if -p flag is used then expected to be an interleaved paired-end fastq file, and fastq2 may be omitted. (can be gzipped)',
```
tmooney: It might be better if the doc refers to the `--paired` option of this tool instead of `-p`.

lib/perl/Genome/Model/Tools/SpeedSeq/Align.pm
```diff
@@ -0,0 +1,78 @@
...
+    push @input_files, $self->reference_fasta;
+    push @input_files, $self->fastq;
+    if ($self->config_file) {
+        push @input_files, $self->config_file;
+    }
+    if ($self->paired) {
+        if ($self->fastq2) {
+            die('Expected single interleaved paired-end FASTQ!');
+        }
+        $options .= ' -p';
```
tmooney: It's a bit strange that one option is passed into `_resolve_options_string`, most are resolved inside there, and this one is tacked on here.

One possibility is that this command could override `_resolve_options_string` to add its own options.  Another might be a system like `GMT::Picard` uses to derive them.

lib/perl/Genome/Model/Tools/SpeedSeq/Align.pm
```diff
@@ -0,0 +1,78 @@
...
+    } else {
+        push @input_files, $self->fastq2;
+        $inputs_string .= ' '. $self->fastq2;
+    }
+
+    # COMMAND
+    my $cmd = $self->speedseq_path .' align '. $options .' '. $inputs_string;
```
apregier: Do we have a use case for align yet, or only realign?  I guess I would favor only wrapping tools we have a use case or request for.

lib/perl/Genome/Model/Tools/SpeedSeq/Align.pm
apregier: I like the idea of using the same pattern as GMT::Picard (add an attributes_have to specify that it is a parameter that gets passed through).  Overall, I think GMT::Picard is a good model for GMTs with multiple tools in the same "package"

lib/perl/Genome/Model/Tools/SpeedSeq/AlignBase.pm
```diff
@@ -0,0 +1,141 @@
...
+        },
+    ],
+    has_param => [
+        output_prefix => {
+            is => 'Text',
+            doc => 'output prefix',
+            is_optional => 1,
+            example_values => ['in1.fq'],
+        },
+        paired => {
```
apregier: This param is only relevant to Align, not Realign, right?

lib/perl/Genome/Model/Tools/SpeedSeq/Realign.pm
```diff
@@ -0,0 +1,67 @@
...
+    }
+    if ($self->config_file) {
+        push @input_files, $self->config_file;
+    }
+
+    # COMMAND
+    my $cmd = $self->speedseq_path .' realign '. $options .' '. $inputs_string;
+
+    # OUTPUTS
+    my $basename = 'in.realign';
```
apregier: Maybe make this a default value for output_prefix, then you don't have to do the conditional?

lib/perl/Genome/Model/Tools/SpeedSeq/Align.t
```diff
@@ -0,0 +1,46 @@
...
+
+isa_ok($align_cmd,$pkg);
+ok($align_cmd->execute,'execute command '. $pkg);
+
+# BAM diff is failing
+
+#for my $output_file ($align_cmd->output_files) {
+#    my ($basename,$dirname) = File::Basename::fileparse($output_file);
+#    my $expected_output_file = $expected_output_dir .'/'. $basename;
+#    compare_ok($output_file,$expected_output_file);
```
apregier: definitely use diag => 0 option to compare_ok when comparing binary files (like bam)

lib/perl/Genome/Model/Tools/SpeedSeq/AlignBase.pm
```diff
@@ -0,0 +1,141 @@
...
+                 .bam
+                 .bam.bai
+                 .discordants.bam
+                 .discordants.bam.bai
+                 .splitters.bam
+                 .splitters.bam.bai
+             /;
+}
+
+sub _resolve_options_string {
```
apregier: If you use the GMT::Picard pattern of marking each param with its speedseq_param_name, then this method could be made very generic and moved up to the SpeedSeq::Base class

lib/perl/Genome/Model/Tools/SpeedSeq/AlignBase.pm
```diff
@@ -0,0 +1,141 @@
...
+    if ($self->config_file) {
+        $params .= ' -K '. $self->config_file;
+    }
+    if ($self->verbose) {
+        $params .= ' -v ';
+    }
+    return $params;
+}
+
+sub _resolve_output_file_paths {
```
apregier: This could probably also be moved to SpeedSeq base, although this refactoring isn't critical until we add more of the speedseq tool wrappers.

lib/perl/Genome/Model/Tools/SpeedSeq/Realign.t
```diff
@@ -0,0 +1,44 @@
...
+   bams => $bam,
+   sort_memory => 3,
+);
+
+isa_ok($realign_cmd,$pkg);
+ok($realign_cmd->execute,'execute command '. $pkg);
+
+# The BAM headers differ since realign uses a temp pipe as the input to the BWAMEM command
+# Figure out a way to diff a BAM excluding the header, it must exist somewhere
+
```
apregier: I am seeing GMT::Sam::Compare, which seems to ignore the headers.  Not sure if that will work?

apregier:
I think we should change the file/package names to Speedseq so that when it is made into command line it looks like "speedseq" instead of "speed-seq"

lib/perl/Genome/Model/Tools/SpeedSeq/Align.pm
jasonwalker80: I took a crack at this and it quickly expanded into the output and input files as well to resolve the speedseq command.  Let me know what you think.

lib/perl/Genome/Model/Tools/SpeedSeq/Align.pm
jasonwalker80: I started with align before I realized I needed to implement realign.  I don't think it hurts to have all the speedseq subcommands wrapped eventually.  I don't think we need to have an align alignment result.  I agree the pipeline steps should be kept as simple as possible.

lib/perl/Genome/Model/Tools/SpeedSeq/AlignBase.pm
jasonwalker80: Fixed.  Thanks for noticing that.

lib/perl/Genome/Model/Tools/SpeedSeq/Realign.pm
jasonwalker80: It's a bit strange how the align and realign commands resolve the output_prefix.  'align' will use the first FASTQ input as the prefix by default.  The 'realign' command will use the string 'in.realign'.  I'm not sure if the GMT version should have any default value.  Can I override the default_value in each subcommand?  Feedback is much appreciated.

lib/perl/Genome/Model/Tools/SpeedSeq/AlignBase.pm
jasonwalker80: Done. (and more)

lib/perl/Genome/Model/Tools/SpeedSeq/AlignBase.pm
jasonwalker80: Most of what I think can move to Base has moved.  Let me know if you disagree.

jasonwalker80:
Renamed classes to Speedseq to match python command usage.

lib/perl/Genome/Model/Tools/SpeedSeq/Align.pm
jasonwalker80: Fixed.

lib/perl/Genome/Model/Tools/SpeedSeq/Align.t
jasonwalker80: Added diag=>0 for binary files (except BAM, see lower comment).

lib/perl/Genome/Model/Tools/SpeedSeq/Realign.t
jasonwalker80: I'm not sure why GMT::Picard::CompareSams doesn't work.  It reports missing and unaligned when the BAM files should be the same.  

GMT::Sam::Compare seems like a workaround by only evaluating the 'Differ' count, but it's the best I can find.  This fails when the output BAM is a truncation of the input or if the BAMs have entirely different read names.

jasonwalker80:
I believe all initial comments have been addressed.  All of my changes have been pushed.  Feel free to give this another once over.  The refactoring of the tool parameters will need additional work.  There is a lot of overlap with Picard and I think that deserves more attention in future PRs.  The redundancy is annoying and there has to be a better way to make that stuff general for all tool wrappers.

lib/perl/Genome/Model/Tools/SpeedSeq/Realign.pm
apregier: Yes, you can override the property in each subcommand with different default_values

lib/perl/Genome/Model/Tools/Speedseq/Base.pm
```diff
@@ -0,0 +1,290 @@
...
+        },
+    ],
+};
+
+sub help_detail {
+    "A flexible framework for rapid genome analysis and interpretation"
+}
+
+my %TOOL_VERSIONS = (
+    'test'     => '/gscmnt/gc2719/halllab/bin/speedseq',
```
susannasiebert: This version fails on very small input files (e.g. for testing). I have worked with Colby on getting it fixed and he has an updated version at ` /gscmnt/gc2719/halllab/users/cchiang/bin/speedseq`. He will also update  the official version at the above path in a week or so once their current alignments have finished. We should make sure that we package a speedseq version that includes commit `6645b100a4d6dc329409d9fe74c41e0269443a4f`.

jasonwalker80: The cchiang version is now used.

lib/perl/Genome/Model/Tools/Speedseq/Align.pm
```diff
@@ -0,0 +1,67 @@
...
+        },
+    ],
+};
+
+sub _tool_subcommand_name {
+    return 'align';
+}
+
+
+#sub execute {
```
apregier: You can go ahead and delete these commented-out methods (in Realign.pm too)

ur
```diff
@@ -1 +1 @@
-Subproject commit 4fffe07b27706f5c6c0bfb007b42a27687566380
+Subproject commit ee4eea5613fa097693b6474d1c427770a4a8fd6d
```
apregier: I don't think you want to actually have this as part of your commit

lib/perl/Genome/Model/Tools/Speedseq/Realign.t
```diff
@@ -0,0 +1,51 @@
...
+    my $expected_output_file = $expected_output_dir .'/'. $basename . $suffix;
+    if ($suffix eq '.bam') {
+          my $cmp = Genome::Model::Tools::Sam::Compare->execute(
+          file1 => $output_file,
+          file2 => $expected_output_file,
+          );
+          ok ($cmp->result, 'compare BAMs'); 
+    } else {
+      #TODO: I'm not sure this is working as intended...
+      #compare_ok($output_file,$expected_output_file,diag=>0);
```
apregier: It might be ok just to compare the bam files and simply test for existence (maybe size?) of the index files.  Those are the only two types of files, right?

lib/perl/Genome/Model/Tools/Speedseq/Base.pm
```diff
@@ -0,0 +1,290 @@
...
+    return grep {$_->can("tool_input_file") && $_->tool_input_file} @property_metas;
+}
+
+# Handle conversion from boolean => 'true'/'false' and general tool arg
+# formatting (argname=argvalue).
+sub _format_tool_arg {
+    my ($type, $name, $value) = @_;
+    return '-'. $name if ($type eq 'Boolean');
+    $value = qq{"$value"} if $type eq 'Text';
+    return sprintf '-%s %s', $name, $value;
```
apregier: it might be good to either make the 'Text' case more general (everything else) or add a die at the end so, for example, that if someone adds an 'Integer' type argument it doesn't just get silently swallowed

lib/perl/Genome/Model/Tools/Speedseq/Base.pm
```diff
@@ -0,0 +1,290 @@
...
+# tool_param_name), return the list of cmdline args that should be passed
+# to the tool based on the values set on the current object.
+#
+# this is where is_many properties get handled.
+sub _tool_param_args_from_meta {
+    my ($self, $meta) = @_;
+
+    my $ur_name = $meta->property_name;
+    # using // instead of ? : below made a lims perl test fail
+    my $type = defined $meta->data_type ? $meta->data_type : 'String';
```
apregier: In _format_tool_arg it is checking for 'Text' type, not 'String'

lib/perl/Genome/Model/Tools/Speedseq/Base.pm
```diff
@@ -0,0 +1,290 @@
...
+# string to be appended after the full command line (e.g., '> foo.out')
+#
+# there was a bug that led to some commands being built like:
+#
+#   java ... foo.jar class ARGS > output_file MORE_ARGS
+#
+# that's bad. this sort of conflicts with the idea of log_file, but commands
+# like "BamIndexStats" that have no native OUTPUT option (i.e., can only spew
+# to stdout) support the existence of this function.
+sub _redirects {
```
apregier: I don't think this applies to speedseq - no need to keep everything from Picard that we don't need.

lib/perl/Genome/Model/Tools/Speedseq/Base.t
```diff
@@ -0,0 +1,24 @@
...
+use warnings;
+
+use above 'Genome';
+use Test::More tests => 3;
+
+my $pkg = 'Genome::Model::Tools::Speedseq::Base';
+use_ok ($pkg);
+
+my $version = 'test';
+my $expected_path = '/gscmnt/gc2719/halllab/users/cchiang/bin/speedseq';
```
apregier: We should ask Colby to update the gsc branch on his github with his fix, and then build our own executable

lib/perl/Genome/Model/Tools/Speedseq/Align.pm
jasonwalker80: Removed.

ur
jasonwalker80: I'm not sure how these commits ended up here.  Likely a `git commit -a`.  Suggestions for removing this from the PR are welcome.

lib/perl/Genome/Model/Tools/Speedseq/Realign.t
jasonwalker80: Since the files are not explicitly defined, the subroutine output_files returns all files output after speedseq runs, I elected to test for the size of the bai files.  If BAI files are not found (or extras are found), the number of tests will not match the expected number.

lib/perl/Genome/Model/Tools/Speedseq/Base.pm
jasonwalker80: I elected to not touch this part of the code since it's copy/paste from Picard.  I think as another PR, the common features between the two tool trees should be refactored into a more generic class.

lib/perl/Genome/Model/Tools/Speedseq/Base.pm
jasonwalker80: Again, I elected to leave this as similar to the Picard code so it could be merged at some point.  I think the intent of 'Text' type was to quote it since it could be a filesystem path.  'String' would not be quoted.

lib/perl/Genome/Model/Tools/Speedseq/Base.pm
jasonwalker80: gone

lib/perl/Genome/Model/Tools/Speedseq/Base.pm
apregier: Ok - I misunderstood this code at first and thought that everything other than Boolean or Text would be skipped, but that is not the case.

ur
apregier: Hmm, I'm not sure what the cleanest way is - you could revert the commit they went in on and then make a new commit only with the files you want, then push both the revert commit and the new commit?  I am guessing this will drop you into merge conflict hell though.

apregier:
+1 when tests pass and ur version change is removed

jasonwalker80:
Replaced by #758

PR #689, opened 2015-04-24
Encode/Decode workflow inputs and outputs when using PTero

davidlmorton:
We were already doing this for steps within a workflow, but some
workflows take objects as inputs and return objects as outputs.  This
allows that to work.


davidlmorton:
Merging this since it's a very small change only affecting PTero and is needed to get these tests to pass:

https://apipe-ci.gsc.wustl.edu/view/Ptero/job/ptero_genome_process_tests/

PR #692, opened 2015-04-24
Set bam_size regardless of _use_alignment_summary_cpp

dufeiyu:
This bug is preventing per_lane_tophat and bsmap alignment results from setting bam_size metric.


tmooney:
+1 :black_small_square:

susannasiebert:
+1

PR #694, opened 2015-04-25
convert Genome to Genome::Config

nnutter:
This is creating a facade (or something) around `%ENV` to separate out this work from the new implementation of `Genome::Config` in genome/genome#635.


nnutter:
I've also run `5.10-reference-alignment` and `5.10-somatic-variation-short` and both passed.

nnutter:
There are some hard-coded references to environment variables that I will be able to remove in genome/genome#635.  There will be some hard-coded references in shell scripts that I'm not sure we can do anything with.

nnutter:
This is basically impossible to review in GitHub so if you want to review it you'll need to [checkout the pull request](https://github.com/genome/docs/wiki/How-do-I-%60git-checkout%60-a-pull-request%3F).  I think realistically you could just look at a couple files and see if you object to the concept.  Any failures that happen due to lack of test coverage are probably not worth the time to try to prevent.

acoffman:
I spot checked a handful of these and they seemed good to me.

+1

brummett:
+1

nnutter:
@acoffman @brummett Thanks for spot checking!

PR #695, opened 2015-04-27
Get bam size from get_bam_file  instead of from alignment_bam_file_paths

susannasiebert:
This should fix issues where we will try to estimate the bam size during merging and the `bam_size` hasn't been set on a deleted per-lane bam.

### DIFF
--- a/lib/perl/Genome/InstrumentData/AlignmentResult.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult.pm
@@ -1070,7 +1070,7 @@ sub _compute_alignment_metrics {
         $self->singleton_read_count         ($res->{singleton});
         $self->singleton_base_count         ($res->{singleton_bp});
 
-        $self->bam_size(stat($bam)->size); #store this for per lane bam recreation
+        $self->set_bam_size($bam); #store this for per lane bam recreation
         Genome::Utility::Instrumentation::inc('alignment_result.read_count', $self->total_read_count);
     }
 
--- a/lib/perl/Genome/InstrumentData/AlignmentResult.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult.pm
@@ -1150,8 +1150,8 @@ sub set_bam_size {
     my ($self, $bam_file) = @_;
     return 1 if $self->bam_size;
 
-    unless (defined $bam_file || -s $bam_file) {
-        $bam_file = $self->get_bam_file unless defined $bam_file;
+    unless (defined $bam_file && -s $bam_file) {
+        $bam_file = $self->get_bam_file;
         unless (-s $bam_file) {
             die $self->error_message('BAM file (%s) does not exist or is empty', $bam_file);
         }
--- a/lib/perl/Genome/InstrumentData/AlignmentResult.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult.pm
@@ -1148,9 +1148,15 @@ sub create_bam_header {
 
 sub set_bam_size {
     my ($self, $bam_file) = @_;
-    $bam_file = $self->get_bam_file unless defined $bam_file;
-
     return 1 if $self->bam_size;
+
+    unless (defined $bam_file || -s $bam_file) {
+        $bam_file = $self->get_bam_file unless defined $bam_file;
+        unless (-s $bam_file) {
+            die $self->error_message('BAM file (%s) does not exist or is empty', $bam_file);
+        }
+    }
+
     $self->bam_size(stat($bam_file)->size);
 
     return 1;
--- a/lib/perl/Genome/InstrumentData/AlignmentResult.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult.pm
@@ -1146,6 +1146,15 @@ sub create_bam_header {
     return 1;
 }
 
+sub set_bam_size {
+    my ($self, $bam_file) = @_;
+    $bam_file = $self->get_bam_path unless defined $bam_file;
+
+    return 1 if $self->bam_size;
+    $self->bam_size(stat($bam_file)->size);
+
+    return 1;
+}
 
 sub _verify_bam {
     my $self = shift;
--- a/lib/perl/Genome/InstrumentData/AlignmentResult.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult.pm
@@ -1156,6 +1156,7 @@ sub set_bam_size {
             die $self->error_message('BAM file (%s) does not exist or is empty', $bam_file);
         }
     }
+    return 1 if $self->bam_size;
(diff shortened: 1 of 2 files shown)


lib/perl/Genome/InstrumentData/AlignmentResult.pm
```diff
@@ -1146,6 +1146,15 @@ sub create_bam_header {
     return 1;
 }
 
+sub set_bam_size {
+    my ($self, $bam_file) = @_;
+    $bam_file = $self->get_bam_path unless defined $bam_file;
```
tmooney: I don't think `get_bam_path` exists.  Both `get_bam_file` and `bam_path` do, though.

lib/perl/Genome/InstrumentData/AlignmentResult.pm
```diff
@@ -1146,6 +1146,15 @@ sub create_bam_header {
     return 1;
 }
 
+sub set_bam_size {
```
tmooney: Should we make the other place that is setting `bam_size` call this as well?

lib/perl/Genome/InstrumentData/AlignmentResult.pm
susannasiebert: I meant to use `get_bam_file`. I pushed a fix.

lib/perl/Genome/InstrumentData/AlignmentResult.pm
tmooney: (The relevant location is being changed in #692.)

susannasiebert: Pushed a commit to change line 1073 to call `set_bam_size` and passing in `$bam`.

ebelter:
+1

susannasiebert:
I merged in #692 and made some changes based on talking to Feiyu.

lib/perl/Genome/InstrumentData/AlignmentResult.pm
```diff
@@ -1146,6 +1146,21 @@ sub create_bam_header {
...
+    my ($self, $bam_file) = @_;
+    return 1 if $self->bam_size;
+
+    unless (defined $bam_file || -s $bam_file) {
```
dufeiyu: I think "and", not "||" should be here.

lib/perl/Genome/InstrumentData/AlignmentResult.pm
```diff
@@ -1146,6 +1146,21 @@ sub create_bam_header {
...
+        $bam_file = $self->get_bam_file unless defined $bam_file;
```
dufeiyu: "unless defined $bam_file" not needed here.

lib/perl/Genome/InstrumentData/AlignmentResult.pm
susannasiebert: You're right. This is now fixed.

lib/perl/Genome/InstrumentData/AlignmentResult.pm
susannasiebert: Apologies for the oversight. It's removed now.

lib/perl/Genome/InstrumentData/AlignmentResult/Merged.pm
```diff
@@ -419,15 +419,14 @@ sub estimated_kb_usage {
         my $bam_size = $alignment->bam_size;
 
         unless (defined $bam_size) {
-            my @aln_bams = $alignment->alignment_bam_file_paths;
-            unless (@aln_bams) {
+            my $aln_bam = $alignment->get_bam_file;
+            unless ($aln_bam) {
                 die $self->error_message("alignment $alignment has no bams at " . $alignment->output_dir);
```
dufeiyu: This error message is not accurate now since get_bam_file would return either revivified bam in temp allocation or copied in-palce bam in temp allocation.

susannasiebert: Do we need to change line 423 to `unless (-s $aln_bam)` and the error message to something along the lines of `die $self->error_message('BAM file (%s) does not exist or is empty', $bam_file);`?

lib/perl/Genome/InstrumentData/AlignmentResult.pm
```diff
@@ -1146,6 +1146,21 @@ sub create_bam_header {
...
+    my ($self, $bam_file) = @_;
+    return 1 if $self->bam_size;
+
+    unless (defined $bam_file && -s $bam_file) {
+        $bam_file = $self->get_bam_file;
+        unless (-s $bam_file) {
+            die $self->error_message('BAM file (%s) does not exist or is empty', $bam_file);
+        }
+    }
+
```
dufeiyu: A lock is probably needed here. Also check bam_size again since revivifying bam of get_bam_file could take long time and bam_size already set
my $lock  = $self->get_bam_lock;
my $guard = Scope::Guard->new(sub { $lock->unlock() });
return 1 if $self->bam_size;

lib/perl/Genome/InstrumentData/AlignmentResult/Merged.pm
```diff
@@ -419,15 +419,14 @@ sub estimated_kb_usage {
...
-            unless (@aln_bams) {
-                die $self->error_message("alignment $alignment has no bams at " . $alignment->output_dir);
-            }
-            for (@aln_bams) {
-                my $size = stat($_)->size;
-                $self->debug_message("BAM has size: " . $size);
-                $bam_size += $size;
+            my $aln_bam = $alignment->get_bam_file;
+            unless (-s $aln_bam) {
+                die $self->error_message('BAM file (%s) for alignment (%s) does not exist or is empty', $bam_file, $alignment);
```
dufeiyu: $bam_file or $aln_bam ?  $alignment->id  or $alignment ?

lib/perl/Genome/InstrumentData/AlignmentResult.pm
```diff
@@ -1146,6 +1146,22 @@ sub create_bam_header {
...
+    my ($self, $bam_file) = @_;
+    return 1 if $self->bam_size;
+
+    unless (defined $bam_file && -s $bam_file) {
+        $bam_file = $self->get_bam_file;
+        unless (-s $bam_file) {
+            die $self->error_message('BAM file (%s) does not exist or is empty', $bam_file);
+        }
+    }
+    return 1 if $self->bam_size;
```
tmooney: If the idea here is to check if another process has set `bam_size` in the meantime, then we'll need to do something to ensure this checks the database again.  (For example, setting our `UR::Context`'s `query_underlying_context` property to `1` temporarily.)

PR #696, opened 2015-04-28
fallback on move if rename is not safe

nnutter:
Since it may be hard to know that rename is safe we'll check for cross-device
renames and fallback to move.

### DIFF
--- a/lib/perl/Genome/Sys.pm
+++ b/lib/perl/Genome/Sys.pm
@@ -1749,9 +1749,22 @@ sub _unpreserved_permissions {
     return 1;
 }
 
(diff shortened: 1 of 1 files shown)


ebelter:
Why not always use File::Path::move?

nnutter:
You're right, I tunneled too much.  `rename` was there to guarantee a `rename` and to fail otherwise.  If `move` is acceptable then that is what the calling code should use.  I'll close this and re-work the fix for [RT#105443](https://rt.gsc.wustl.edu/Ticket/Display.html?id=105443).

PR #697, opened 2015-04-28
improve error message when trying to rename across devices

nnutter:


ebelter:
+1 Not sure why anyone would use rename vs. File::Copy::move but this gives a great error.

nnutter:
You would use `rename` when you needed to ensure an atomic operation.  `File::Copy::move` falls back to a copy and delete when a rename isn't possible (which is usually when the move is cross device).

ebelter:
File::Copy::move tries rename first.

nnutter:
And falls back on a non-atomic operation.

tmooney:
+1 :cherry_blossom:

PR #698, opened 2015-04-28
add code to help debug MutationSpectrum failures.

gatoravi:
Some of the test-model builds don't diff cleanly because of missing p-values from the `prop.test`. This is an intermittent issue which is yet to be reproduced outside a test-build. This PR is an attempt to find the cause of this error.


apregier:
Is there some way of printing out all of the versions of all of the R modules that are loaded to the log?  Since we suspect that might be the culprit, it would be a place to start.

gatoravi:
Great idea. I think `sessionInfo()` is what we need.

gatoravi:
I've added this. Also, I've changed the code to write the p-value columns to a new file since that seemed like a good idea. Is this ok, or should we just leave as is and write to the same file?

malachig:
Other ways to get info on an R build that might be relevant to compatibility issues:

`capabilities()`
`.libPaths()`

Also, in the sGMS when we were having problems with this I created a list of R packages that we might need and tested the ability to load them as follows:

```R
#List of packages installed properly (by install.packages or biocLite)
installed_package_list = c('bitops', 'Cairo', 'caTools', 'colorspace', 'doMC', 'e1071', 'filehash', 'foreach',
                           'fpc', 'ggplot2', 'Hmisc', 'intervals', 'itertools', 'Matrix', 'mgcv', 'mixdist',
                           'mixtools', 'multicore', 'nortest', 'h5r', 'plotrix', 'proto', 'RColorBrewer',
                           'scatterplot3d', 'SKAT', 'statmod', 'xtable', 'getopt', 'gtools', 'gdata',
                           'affy', 'Biobase', 'cummeRbund', 'DNAcopy', 'edgeR', 'limma', 'gcrma', 'genefilter',
                           'geneplotter', 'GenomeGraphs', 'lumi', 'methylumi', 'multtest', 'preprocessCore', 'ChemmineR',
                           'boot','copyCat', 'gplots', 'pbh5')

package_count = length(installed_package_list)
package_pass_count = 0
package_fail_count = 0

#Test packages installed properly (by install.packages or biocLite)
for (package_name in installed_package_list){
  if (!package_name %in% rownames(installed.packages())){
    message=paste("ERROR: ", package_name, " package not installed successfully", sep="")
    print(message)
    package_fail_count = package_fail_count+1
  }else{
    message=paste(package_name, " loaded", sep="")
    print(message);
    package_pass_count = package_pass_count+1
  }
}

if ((package_pass_count == package_count) & (package_fail_count == 0)){
  message=paste("R PACKAGE INSTALL SUCCESS. Installed ", package_count, " packages and their dependencies", sep="")
  print(message)
}

```

gatoravi:
good call, I've added those two as well. I suspect `.libPaths()` will be helpful here, running a build to see if the information we need is being captured.

gatoravi:
I made some changes to the way I call these diagnostic R functions. 

(These are now called from the CallR.pm module. This module uses the Statistics::R module which has a method to read the output from R, if the print is in the middle of a different function instead its not clear how to get that output.)

It looks like we've moved away from this way of calling R scripts in genome to an `Rscript` based approach. That might be another cryptic cause for the failures we are seeing.

gatoravi:
Selfishly merging this in so that we can debug the sporadic build failures better.

PR #700, opened 2015-04-29
Use the labelled merged result in SomaticValidation's Identify DNP

tmooney:
The "refinements" each produce separate merged BAMs, so picking the first one we run across is not guaranteed to get the final BAM.  Thankfully we already have an accessor for choosing the proper one.


acoffman:
+1 much simpler!

susannasiebert:
+1

PR #701, opened 2015-04-29
Bump FastQC's JVM memory request to 512 MB.

tmooney:
* This is still well inside the LSF memory request.
* 51a4d27df3d515bf5697032aad5a0584d229ac91 claims this value should scale with threads.  (I don't know that we're currently scaling our request, however.  The failure case I'm addressing was run with one thread.)


nnutter:
Well 51a4d27df3d515bf5697032aad5a0584d229ac91 isn't a very helpful commit message!

:+1:

ebelter:
+1

PR #702, opened 2015-04-29
Add tools to QC framework

susannasiebert:
This is the last set of commits needed to start using the QC framework in our builds. 

This PR adds qc tool wrappers to a minimum set of picard tools needed for a basic level of qc. It also sets a default qc config that uses these tools. It also fixes a few minor bugs.


lib/perl/Genome/Qc/Tool.pm
```diff
@@ -50,4 +50,14 @@ sub output_file_accessor {
...
 }
 
+sub reference_build {
+    my $self = shift;
+    return $self->alignment_result->reference_build;
+}
+
+sub reference_sequence {
+    my $self = shift;
+    my $build =  $self->reference_build->full_consensus_path('fa');
```
tmooney: This might be clearer as something like:
```perl
my $build = $self->reference_build;
return $build->full_consensus_path('fa');
```
or just:
```perl
return $self->reference_build->full_consensus_path('fa');
```

The current version happens to return the FASTA.  It is a little misleading, though, since `$build` isn't a build and the variable isn't used.

lib/perl/Genome/Qc/Tool/BamFile.pm
```diff
@@ -18,7 +14,11 @@ sub supports_streaming {
 
 sub cmd_line {
     my $self = shift;
-    return (qw(samtools view), $self->input_file);
+    return (qw(samtools view), $self->gmt_params->{input_file});
```
tmooney: This should probably be concerned with using a particular version via `Genome::Model::Tools::Sam`.

susannasiebert:
Yes, they can be paired down. I'll limit them to maybe a single chromosome.

These files were created from one of our feature lists (I forget which one exactly) which I thought would be ok to use.

lib/perl/Genome/Qc/Tool/BamFile.pm
susannasiebert: I talked to @apregier about this suggestion and we believe that samtools is stable enough that we are safe to call samtools directly. Moreover, `G::M::T::Sam::BamToSam` also just calls `samtools view` directly and ignores the `user_version` parameter.

tmooney: Sounds like we should change `gmt sam bam-to-sam`! :smile:

tmooney:
+1 :rice_cracker:

It might be nice to rebase the commit that pares down the files into the other one so the big versions don't make it into the main repository history.

susannasiebert:
Yes, I will rebase before merging.

apregier:
It would be great to get a review from @ebelter and/or @dufeiyu - could you guys look at this, since you have had experience with previous implementations of qc?

ebelter:
+1 This seems reasonable.

susannasiebert:
I made the rebase as suggested by @tmooney and will merge once the tests pass.

PR #703, opened 2015-04-29
Open fewer simultaneous filehandles in AlignmentSummaryV2

tmooney:
AlignmentSummaryV2 splits the BED into one per chromosome. For the human reference, this isn't too many, but for other combined references this may exceed the maximum number of allowed filehandles.  This PR closes the previous ones if it would open a 101st and lets them be reopened as needed.

### DIFF
--- a/lib/perl/Genome/Model/Tools/BioSamtools/AlignmentSummaryV2.pm
+++ b/lib/perl/Genome/Model/Tools/BioSamtools/AlignmentSummaryV2.pm
@@ -568,7 +568,7 @@ sub _prepare_split_sorted_bed_files {
     # Make a file for each chromosome, and sort by the start and end position 
 
     my $original = IO::File->new($self->bed_file, 'r') || die "Can't open bed file ".$self->bed_file." for reading: $!";
-    my $temp_dir = File::Temp::tempdir( CLEANUP => 1 );
+    my $temp_dir = Genome::Sys->create_temp_directory;
 
     # First, bin each line by chromosome and make unsorted files
     my %open_handles;
--- a/lib/perl/Genome/Model/Tools/BioSamtools/AlignmentSummaryV2.pm
+++ b/lib/perl/Genome/Model/Tools/BioSamtools/AlignmentSummaryV2.pm
@@ -577,8 +577,13 @@ sub _prepare_split_sorted_bed_files {
         my $chrom = $1;
 
         unless ($open_handles{$chrom}) {
+            if(scalar keys %open_handles >= 100) {
+                $_->close foreach values %open_handles;
+                %open_handles = ();
+            }
+
             my $filename = $self->_unsorted_bed_file_name_for_chrom($temp_dir, $chrom);
-            $open_handles{$chrom} = IO::File->new($filename, 'w');
+            $open_handles{$chrom} = IO::File->new($filename, 'a');
             $open_handles{$chrom} || die "Can't open file $filename for writing: $!";
         }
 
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Model/Tools/BioSamtools/AlignmentSummaryV2.pm
```diff
@@ -577,8 +577,13 @@ sub _prepare_split_sorted_bed_files {
         my $chrom = $1;
 
         unless ($open_handles{$chrom}) {
+            if(scalar keys %open_handles >= 100) {
+                $_->close foreach values %open_handles;
+                %open_handles = ();
```
nnutter: Does this break next for loop where it is "Now sort each file"?

lib/perl/Genome/Model/Tools/BioSamtools/AlignmentSummaryV2.pm
```diff
@@ -577,8 +577,13 @@ sub _prepare_split_sorted_bed_files {
         my $chrom = $1;
 
         unless ($open_handles{$chrom}) {
+            if(scalar keys %open_handles >= 100) {
+                $_->close foreach values %open_handles;
```
nnutter: Maybe it should just cache a filehandle for consecutive writes for the same chromosome but close and open anytime the chromosome changes?  I don't know what the data looks like.

lib/perl/Genome/Model/Tools/BioSamtools/AlignmentSummaryV2.pm
tmooney: Quite right you are!  I'll have to change this.

lib/perl/Genome/Model/Tools/BioSamtools/AlignmentSummaryV2.pm
tmooney: A lot of our input files are not sorted at all and jump around from chromosome to chromosome.  I had also thought about sorting the file in advance of this loop, so we knew we only needed one filehandle at a time.

lib/perl/Genome/Model/Tools/BioSamtools/AlignmentSummaryV2.pm
tmooney: I've added a separate hash to track all the chromosomes we've seen previously.

nnutter:
:+1:

brummett:
+1

If you're interested, the FileCache module basically does the same thing and it comes with Perl.

nnutter:
[FileCache](http://perldoc.perl.org/5.8.9/FileCache.html)++

tmooney:
`FileCache` is quite odd:
> The module functionality relies on symbolic references, so things will break under 'use strict' unless 'no strict "refs"' is also specified.

This code works:
```perl
package This::Is::A::Test;
use strict;
use warnings;
use FileCache maxopen => 16;

my $filename = "/tmp/test";
cacheout($filename) or die "Can't open file $filename for writing: $!";
my $fh = __PACKAGE__ . '::' . $filename;
no strict 'refs';
print $fh "This is a test.\n";
```

This code does not:
```perl
package This::Is::A::Test;
use strict;
use warnings;
use FileCache maxopen => 16;

my $filename = "/tmp/test";
my $fh = cacheout($filename) or die "Can't open file $filename for writing: $!";
no strict 'refs';
print $fh "This is a test.\n";
```
(This example produces `print() on unopened filehandle /tmp/test at - line 9.` and thus no output in the file.)

nnutter:
[FileCache](http://perldoc.perl.org/5.8.9/FileCache.html)--

acoffman:
Personally, I think the extra clarity that comes from using FileCache isn't worth turning strict refs off, but 
I'm fine with either approach.

+1 either way.

brummett:
Looks like FileCache is a relic from Perl4 or early Perl5.  Forget I even suggested it...

My previous +1 still stands

PR #705, opened 2015-04-29
show stderr when trying to load module for assert_module_found

nnutter:
If there is an error trying to use the module then this will show it.


acoffman:
+1

PR #706, opened 2015-04-29
PTero can run Processes without a workflow slot

davidlmorton:
This makes it possible to run a `Genome::Process` using PTero that does not require a workflow lsf queue slot.  The only thing that the process using that lsf slot was doing, was to update the status on the `Genome::Process`.

This PR wraps the work you want to do in an outer workflow that has steps that update the `Genome::Process`s status.


lib/perl/Genome/Process.pm
```diff
@@ -225,6 +231,24 @@ sub _schedule_process {
...
+        die sprintf("Failed to submit process (%s): %s",
+            $self->id, $transaction->error_message || 'Reason Unknown');
+    }
+
+    my $cmd = Genome::Process::Command::Run->create(
+        process => $self,
+    );
+    my $wf_proxy = $cmd->submit();
+    $self->status_message("Successfully launched process (%s) and ".
+        "submitted ptero workflow (%s)",
```
tmooney: Should we consistently capitalize `PTero`?

tmooney:
`Genome::Process::Command::SetStatus` might benefit from a minimal LSF resource specification (possibly specifying less than the default 4GB of RAM and `$ENV{GENOME_LSF_QUEUE_SHORT}` as the queue).

davidlmorton:
`Genome::Process::Command::SetStatus` should never execute on lsf.  It is only specified to run on the shell-command service, although nothing is currently preventing it from being included by somebody in a flow/legacy workflow where it could fail (exit 1) and then be dispatched on lsf.

lib/perl/Genome/Ptero/ToJSONForProcess.t
```diff
@@ -0,0 +1,67 @@
...
+    $ENV{UR_USE_DUMMY_AUTOGENERATED_IDS} = 1;
+};
+
+
+for my $test_directory (glob test_data_directory('*')) {
+    my $test_name = basename($test_directory);
+    my $workflow = Genome::WorkflowBuilder::DAG->from_xml_filename(workflow_xml_file($test_name));
+
+    my $json_filename = do {
+        local %ENV; # otherwise the JSON contains the current user's environment.
```
tmooney: Would it be worthwhile to include a small environment here that includes a defined and an undefined variable?

tmooney:
Re: `SetStatus`... Fair enough!  It's probably not necessary; nobody ought to be doing that.

brummett:
+1

tmooney:
+1 :gem: Go, go, PTero!

iferguson90:
+1 :bird:

PR #707, opened 2015-04-29
Reallocate temp allocation for revivified bam or in-place bam copy

dufeiyu:
Reallocation is needed since revivified bam or in-place bam copy will exist for a while during merge and the temp allocation could be over-allocated if bam_size is not set.

### DIFF
--- a/lib/perl/Genome/InstrumentData/AlignmentResult.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult.pm
@@ -13,6 +13,7 @@ use File::stat;
 use Carp qw(confess);
 use File::Basename;
 use Scope::Guard qw();
+use Try::Tiny qw(try catch);
 
 use Genome::Sys::LockMigrationProxy qw();
 use Genome::Utility::Instrumentation;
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/InstrumentData/AlignmentResult.pm
```diff
@@ -1736,6 +1736,7 @@ sub get_bam_file {
                     Genome::Sys->copy_file($file, $dest_file);
                 }
             }
+            eval{$temp_allocation->reallocate};
```
nnutter: We should probably use Try::Tiny try to avoid leaving `$@` set.

nnutter: I think I'm wrong.

nnutter:
:+1:

brummett:
You may not need to use Try::Tiny, but you do still need to ```local $@;``` before each eval.

You don't need to jump through hoops to get the exception out of the block, since you're just masking any exception within the eval, and not trying to throw one.

dufeiyu:
Thanks Tony. I actually made a change to use Try::Tiny.

brummett:
+1

dufeiyu:
Added return;

PR #709, opened 2015-04-29
remove `local $@`

nnutter:
I misunderstood one of the issues with `$@` so am trying to cleanup a bit.  I
looked for all `local $@` because `$@` has to be scoped very narrowly to avoid
causing problems when using `local $@`.  See [Localizing silently masks errors][1].

[1]: https://metacpan.org/pod/Try::Tiny#Localizing-silently-masks-errors


tmooney:
+1 :pushpin:

brummett:
+1

PR #710, opened 2015-04-30
Reduce AlignmentSummary (V1) cache size based on contig count

tmooney:
This uses a coarser coordinate index as the number of contigs (chromosomes) increases.  We were running out of memory on a reference with > 30k contigs.  With this change it succeeded with a runtime still under an hour.

(In general there might be better ways of going about this--that's why `AlignmentSummaryV2` exists!  But we run both of them, since they produce slightly different results.)


brummett:
+1

acoffman:
+1

PR #711, opened 2015-04-30
Match per lane bam sort order to merge bam sort order

dufeiyu:
Currently build will fail at picard ReplaceSamHeader if sort orders are different between per lane and merge bams. This PR is to match per lane bam sort order to that of merge bam.


tmooney:
+1 :melon:

ebelter:
+1 Would like to see a test, but it appears it should work.

PR #712, opened 2015-04-30
InstData Import: make work flow inputs and source files objects. This is a step to making the import a genome process.

ebelter:


lib/perl/Genome/InstrumentData/Command/Import/Basic.pm
```diff
@@ -104,6 +98,7 @@ sub execute {
     my $inputs = $self->_gather_inputs_for_workflow;
     return if not $inputs;
 
+    $DB::single=1;
```
tmooney: Hark! A breakpoint lurks in the shadows, ready to leap out at unsuspecting debuggers by.

lib/perl/Genome/InstrumentData/Command/Import/Basic.pm
```diff
@@ -112,72 +107,32 @@ sub execute {
...
 
-    my $instrument_data_properties =  $instdata_props_processor->resolved_instrument_data_properties;
-    if ( not $instrument_data_properties->{original_data_path} ) {
-        $instrument_data_properties->{original_data_path} = join(',', $self->source_files);
-    }
+    return $self->inputs($instdata_props_processor);
+}
 
-    return $self->_instrument_data_properties($instrument_data_properties);
+sub _die_if_original_data_path_previously_imported {
```
tmooney: How hard would it be to check something like the md5sum of the instrument data rather than its path?

lib/perl/Genome/InstrumentData/Command/Import/Basic_fails.t
```diff
@@ -22,82 +25,97 @@ my $library = Genome::Library->create(
...
+            analysis_project => $analysis_project,
+            library => $library,
+            source_files => [ $source_file ],
+            import_source_name => 'broad',
+            downsample_ratio => 0.25,
+            instrument_data_properties => [qw/ downsample_ratio=0.24 /],
+        );
+    },
+    qr//,
+    'Fails w/ conflicting cmd and instdata properties instrument_data_properties',
```
tmooney: Not something you changed, but this says `instdata properties instrument_data_properties`.

lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/SourceFile.pm
```diff
@@ -0,0 +1,121 @@
...
+    else {
+        return 'local disk';
+    }
+}
+
+sub file_size {
+    my $self = shift;
+
+    my $source_file = $self->path;
+    if ( $source_file !~ /^http/ ) {
```
tmooney: This might be better as `if ($self->retrieval_method eq 'local disk') {`.

lib/perl/Genome/InstrumentData/Command/Import/Basic.pm
ebelter: This is just a quick check. A proper MD5 is run as the first step of import to ensure it had not been imported before. It is in the work flow because files can be large.

lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/SourceFiles.pm
```diff
@@ -0,0 +1,105 @@
...
+
+sub kilobytes_required_for_processing {
+    my $self = shift;
+
+    my $kb_required = 0;
+    for my $source_file ( $self->source_files ) {
+        $kb_required += $source_file->kilobytes_required_for_processing;
+    }
+
+    die $self->error_message('Failed to calculatge kilobytes required fo processing!') if not $kb_required;
```
apregier: 2 typos (calculatge fo)

lib/perl/Genome/InstrumentData/Command/Import/Basic.pm
apregier: I wonder if we need to also add some locking to be sure an import isn't in progress

ebelter: It is true that one could start while another is running, and 2 copies could ultimately exist. There are 2 md5 checks, one at the beginning and one again at the end when the instdata are created that prevent it happening most of the time. Locking can be added, but that would be another PR.

PR #713, opened 2015-04-30
GMT Integration of VCF Evaluation (part 3)

indraniel:
This is part 3 of a roll out to integrate the [VCF Evaluation scripts][1] within the genome codebase. The aim of this pull request is to address minor differences with recent changes to [evaluate_vcf.pl][2] and minor annoyances with the final summary report.

* use `NaN` instead of `0` for divde by zero cases
* use `has_transient_optional_output` instead of `has_output` for the transient output hash storing the statistical results
* store the statistical results to a JSON file

See internal jira issue [BIO-1387][3] for details.

[1]: https://github.com/genome/vcf-evaluation
[2]: https://github.com/genome/vcf-evaluation/blob/master/evaluate_vcf.pl
[3]: https://jira.gsc.wustl.edu/browse/BIO-1387

### DIFF
--- a/lib/perl/Genome/Model/Tools/Vcf/EvaluateVcf.pm
+++ b/lib/perl/Genome/Model/Tools/Vcf/EvaluateVcf.pm
@@ -472,8 +472,11 @@ sub stat_sensitivity_exact {
     my $results = $self->get_rawstats();
 
     my $total_true = $self->stat_total_true_positive_exact;
-    if ($total_true == 0) {
-        return 0;
+    {
+        no warnings;
+        if ($total_true == 0) {
+            return "NaN";
+        }
     }
 
     my $stat = $results->{true_positive_exact} / $total_true;
@@ -500,8 +503,11 @@ sub stat_sensitivity_partial {
     my $results = $self->get_rawstats();
 
     my $total_true = $self->stat_total_true_positive_partial;
-    if ($total_true == 0) {
-        return 0;
+    {
+        no warnings;
+        if ($total_true == 0) {
+            return "NaN";
+        }
     }
 
     my $stat = $results->{true_positive_partial} / $total_true;
@@ -537,8 +543,11 @@ sub stat_exact_specificity {
     # may be significantly smaller than the target space ROI
     
     my $tn = $self->stat_true_negatives();
-    if ($tn == 0) {
-        return 0;
+    {
+        no warnings;
+        if ($tn == 0) {
+            return "NaN";
+        }
     }
 
     my $stat =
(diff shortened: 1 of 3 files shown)


lib/perl/Genome/Model/Tools/Vcf/EvaluateVcf.pm
```diff
@@ -472,8 +472,11 @@ sub stat_sensitivity_exact {
     my $results = $self->get_rawstats();
 
     my $total_true = $self->stat_total_true_positive_exact;
-    if ($total_true == 0) {
-        return 0;
+    {
+        no warnings;
```
tmooney: What warnings are you trying to avoid?  It's probably better to only disable the warnings you expect to see, e.g.:
```perl
no warnings qw(uninitialized);
```
although another way to avoid `undef` here could be:
```perl
my $total_true = $self->stat_total_true_positive_exact // 0;
```

lib/perl/Genome/Model/Tools/Vcf/EvaluateVcfs.pm
```diff
@@ -5,6 +5,7 @@ use warnings;
 
 use Cwd;
 use Path::Class;
+use JSON::XS;
```
tmooney: If you `use JSON;` it will use the XS version when it is available, but also fall back on the pure-Perl implementation otherwise.

From [the JSON::XS docs](http://search.cpan.org/~mlehmann/JSON-XS-3.01/XS.pm#DESCRIPTION):
>Beginning with version 2.0 of the JSON module, when both JSON and JSON::XS are installed, then JSON will fall back on JSON::XS (this can be overridden) with no overhead due to emulation (by inheriting constructor and methods). If JSON::XS is not available, it will fall back to the compatible JSON::PP module as backend, so using JSON instead of JSON::XS gives you a portable JSON API that can be fast when you need and doesn't require a C compiler when that is a problem.

lib/perl/Genome/Model/Tools/Vcf/EvaluateVcf.pm
indraniel: The variable may not necessarily be a numerical zero.  It could be `undef`; a string like `NaN`, `nan`, or `NA`, or some other potentially non-numerical result. Unfortunately, I don't exactly know what the other non-numerical results could be at the moment, but I know that perl will evaluate it to be zero which is what I would like the code to do.

tmooney: For that case you could use:
```perl
no warnings qw(numeric);
```

`nan` (also `NaN`) is a numeric value, but it does not equal zero.  `NA` is not a numeric value, but it does equal zero.  If the set of non-numerical results could be any string, you cannot know if it will be equal to zero.  (For example, `2FA` is not numeric, but equals `2`.)   It might be better to use `Scalar::Util`'s `looks_like_number` and handle non-numbers specifically.

indraniel: After thinking about this and conversing on the side via IM, I think I'll keep these "no warnings" sections as is for the moment.

indraniel:
Would it be okay to merge in this pull request now?

acoffman:
+1

PR #715, opened 2015-05-01
avoid tight coupling to last error_message

nnutter:
These tests fails for genome/UR#75 because a later error message is sent that
differs from the one these are testing.  The ones they are testing are still
sent but are not the last `error_message`.  I've changed the tests to try to
avoid tight coupling with the last `error_message`.


brummett:
+1

You could have grepped through the error_messages() strings instead of the error Tag objects, but this works, too.

tmooney:
+1 :doughnut:

PR #717, opened 2015-05-04
try to improve consistency of unarchiving

nnutter:
I found 391 allocations that had their `status` still set to 'archived' but
their `mount_path` was not a `/gscarchive` path.  These changes are possible
fixes for what may be causing that inconsistency.

### DIFF
--- a/lib/perl/Genome/Disk/Detail/Allocation/Unarchiver.pm
+++ b/lib/perl/Genome/Disk/Detail/Allocation/Unarchiver.pm
@@ -57,6 +57,7 @@ sub unarchive {
     my $tar_path = $allocation_object->tar_path;
     my $cmd = "tar -C $target_path -xf $tar_path";
 
+    my $tx = UR::Context::Transaction->begin();
     try {
         # It's very possible that if no commit is on, the volumes/allocations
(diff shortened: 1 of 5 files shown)


lib/perl/Genome/Disk/Detail/Allocation/Unarchiver.pm
```diff
@@ -126,33 +116,54 @@ sub unarchive {
...
-            $allocation_object);
+        Genome::Timeline::Event::Allocation->unarchived($self->reason, $allocation_object);
         $allocation_object->status('active');
+
+        unless ($tx->commit() && $allocation_object->_commit_unless_testing) {
+            die 'failed to commit';
+        }
+    }
+    catch {
+        $tx->rollback();
```
tmooney: If `$tx->commit()` succeeded this could result in:
```
Cannot rollback a transaction that is committed. at /.../UR/Context/Transaction.pm line 150
```

nnutter:
CI, test this.

nnutter:
@tmooney and @brummett thanks for catching those.  Your concerns should be addressed in the latest revision.

nnutter:
I'm going to merge because I believe I've addressed concerns.

PR #718, opened 2015-05-04
Set validation references as being rederivable.

tmooney:
This adds the `is_rederivable` flag to the references created in these GMTs, which makes them like the ones created in the SomaticValidation pipeline.  (The practical effect is that such references will not be stored in the location to which `Genome::Config::get('disk_group_references')` points.)


acoffman:
+2

tmooney:
CI, test this.

nnutter:
:+1:

PR #719, opened 2015-05-05
Set keep_dbh_connection_open => 1

dufeiyu:
The command "samtools view -H <bam>" runs very fast (about several seconds).


nnutter:
:+1: Was this an "just" an optimization or did you experience some problem?

dufeiyu:
If without this, it sometimes gives out the following error:
UR::DBI::db=HASH(0x3c6d7b8)->disconnect invalidates 1 active statement handle (either destroy statement handles or call finish on them before disconnecting)

PR #720, opened 2015-05-05
ClinSeq caTissue Timepoint

jasonwalker80:
Add support for Surgery timepoints to clinseq.


jasonwalker80:
@malachig can you look this over?

lib/perl/Genome/Model/ClinSeq/Util.pm
```diff
@@ -1264,27 +1264,30 @@ sub get_ref_align_builds{
...
   foreach my $name (sort {$ref_builds{$a}->{time_point} cmp $ref_builds{$b}->{time_point}} keys %ref_builds){
-    my $day = $ref_builds{$name}{day};
-    if ($day =~ /day(\d+)/){
-      my $day_number = $1;
-      $timepoint_positions{$day_number}{position} = 0;
-      $ref_builds{$name}{day_number} = $day_number;
-    }else{
-      die $self->error_message("could not parse day value from sample attribute (caTissue timepoint): $day");
+    my $time_point_string = $ref_builds{$name}{time_point_string};
+    if ($time_point_string =~ /(day|Surgery)(\d+)/){
```
tmooney: Are we likely to see other time point units? We could make this a bit more generic as: `/([a-z]+)(\d+)/i`.

tmooney:
Does the `{day}` [referenced here](https://github.com/jasonwalker80/genome/blob/7485df6406a833d42a9f7ac41b8a17d9e1d71b83/lib/perl/Genome/Model/ClinSeq/Command/Converge/SnvIndelReport.pm#L1377) also need to be changed to `time_point_value`?

jasonwalker80:
@malachig and/or @gatoravi Any comments?  I see a comment in the code:
`#TODO: Right now, this script will only work for a very particular situation (normal, dayX_tumor, dayY_tumor) - Make it more flexible ...`

gatoravi:
Hi Jason, yes, the `{day}` reference that Tom points out will also need to be changed since it uses the hash reference obtained from `get_ref_align_builds` which now has a modified key.
I think overall the change looks good to me but it'd be nice if @malachig could have a look over as well.

jasonwalker80:
Since day is no longer a part of the hash, I updated the referenced code to use time_point_unit and time_point_value concatenated as the new time point name.

lib/perl/Genome/Model/ClinSeq/Util.pm
jasonwalker80: I am ok with the suggestion; however, it could result in unexpected values in downstream plots.  Regardless, I don't think this is the right place to attempt to control the time point vocabulary.  I think it's ok to use the values found in the database as-is.  @malachig any thoughts on how open you want ClinSeq to be to new caTissue time point values?

malachig:
I believe this is all fine.  Thanks @jasonwalker80

PR #721, opened 2015-05-05
add Genome::Sys::LockProxy::unlock_guard()

nnutter:
It doesn't really save a lot of code but it still seems like an improvement.

### DIFF
--- a/lib/perl/Genome/Sys/LockProxy.pm
+++ b/lib/perl/Genome/Sys/LockProxy.pm
@@ -8,6 +8,8 @@ use Genome::Sys::Lock qw();
 use Params::Validate qw(validate_with);
 use Scalar::Util qw(blessed);
 
+require Scope::Guard;
+
 =item new()
 
 Keyword Arguments:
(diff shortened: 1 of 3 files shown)


acoffman:
It seems a little odd to have identical implementations in both Lock and LockProxy, but on the other hand, I'm not sure a single 2 line method is really worth adding an extra module and mixin for either. 

+1

lib/perl/Genome/Sys/LockProxy.pm
```diff
@@ -8,6 +8,8 @@ use Genome::Sys::Lock qw();
 use Params::Validate qw(validate_with);
 use Scalar::Util qw(blessed);
 
+require Scope::Guard;
```
dufeiyu: Does LockMigrationProxy.pm require Scope::Guard too ?

nnutter: > Does LockMigrationProxy.pm require Scope::Guard too ?

Yep!  Added.

dufeiyu:
+1  assume tests pass

PR #722, opened 2015-05-07
update clinseq blessed build

gatoravi:
This build adds the TDG clinical trial to the DGIdb sources used in ClinSeq.
Ref - https://github.com/genome/dgi-db/commit/d5c1d813bc84febc37dfffd5a807b9f17bb8b3ca


acoffman:
+1, good catch!

dufeiyu:
+1   this can be merged early and fix the clinseq build fail

PR #723, opened 2015-05-07
clone from existing genome repo

nnutter:
Clone from the existing genome repo instead of getting an origin/master.


nnutter:
This should, for example, allow the tests to pass on genome/genome#635.

acoffman:
+1

brummett:
+1

PR #724, opened 2015-05-07
Ensure that errors are able to be displayed in G::M::B::Abandon

davidlmorton:


acoffman:
+1

lib/perl/Genome/Model/Build/Command/Abandon.pm
```diff
@@ -52,12 +53,23 @@ sub execute {
...
+        if ($successful) {
+            if ($transaction->commit) {
+                $self->successfully_abandoned_callback($build);
+            } else {
+                $self->append_error($build->__display_name__, "Failed to commit transaction to abandon build");
+                $transaction->rollback;
+            }
+        } else {
+            $self->append_error($build->__display_name__, "Failed to abandon build for some reason.");
             $transaction->rollback;
```
nnutter: I think this is a second call to `rollback()` in the case where `abandon()` dies which I believe will cause an exception.  I think it would also append two errors in that case.

lib/perl/Genome/Model/Build/Command/Abandon.pm
```diff
@@ -52,12 +53,23 @@ sub execute {
...
-            $self->status_message( "Successfully abandoned build (" . $build->__display_name__ . ")." );
-        }
-        else {
-            $self->append_error($build->__display_name__, "Failed to abandon build: $@.");
+
+        my $successful = try {
+            $build->abandon($self->header_text, $self->body_text);
+        } catch {
+            $self->append_error($build->__display_name__, "Failed to abandon build: $_.");
+            $transaction->rollback;
```
nnutter: `catch` needs to `return;` otherwise `$successful` will be set to the value of `$transaction->rollback`.

lib/perl/Genome/Model/Build/Command/Abandon.pm
```diff
@@ -52,12 +53,23 @@ sub execute {
...
+
+        if ($successful) {
+            if ($transaction->commit) {
+                $self->successfully_abandoned_callback($build);
+            } else {
+                $self->append_error($build->__display_name__, "Failed to commit transaction to abandon build");
+                $transaction->rollback;
+            }
+        } else {
+            $self->append_error($build->__display_name__, "Failed to abandon build for some reason.");
```
nnutter: You could consolidate this into the `try` block, e.g.

```
try {
    $build->abandon(...)
        or die 'abandon failed for unknown reason';
} ...
```

lib/perl/Genome/Model/Build/Command/Abandon.pm
```diff
@@ -52,12 +53,23 @@ sub execute {
...
+
+        my $successful = try {
+            $build->abandon($self->header_text, $self->body_text);
+        } catch {
+            $self->append_error($build->__display_name__, "Failed to abandon build: $_.");
+            $transaction->rollback;
+        };
+
+        if ($successful) {
+            if ($transaction->commit) {
```
nnutter: You could consolidate this into the `try` block, e.g.

```
try {
    $build->abandon(...);
    $transaction->commit()
        or die 'commit failed';
} ...
```

nnutter:
Good job cleaning up the duplication.

nnutter:
:+1:

PR #725, opened 2015-05-08
WIP: Speedseq alignment strategy and software results

susannasiebert:
This is a prototype and @apregier and I would appreciate some feedback about this implementation.

The speedseq aligner does not create intermediate per-lane alignments but just creates the merged alignment. This implementation makes it so that the alignment strategy triggers an `AlignAndMerge` command which first creates the merged speedseq alignment result and then creates per-lane alignment results for each instrument data in order to fit our current alignment paradigm. During the creation of the per-lane alignment result the header, flagstat, and md5 files are created by "fake revivifying" the per-lane bams from the merged result and creating those files from it. This ensures that we have per-lane alignment results as well as everything needed to create the per-lane bams on demand which will be needed for further analysis since our system assumes that a merged alignment results always has matching per-lane alignment results, which are used all over the place.

There are a few drawbacks to this solution, mainly that we can't have `G::ID::AR::M::Speedseq` inherit from `G::ID::AR::M`. When we call `$class->SUPER::create(@_);`  it would call the `G::ID::AR::M` create method which would fail for the speedseq case. We thus have to inherit directly from `G::ID::ABR::M' and 'G::SR::WithNestedResults' which forces us to copy a few subroutine implementations. We probably want to make a joint class for those subroutines that both `G::ID::AR::M::Speedseq` and `G::ID::AR::M` can use/inherit from (depending on implementation). I'm open to suggestions on how best to resolve this.


susannasiebert:
I pushed a commit to remove the code duplication between `G:ID::AR::Merged` and 'G::ID::AR::Merged::Speedseq` by adding a helper module.

lib/perl/Genome/InstrumentData/Composite/Strategy.pm
```diff
@@ -90,6 +92,9 @@ sub grammar {
         | "aligned" "to" reference "using" aligner
             { $return = $item[5]; $item[5]->{reference} = $item[3]; $item[5]->{type} = 'align'; }
 
+        align_and_merge: "align" "and" "merge" "to" reference "using" aligner
```
tmooney: All of the other actions in the alignment strategies are `-ed` verbs.  One option might be: `"both" "aligned" "to" reference "and" "merged" "using" aligner`.

tmooney:
Will the Speedseq results support read trimming like other alignment results?

lib/perl/Genome/Model/Tools/Speedseq/AlignBase.pm
```diff
@@ -0,0 +1,141 @@
...
+}
+
+sub _resolve_options_string {
+    my $self = shift;
+    my $params = shift;
+
+    unless ($params) {
+        $params = '';
+    }
+    if ($self->output_prefix) {
```
dufeiyu: I would set a parameter map hash like:
%hash = (o => "output_prefix", t => "threads" ...);
then loop through the hash to set parameter option (also considering boolean or not) if defined that parameter.

lib/perl/Genome/InstrumentData/Composite/Workflow/Generator.pm
```diff
@@ -53,7 +53,13 @@ sub generate {
     for my $group (keys %$objects_by_group) {
         my $alignment_objects = $objects_by_group->{$group};
 
-        my ($next_object_workflows, $next_object_inputs) = Genome::InstrumentData::Composite::Workflow::Generator::Align->generate( $tree, $input_data, $alignment_objects);
+        my ($next_object_workflows, $next_object_inputs);
+        if ($tree->{action}->[0]->{type} eq 'align_and_merge') {
```
dufeiyu: Is there any way to resolve class name camel case here ?

lib/perl/Genome/InstrumentData/Command/AlignAndMerge.pm
```diff
@@ -0,0 +1,175 @@
...
+
+    return $result;
+}
+
+sub _process_alignments {
+    my $self = shift;
+    my $mode = shift;
+
+    my %params = $self->_alignment_params;
+    my $result = Genome::InstrumentData::AlignmentResult::Merged::Speedseq->$mode(
```
dufeiyu: Hardcode speedseq here ?

lib/perl/Genome/InstrumentData/Command/AlignAndMerge.pm
```diff
@@ -0,0 +1,175 @@
...
+}
+
+sub _process_per_lane_alignments {
+    my $self = shift;
+    my $mode = shift;
+
+    my %params = $self->_alignment_params;
+    my @per_lane_results;
+    for my $instrument_data ($self->instrument_data) {
+        push @per_lane_results, Genome::InstrumentData::AlignmentResult::Speedseq->$mode(
```
dufeiyu: Hardcode speedseq ?

lib/perl/Genome/InstrumentData/AlignmentResult/Speedseq.pm
```diff
@@ -0,0 +1,101 @@
...
+    return $self->SUPER::_check_read_count($filtered_bam_rd_ct);
+}
+
+sub _promote_data {
+    my $self = shift;
+
+    Genome::Sys->copy_file($self->revivified_alignment_bam_file_path . '.flagstat', File::Spec->join($self->temp_staging_directory, 'all_sequences.bam.flagstat'));
+    Genome::Sys->copy_file($self->revivified_alignment_bam_file_path . '.md5', File::Spec->join($self->temp_staging_directory, 'all_sequences.bam.md5'));
+
+    return $self->SUPER::_promote_data;
```
dufeiyu: Here just want to make sure per lane speedseq bam does not exist in temp_staging_directory. We don't keep that per lane bam in result output dir since it can always be revivified.

lib/perl/Genome/Model/Tools/Speedseq/AlignBase.pm
susannasiebert: This comment would be better directed to #688. I pulled in that PR into this branch because my code changes depend on the speedseq wrappers.

dufeiyu: I see. Jason actually made the change on this.

lib/perl/Genome/InstrumentData/AlignmentResult/Speedseq.pm
susannasiebert: Since we are using revivification to create the per-lane bam, it should never exist in the temp_staging_directory. Do you want me to add a check to make sure that that is indeed the case?

dufeiyu: It's probably better to test bam not existing in G::I::AR::Speedseq.t.

lib/perl/Genome/InstrumentData/AlignmentResult/Speedseq.pm
```diff
@@ -0,0 +1,101 @@
...
+    my $self = shift;
+    return $self->get_merged_bam_to_revivify_per_lane_bam;
+}
+
+#Don't create flagstat during revivification - postprocess_bam_file will take
+#care of that
+sub create_bam_flagstat_and_revivify {
+    my ($self, $merged_bam, $revivified_bam) = @_;
+
+    my $cmd = Genome::InstrumentData::AlignmentResult::Command::RecreatePerLaneBam->create(
```
dufeiyu: comparison_flagstat is probably needed again after per lane speedseq AR is created and .bam.flagstat now exists in result output dir.

lib/perl/Genome/InstrumentData/AlignmentResult/Speedseq.pm
susannasiebert: Ok I added a check for the per-lane bam in the temp_staging_dir_ and output_dir in G::ID::Command::AlignAndMerge.t.

lib/perl/Genome/InstrumentData/AlignmentResult/Speedseq.pm
susannasiebert: I pushed a commit that will pass this file along to the `comparison_flagstat` param of `RecereatePerLaneBam`, if it exists.

lib/perl/Genome/InstrumentData/Command/AlignAndMerge.pm
susannasiebert: I pushed a commit to create the class names more generically from the `name` param.

lib/perl/Genome/InstrumentData/AlignmentResult/Merged/Speedseq.pm
```diff
@@ -0,0 +1,120 @@
...
+    my %params = (
+        bams => [ map {$_->bam_path} $self->instrument_data ],
+        reference_fasta => $self->reference_build->full_consensus_path('fa'),
+        output_prefix => File::Spec->join($self->temp_staging_directory, $self->id),
+        version => $self->aligner_version,
+        sort_memory => $self->sort_memory,
+        temp_directory => $temp_directory,
+    );
+    my $command = Genome::Model::Tools::Speedseq::Realign->create(%params);
+    $command->execute;
```
dufeiyu: return value check ?

dufeiyu:
+1  tests should pass

PR #727, opened 2015-05-11
Remove length restriction on fail_count.

tmooney:
The backing column's type is an `integer` itself.

(CQID is currently failing because it's trying to set a fail_count of 10,000.)


acoffman:
+1

Weird.

PR #732, opened 2015-05-12
Remove tigra-sv version 20110321

tmooney:
This version only lived in a home directory, which has been removed.

Version 0.1 is the "default" specified [in `Genome::Model::Tools::DetectVariants2::Filter::TigraValidation`](https://github.com/genome/genome/blob/d50a74cd0998c9a1618c789609ff62502119b2b1/lib/perl/Genome/Model/Tools/DetectVariants2/Filter/TigraValidation.pm#L77).


brummett:
+1

acoffman:
+1

tmooney:
Merging early to fix tests--it's not like the removed version works, anyway!

PR #734, opened 2015-05-12
adding mutect support, cleaning trailing whitespace

chrisamiller:


tmooney:
+1 :speak_no_evil: That subroutine is scary, though.

chrisamiller:
Ain't that the truth...  Maybe someday we'll have better hooks into individual mutation caller results.

PR #735, opened 2015-05-12
Search base references for CopyCat annotation data

tmooney:
If there is no annotation data specific to our reference, try looking at references from which ours is derived.


chrisamiller:
Looks great! My only suggestion would be to have it spit out a message to the log explicitly stating something like:
"No annotation data found for reference_build XXX, using annotations from reference_build XXX, from which it was derived" 
+1

chrisamiller:
+1 on the change as well.

acoffman:
+1

PR #736, opened 2015-05-12
Add wrapper for bedtools pairToPair

apregier:


lib/perl/Genome/Model/Tools/BedTools/PairToPair.pm
```diff
@@ -0,0 +1,110 @@
...
+        },
+        input_file_b => {
+            is => 'Text',
+            doc => 'The file in which to find overlaps for each line of input-file-a.',
+            shell_args_position => 2,
+        },
+        intersection_type => {
+            is => 'Text',
+            doc => 'The results to output: "a-only" returns those regions in file A but not overlapped in B; "unique" returns one line for each region in file A that is matched;"overlap_both" returns original A and B entries plus the number of base pairs of overlap between the two features. "v" means Only report those entries in A that have _no overlaps_ with B.',
+            valid_values => ['neither', 'either', 'both', 'notboth'],
```
tmooney: The documentation doesn't seem to match the `valid_values`.

apregier: Thanks, I missed that one!  Updated now

lib/perl/Genome/Model/Tools/BedTools/PairToPair.pm
```diff
@@ -0,0 +1,110 @@
+package Genome::Model::Tools::BedTools::PairToPair;
+
+use strict;
+use warnings;
+
+use Genome;
+
+my $DEFAULT_INTERSECTION_TYPE = 'both';
+
```
ebelter: Unclear to me, it is, the purpose of declaring this variable.

ebelter:
+1

lib/perl/Genome/Model/Tools/BedTools/PairToPair.pm
```diff
@@ -0,0 +1,110 @@
...
+    if ($self->slop_strand ne '+-') {
+        $options .= " -ss ".$self->slop_strand;
+    }
+    if ($self->ignore_strand) {
+        $options .= " -is";
+    }
+    if ($self->require_different_names) {
+        $options .= " -rdn";
+    }
+    my $cmd = $self->bedtools_path .'/bin/pairToPair '. $options .' -a '. $self->input_file_a .' -b '. $self->input_file_b .' > '. $self->output_file;
```
tmooney: We may want to quote the filenames.  There is `Genome::Sys->quote_for_shell`, but that should probably be updated to use `String::ShellQuote` instead of rolling its own.

tmooney:
+1 :house_with_garden:

lib/perl/Genome/Model/Tools/BedTools/PairToPair.pm
apregier: Good catch, thanks!

PR #737, opened 2015-05-12
Heartbeat check only if operation_type exists.

tmooney:
An instance without a parallel_index for parallel_by jobs doesn't point back to the operation the way the jobs with the parallel_index do.  It doesn't refer to a job that will be scheduled to LSF, so there's no need to investigate it further.

### DIFF
--- a/lib/perl/Genome/Model/Build.pm
+++ b/lib/perl/Genome/Model/Build.pm
@@ -2679,8 +2679,8 @@ sub _heartbeat {
         }
 
         # only certain operation types would have LSF jobs and everything below is inspecting LSF status
-        my $operation_type = $wf_instance_exec->operation_instance->operation->operation_type;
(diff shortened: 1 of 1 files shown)


ebelter:
+1

nnutter:
@tmooney I don't understand the body on the PR in the context of the code.  Why would an operation not have a type?

tmooney:
@nnutter The instance (`Workflow::Operation::Instance`) doesn't point to an operation (`Workflow::Operation`).  (Current master fails with `Can't call method "operation_type" on an undefined value` because of this.)  Using the indirect `operation_type` accessor on the instance allows us to get `undef` instead.

I believe this is related to the way `Flow` constructs `parallel_by` instances.  It is possible we could make a change there instead of this PR.

nnutter:
:+1: Fixing the crash is useful even if it's maybe better fixed elsewhere.

PR #738, opened 2015-05-12
Don't report idle parents of running children in heartbeat check.

tmooney:
DV2 dispatchers and other jobs that spawn workers and wait on them will spuriously report no log updates in 48 hours under `Flow`.  Skip these steps instead and report on their children's status.

### DIFF
--- a/lib/perl/Genome/Model/Build.pm
+++ b/lib/perl/Genome/Model/Build.pm
@@ -2747,6 +2747,11 @@ sub _heartbeat {
             }
         }
 
+        my @parents = grep { $_->status eq 'running' } grep { defined($_->parent_execution_id) and $_->parent_execution_id eq $wf_instance_exec_id } @wf_instances;
+        if (@parents > 0) {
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Model/Build.pm
```diff
@@ -2747,6 +2747,11 @@ sub _heartbeat {
             }
         }
 
+        my @parents = grep { $_->status eq 'running' } grep { defined($_->parent_execution_id) and $_->parent_execution_id eq $wf_instance_exec_id } @wf_instances;
+        if (@parents > 0) {
+            next WF; #this is a parent, so we wouldn't expect its output to be updated while its children are running.  Let them determine the heartbeat status.
```
nnutter: Should it check that it still has children?

nnutter: I guess it will on the next loop.

nnutter:
:+1:

acoffman:
+1

PR #739, opened 2015-05-12
Periodically report progress through the input VCF.

tmooney:
This should help reassure that progress is being made, both for manual inspection and the heartbeat check.

Is every 100,000 lines a reasonable interval at which to report?  One build I looked at had 3.6 million lines, so this would be 36 updates for that one.


brummett:
+1

acoffman:
+1

PR #740, opened 2015-05-13
fix circular dependency

nnutter:
Can't require Genome::Model::Build at top of because it creates a circular
dependency Genome::Model::Build uses Genome which uses Genome::Sys::Log which
uses Genome::Model::Build.


tmooney:
+1 :deciduous_tree:

brummett:
+1

iferguson90:
+1

PR #741, opened 2015-05-13
add valid_signals

nnutter:
An upcoming change to UR fixes a bug in the signal validation code that 
exposes cases where a class did not define `valid_signals` nor a
`validate_subscription` method.  This is one such example.


brummett:
+1

acoffman:
+1

dufeiyu:
+2  this bug is causing temp allocation not cleaned up  damn

PR #742, opened 2015-05-13
Use String::QuoteShell to do quoting for shell.

tmooney:
The require is included inside the subroutine because we don't have the module installed for Perl 5.8.  (I generally believe that it is safe to stop supporting Perl 5.8, but there's no reason to force the issue here.)


nnutter:
:+1:

brummett:
+1

PR #743, opened 2015-05-13
Evaluate bedpe

apregier:
@indraniel Would appreciate a look from you!  This is analogous to EvaluateVcf.  Not sure if we will need to tie them both into one EvaluateVcfs-like module or just add something to print out the stats for one file.


lib/perl/Genome/Model/Tools/Bedpe/EvaluateBedpe.pm
```diff
@@ -0,0 +1,81 @@
...
+        $self->common_params,
+        output_file => $output_file,
+        input_file_a => $file_a,
+        input_file_b => $file_b,
+        intersection_type => $type,
+    );
+    return Genome::Sys->line_count($output_file);
+}
+
+sub _set_derivitive_stats {
```
indraniel: This is a minor nitpick, but I'm sure you probably meant `_set_derivative_stats`.

lib/perl/Genome/Model/Tools/Bedpe/EvaluateBedpe.pm
```diff
@@ -0,0 +1,81 @@
...
+        gold_bedpe => {
+            is => 'Path',
+            doc => 'bedpe file with gold standard breakpoints'
+        },
+        bedtools_version => {
+            is => 'Text',
+        },
+        slop => {
+            is => 'Integer',
+            doc => 'The amount of slop (in b.p.). to be added to each footprint of A. *Note*: Slop is subtracted from start1 and start2 and added to end1 and end2.',
```
tmooney: It's not really clear from the interface what `A` will be.  (I see in the code that it changes depending on the stat.)

indraniel:
Does the version of `bedtools` need to be specified as an option?

indraniel:
It might be good to have an object method that will printout the stats.  It may not need to be called in the `execute` method.

lib/perl/Genome/Model/Tools/Bedpe/EvaluateBedpe.pm
apregier: Thanks - I had just copied the doc from pairToPair.  I have now tried to clarify it for this context.

tmooney:
+1 :dog2:

indraniel:
:+1:

PR #744, opened 2015-05-14
fix (possible) bugs revealed by upcoming UR change

nnutter:


nnutter:
I'm labeling this a _work in progress_ because I'm not sure this is the correct thing to do yet as I don't yet understand what caused the change in behavior.

brummett:
These changes are needed because creating an observer now always starts a software transaction.

[this change](https://github.com/genome/UR/pull/78/files#diff-2a4eac961f1cc71c7d72fbb0d4392a09R275) explains why get() needs to work in Genome::Report.

For the other two, when the first software transaction is started, it sets ```$UR::Context::Transaction::log_all_changes``` to true, but it never gets set back to false when all software transactions are done.  Since observers are created as Genome starts up, all changes get UR::Change objects created for them.  UR::Context::Transaction should probably set ```$log_all_changes``` back to false when the last software transaction gets committed or rolled back.

nnutter:
I added a commit to genome/UR#78 to disable `log_all_changes` when last transaction is rolled back or committed.  I think these Genome changes still should be merged.  I still stand by the `Genome::Report` commit message; either removing the `get()` override is acceptable or it should not be a `UR::Object`.  For the others the UR change would avoid the issue but I believe it revealed that they had a bug that.

tmooney:
+1.  Calling `get()` for `Genome::Report` doesn't really make sense as a user, but it seems fine to let the system recall an object it knows about that way.  Maybe `get()` should be noisy if someone tries to `get()` by something other than ID (sort of like `Genome::SoftwareResult` but for different reasons)?

nnutter:
@tmooney, I was thinking about that too.  I feel like it wouldn't take long for someone to figure out `get()` doesn't work.

acoffman:
+1

nnutter:
Merging earlier because this is required to get the lane-qc test passing.

PR #745, opened 2015-05-14
Allow specifying the command as a list in Genome::Sys->shellcmd

tmooney:
This skips the `bash -c` and eliminates the need to worry about properly quoting arguments!  (The old way is still available for using commands that involve pipes and the like.)

* Should whether or not to use `bash -c` when supplying an array be a separate option?

### DIFF
--- a/lib/perl/Genome/Sys.pm
+++ b/lib/perl/Genome/Sys.pm
@@ -1533,7 +1533,7 @@ sub shellcmd {
                         @cmdline = ('bash', '-c', "$shellopts_part $cmd");
                     }
 
-                    exec(@cmdline)
+                    exec { $cmdline[0] } (@cmdline)
                         or do {
                             print STDERR "Can't exec: $!\nCommand line was: ",join(' ', @cmdline),"\n";
                             exit(127);
--- a/lib/perl/Genome/Sys.pm
+++ b/lib/perl/Genome/Sys.pm
@@ -1405,6 +1405,12 @@ sub shellcmd {
 
     my ($t1,$t2,$elapsed);
 
+    my @cmdline;
+    if (ref($cmd) and ref($cmd) eq 'ARRAY') {
+        @cmdline = @$cmd;
+        $cmd = join(' ', map $self->quote_for_shell($_), @cmdline);
+    }
+
     # Go ahead and print the status message if the cmd is shortcutting
     if ($output_files and @$output_files) {
         my @found_outputs = grep { -e $_ } grep { not -p $_ } @$output_files;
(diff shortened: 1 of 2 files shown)


lib/perl/Genome/Sys.pm
```diff
@@ -1409,6 +1409,12 @@ sub shellcmd {
 
     my ($t1,$t2,$elapsed);
 
+    my @cmdline;
+    if (ref($cmd) and ref($cmd) eq 'ARRAY') {
+        @cmdline = @$cmd;
+        $cmd = join(' ', map $self->quote_for_shell($_), @cmdline));
```
nnutter: Extra paren here.

nnutter: It's sorta formal but I think you should rebase or merge `master` to get your `quote_for_shell` change in to this branch.

lib/perl/Genome/Sys.pm
```diff
@@ -1409,6 +1409,12 @@ sub shellcmd {
 
     my ($t1,$t2,$elapsed);
 
+    my @cmdline;
+    if (ref($cmd) and ref($cmd) eq 'ARRAY') {
```
nnutter: Should this `croak` if `set_pipefail` was passed in as an argument?  Hopefully it's obvious to developers that the list version doesn't support piping and so doesn't support the `set_pipefail` option?

nnutter:
I'm glad to see this!  I've been wanting to do this for awhile but had convinced myself it would be hard.

nnutter:
:+1: with the possible caveat that I think this really does depend on the improved `quote_for_shell`.

lib/perl/Genome/Sys.pm
tmooney: Hah, this will be the third time I've removed this paren.  (It came back in a previous rebase I did--doh!)  You're right, the `quote_for_shell` change is important for this.  I will rebase.

lib/perl/Genome/Sys.pm
tmooney: Seems reasonable; I've added a check for this.

tmooney:
I've changed the `exec` line to use the "indirect object" syntax to avoid possibly spawning a shell if the `@cmdline` is only a single item.

lib/perl/Genome/Sys.pm
```diff
@@ -1518,8 +1528,12 @@ sub shellcmd {
...
                     # pipefail option useless.
                     local $SIG{PIPE} = 'DEFAULT';
-                    my @cmdline = ('bash', '-c', "$shellopts_part $cmd");
-                    exec(@cmdline)
+
+                    unless (@cmdline) {
+                        @cmdline = ('bash', '-c', "$shellopts_part $cmd");
+                    }
+
+                    exec { $cmdline[0] } (@cmdline)
```
nnutter: You could also do,

```
exec shift @cmdline, @cmdline;
```

nnutter: Never mind,

```
$ perl -e 'use strict; use warnings; my @a = (q(time echo foo)); exec { $a[0] } @a;'
Can't exec "time echo foo": No such file or directory at -e line 1.

$ perl -e 'use strict; use warnings; my @a = (q(time echo foo)); exec shift @a, @a;'
foo
0.00user 0.00system 0:00.00elapsed ?%CPU (0avgtext+0avgdata 1836maxresident)k
0inputs+0outputs (0major+75minor)pagefaults 0swaps
```

nnutter: > I've changed the `exec` line to use the "indirect object" syntax to avoid possibly spawning a shell if the `@cmdline` is only a single item.

Or as the previous comment shows, falling back on splitting words and sending as shell command.

acoffman:
+1

PR #746, opened 2015-05-14
cleaning up `genome model build start` a bit

nnutter:


lib/perl/Genome/Model/Build/Command/Start.pm
```diff
@@ -137,69 +139,52 @@ sub create_and_start_build {
...
+    catch {
+        $start_transaction->rollback();
+        $self->append_error($model->__display_name__, 'Build (' . $build->id . ') ' . $_);
+        $build->status('Unstartable');
+        $build->add_note(
+            header_text => 'apipe_cron_status',
+            body_text => $_,
+            auto_truncate_body_text => 1,
+        );
+        $build->model->build_requested(0);
```
tmooney: This isn't something you changed, but I think there's still a small hole in the `build_requested` locking.  Since the commit that sets `build_requested` to true may also contain updates to the model, any build we've already `create`d might be wrong by the time the lock is acquired either inside `$build->start` or here.

tmooney:
+1 :ice_cream: The new version is much easier to follow!

lib/perl/Genome/Model/Build/Command/Start.pm
nnutter: Yep, if I address that it will be a separate PR.

PR #747, opened 2015-05-14
Make Genome::Sys->create_symlink() idempotent

brummett:
There have been several builds in rt 104974 crashing because a call to create_symlink() fails due to the link already existing, but the existing link already has the same target as is being requested, and I can't figure out where the link is being created from.

It seems safe to allow this condition to succeed.

### DIFF
--- a/lib/perl/Genome/Sys.pm
+++ b/lib/perl/Genome/Sys.pm
@@ -899,7 +899,8 @@ sub create_symlink {
 
     unless (symlink($target, $link)) {
         if ($! == Errno::EEXIST) {
-            if (readlink($link) ne $target) {
+            my $current_target = readlink($link);
+            if (defined($current_target) and $current_target ne $target) {
                 Carp::croak("Link ($link) for target ($target) already exists.");
             }
         } else {
--- a/lib/perl/Genome/Sys.pm
+++ b/lib/perl/Genome/Sys.pm
@@ -901,7 +901,7 @@ sub create_symlink {
         my $symlink_error = $!;
         if ($symlink_error == Errno::EEXIST) {
             my $current_target = readlink($link);
-            if (defined($current_target) and $current_target ne $target) {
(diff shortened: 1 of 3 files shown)


lib/perl/Genome/Sys.pm
```diff
@@ -889,26 +889,23 @@ sub gidgrnam {
...
-
-    if ( -l $link ) { # the link exists, but does not point to something
-        Carp::croak("Link ($link) for target ($target) is already a link.");
-    }
-
-    unless ( symlink($target, $link) ) {
-        Carp::croak("Can't create link ($link) to $target\: $!");
+    unless (symlink($target, $link)) {
+        if ($! == Errno::EEXIST) {
+            if (readlink($link) ne $target) {
```
tmooney: This will create a warning if `$link` exists but isn't a symlink.

nnutter: An _uninitialized value_ warning.

lib/perl/Genome/Sys.pm
```diff
@@ -889,26 +889,23 @@ sub gidgrnam {
...
-    }
-
-    if ( -l $link ) { # the link exists, but does not point to something
-        Carp::croak("Link ($link) for target ($target) is already a link.");
-    }
-
-    unless ( symlink($target, $link) ) {
-        Carp::croak("Can't create link ($link) to $target\: $!");
+    unless (symlink($target, $link)) {
+        if ($! == Errno::EEXIST) {
```
nnutter: We should probably save `$!` to a variable.

brummett:
Added new commits to address both concerns

nnutter:
:+1:

brummett:
One more commit to handle the case where it's asked to stomp over something that's not a link

nnutter:
Good catch!

brummett:
CI, test this

acoffman:
+1

PR #748, opened 2015-05-15
Small lumpy fixes

apregier:
I am hoping to run lumpy in a build - putting in some remaining small fixes to try it out.

### DIFF
--- a/lib/perl/Genome/Model/Tools/DetectVariants2/Lumpy.pm
+++ b/lib/perl/Genome/Model/Tools/DetectVariants2/Lumpy.pm
@@ -17,7 +17,12 @@ class Genome::Model::Tools::DetectVariants2::Lumpy {
             calculate_from => ['_temp_staging_directory'],
             calculate => q{ File::Spec->join($_temp_staging_directory, 'legend.tsv'); },
         },
-    ]
+    ],
+    has_param => [
+        lsf_resource => {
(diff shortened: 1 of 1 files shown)


ebelter:
+1

lib/perl/Genome/Model/Tools/DetectVariants2/Lumpy.pm
```diff
@@ -17,7 +17,12 @@ class Genome::Model::Tools::DetectVariants2::Lumpy {
             calculate_from => ['_temp_staging_directory'],
             calculate => q{ File::Spec->join($_temp_staging_directory, 'legend.tsv'); },
         },
-    ]
+    ],
+    has_param => [
+        lsf_resource => {
+            default_value => "-M 16000000 -R 'select[mem>16000] rusage[mem=16000]'",
```
tmooney: This should probably include a request for some temp space.

apregier: Yeah, my test build ran into that problem!  I am also going to investigate whether we can use StreamGraph to optimize this a little bit and not write out some of the files.  But we will still need temp space.

apregier:
New versions of lumpy are much easier to run, so I don't think it is worth the effort to get this version working.

PR #750, opened 2015-05-15
acquire build_requested lock before creating builds to start

nnutter:
@tmooney pointed out that if something else were updating the model between
when we created the build and turned off `build_requested` then we would turn
off `build_requested` prematurely.  This changes it so we disable
`build_requested` (within a transaction) prior to even creating the build so
that the `build_requested` locking prevents this race.


ebelter:
+1

tmooney:
+1 :octopus:

PR #751, opened 2015-05-16
avoid side effects if an allocation is rolled back

nnutter:
When an allocation is rolled back UR would, for example, run
`$a->archivable(0)` which since `archivable` is an overriden property would
have unintended side-effects.

### DIFF
--- a/lib/perl/Genome/Disk/Allocation-rollback.t
+++ b/lib/perl/Genome/Disk/Allocation-rollback.t
@@ -22,6 +22,6 @@ subtest 'regression test for RT 105743' => sub {
     my $a_iter = Genome::Disk::Allocation->create_iterator();
     my $a = $a_iter->next;
     ok($a, 'got an allocation');
-    ok(UR::Context->rollback, 'rolled back again to induce arhivable side-effects');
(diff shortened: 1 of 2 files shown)


lib/perl/Genome/Disk/Allocation-rollback.t
```diff
@@ -0,0 +1,27 @@
...
+    # destroyed during UR::Context->rollback().  "Random" hash ordering would
+    # sometimes cause the destruction of observers before other, dependent
+    # rollback actions ran.  In this test we rollback twice to not depend on
+    # the hash ordering.
+    plan tests => 4;
+    ok(UR::Context->rollback, 'rolled back to induce destruction of necessary observers');
+    my $a_iter = Genome::Disk::Allocation->create_iterator();
+    my $a = $a_iter->next;
+    ok($a, 'got an allocation');
+    ok(UR::Context->rollback, 'rolled back again to induce arhivable side-effects');
```
tmooney: Typo: `arhivable`.

tmooney:
+1 modulo typo :musical_keyboard:

acoffman:
+1

PR #752, opened 2015-05-16
validate values to hash

nnutter:
We had a case where no sponsor was found for a build which caused a warning
about odd number of elements in the hash and then failed downstream.

### DIFF
--- a/lib/perl/Genome/SoftwareResult/User.pm
+++ b/lib/perl/Genome/SoftwareResult/User.pm
@@ -191,10 +191,18 @@ sub _role_for_type {
 sub user_hash_for_build {
     my $class = shift;
     my $build = shift;
+    unless ($build) {
+        Carp::croak q(user_hash_for_build requires 'build' as an argument);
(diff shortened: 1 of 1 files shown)


ebelter:
+1

lib/perl/Genome/SoftwareResult/User.pm
```diff
@@ -191,10 +191,18 @@ sub _role_for_type {
 sub user_hash_for_build {
     my $class = shift;
     my $build = shift;
+    unless ($build) {
+        Carp::croak q(user_hash_for_build requires 'build' as an argument);
+    }
+
+    my $sponsor = $build->model->analysis_projects // Genome::Sys::User->get(username => $build->model->run_as);
+    unless ($sponsor) {
+        die q(unable to determine sponsor for build);
```
tmooney: It might be useful if this message included the build ID.

tmooney:
+1 :cool:

PR #753, opened 2015-05-16
fix `register_callback` call because subject is not valid for that API

nnutter:
This was not caught by testing on the PR because `GENOME_NESSY_SERVER` isn't
set on PR tests (yet).


nnutter:
Merging to fix `master`.  Please audit.

tmooney:
+1 :+1: I could successfully `get` a `UR::Context::Process` by class and ID, just to be sure.

PR #757, opened 2015-05-18
Remove alert from `genome disk group available-space` report

tmooney:
I'm migrating this to a Jenkins task, so that will be responsible for alerting if trouble arises.


nnutter:
Depending on how you want to integrate it you might want it to exit non-zero when it would have previously alerted.

tmooney:
Good idea--I've changed this to exit non-zero if any groups are low on space.

I'm currently using the Text Finder plugin marking it Unstable on low space. It'll be simpler once [this issue](https://issues.jenkins-ci.org/browse/JENKINS-23786) is resolved!

nnutter:
:+1:

acoffman:
+1

PR #760, opened 2015-05-19
Builds use processes

davidlmorton:
This makes it so that when server_dispatch != inline, the `Build` uses a `Genome::Process` to execute the workflow.  There should be no changes for server_dispatch == inline.  This is a necessary step towards full PTero integration, and should make it possible to run builds with PTero in the same way that you can run processes with Ptero.

Model tests that I ran by hand diffed clean (rnaseq, somatic-validation and somatic-variation).

One side-effect of this change is that every build will use Flow unless you explicitly set WF_USE_FLOW=0.  This is because the `Genome::Process` framework defaults to using Flow unless it is already set.

### DIFF
--- a/lib/perl/Genome/Model/Build/Command/Start.pm
+++ b/lib/perl/Genome/Model/Build/Command/Start.pm
@@ -160,7 +160,7 @@ sub create_and_start_build {
 
     my $start_transaction = UR::Context::Transaction->begin();
     my $build_started = try {
-        $build->start(%{$self->_start_params});
+        $build->start(%{$self->_start_params}) || die "Couldn't start build";
         $start_transaction->commit();
         return 1;
     }
--- a/lib/perl/Genome/Model/Build/Command/Start.pm
+++ b/lib/perl/Genome/Model/Build/Command/Start.pm
@@ -161,7 +161,7 @@ sub create_and_start_build {
     my $start_transaction = UR::Context::Transaction->begin();
     my $build_started = try {
         my $rv = $build->start(%{$self->_start_params});
-        $start_transaction->commit();
+        $start_transaction->commit() or die "Cannot commit 'build start' transaction";
         return $rv;
     }
     catch {
--- a/lib/perl/Genome/Model/Build/Command/Start.pm
+++ b/lib/perl/Genome/Model/Build/Command/Start.pm
@@ -160,9 +160,9 @@ sub create_and_start_build {
 
     my $start_transaction = UR::Context::Transaction->begin();
     my $build_started = try {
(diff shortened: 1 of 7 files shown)


tmooney:
With this change, will `$build->the_master_event->lsf_job_id` no longer be filled in?  (If it won't be, it might be worth a `CHANGELOG` entry, since that's something that shows up in the webviews.)

lib/perl/Genome/Model/Build/Command/Start.pm
```diff
@@ -160,7 +160,7 @@ sub create_and_start_build {
 
     my $start_transaction = UR::Context::Transaction->begin();
     my $build_started = try {
-        $build->start(%{$self->_start_params});
+        $build->start(%{$self->_start_params}) || die "Couldn't start build";
```
tmooney: `$build->start` currently does its own `eval` and sets its status to `Unstartable` if anything goes wrong.  Adding a `die` here will abort that process and remove the note that explains what happened.

davidlmorton:
That's correct, Tom, that field will not longer be filled in.  I'll work on a CHANGELOG... anything else that ought to be in it?

ebelter:
CI, test this. Build timed out.

davidlmorton:
I don't quite know how yet but the failing test is trying to use LSF:

***** GENOME_DEV_MODE (blade8-1-5.gsc.wustl.edu) ***** at /gscmnt/gc7008/info/jenkins/workspace/genome_genome/2070/lib/perl/Genome.pm line 30.
Workflow LSF Dispatcher issuing command: bsub -R 'select[ncpus>=1] span[hosts=1]' -n 1 -q apipe -o /tmp/tempYiwh2/info/model_data/2772828715/build6b2cfcc241da427cae25aab4d9ae4caf/logs/-92762001/-92762008.out

But that clearly won't work when ids are fake and files are in /tmp.

tmooney:
[This section](https://github.com/genome/genome/blob/b245abb1d599e4717be69fef19b6e1ab281e8432/lib/perl/Genome/Model/Build.pm#L1529-L1532) in the current master skips the `bsub` when no-commit is turned on.

davidlmorton:
I confirmed that @nnutter was correct about build->stop using the lsf_job_id on the master_event.  To support that, I added a note to processes that bsub a workflow that contains the lsf job_id.  I confirmed that you can stop a build (and it kills the workflow) by hand.

davidlmorton:
CI, test this. Build timed out.

davidlmorton:
CI, test this. Bugs that were in Master branch failed tests.

acoffman:
+1

its happening gif

ebelter:
+1

PR #761, opened 2015-05-19
Sync: LIMS organism classes names are optional

ebelter:


acoffman:
+1

nnutter:
:+1: I think this is a correct change even though I think the underlying issue is a bug in a recent change I made in UR.  A fix for the underlying bug will be developed on genome/UR#83.

ebelter:
@nnutter Agreed.

PR #764, opened 2015-05-20
Model Summary: Allow specification of max fail count

tmooney:
* Count all builds when deciding if there are too many failures.  (We regularly abandon failures preceding previous successes, so these would usually all be "newer" than the last success.)
* Allow the maximum fail count to be specified.  (I currently believe 5 might be on the high side, but that could be mitigated by improving the `model_has_progressed` check.)

### DIFF
--- a/lib/perl/Genome/Model/Command/Admin/ModelSummary.pm
+++ b/lib/perl/Genome/Model/Command/Admin/ModelSummary.pm
@@ -43,7 +43,7 @@ class Genome::Model::Command::Admin::ModelSummary {
         max_fail_count => {
             is => 'Number',
             default => 5,
-            doc => 'Rebuild will not be recommended for models that have failed this many times or more.',
+            doc => 'Rebuild will not be recommended for models that have failed this many times or more. (If undefined, any fail count will be allowed.)',
         },
     ],
 };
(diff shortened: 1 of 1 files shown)


nnutter:
:+1:

lib/perl/Genome/Model/Command/Admin/ModelSummary.pm
```diff
@@ -236,18 +244,18 @@ sub latest_build_succeeded {
...
+    return unless defined $max_fails;
 
-    my $n = (@builds >= $max_fails ? ($max_fails - 1) : $#builds);
-    my @last_n_builds = @builds[0..$n];
+    my @builds = $model->builds;
+    return unless @builds;
 
-    my @failed_builds = grep { $_->status eq 'Failed' } @last_n_builds;
-    return (@failed_builds == $max_fails );
+    my @failed_builds = grep { $_->status eq 'Failed' } @builds;
```
ebelter: Could $model->builds(status => 'Failed') be used above to eliminate this grep?

tmooney: ModelSummary currently pre-loads all the builds into memory.  After this and #765 I'd like to try to make many changes to be more careful about loading.  I'll make sure taking out this `grep` is one of those changes!

ebelter:
+1

PR #765, opened 2015-05-20
ModelSummary uses DetermineError to decide build progress

tmooney:
* Instead of doing its own workflow inspection, decide if it failed the same way twice by the error found by the `genome model build determine-error` command.
* Fix the `previous_build` finder to not find the first build instead!


acoffman:
+1

jweible:
+1

PR #767, opened 2015-05-20
Expose gmt bedpe on the command line

apregier:


lib/perl/Genome/Model/Tools/Bedpe.pm
```diff
@@ -0,0 +1,30 @@
+package Genome::Model::Tools::Bedpe;
+
+use strict;
+use warnings;
+
+use Genome;
+
+class Genome::Model::Tools::Bedpe {
+    is => ['Command'],
```
tmooney: Should this be a `Command::Tree`?

apregier: Maybe?  Seems to work as is though:
> gmt bedpe -h
Using libraries at /gscuser/aregier/git/genome/copy/lib/perl
Sub-commands for gmt bedpe:
 evaluate-bedpe  --bedpe=? --gold-...  no description!!!: define 'doc' in the class definition for
                                        Genome::Model::Tools::Bedpe::EvaluateBedpe

tmooney: Aye, `Command` is `Command::V1`.  For `Command::V2`, the concept of "parent of subcommands" was split out into `Command::Tree`.  (In `V1` we could often have something be both the base class for a tree of commands and the parent of it, but now they must be distinct.)

lib/perl/Genome/Model/Tools/Bedpe.pm
```diff
@@ -0,0 +1,30 @@
...
+
+};
+
+sub help_brief {
+    "Tools related to BEDPE files.",
+}
+
+sub help_synopsis {
+    my $self = shift;
```
dufeiyu: unneeded $self

apregier:
Pushed commits to address the comments, thanks!

dufeiyu:
+1

tmooney:
+1 :chestnut:

PR #768, opened 2015-05-21
Remove old `gmt gatk germline-snv` tool.

tmooney:
One of the required files to run this tool has been deleted from the system, so it will no longer run.  (It wasn't really ever updated when we switched away from build36 anyway.)


acoffman:
+1

Yay deletion!

PR #769, opened 2015-05-21
delete allocation rollback test

nnutter:
This test isn't very clear about what it's testing and I feel like making it
clearly would too strongly couple it to implementation details and UR tests
the actual underlying cause of this problem so I think it's fair to just
delete this test.

The motivation to delete it is that getting an iterator for all allocations
against the production database is extremely expensive.


tmooney:
+1 :light_rail:

brummett:
+1

PR #770, opened 2015-05-22
convert a couple more values to configuration

nnutter:


nnutter:
CI, test this.

davidlmorton:
Was the change from 1280 to 1024 intentional?

nnutter:
Yes, for us, it's still configured to be 1280 but I changed the default to 1024 because the original `syslog` has a 1KB limit.  Even 1024 might be too big depending on the formatting of the output but it seems less wrong than 1280 as a default.

davidlmorton:
+1

tmooney:
+1, I'd be tempted to rename the configuration variable to `max_syslog_message_length`.

PR #771, opened 2015-05-22
update broken test

nnutter:
Test was broken due to 'sys_umask' configuration value changing from
'0002' to '0007'.  Test is updated to test that write permissions are removed
not to test specific mode.


davidlmorton:
+1

PR #773, opened 2015-05-26
Unload LIMS entities outside of the transaction.

tmooney:
We don't need to bother registering 100s of thousands of "unload" records as changes inside the transaction.  Since transactions only verify changed objects, this will save tons of unnecessary work.


ebelter:
+2

PR #775, opened 2015-05-27
add support for getting multiple configuration keys

nnutter:
I am not sure if this is useful or not yet.  As documented for
`Genome::Config::get_first()` the two use cases I am anticipating are:

- `Genome::Config::get_first($my_key, $parent_key, $grandparent_key)`
- `Genome::Config::get_first($new_key, $old_key)`


nnutter:
I think I'll try something different instead.  I'll add a `default_from` to `Genome::ConfigSpec` to inherit a `default_value` from another spec.

PR #777, opened 2015-05-28
add `default_from` to config

nnutter:
This add a `default_from` field to the configuration spec allowing the default
value to be found via the specificed key.

This could allow a specific tool to use defaults for a common parent, e.g. all
the `gmt picard` commands could share a default value while allow each specific
command to be overridden in configuration.

It could also be used to help migrate configuration variables names (keys)
while coupling the `default_value`.


ebelter:
+1

acoffman:
+1

PR #779, opened 2015-05-28
Call renamed accessor in search index command.

tmooney:
The property name was changed in 9a7b89d0ce79eb135d72df0586a9faf41d654dac.


davidlmorton:
+1 a quick grep shows that there are no more references to 'log_dispatch' in the codebase.

nnutter:
+1

PR #780, opened 2015-05-28
make lsf_resource configurable

nnutter:
The goal of this is to demonstrate how we can make `lsf_resource`, etc. configurable so that we no longer need to make custom snapshots just to boost the memory request.


nnutter:
```
$ genome-env -D genome-perl -MGenome -e 'use Genome::Model::Tools::DetectVariants2::Breakdancer; print "---\n"; printf "%s\n", Genome::Model::Tools::DetectVariants2::Breakdancer->lsf_resource;'
...
---
-M 25000000 -R 'select[mem>25000] rusage[mem=25000]'
```

```
$ XGENOME_DV2_BREAKDANCER_LSF_RESOURCE=foo genome-env -D genome-perl -MGenome -e 'use Genome::Model::Tools::DetectVariants2::Breakdancer; print "---\n"; printf "%s\n", Genome::Model::Tools::DetectVariants2::Breakdancer->lsf_resource;'   
...
---
foo
```

PR #781, opened 2015-05-28
don't need to use Genome here

nnutter:
I'm trying to avoid just blindly using Genome which can trigger accidental
circular dependencies if packages such as this are used by any of the packages
that are loading during `use Genome`.


ebelter:
+2

PR #785, opened 2015-05-29
Support new versions of vep cache

apregier:
Since the release of GRCh38, we need to specify which reference version we want when downloading the vep cache.


lib/perl/Genome/Db/Ensembl/Api.pm
```diff
@@ -27,19 +27,21 @@ sub create {
     my @package_names = $self->package_names;
     #OLD CVS location
     #my $base_url = "'http://cvs.sanger.ac.uk/cgi-bin/viewvc.cgi/PACKAGENAME.tar.gz?root=ensembl&only_with_tag=branch-ensembl-VERSION&view=tar'";
-    my $base_url = "https://github.com/Ensembl/PACKAGENAME/archive/release/VERSION.zip";
+    #OLD github location
+    #my $base_url = "https://github.com/Ensembl/PACKAGENAME/archive/release/VERSION.zip";
```
tmooney: Do we need to keep these old locations around?

tmooney:
+1 :construction:

acoffman:
+1 Looks good to me!

lib/perl/Genome/Db/Ensembl/Api.pm
apregier: Good point - looks like they now have git branches for all legacy releases, so I removed this old code.  (It's always there in the history anyway!)

PR #788, opened 2015-05-29
minor update to clinseq to handle some edge cases

malachig:
Two minor things being committed:

1.) Allow users to run converge tools on a set of clin-seq build that span across multiple individuals (but only if they use the `--force` option because this should only be done carefully)

2.) When searching for existing genotype microarray models, now do this two ways.  First, obtain all genotype microarray models that are using the default_genotype_data for a sample.  Second, obtain all genotype microarray models that are associated with the sample (may or not be using the default_genotype_data).  

This is needed because occasionally the default_genotype_data for a sample is actually generated using *another* sample.  But we can't always use the first method because sometimes we have generated new genotype microarray data (and models) that we want to use, but the default_genotype_data didn't get updated...

### DIFF
--- a/lib/perl/Genome/Model/ClinSeq/Command/Converge/Base.pm
+++ b/lib/perl/Genome/Model/ClinSeq/Command/Converge/Base.pm
@@ -36,10 +36,10 @@ class Genome::Model::ClinSeq::Command::Converge::Base {
               doc => 'minimum base quality of bases in reads to be considered',
               default => '20',
         },
-        force => { 
+        allow_multiple_individuals => { 
                is => 'Boolean',
                default_value => 0,
-               doc => 'allow unusual situtations (combining clin-seq builds across multiple individuals)',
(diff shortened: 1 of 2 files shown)


acoffman:
+1

lib/perl/Genome/Model/ClinSeq/Command/Converge/Base.pm
```diff
@@ -36,6 +36,11 @@ class Genome::Model::ClinSeq::Command::Converge::Base {
               doc => 'minimum base quality of bases in reads to be considered',
               default => '20',
         },
+        force => {
```
tmooney: I would encourage using a descriptive name for what this boolean allows (e.g., `--allow-multiple-individuals`) instead of a generic `--force`.

tmooney:
+1 :airplane:

malachig:
@tmooney yeah good idea.  Added that with a minor commit.  Once tests pass I think I'll merge this one now.

PR #789, opened 2015-06-01
Fix builds use process

davidlmorton:


lib/perl/Genome/Model/Build.pm
```diff
@@ -1359,8 +1361,8 @@ sub set_build_id {
 }
 
 sub _launch {
-    my $self = shift;
-    my %params = @_;
+    my ($self, $params, $workflow_xml) = @_;
+    my %params = %$params;
```
tmooney: `%params` is mostly passed by reference below.  Those could be changed to use `$params` instead and then maybe this variable isn't necessary.  (Should the `$workflow_xml` be added as a param instead of passed separately?  Or validated some other way?)

tmooney:
+1 assuming tests pass :boat:

davidlmorton:
Merging early because these changes may fix model-tests that are currently failing on master.

PR #790, opened 2015-06-01
make search queue deduplication use less memory

nnutter:


lib/perl/Genome/Search/Queue.pm
```diff
@@ -79,4 +82,70 @@ sub default_priority {
...
+              'specifies subject_id' => sub { shift->rule->value_for('subject_id') },
+              'specifies subject_class' => sub { shift->rule->value_for('subject_class') },
+          },
+        },
+    );
+
+    my $m_iter = $set->member_iterator;
+    my $q = $m_iter->next;
+    my $duplicate_count = 0;
+    while (my $q = $m_iter->next) {
```
tmooney: It strikes me as odd that both the one to keep and the ones to delete are kept in `my $q`.

nnutter: I've pushed a commit to address this.

tmooney:
+1 :+1:

ebelter:
+1

PR #791, opened 2015-06-02
QC bugfix

susannasiebert:
This fixes a bug in the Picard CollectGcBiasMetrics qc tool where the chart_output param would default to the cwd. We're not interested in the contents of this file for qc so we just write it to a temp file.

I also did some subroutine renaming for clarity.


tmooney:
+1 :city_sunset:

acoffman:
+1

susannasiebert:
Merging early to fix bug

PR #792, opened 2015-06-02
Print the resource in the debug message, not the object hash.

tmooney:
The message will now look like:
```
DEBUG: Cleaning up lock genome/Genome::Example/1234567890abcdef1234567890abcdef...
```
rather than:
```
DEBUG: Cleaning up lock Genome::Sys::LockProxy=HASH(0x6543210)...
```

### DIFF
--- a/lib/perl/Genome/SoftwareResult.pm
+++ b/lib/perl/Genome/SoftwareResult.pm
@@ -1012,7 +1012,7 @@ sub _resolve_param_value_from_text_by_name_or_id {
 sub _release_lock_or_die {
     my ($class, $lock, $error_message) = @_;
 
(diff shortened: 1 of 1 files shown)


nnutter:
:+1:

acoffman:
+1 - does it make sense to just implement display_name for lock to be the resource name?

tmooney:
@acoffman I think a `__display_name__` for locks should include the `scope` as well.   But that might be an even better choice than making this mirror the messages in `get_with_lock`.

acoffman:
Yeah, good point - the scope probably should be in there for __display_name__. I'm fine merging this as is, or doing it that way.

PR #793, opened 2015-06-02
Get failed/unstartable builds directly.

tmooney:
Unstartable builds no longer generate Unstartable events, so this tool has been missing them entirely from its reports.

(Status is stored directly on builds these days, too, so there's no need to use the events as an intermediary.)


ebelter:
+1

acoffman:
+1

PR #794, opened 2015-06-02
Reduce Build Loading in ModelSummary

tmooney:
As promised, here's the next phase of ModelSummary revisions.  We now use sets and iterators to avoid loading most builds.

Some of the functionality has changed. Many places are more selective about which builds are considered.  (`Abandoned` builds are more consistently ignored; deciding whether a model "has progressed" now considers the previous failure (whether `Failed` or `Unstartable`)).

I'm not sure this works properly for builds with `Unknown` status...

### DIFF
--- a/lib/perl/Genome/Model/Command/Admin/ModelSummary.pm
+++ b/lib/perl/Genome/Model/Command/Admin/ModelSummary.pm
@@ -295,6 +295,8 @@ sub determine_error_for_build {
 sub find_first_nondone_step {
     my $build = shift;
 
+    return if grep { $_ eq $build->status } ('Succeeded', 'Scheduled', 'Unstartable', 'New');
+
     # The following is wrapped in a transaction and try to protect against
     # "corrupt" Workflow::Models.
     my $tx = UR::Context::Transaction->begin();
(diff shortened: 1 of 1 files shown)


ebelter:
+1

lib/perl/Genome/Model/Command/Admin/ModelSummary.pm
```diff
@@ -202,6 +184,21 @@ sub execute {
     return 1;
 }
 
+sub _build_and_status_for_model {
```
acoffman: The first half of this method looks very similar to `current_build` on `Model.pm`, except that `current_build` also checks if the inputs are up to date. I imagine doing that would slow this down way too much to use in bulk processing? `status_with_build` in `Model.pm` also seems to have some overlap with this method.

tmooney: If genome/UR#85 were merged, this would look less like both of those, as we could use `-limit` in a `get()` to just get the one we need.  `current_build` would still need an iterator.

It might make sense to change `Genome::Model->latest_build` to accept an additional filter.

acoffman:
+1

PR #795, opened 2015-06-02
Fix wrongly labeled statistic

apregier:
How embarrassing - I was looking at the results of my script and realized it didn't make sense at all - because I labeled the statistic wrong.


tmooney:
+1 :heavy_division_sign:

acoffman:
+1

PR #796, opened 2015-06-03
AlignAndMerge only creates per-lane alignment objects without revivification

susannasiebert:
Previously the AlignAndMerge step would create a merged alignment object (e.g. run Speedseq) and then use it to create the per-lane alignment objects. We would use revivification to create the per-lane bam so that we could create a .header and .flagstat file (and do some other per-lane alignment processing). As a result of having to revivifiy each per-lane bam for all the instrument data (in serial), the AlignAndMerge step would run extremely long.

This PR makes it so that the AlignAndMerge step will still run the aligner to create the merged alignment object but this step will only create the per-lane alignment objects without any further processing. .header and .flagstat file creation are postponed until the first call to get_bam_file. Calling this subroutine will run revivification anyway and we then do all the necessary post-processing at this point.

### DIFF
--- a/lib/perl/Genome/InstrumentData/AlignmentResult/Speedseq.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult/Speedseq.pm
@@ -104,6 +104,14 @@ sub _check_read_count {
     return $self->SUPER::_check_read_count($filtered_bam_rd_ct);
 }
 
+sub _create_bam_md5 {
+    return 1;
+}
+
+sub _create_bam_index {
+    return 1;
+}
+
 sub _promote_data {
     my $self = shift;
 
--- a/lib/perl/Genome/InstrumentData/AlignmentResult/Speedseq.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult/Speedseq.pm
@@ -55,15 +55,17 @@ sub _inititalize_revivified_bam {
     $self->_create_disk_allocation;
 
     $self->_prepare_working_and_staging_directories;
-    $self->_prepare_output_directory;
-    my $bam_file = $self->SUPER::get_bam_file;
-    $self->postprocess_bam_file;
-    $self->_compute_alignment_metrics();
 
     $self->debug_message("Preparing the output directory...");
     $self->debug_message("Staging disk usage is " . $self->_staging_disk_usage . " KB");
     my $output_dir = $self->output_dir || $self->_prepare_output_directory;
     $self->debug_message("Alignment output path is $output_dir");
+
+    my $bam_file = $self->SUPER::get_bam_file;
+
+    $self->postprocess_bam_file;
+
+    $self->_compute_alignment_metrics;
 
     $self->debug_message("Moving results to network disk...");
     $self->_promote_data;
--- a/lib/perl/Genome/InstrumentData/AlignmentResult/Speedseq.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult/Speedseq.pm
@@ -13,12 +13,7 @@ class Genome::InstrumentData::AlignmentResult::Speedseq {
     ],
 };
 
-sub _run_aligner {
-    my $self = shift;
-
-    #Run get_bam_file to fake revivify the per-lane bams
-    $self->_prepare_output_directory;
-    $self->get_bam_file;
+sub post_create {
     return 1;
 }
 
--- a/lib/perl/Genome/InstrumentData/AlignmentResult/Speedseq.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult/Speedseq.pm
@@ -57,10 +57,7 @@ sub _inititalize_revivified_bam {
     $self->_prepare_working_and_staging_directories;
     $self->_prepare_output_directory;
     my $bam_file = $self->SUPER::get_bam_file;
-    unless ($self->postprocess_bam_file()) {
-        $self->error_message("Postprocess BAM file failed");
-        die $self->error_message;
-    }
+    $self->postprocess_bam_file;
     $self->_compute_alignment_metrics();
 
     $self->debug_message("Preparing the output directory...");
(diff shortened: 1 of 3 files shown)


lib/perl/Genome/InstrumentData/AlignmentResult/Speedseq.pm
```diff
@@ -13,13 +13,82 @@ class Genome::InstrumentData::AlignmentResult::Speedseq {
...
+    $self->_compute_alignment_metrics();
+
+    $self->debug_message("Preparing the output directory...");
+    $self->debug_message("Staging disk usage is " . $self->_staging_disk_usage . " KB");
+    my $output_dir = $self->output_dir || $self->_prepare_output_directory;
+    $self->debug_message("Alignment output path is $output_dir");
+
+    $self->debug_message("Moving results to network disk...");
+    my $product_path;
+    unless($product_path= $self->_promote_data) {
```
tmooney: `$product_path` is unused.

lib/perl/Genome/InstrumentData/AlignmentResult/Speedseq.pm
```diff
@@ -13,13 +13,82 @@ class Genome::InstrumentData::AlignmentResult::Speedseq {
...
+    my $guard = $self->get_speedseq_bam_lock->unlock_guard();
+    unless ($self->disk_allocations) {
+        return $self->_inititalize_revivified_bam;
+    }
+    else {
+        return $self->SUPER::get_bam_file;
+    }
+}
+
+sub get_speedseq_bam_lock {
```
dufeiyu: This is the exact copy of get_bam_lock in base class.

lib/perl/Genome/InstrumentData/AlignmentResult/Speedseq.pm
```diff
@@ -13,13 +13,82 @@ class Genome::InstrumentData::AlignmentResult::Speedseq {
...
+    my $self = shift;
+
+    $self->_create_disk_allocation;
+
+    $self->debug_message("Prepare working directories...");
+    $self->_prepare_working_and_staging_directories;
+    $self->debug_message("Staging path is " . $self->temp_staging_directory);
+    $self->debug_message("Working path is " . $self->temp_scratch_directory);
+
     $self->_prepare_output_directory;
```
dufeiyu: line 77 also has $self->_prepare_output_directory;

lib/perl/Genome/InstrumentData/AlignmentResult/Speedseq.pm
tmooney: True, but `__PACKAGE__` will have a different value here!

lib/perl/Genome/InstrumentData/AlignmentResult/Speedseq.pm
```diff
@@ -13,13 +13,82 @@ class Genome::InstrumentData::AlignmentResult::Speedseq {
...
-    $self->get_bam_file;
-    return 1;
+
+    my $bam_file = $self->SUPER::get_bam_file;
+
+    $self->debug_message("Postprocessing & Sanity Checking BAM file (if necessary)...");
```
dufeiyu: Below line 66 - 89 are duplicates of lower part of create method in base class AR.pm except for commented out step description. It would be better to reduce the redundant code.

lib/perl/Genome/InstrumentData/AlignmentResult/Speedseq.pm
brummett: Would ```$self->class``` replace ```__PACKAGE__```?

tmooney: We need both locks.  This lock is to protect the initial speedseq bam production process; the other lock is for the usual revivification.

lib/perl/Genome/InstrumentData/AlignmentResult/Speedseq.pm
susannasiebert: I disagree. That portion of the code is pretty much just calls to helper methods with some added debug and error messages. I don't think that further deduplication is very useful here. Plus, the _prepare_output_directory call on line 77 will go away/be moved up to line 62 so the duplication here will be minimal.

lib/perl/Genome/InstrumentData/AlignmentResult/Speedseq.pm
dufeiyu: Do we really need "Speedseq" in lock resource since alignment result id is included there ?

lib/perl/Genome/InstrumentData/AlignmentResult/Speedseq.pm
susannasiebert: Maybe a better course of action would be to move the debug messages inside of the subroutine calls.

lib/perl/Genome/InstrumentData/AlignmentResult/Speedseq.pm
tmooney: The lock resource must be different from the one used by revivification.  Otherwise they will conflict with each other yielding a `Tried to lock resource more than once` error, since revivification is called as part of the Speedseq creation process.  If it would reduce confusion, they could be made less similar.

dufeiyu: How about in line27:
my $guard = $self->get_bam_lock(__PACKAGE__)->unlock_guard();
and change "get_bam_lock" in AR.pm to take argument.

lib/perl/Genome/InstrumentData/AlignmentResult/Speedseq.pm
susannasiebert: I pushed a commit to fix this.

lib/perl/Genome/InstrumentData/AlignmentResult/Speedseq.pm
susannasiebert: I moved debug messages to their respective subroutines, where appropriate. I also removed a bunch of error checking for subroutines that never return 0.

lib/perl/Genome/InstrumentData/AlignmentResult/Speedseq.pm
susannasiebert: I think that's a good solution. I pushed a commit for this.

susannasiebert:
Yes, I was debating between these two implementations and ultimately decided on the current one because it's more explicit.

dufeiyu:
+1

acoffman:
+1

PR #797, opened 2015-06-03
Set a default tmp space request for SomaticVariation's CreateReport step

tmooney:
A recent build failed because `/tmp` was full.  Since this step uses `/tmp`, request some!


brummett:
+1

jweible:
+1

PR #798, opened 2015-06-03
Remove second lsf_resource property in AnnotateAndUploadVariants

tmooney:
Defining the request once ought to be enough!


brummett:
+1

ebelter:
+2

PR #799, opened 2015-06-04
Ensure that _teardown_logging happens

mkiwala:
If `_setup_logging()` is called and succeeds, then the `_teardown_logging()` function must be called or the wrapper will perpetually hang while waiting for annotate-log processes to spontaneously end (they never will).

Before this PR, try/catch was being used to safely handle the command-class, and then exceptions were being re-thrown with `croak()` to give a stack trace.  It was possible for `execute()` to exit early if `_instantiate_command()` or `_run_command()` threw an exception.  The re-thrown exception would filter up past `execute()` without doing `_teardown_logging()` first.

In place of `croak()`, this PR uses `cluck()` to log a stack trace at the point of catching the exception.  Instead of re-throwing, return codes are used to communicate up the stack that a failure occurred.  `_teardown_logging()` is called whether or not the command class method succeeded, then the wrapper either succeeds or dies based on the return code value.

### DIFF
--- a/lib/perl/Genome/Ptero/Wrapper.pm
+++ b/lib/perl/Genome/Ptero/Wrapper.pm
@@ -60,11 +60,14 @@ sub execute {
     # this will get logged to the log files
     $self->_log_execution_information;
 
-    my $command = $self->_instantiate_command(decode($self->execution->inputs));
-
-    $self->_run_command($command);
+    my ($status_ok, $command) = $self->_instantiate_and_run_command;
 
     $self->_teardown_logging;
+
+    unless ($status_ok) {
+        Carp::croak sprintf("Failed to instantiate and run (%) on (%s)\n",
+            $self->method, $self->command_class);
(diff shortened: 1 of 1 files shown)


davidlmorton:
I don't really like to mix return codes and exception catching.  I think it would be more readable to just have a cleanup function that we call in any catch block.  Another option is to put a try/catch up at the top level, for everything after the setup_logging step, and in that catch block call teardown_logging.

mkiwala:
I don't like mixing return codes and exception catching paradigms either.  I don't consider this a mix of paradigms since all of the methods that yield return codes are designed to not throw exceptions.

Before creating this PR, I had considered the options you propose.  I like calling the cleanup method option better than single try/catch option.  Since we both find the first option acceptable, I'll close this PR, in favor of this alternative approach.

mkiwala:
This PR is superseded by #800

PR #800, opened 2015-06-05
_teardown_logging when re-throwing exceptions

mkiwala:
If `_setup_logging()` is called and succeeds, then the `_teardown_logging()` function must be called or the wrapper will perpetually hang while waiting for annotate-log processes to spontaneously end (they never will).

### DIFF
--- a/lib/perl/Genome/Ptero/Wrapper.pm
+++ b/lib/perl/Genome/Ptero/Wrapper.pm
@@ -12,7 +12,7 @@ use Genome::Utility::Text qw(
 use Genome::Utility::Inputs qw(encode decode);
 use Data::Dump qw(pp);
 use Ptero::Proxy::Workflow::Execution;
-use Try::Tiny qw(try catch);
+use Try::Tiny qw(try catch finally);
 
 class Genome::Ptero::Wrapper {
     is => 'Command::V2',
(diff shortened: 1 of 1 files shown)


davidlmorton:
We could 'cluck' before tearing down logging, and then croak.  It wouldn't make me sad to have the error both in the application's log files as well as on the workflow's execution data.  And since it is in a few places, a 'graceful_die' method would make sense.

mkiwala:
> Does the finally block get executed if you die in the catch block?

I tested the behavior of Perl's try/catch/finally with a little test script:
```perl
#!/usr/bin/env perl

use strict;
use warnings FATAL => 'all';

use Try::Tiny;

try {
    print "Try\n";
    die "throw" if $ARGV[0];
}
catch {
    print "Catch\n";
    die "Rethrow: $_";
}
finally {
    print "Finally\n";
};

print "Done\n";
```

When no exception is thrown, the `try` and `finally` blocks are executed, and then the code after the try/catch/finally is executed.
```
mkiwala@linus82:tmp.yCqGO52dQ3 $ perl test-try.pl 0
Try
Finally
Done
```

When no exception is thrown in the `try`, the `catch` is executed, followed by the `finally`.  The program exits without executing code after the try/catch/finally.
```
mkiwala@linus82:tmp.yCqGO52dQ3 $ perl test-try.pl 1
Try
Catch
Rethrow: throw at test-try.pl line 10.
Finally
```

tmooney:
+1.  Since the `catch` block doesn't do anything, this could also be a good candidate for a `Scope::Guard`.

mkiwala:
@tmooney thanks for pointing me at `Scope::Guard`.  @davidlmorton and I will merge for now, and consider changing to `Scope::Guard` in the future.

PR #802, opened 2015-06-08
Have speedseq per-lane alignment header only include matching instrument data's RG

susannasiebert:
Previously the per-lane alignment header would just be the same as the merged alignment header. Now it will only contain the @RG tag for the matching instrument data.

This results in certain QC tools not outputting multiple metrics for each RG group so the library name check can be removed.


susannasiebert:
This includes all the commits from #796. I will rebase once 796 has been merged.

susannasiebert:
I rebased this PR so it's now clean and ready to be reviewed.

acoffman:
+1

It was initially a little confusing to see `prepare_scratch_sam_file` get called and then the result treated like the header, but then I saw that you've overridden `scratch_sam_file_path` to just return the header. It probably indicates that these methods need better names, but we don't really want to break API compatibility with other aligners, so I guess we're good for now!

lib/perl/Genome/InstrumentData/AlignmentResult/Speedseq.pm
```diff
@@ -61,13 +63,18 @@ sub _inititalize_revivified_bam {
...
 
-#Use merged bam for header since we don't have an original per-lane bam
-sub source_bam_path_for_header {
+sub create_bam_header {
     my $self = shift;
-    return $self->get_merged_bam_to_revivify_per_lane_bam;
+
+    return $self->bam_header_path if -s $self->bam_header_path;
+
+    $self->prepare_scratch_sam_file;
```
dufeiyu: Regard to Adam's comment, an option here is to extract line 591 - 616 of AlignmentResult.pm to a new method prepare_scratch_sam_header_file, and then use $self->prepare_scratch_sam_header_file here.

susannasiebert: I like that solution. I just pushed a commit for it.

dufeiyu:
+1

PR #804, opened 2015-06-09
Get dindel working again

brummett:
The volume it used to live on was removed.  This makes it use the packaged
version.

### DIFF
--- a/lib/perl/Genome/Model/Tools/Dindel/Base.pm
+++ b/lib/perl/Genome/Model/Tools/Dindel/Base.pm
@@ -50,24 +50,34 @@ sub create_output_directory {
     }
 }
 
-sub dindel_install_dir {
-    my $self = shift;
-    return "/gscmnt/gc2146/info/medseq/dindel";
-}
-
 sub dindel_executable {
     my $self = shift;
-    return File::Spec->join($self->dindel_install_dir, "binaries", "dindel-1.01-linux-64bit");
+    return '/usr/bin/dindel-tgi1.01-wu1';
 }
 
 sub python_script {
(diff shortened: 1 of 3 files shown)


lib/perl/Genome/Model/Tools/Dindel/Base.pm
```diff
@@ -50,24 +50,34 @@ sub create_output_directory {
...
 
 sub python_script {
     my ($self, $name) = @_;
-    my $path = File::Spec->join($self->dindel_install_dir, 'dindel-1.01-python', $name . '.py');
-    if (-s $path) {
-        return $path;
-    } else {
-        die "Couldn't find Dindel python script named '$name' at: $path\n";
-    }
+$DB::single=1;
```
davidlmorton: Accidental commit?

lib/perl/Genome/Model/Tools/Dindel/Base.pm
```diff
@@ -50,24 +50,34 @@ sub create_output_directory {
...
+sub python_dir {
+    return '/usr/share/dindel-tgi1.01-wu1';
+}
+
+sub run_python_shellcmd {
+    my $self = shift;
+
+    # This is to get around Dindel's python scripts unconventional package
+    # management strategy
+    #mkdir 'utils/';
```
davidlmorton: I think this directory existing is needed for Dindel's python scripts to work, if not then we should remove the unlink below as well.

brummett:
Add commits to fix both problems

brummett:
Merging early to get tests to pass

PR #805, opened 2015-06-10
Print out diag info when the test fails

brummett:
This test is consistently failing in Jenkins, but I can't reproduce it myself.


tmooney:
+1, even though this passed tests.

brummett:
Merging early to run it with the full set of tests

ebelter:
+1

PR #806, opened 2015-06-11
Remove Doubled-Up SmallRNA Workflow Links

tmooney:
WorkflowBuilder was failing to validate these workflows, because they strangely were mapping several things into the final `result`.  This deduplicates the ones in the outer workflow and simply removes them from the inner workflow.

(This is another case where `?w=1` comes in handy viewing the diffs!)


ebelter:
+1

acoffman:
+1

PR #807, opened 2015-06-11
Only schedule current dependent reference alignments.

tmooney:
This should eliminate the issue of new microarray data triggering rebuilds of old/outdated models.

### DIFF
--- a/lib/perl/Genome/Config/AnalysisProject.pm
+++ b/lib/perl/Genome/Config/AnalysisProject.pm
@@ -5,6 +5,8 @@ use warnings;
 
 use Genome;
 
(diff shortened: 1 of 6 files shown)


acoffman:
To be sure I understand, we still want this processing to occur for analysis projects that are on hold or still pending correct?

tmooney:
@acoffman Yes, I believe so. (At least... we can't rule it out, since) those two states can both transition into "In Progress".

acoffman:
Cool, that's what I thought, I just wanted to be sure.

+1

susannasiebert:
+1

PR #808, opened 2015-06-11
EvaluateBedpe tweaks

apregier:
-a couple tweaks to the stats produced by evaluate-bedpe
-a new command, evaluate-bedpes to take in a matrix of evaluate-bedpe experiments to be performed, and write out a json file with the resulting stats


tmooney:
+1

PR #809, opened 2015-06-11
Also test for directory existence before trying to create it.

tmooney:
Unmounted volumes will not return EEXIST from mkdir().  Checking for
directory existence triggers the automounter to ensure the volume is
mounted before trying to create it.

### DIFF
--- a/lib/perl/Genome/Sys.pm
+++ b/lib/perl/Genome/Sys.pm
@@ -841,13 +841,13 @@ sub create_directory {
 sub make_path {
     my ($path) = validate_pos(@_, {type => SCALAR});
 
+    return 1 if -d $path; #This also triggers the automounter so mkdir() gets EEXIST instead of EACCES.
+
     my $gid = gidgrnam(Genome::Config::get('sys_group'));
 
     my @dirs = File::Spec->splitdir($path);
     for (my $i = 0; $i < @dirs; $i++) {
         my $subpath = File::Spec->catdir(@dirs[0..$i]);
-
-        next if -d $subpath;
 
         my $rv = mkdir $subpath;
         my $mkdir_errno = $!;
--- a/lib/perl/Genome/Sys.pm
+++ b/lib/perl/Genome/Sys.pm
@@ -847,6 +847,8 @@ sub make_path {
     for (my $i = 0; $i < @dirs; $i++) {
         my $subpath = File::Spec->catdir(@dirs[0..$i]);
 
+        next if -d $subpath;
+
         my $rv = mkdir $subpath;
         my $mkdir_errno = $!;
         if ($rv) {


tmooney:
@davidlmorton In 9e7ed41a8d1fafc7ee9c0f71d226616285d29441 you resolved a race condition that involved removing a previous `-d` check.  I'm not sure what that race condition was--do you remember (and does this inadvertently reintroduce it)?

It would be sufficient to fix the present bug by `stat`ting the potential directory before attempting to `mkdir`, but if we can it seems better to skip to the next subdirectory.

mkiwala:
> Unmounted volumes will not return EEXIST from mkdir().

@tmooney -- if a directory exists on an unmounted volume and then mkdir() is called, what is the behavior of mkdir()?  I think you are saying that mkdir() returns false and sets errno to something other than EEXIST.  Is that true?

tmooney:
Yes, the `mkdir` system call returns `-1 EACCES (Permission denied)` if the volume is not mounted.

mkiwala:
By looking at the old code, I think the race condition that @davidlmorton is referring to is the [TOCTTOU](https://en.wikipedia.org/wiki/Time_of_check_to_time_of_use) of stat()ing to decide whether to mkdir().  Unfortunately, the mkdir() system call doesn't seem to work well under autofs, as you've pointed out.

Common workarounds that I've seen are the stat()ing or chdir()ing to force automount.  I prefer stat()ing since that is thread-safe (not that we use threads).

> It would be sufficient to fix the present bug by statting the potential directory before attempting to mkdir, but if we can it seems better to skip to the next subdirectory.

I agree that it would be sufficient to fix the present bug by stating the potential directory on line 843 or so.  In fact, if the potential directory does already exist, then you could shortcut out of the entire rest of the function.  I find it unnecessary to stat() then mkdir(), since the result of mkdir() will tell you whether or not the directory previously existed and create the directory in one system call (assuming the underlying volume has already been mounted).

mkiwala:
Also, do you know whether `-d $subpath` gives a true or false result for a directory which exists on an unmounted volume?  Before 9e7ed41, we would chown every directory unless `-d $subpath` was true.  There may be a difference in the directories that we chown depending on whether the volume is mounted at the time the make_path() function is called.

tmooney:
`-d` performs a `stat()` system call, and it will produce the correct result for a directory that exists on a volume that is automounted as a result of the `-d`.

mkiwala:
> -d performs a stat() system call, and it will produce the correct result for a directory that exists on a volume that is automounted as a result of the -d.

Ok, that is good news.  I recommend moving the `next if -d` up to line 843 and adding a brief comment that its purpose is to force mounting on autofs filesystems.

tmooney:
Okie-dokie; I've pulled the `-d` outside of the loop.  How does that look?

mkiwala:
+1

acoffman:
+1

PR #810, opened 2015-06-12
Ptero build view

davidlmorton:
Since this includes a code change in Genome::Model::Build, expect almost all of the tests to be run in the quick-tests phase.

### DIFF
--- a/lib/perl/Genome/Model/Build.pm
+++ b/lib/perl/Genome/Model/Build.pm
@@ -107,6 +107,10 @@ class Genome::Model::Build {
             calculate => q( return $self->newest_workflow_instance(); ),
         },
         software_revision => { is => 'Text', len => 1000 },
+        process => {
+            is => 'Genome::Model::Build::Process',
+            reverse_as => 'build',
+        }
     ],
     has_many_optional_deprecated => [
         # the_* ?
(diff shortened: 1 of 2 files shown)


lib/perl/Genome/Model/Build.pm
```diff
@@ -1309,12 +1309,18 @@ sub _kill_job {
     return 1;
 }
 
+sub process {
```
tmooney: It might be nice for this to be a property of builds, like:

```perl
process => {
  is => 'Genome::Model::Build::Process',
  reverse_as => 'build',
}
```

Then we could let `genome model build view` work given a process ID or something similar.

lib/perl/Genome/Model/Build/Command/View.pm
```diff
@@ -94,8 +95,12 @@ sub write_report {
     }
 
     my $workflow = $self->build->newest_workflow_instance;
-    $self->_display_workflow($handle, $workflow);
-    $self->_display_logs($handle, $workflow);
+    if (defined $workflow) {
+        $self->_display_workflow($handle, $workflow);
+        $self->_display_logs($handle, $workflow);
+    } else {
+        $self->_display_ptero_workflow($handle, $self->build->process->workflow_name);
```
tmooney: `$build` is available in this scope, if you'd like to save an `->`.

mkiwala:
+1

acoffman:
+1

PR #812, opened 2015-06-12
Change property from variant_list_build to dbsnp_build in genotype model define  command.

ebelter:


malachig:
+1.  Awesome.  Thanks for the quick fix.

tmooney:
+1. :rooster:

PR #813, opened 2015-06-15
Only print an extra newline if output was not redirected.

tmooney:
Only three modules are currently using `redirect_stdout`, two of which send it to `/dev/null`, and `Genome::InstrumentData::Command::Import::WorkFlow::SraToBam` which is using it while writing a BAM file.  Seems like an extra newline isn't desirable there.

This reverts commit ba557795fe0e72ca1d4a691230206d9e39e7e80e.  (Is there a reason to keep the extra newline that I'm missing?)


ebelter:
+1

acoffman:
+1

PR #814, opened 2015-06-15
Genotype Microarray Dependent Cron RefAlign

ebelter:
Allow getting of dependent cron ref align models w/o reference first, then
filter by if the reference is compatible.


tmooney:
+1 :telescope:

PR #815, opened 2015-06-15
Remove other locking now that SR::User is ObjectWithLockedConstruction.

tmooney:
This should be merged only after no running builds are on a snapshot prior to `genome-3607`.

### DIFF
--- a/lib/perl/Genome/SoftwareResult/User.pm
+++ b/lib/perl/Genome/SoftwareResult/User.pm
@@ -116,23 +116,10 @@ sub _register_users {
         aspect => 'precommit',
         once => 1,
         callback => sub {
-            my @locks;
             for my $params (@param_sets) {
                 next if grep { $params->{$_}->isa('UR::DeletedRef') } qw(user software_result);
-                push @locks, $class->_lock_and_create_if_needed($params);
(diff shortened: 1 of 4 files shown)


lib/perl/Genome/SoftwareResult/User.pm
```diff
@@ -140,28 +127,11 @@ sub _register_users {
...
-
-    return $lock;
-}
-
-sub _load_or_create {
-    my $class = shift;
-    my $params = shift;
-
-    my $self = $class->load(%$params);
+    my $self = $class->get(%$params);
```
acoffman: Does this need to try `get()` before calling `create()`? I think the `ObjectWithLockedConstruction` mixin already attempts a `get()` before locking and creating.

tmooney: Nope, it sure doesn't.  It makes me think it would be nice to eventually unify `Genome::Utility::ObjectWithLockedConstruction` and `Genome::SoftwareResult`.  I'll just remove `get_or_create` entirely.

acoffman:
+1 (once tests pass)

ebelter:
+1

PR #816, opened 2015-06-16
New command: failed-model-ticket-status

tmooney:
Conveniently examine existing tickets from the command-line.  Assuming that the ticket body was created with the model ID at the start of each line (as in the output format of `failed-model-tickets`) and that any additions are replies and not comments, this command will list the current status of the listed models.

(This is how I've been going through and closing tickets when rebuilds from `model-summary` were successful.)

### DIFF
--- a/lib/perl/Genome/Model/Command/Admin/FailedModelTicketBase.pm
+++ b/lib/perl/Genome/Model/Command/Admin/FailedModelTicketBase.pm
@@ -1,0 +1,101 @@
+package Genome::Model::Command::Admin::FailedModelTicketBase;
+
+use strict;
+use warnings;
+
+use Error qw(:try);
+require RT::Client::REST;
+require RT::Client::REST::Ticket;
+require WWW::Mechanize;
+
+class Genome::Model::Command::Admin::FailedModelTicketBase {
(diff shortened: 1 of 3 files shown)


lib/perl/Genome/Model/Command/Admin/FailedModelTicketBase.pm
```diff
@@ -0,0 +1,96 @@
...
+    my $host = $uri->host;
+    if ($host ne 'sso.gsc.wustl.edu') {
+        return;
+    }
+
+    $mech->submit_form (
+        form_number =>  1,
+        fields =>  {
+            j_username => 'limsrt',
+            j_password => 'Koh3gaed',
```
iferguson90: Are we okay with having this available in a public repository?

iferguson90:
+1

lib/perl/Genome/Model/Command/Admin/FailedModelTicketBase.pm
ebelter: I made a JIRA issue (https://jira.gsc.wustl.edu/browse/APIPE-3223) to move the RT credentials to config.

ebelter:
+1

PR #818, opened 2015-06-16
Ensure reason for error is written to the log file

davidlmorton:
The die message in the catch block was not being written to the log
file, instead it was only visible on the PTero Execution object.

### DIFF
--- a/lib/perl/Genome/Ptero/Wrapper.pm
+++ b/lib/perl/Genome/Ptero/Wrapper.pm
@@ -66,9 +66,10 @@ sub execute {
         $self->_run_command($cmd);
         return $cmd;
     } catch {
-        Carp::cluck($_);
+        my $error = $_;
+        Carp::cluck($error);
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Ptero/Wrapper.pm
```diff
@@ -65,11 +66,13 @@ sub execute {
         $self->_run_command($cmd);
         return $cmd;
     } catch {
-        die $_;
-    } finally {
+        Carp::cluck($_);
         $self->_teardown_logging;
+        die $_;
```
ebelter: It might be prudent to capture $_ in a separate variable in case something in _teardown_logging overwrites it.

iferguson90:
+1 after `$_` is captured in a separate variable.

susannasiebert:
I agree about the `$_`. +1 after that's fixed.

PR #821, opened 2015-06-16
Add options to clinseq update-analysis to use validation data as exome data

kkrysiak:
This pull request added 2 options to clin-seq update-analysis. 1) --validation-as-exome allows users to combine exome and validation instrument data when setting up reference alignments. 2) --my-trsn allows the user to specify which instrument data (validation or exome) to use for the target-set-region-name and region-of-interest options when using --validation-as-exome.

### DIFF
--- a/lib/perl/Genome/Model/ClinSeq/Command/UpdateAnalysis.pm
+++ b/lib/perl/Genome/Model/ClinSeq/Command/UpdateAnalysis.pm
@@ -871,11 +871,12 @@ sub get_trsn{
   foreach my $instrument_data (@instrument_data){
     my $trsn = $instrument_data->target_region_set_name;
     if ($trsn){
+      #Force exome trsn when validation-as-exome-my-trsn option is used with validation-as-exome option
       if ($self->validation_as_exome && $self->validation_as_exome_my_trsn) {
         my $fl = Genome::FeatureList->get(name => $trsn);
         if ($fl->content_type eq 'exome') {
           $trsns{$trsn}=1;
-          $trsn_ref = 'SeqCap EZ Human Exome v2.0';
+          $trsn_ref = $trsn;
         }else{
           $trsns{$trsn}=1;
         }
--- a/lib/perl/Genome/Model/ClinSeq/Command/UpdateAnalysis.pm
+++ b/lib/perl/Genome/Model/ClinSeq/Command/UpdateAnalysis.pm
@@ -684,23 +684,18 @@ sub get_dna_instrument_data{
 
   foreach my $instrument_data (@sample_instrument_data){
     my $trsn = $instrument_data->target_region_set_name;
-    if ($self->validation_as_exome) {
-      if ($trsn){
-        my $fl = Genome::FeatureList->get(name => $trsn);
+    if ($trsn){
+      my $fl = Genome::FeatureList->get(name => $trsn);
+      if ($self->validation_as_exome) {
         if (not $fl or not $fl->content_type) {
           push @unknown, $instrument_data;
         }elsif ($fl->content_type eq 'exome' || $fl->content_type eq 'validation') {
           push @exome, $instrument_data;
           $trsns{$trsn}=1;
-        }else {
+        }else{
           push @other, $instrument_data;
         }
       }else{
-        push @wgs, $instrument_data;
-      }
-    }else{
-      if ($trsn){
-        my $fl = Genome::FeatureList->get(name => $trsn);
         if (not $fl or not $fl->content_type) {
           push @unknown, $instrument_data;
         }elsif ($fl->content_type eq 'exome') {
@@ -709,9 +704,9 @@ sub get_dna_instrument_data{
         }else {
           push @other, $instrument_data;
         }
-      }else{
-        push @wgs, $instrument_data;
       }
+    }else{
+      push @wgs, $instrument_data;
     }
   }
 
--- a/lib/perl/Genome/Model/ClinSeq/Command/UpdateAnalysis.pm
+++ b/lib/perl/Genome/Model/ClinSeq/Command/UpdateAnalysis.pm
@@ -842,7 +842,7 @@ sub get_trsn{
   my $trsn_ref;
   
   #When using validation-as-exome and my-trsn options, use user-defined target region set name (exome or validation)
-  if ($self->validation_as_exome && $self->my_trsn) {
+  if ($self->my_trsn) {
     my $user_trsn = $self->my_trsn;
     $self->warning_message("Using '$user_trsn' target region specified by my-trsn.");
     
@@ -857,7 +857,12 @@ sub get_trsn{
         }
       }
     }
-  
+    
+    #Checks that $trsn_ref has been initialized
+    if(!defined $trsn_ref) {
+      die $self->error_message("Cannot provide --my-trsn specified target region set name. Feature list of the target region set name does not have expected 'exome' or 'validation' content type.");
+    }
+
   #Default behavior for choosing target-region-set-name   
   }else{
     foreach my $instrument_data (@instrument_data){
--- a/lib/perl/Genome/Model/ClinSeq/Command/UpdateAnalysis.pm
+++ b/lib/perl/Genome/Model/ClinSeq/Command/UpdateAnalysis.pm
@@ -255,6 +255,14 @@ sub execute {
     $self->status_message("\nExiting...  --display-defaults mode is used for summarizing purposes only\n\n");
     return 1;
   }
+  
+  #If the user selected the --validation-as-exome-my-trsn option, make sure --validation-as-exome option is used
+  if ($self->validation_as_exome_my_trsn) {
+    unless ($self->validation_as_exome) {
+      $self->error_message("Exiting... validation_as_exome_my_trsn requires validation_as_exome option");
+      exit 1;
+    }
+  }
 
   #Make sure an individual is defined
   unless ($self->individual){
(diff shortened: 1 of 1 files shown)


kkrysiak:
Please give suggestions for ways I can improve my code.

lib/perl/Genome/Model/ClinSeq/Command/UpdateAnalysis.pm
```diff
@@ -820,12 +839,42 @@ sub get_trsn{
...
-    if ($trsn){
-      $trsns{$trsn}=1;
-      $trsn_ref = $trsn;
+  
+  #When using validation-as-exome and my-trsn options, use user-defined target region set name (1=exome or 2=validation)
+  if ($self->validation_as_exome && $self->my_trsn) {
+    my $user_trsn = $self->my_trsn;
+    
+    #Check if a valid option for my-trsn and use or warn user and exit
+    if ($user_trsn == 1){
```
acoffman: Is there a reason you're using numbers here? Depending on if you need the numeric values or not, you could let UR handle this validation for you.

Instead of 
````perl
        my_trsn => {
              is => 'Number',
              doc => 'If using validation-as-exome option, option to specify what instrument data to derive the target-set-region-name and region-of-interest from. 1=exome or 2=validation', 
        }
````

You could do something like:

````perl
    my_trsn => {
              is => 'Text',
             doc => 'If using validation-as-exome option, option to specify what instrument data to derive the target-set-region-name and region-of-interest from.', 
            valid_values => ['exome', 'validation']
        },
````

Which will constrain the values to either `exome` or `validation` for you. Then you could change your `if` block to be a little simpler:

````perl

#no need to double check that `my_trsn` is set here as you're checking it near line 260
if ($self->validation_as_exome ) {
    my $user_trsn = $self->my_trsn;
    $self->status_message("Using $user_trsn target region specified by my-trsn");
}
````

File: lib/perl/Genome/Model/ClinSeq/Command/UpdateAnalysis.pm [hunk 2]
```diff
@@ -820,12 +839,42 @@ sub get_trsn{
...
+      $user_trsn = 'exome';
+      $self->warning_message("Using '$user_trsn' target region specified by my-trsn.");
+    }elsif($user_trsn == 2){
+      $user_trsn = 'validation';
+      $self->warning_message("Using '$user_trsn' target region specified by my-trsn.");
+    }else{
+      die $self->error_message("Not a valid argument for my_trsn. 1='exome' or 2='validation'");
+    }
+    
+    #Assigns the target-region-set-name to the value defined by the user
```
susannasiebert: You might want to add a check after this foreach block in case none of the feature lists are of the correct content type which would result in `$trns_ref` not getting set at all.

lib/perl/Genome/Model/ClinSeq/Command/UpdateAnalysis.pm
kkrysiak: Great suggestions. I've made a commit with these improvements. I didn't realize I could constrain the text values in that way. Thanks!

lib/perl/Genome/Model/ClinSeq/Command/UpdateAnalysis.pm
kkrysiak: That's a good idea. I'm going to have to think about how to handle such a situation from a user perspective before implementing a fix.

lib/perl/Genome/Model/ClinSeq/Command/UpdateAnalysis.pm
```diff
@@ -820,12 +840,32 @@ sub get_trsn{
...
   my $trsn_ref;
-
-  foreach my $instrument_data (@instrument_data){
-    my $trsn = $instrument_data->target_region_set_name;
-    if ($trsn){
-      $trsns{$trsn}=1;
-      $trsn_ref = $trsn;
+  
+  #When using validation-as-exome and my-trsn options, use user-defined target region set name (exome or validation)
+  if ($self->validation_as_exome && $self->my_trsn) {
```
acoffman: I haven't tested it, so you'll likely want to double check my work, but I think now that lines 840-870 can be condensed down into something like this (getting rid of the extra `for` loop and some of the nesting):

````perl
my %trsns;
my $trsn_ref;
my $user_trsn;

if ($self->validation_as_exome && $self->my_trsn) {
    $user_trsn = $self->my_trsn;
    $self->warning_message("Using '$user_trsn' target region specified by my-trsn.");
}

for my $instrument_data (@instrument_data){
    if (my $trsn =  $instrument_data->target_region_set_name) {
        $trsns{$trsn} = 1;
        if ($user_trsn && Genome::FeatureList->get(name => $trsn, content_type => $user_trsn)) {
            $trsn_ref = $trsn;
        } else {
            $trsn_ref = $trsn;
        }
    }
}
````

A quick thought as well: this will blow up (both my implementation and the already existing one) if there is more than one feature list with the same trsn and content type. I assume that shouldn't happen? It looks like this logic exists to basically make sure that a feature list with the user specified content type and trsn exists before continuing, but we don't actually use the feature list at this point in the code, is that correct?

Comment (kkrysiak) [inline lib/perl/Genome/Model/ClinSeq/Command/UpdateAnalysis.pm]: To answer your questions:
I am also assuming that there should not be more than one feature list with the same trsn and content type. I think that would cause other problems in the pipeline and at this time, I'm not really sure I know how to fix it.

The feature list is only used in reference to the $trsn but is grabbed for each piece of instrument data for different purposes in the code.

Condencing the code:
Unfortunately, I don't think the condensation of the code can happen this way. The way it used to work is it iterated through the exome instrument data and just returned the last trsn (but also warned the user if there was more than 1 type of trsn). With my new implementation, I'm combining two different data types (validation and exome) but I only want it to assign the trsn of the data type of the user's choice. If I combine data types like this and do the old method, it will only use the last trsn read in which will not be consistent across my cohort of samples. I still want it to assign 1 to %trsns for both types of instrument data for counting purposes but only the specified trsn to the $trsn_ref that will be returned at the end of the subroutine. I think your method will still assign the trsn of the last piece of instrument data to $trsn_ref. That is, if I understand your code.

File: lib/perl/Genome/Model/ClinSeq/Command/UpdateAnalysis.pm (hunk 2 above)
Comment (kkrysiak) [inline lib/perl/Genome/Model/ClinSeq/Command/UpdateAnalysis.pm]: I believe I've addressed this in a satisfactory way now 0210f897ac1193377557504e17d25632992faafd. Not returning a trsn to the user is definitely not the behavior we want when using the --my-trsn option. Thanks!

File: lib/perl/Genome/Model/ClinSeq/Command/UpdateAnalysis.pm (hunk 3 above)
Comment (acoffman) [inline lib/perl/Genome/Model/ClinSeq/Command/UpdateAnalysis.pm]: Ah yes, you are very correct! Good catch :)

File: lib/perl/Genome/Model/ClinSeq/Command/UpdateAnalysis.pm [hunk 4]
```diff
@@ -194,6 +194,15 @@ class Genome::Model::ClinSeq::Command::UpdateAnalysis {
...
         },
+        validation_as_exome => {
+              is => 'Boolean',
+              doc => 'Treat validation data as exome data. Includes instrument data from validation capture in exome reference alignment model.',
+        },
+        my_trsn => {
+              is => 'Text',
+              doc => 'If using validation-as-exome option, option to specify what instrument data to derive the target-set-region-name and region-of-interest from. valid options: exome or validation', 
+              valid_values => ['exome','validation']
+        },
```
bainscou: Would it be better to combine `validation_as_exome` and `my_trsn` as one parameter with the option of null, exome or validation. That way the user can never call `my_trsn` without enabling the `validation_as_exome` option. @acoffman  is it possible in UR to accept valid values of nothing and something (exome, validation)? Also I think that the name 'validation as exome' is a little confusing, perhaps `add_validation`?

acoffman: Someone else may know better, but I believe the way to accomplish that would be to make it an optional parameter with `valid_values` of `exome` and `validation`. That would allow for it to be either exome, validation, or undef. It does seem to appear that this parameter is currently optional, so I think it could be used in that manner already.

kkrysiak: I combined the parameters and updated the name. It now requires the user to specify exome or validation when using this option but I think that is reasonable.

lib/perl/Genome/Model/ClinSeq/Command/UpdateAnalysis.pm
```diff
@@ -1141,7 +1175,7 @@ sub check_ref_align_models{
     next unless ($self->determine_model_data_type('-model'=>$model) eq $data_type);
 
     #If the desired $data_type is exome.  Check that the TRSN and ROI have been set correctly, exclude models that are not
-    next unless ($self->check_model_trsn_and_roi('-model'=>$model, '-data_type'=>$data_type));
+#    next unless ($self->check_model_trsn_and_roi('-model'=>$model, '-data_type'=>$data_type));
```
tmooney: Is commenting out this line intentional?  If so, this is the only place that calls `check_model_trsn_and_roi` and thus that subroutine (and this commented-out line) could be removed completely.

kkrysiak: This was not intentional, I had commented this out as a test and forgot to return it to its original state. Thanks for the catch.

tmooney:
+1

bainscou:
+1

PR #822, opened 2015-06-17
QC tool wrappers to run verify bam id

susannasiebert:
This PR implements QC tool wrappers for verify bam id as well as bedtools intersect, which is needed for capture data in order to pair down the input vcf to the target set region of the bam file.

This PR does not implement any QC config changes because it's not quite certain yet which vcf will be used for CCDG data. At this point this PR implements a WithVariationListVcf QC tool class which allows us to specify any imported variation list id in the QC config and it will get the vcf from it. Ultimately we will want to create a vcf based on the dbsnp vcf, pruned down by specific criteria (see https://jira.gsc.wustl.edu/browse/BIO-1465 for more details). Once this vcf has been created we would import into our database as a ImportedVariationList and we can the use it as described above.


lib/perl/Genome/Qc/Tool/Bedtools/Intersect.t
```diff
@@ -0,0 +1,48 @@
...
+
+use above "Genome";
+use Test::More;
+
+my $pkg = 'Genome::Qc::Tool::Bedtools::Intersect';
+use_ok($pkg);
+
+my $data_dir = __FILE__.".d";
+
+# my $output = File::Spec->join($data_dir, 'out');
```
tmooney: Is this left over from generating the expected data?

lib/perl/Genome/Qc/Tool/WithVariationListVcf.pm
```diff
@@ -0,0 +1,32 @@
...
+        variation_list_build_id => {
+            is => 'Text',
+            is_optional => 1,
+        }
+    ],
+};
+
+sub variation_list_vcf_file {
+    my $self = shift;
+    my $variation_list_build = Genome::Model::Build::ImportedVariationList->get($self->variation_list_build_list);
```
tmooney: What is the `variation_list_build_list`?  (Perhaps `variation_list_build` could be added as a property of this class with `id_by => 'variation_list_build_id'`?)

susannasiebert: Oops, that was supposed to be `variation_list_build_id`. I implemented your suggestion.

lib/perl/Genome/Qc/Tool/Bedtools/Intersect.t
susannasiebert: I removed this comment.

lib/perl/Genome/Qc/Tool/WithVariationListVcf.pm
```diff
@@ -0,0 +1,37 @@
...
+    ],
+};
+
+sub variation_list_vcf_file {
+    my $self = shift;
+
+    my $vcf_file = Genome::Sys->create_temp_file_path;
+    my $cmd = Genome::Model::GenotypeMicroarray::Command::ExtractToVcf->create(
+        variation_list_build => $self->variation_list_build,
+        sample => $self->sample,
```
tmooney: Where is `$self->sample` defined?  (Or... should this be a `Genome::Qc::Tool`?)

susannasiebert: It's a subroutine on `Genome::Qc::Tool`. `Genome::Qc::Tool::WithVariationListVcf` is supposed to work as an interface that qc tools can inherit from, in addition to inheriting from `Genome::Qc::Tool`. `Genome::Qc::Tool::WithVariationListVcf` is supposed to provide additional functionality for certain qc tool child classes. I guess I could make  `Genome::Qc::Tool::WithVariationListVcf` inherit from `Genome::Qc::Tool` and have the appropriate tools (like `Genome::Qc::Tool::Bedtools::Intersect`) inherit only from `Genome::Qc::Tool::WithVariationListVcf` and not from both `[Genome::Qc::Tool, Genome::Qc::Tool::WithVariationListVcf]`. Do you prefer that inheritance hierarchy?

tmooney: Yes; I prefer that this make explicit its dependence on `Genome::Qc::Tool` in some way.

susannasiebert: Ok, I just pushed a commit to change the hierarchy.

lib/perl/Genome/Qc/Tool/WithVariationListVcf.pm
```diff
@@ -0,0 +1,37 @@
+package Genome::Qc::Tool::WithVariationListVcf;
+
+use strict;
+use warnings;
+use Genome;
+
+class Genome::Qc::Tool::WithVariationListVcf {
+    is => 'Genome::Qc::Tool',
```
tmooney: This should probably still be abstract as well.

susannasiebert: Yes, you're right. It's now fixed.

lib/perl/Genome/Qc/Tool/Bedtools/Intersect.pm
```diff
@@ -0,0 +1,35 @@
...
+        }
+    }
+    return @cmd;
+}
+
+sub get_metrics {
+    return ();
+}
+
+sub target_region_bed_file {
```
apregier: Is this used in the actual config?  I don't see it used anywhere in this pr

susannasiebert: This PR doesn't implement the QC config changes necessary to run verify bam id (since the exact setup isn't clear yet). Bedtools intersect will be used for capture data which the `target_region_bed_file` as `input-file-b`.

apregier:
+1

lib/perl/Genome/Qc/Tool/VerifyBamId.pm
```diff
@@ -0,0 +1,59 @@
...
+}
+
+sub get_metrics {
+    my $self = shift;
+
+    my $file = $self->qc_metrics_file . '.selfSM';
+    my $reader = Genome::Utility::IO::SeparatedValueReader->create(
+        input => $file,
+        separator => '\t',
+        is_regex => 1,
```
tmooney: This needn't be `is_regex` if the separator is `"\t"` instead of `'\t'`.  (Since the current implementation uses a regular expression either way, I wonder which one is more performant.)

tmooney:
+1 :musical_score:

PR #824, opened 2015-06-17
Update speedseq to not pass number threads along as an aligner_param

susannasiebert:
Previously, we would specify `threads => 8` in the aligner params portion of the alignment strategy for speedseq. This would get passed into the speedseq wrapper and would set the `-t 8` parameter which would allow speedseq to run with 8 threads (or whatever number was specified in the alignment strategy). 

This PR makes it so that the number of threads are not being defined in the alignment strategy anymore. Instead the `-t` parameter is being set by calling `_available_cpu_count` which under the hood examines the LSF resources of the current job. 

This ensures that the `-t` parameter of speedseq is the same value as the number of cpus requested in LSF and it also makes it so that the number of threads aren't part of the lookup hash of the software result anymore. Speedseq software results will now shortcut no matter the number of threads the data was processed with.

We will need to set test names on existing speedseq software results that have threads specified in the aligner_params. We will also need to change the name of any processing profiles that use threads in the aligner params to deprecate them.

### DIFF
--- a/lib/perl/Genome/InstrumentData/AlignmentResult/Merged/Speedseq.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult/Merged/Speedseq.pm
@@ -84,6 +84,7 @@ sub create {
         temp_directory => $temp_directory,
     );
     my %aligner_params = eval($self->aligner_params);
+    $aligner_params{threads} = $self->_available_cpu_count;
     my $command = Genome::Model::Tools::Speedseq::Realign->create(%params, %aligner_params);
     unless ($command->execute) {
         die $self->error_message('Failed to execute Speedseq realign for instrument data: ' . join(', ', map {$_->id} $self->instrument_data));
--- a/lib/perl/Genome/InstrumentData/Composite/Workflow.t
+++ b/lib/perl/Genome/InstrumentData/Composite/Workflow.t
@@ -197,7 +197,7 @@ subtest 'simple align_and_merge strategy' => sub {
             force_fragment => 0,
             result_users => $result_users,
         },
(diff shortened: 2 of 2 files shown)


iferguson90:
Any reason to not just update the lookup hash of existing speedseq software results? (And if two different software results are updated to have the same lookup hash, then set a test-name on one of them…)

iferguson90:
Otherwise +1!

acoffman:
+1

susannasiebert:
Our concern is that these software results were created by builds with soon-to-be deprecated processing profiles. It seems "more correct" to invalidate results that were created with deprecated processing profiles but I'm not sure what our usual modus operandi is in this case. Simply updating the lookup hash certainly has the advantage that we wouldn't re-process things (and waste resources). @jasonwalker80 can you please make an executive decision on what to do with the software results in this case?

susannasiebert:
Also, I suspect that all build for models with the deprecated processing profiles should be abandoned.

jasonwalker80:
If there were data that had run in production that we wanted to keep around, we would want to update the lookup hashes to preserve the data.  I this case I'm fine with deprecating the processing profiles, setting test names on all speedseq results and abandoning succeeded builds.  The data used in models to-date have all been testing data sets from NA12878.  The only model that we would use again is likely the genome model test build which could be rerun.

susannasiebert:
Ok, software results that use `threads` in the  `aligner_params` have a test name set. All processing profiles that use this configuration have been renamed to <>-DEPRECATED. All models that use these processing profiles have been renamed to <>-DEPRECATED. All builds for these models have been abandoned. I will merge this PR.

PR #825, opened 2015-06-17
InstData Update Command

ebelter:
remove the only_if_null to allow values to be updated and expand the exclude list
of things that should not be easily modifiable.


acoffman:
+1

PR #826, opened 2015-06-18
Dynamic resource requirements for AlignAndMerge (Speedseq)

acoffman:
Previously, lsf resource usage for `AlignAndMerge` was hardcoded (and did not include gtmp). Now tmp usage will be calculated dynamically based on the size of the input instrument data.

Unfortunately, that necessitates the sort of kludgey approach taken here. The `lsf_resource` param on the command does not have access to the particular command instance of a given run, and thus cannot know how many instrument data are attached. Instead, we staple the lsf resource requirements in at workflow construction time (this is done in other places as well, so it isn't unprecedented in our codebase).

### DIFF
--- a/lib/perl/Genome/InstrumentData/Command/AlignAndMerge.pm
+++ b/lib/perl/Genome/InstrumentData/Command/AlignAndMerge.pm
@@ -139,9 +139,9 @@ sub _process_alignments {
     my $self = shift;
     my $mode = shift;
 
-    my $class = $self->class->merged_result_class($self->name);
+    my $merged_class = $self->class->merged_result_class($self->name);
     my %params = $self->_alignment_params;
-    my $result = $class->$mode(
+    my $result = $merged_class->$mode(
         instrument_data => [$self->instrument_data],
         %params,
     );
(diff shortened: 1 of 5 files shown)


lib/perl/Genome/InstrumentData/Composite/Workflow/Generator/AlignAndMerge.pm
```diff
@@ -72,6 +82,40 @@ sub generate {
     return $workflows, $inputs;
 }
 
+sub _get_lsf_resource_string_for_aligner_and_instrument_data {
+    my $class = shift;
+    my $aligner_name = shift;
+    my @instrument_data = @_;
+
+    my $merged_result_class = Genome::InstrumentData::Command::AlignAndMerge->merged_result_class($aligner_name);
+    my $estimated_gtmp_bytes = sum(map { $merged_result_class->estimated_gtmp_for_instrument_data($_) } @instrument_data);
```
susannasiebert: I think we should add some error checking here in case `estimated_gtmp_for_instrument_data` isn't defined on the `$merged_result_class`. It might be sufficient to just define an abstract version of `estimated_gtmp_for_instrument_data` on `G::ID::AR::Merged` and have that error out when it's not defined on the subclass.

acoffman: Good call, I'll add that!

lib/perl/Genome/InstrumentData/Command/AlignAndMerge.pm
```diff
@@ -162,7 +139,7 @@ sub _process_alignments {
     my $self = shift;
     my $mode = shift;
 
-    my $class = $self->merged_result_class;
+    my $class = $self->class->merged_result_class($self->name);
```
tmooney: The introduction of the intermediate `class` in this chain makes the variable name `$class` seem more confusing.  Perhaps it could be `$merged_class` (and later `$per_lane_class`)?

tmooney:
+1 with the promised fancier error-handling!

acoffman:
@tmooney @susannasiebert I believe I've addressed both of your comments!

susannasiebert:
Looks good. +1

PR #827, opened 2015-06-18
Minor CQID updates

ebelter:


acoffman:
+1

lib/perl/Genome/Config/Command/ConfigureQueuedInstrumentData.pm
```diff
@@ -154,9 +155,10 @@ sub _mark_sync_status {
 }
 
 sub should_skip {
-    my ($self, $inst_data) = @_;
+    my ($self, $anp_instdata_bridge) = @_;
 
-    return 'ignored flag is set on instrument data' if $inst_data->ignored;
+    my $inst_data = $anp_instdata_bridge->instrument_data;
+    return sprintf('Instrument data (%s) has ignored flag is set, skipping!', $inst_data) if $inst_data->ignored;
```
tmooney: * I would either replace the ` has` with a `:` or remove the `is` from the message.
* The value supplied for the `%s` should probably be a string rather than the `$inst_data` object itself.

PR #828, opened 2015-06-19
AnP Config for InstData: require_user_verify => 0 for instdata

ebelter:


tmooney:
I personally like that this will stop me from accidentally listing over nine thousand instrument data.  But if people find it annoying, then +1 :smile:

acoffman:
+1

susannasiebert:
Just out of curiosity, what's the reasoning behind making this change?

ebelter:
@susannasiebert The default was to require the user to select instrument data. When I use the command, I have already selected the instrument data I am curious about to give to the command. I have not merged it because I have been thinking about Tom's comment. Either way, someone will be annoyed :smile: . Do you have an opinion? [about this]

susannasiebert:
I don't have very strong feelings about this option one way or another but I do think it would be nice to limit listing thousands of instrument data if somehow possible. I'm thinking something along the lines of not requiring user verification if the instrument data was pulled by id but require verification if you use any other property. Not sure if that's possible though and if that would solve your problem.

ebelter:
I am going to close this as it is NBD to filter.

PR #829, opened 2015-06-22
More tweaks to compare-bedpes

apregier:


lib/perl/Genome/Model/Tools/Bedpe/EvaluateBedpes.pm
```diff
@@ -13,7 +13,7 @@ class Genome::Model::Tools::Bedpe::EvaluateBedpes {
         },
         config_file => {
             is => 'Path',
-            doc => 'A tsv file which contains the following columns: caller_name, gold_name, data_name, bedpe, gold_bedpe, slop',
+            doc => 'A tsv file which contains the following columns: bedpe, gold_bedpe, slop (along with any other columns of metadata you wish to include',
```
tmooney: This doc line needs a `)`!

lib/perl/Genome/Model/Tools/Bedpe/EvaluateBedpes.pm
```diff
@@ -36,14 +36,12 @@ sub execute {
...
-            caller_name => $line->{caller_name},
-            gold_name => $line->{gold_name},
-            data_name => $line->{data_name},
-            stats => $stats,
-            slop => $line->{slop},
-        };
-        push @output, $entry;
+        for my $key (keys %$stats) {
+            $line->{$key} = $stats->{$key};
+        }
```
tmooney: This loop could be replaced by `@$line{keys %$stats} = values %$stats;`.

ebelter:
+1

lib/perl/Genome/Model/Tools/Bedpe/EvaluateBedpes.pm
iferguson90: Is there a risk of `$stats` having a key that already exists in `$line`? For example, would this be necessary?
```perl
for my $key (keys %$stats) {
    die "Duplicate key $key: (descriptive message here)" if exists $line->{$key};
    $line->{$key} = $stats->{$key};
}
```

iferguson90:
See comment, otherwise +1! :ear_of_rice:

apregier:
Thanks for the comments, I made the minor changes and will merge after tests pass.

PR #832, opened 2015-06-23
RefAlign

ebelter:
add error message when getting default model name for a lane qc model w/o instdata assigned
update tests


lib/perl/Genome/Model/ReferenceAlignment.pm
```diff
@@ -260,17 +261,6 @@ sub _calculate_library_count {
     return scalar($self->libraries);
 }
 
-sub complete_build_directory {
```
tmooney: This was redundant to `last_complete_build_directory` on `ModelDeprecated` anyway!

tmooney:
+1 :seedling:

iferguson90:
+1 :gem:

PR #834, opened 2015-06-23
RefSeq Converter: add and use valid algorithms

ebelter:


lib/perl/Genome/Model/Build/ReferenceSequence/Converter.pm
```diff
@@ -5,6 +5,8 @@ use warnings;
 use strict;
 use Sys::Hostname;
 
+# README - If adding an alogrithm, please add to the list of valid algorithms
```
tmooney: `algorithm` has a transposed `o` and `g`.

lib/perl/Genome/Model/ReferenceSequence/Command/Converter/Create.pm
```diff
@@ -5,13 +5,14 @@ use warnings;
...
 
+use Genome::Model::Build::ReferenceSequence::Converter;
 
 class Genome::Model::ReferenceSequence::Command::Converter::Create {
     is => 'Command::V2',
     has_input => [
         source_reference => { is => 'Genome::Model::Build::ReferenceSequence', doc => 'the reference to convert from' },
         destination_reference => { is => 'Genome::Model::Build::ReferenceSequence', doc => 'the reference to convert to' },
-        algorithm => { is => 'Text', doc => 'method to do the conversion (valid values are methods of Genome::FeatureList::Converter)' },
+        algorithm => { is => 'Text', doc => 'method to do the conversion', valid_values => $Genome::Model::Build::ReferenceSequence::Converter::VALID_ALGORITHMS, },
```
tmooney: This could refer to `Genome::Model::Build::ReferenceSequence::Converter->__meta__->property(name => 'algorithm')->valid_values` rather than requiring a separate `our` variable.

ebelter: Yeah - I wasn't sure what was preferred. I will update.

lib/perl/Genome/Model/ReferenceSequence/Command/Converter/Create.pm
```diff
@@ -5,13 +5,14 @@ use warnings;
 
 use Genome;
 
+use Genome::Model::Build::ReferenceSequence::Converter;
```
tmooney: Now that this uses `valid_values`, it should get loaded by the autoloader without being `use`d here.

tmooney:
+1 :bookmark:

acoffman:
+1

PR #835, opened 2015-06-24
RT Auth: use config for RT login/password

ebelter:


acoffman:
+1

etc/genome/spec/rt_auth.yaml
```diff
@@ -0,0 +1,4 @@
+# The 'rt_auth' configuration variable holds the login password
+# to connect to RT @ MGI.
```
tmooney: Are these configuration variables really MGI-specific?

ebelter: Nope! I can remove that comment.

tmooney:
+1, assuming that the change to add the values to the `config.yaml` will be coming!

ebelter:
@tom  What would our team be without trust :smile:

ebelter:
The new ENVs that were implemented in puppet break the CLE validation snapshot. I have added defaults values to allow this PR to be merged.

@tmooney @acoffman Please review.

tmooney:
Could we add the new values to the `config.yaml` but *not* create `GENOME_*` environment variables (in order to not break CLE)?  Those environment variables don't do anything, since nothing in any snapshot looks for them.

ebelter:
Review welcome: https://github.com/genome/puppet-modules/pull/4

ebelter:
Puppet changes merged and applied. Merging.

PR #836, opened 2015-06-24
CHANGELOG: bump clin-seq blessed build

ebelter:


ebelter:
@gatoravi New clin-seq blessed build.

ebelter:
No tests to run, so the quick-tests is failing.

gatoravi:
+2 I assume this is because of https://github.com/genome/genome/pull/720

ebelter:
Yep. I think @jasonwalker80 wants to take a closer look at why these files were not present in the previous build.

jasonwalker80:
+1

PR #839, opened 2015-06-29
Add additional qc metrics

susannasiebert:
See Jira issue https://jira.gsc.wustl.edu/browse/APIPE-3242


lib/perl/Genome/Qc/Config.pm
```diff
@@ -45,6 +45,12 @@ sub get_commands_for_alignment_result {
                 use_version => 1.123,
             },
         },
+        samtool_flagstat => {
```
tmooney: Should this be `samtools_flagstat`?

susannasiebert: Yes, it's a typo (although this hash key doesn't have to match the tool; it can really be anything). It's now fixed.

tmooney:
+1, assuming that it would be silly for this to ever be run on something with 0 `total_reads`.

mkiwala:
+1

PR #840, opened 2015-06-30
Add exclusive flag to vaf-filter

apregier:


tmooney:
+1 :sparkler:

susannasiebert:
I like this implementation with the `inclusive` subroutine. +1

apregier:
@ernfrid This should be available once it gets into a snapshot

PR #842, opened 2015-07-01
adding option to omit the source build, since variant data can be imported

chrisamiller:
from external sources


chrisamiller:
Adding a note here - @ebelter suggested that maybe there should be a separate command for imported data, instead of making source build optional in the way I've done here. Willing to entertain feedback on the idea.

tmooney:
I'm inclined to agree--if the result is being made completely manually, it'd be useful to be able to specify all of the things that are normally derived from the source build.

ebelter:
A separate command would allow for additional information added to the result, and not confuse the current command used w/  builds.

jeldred:
we will certainly entertain future discussion on this topic and/or new PR's and/or additions to this one.

chrisamiller:
Fair enough - fixing this is still on my todo list (but obviously not very
high on it, sorry :-/ )

PR #843, opened 2015-07-02
CQID: Trust the configuration to know what instrument data to assign.

tmooney:
CQID tries to assign instrument data for models that were gotten or created by the configuration matching that instrument data.  It doesn't make sense to then not assign that instrument data to the model, so ensure the assignment always happens.


chrisamiller:
+1

ebelter:
+1

PR #844, opened 2015-07-02
Making annotation index process optional

dufeiyu:
Annotation index is not required since annotation_build is optional. The bug is detected in RT#106815. CreateAlignerIndex.pm needs some refactoring that will be done by the future PR.


ebelter:
+1

tmooney:
+1 :+1:

PR #845, opened 2015-07-02
Add new method get_superseding_results to Merged.pm

dufeiyu:
This new method will get supersede merged results for merged result. Refer to https://jira.gsc.wustl.edu/browse/APIPE-3231.

### DIFF
--- a/lib/perl/Genome/InstrumentData/AlignmentResult/Merged.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult/Merged.pm
@@ -9,6 +9,7 @@ use File::Path 'rmtree';
 use List::MoreUtils qw{ uniq };
 use List::Util qw(shuffle);
 use Try::Tiny qw(try catch finally);
+use Set::Scalar;
 
 use Genome;
 use Genome::Utility::Text; #quiet warning about deprecated use of autoload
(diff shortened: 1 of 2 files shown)


lib/perl/Genome/InstrumentData/AlignmentResult/Merged.pm
```diff
@@ -739,4 +740,46 @@ sub estimated_gtmp_for_instrument_data  {
     );
 }
 
+
+sub get_supersede_merged_alignment_results {
```
tmooney: I would call this `get_superseded_merged_alignment_results`, since it returns results that have been superseded by others rather than results that supersede others.

tmooney: Oops, I got that exactly backwards!  How about `get_superseding_merged_alignment_results`?

tmooney: Or, maybe, since this is a method on `Genome::InstrumentData::AlignmentResult::Merged`, we could get away with `get_superseding_results`?

dufeiyu: hmm, I like get_superseding_results. I will make the change.

lib/perl/Genome/InstrumentData/AlignmentResult/Merged.pm
```diff
@@ -739,4 +740,46 @@ sub estimated_gtmp_for_instrument_data  {
...
+    }
+
+    for my $supmr_id (keys %ct) {
+        next unless $ct{$supmr_id} == $ars->size;
+        next if $supmr_id eq $self->id;
+
+        my $supmr  = __PACKAGE__->get($supmr_id);
+
+        unless ($test_name_ok) {
+            next if $supmr->test_name;
```
tmooney: Both [`get_merged_alignments`](https://github.com/genome/genome/blob/be14ea7b97b765d273a0268a55b47ff00d789e5b/lib/perl/Genome/InstrumentData/AlignmentResult.pm#L1927) and [`collect_individual_alignments`](https://github.com/genome/genome/blob/be14ea7b97b765d273a0268a55b47ff00d789e5b/lib/perl/Genome/InstrumentData/AlignmentResult/Merged.pm#L359) respect the test name on the result upon which they're called, so this option (when false) is equivalent to returning immediately if `$self` has a `test_name` set.

dufeiyu:
CI, test this

dufeiyu:
https://github.com/genome/genome/pull/989 replaces this PR. Close.

PR #846, opened 2015-07-02
Set up a QC configuration management system

susannasiebert:
This PR depends on https://github.com/genome/genome-sqitch/pull/3


lib/perl/Genome/Qc/Command/Config/Add.pm
```diff
@@ -0,0 +1,75 @@
...
+    my $config_item = Genome::Qc::Config->create(
+        name => $self->name,
+        type => $self->type,
+        config => encode_json($config),
+    );
+
+    return 1;
+}
+
+sub _validate_config {
```
tmooney: Would it be useful for this validation logic to exist in `Genome::Qc::Config` itself (possibly in `__errors__`)?

lib/perl/Genome/Qc/Config.pm
```diff
@@ -3,73 +3,54 @@ package Genome::Qc::Config;
...
-    else {
-        $config{picard_collect_wgs_metrics} = {
-            class => 'Genome::Qc::Tool::Picard::CollectWgsMetrics',
-            params => {
-                input_file => 'bam_file',
-                reference_sequence => 'reference_sequence',
-                use_version => 1.123,
-            },
-        };
+    unless ($is_capture && $self->_config_types_for_capture_data->contains($self->type)) {
```
tmooney: Should this check some sort of `_config_types_for_wgs_data` when `$is_capture` is false instead of always failing?

lib/perl/Genome/Qc/Command/Config/Add.pm
susannasiebert: I moved these checks to `__errors__` like you suggested.

lib/perl/Genome/Qc/Config.pm
susannasiebert: This bug is now fixed and I added some tests.

davidlmorton:
+1 I'm assuming tests will pass with DB schema updates.

I'm not a fan of the number of different configuration schemes we've got in Genome at the moment.  We have Processing Profiles, Analysis Project Config, numerous different strategies (DV2, Alignment, Fusion Detection, ect), and Variant Reporting Plans already.  I wouldn't block this for that reason, I do think we want to sanitize these though, it is confusing for users (i.e. error prone) and a lot of maintenance burden.

ebelter:
+1

lib/perl/Genome/Config/Parser/YAML.pm
```diff
@@ -19,4 +20,11 @@ sub parse {
     return LoadFile($filename);
 }
 
-1;
\ No newline at end of file
+sub write {
+    my ($self, $filename, $data) = validate_pos(@_, 1, 1, 1);
+    die('Failed to provide filename!') unless $filename;
+    die('Failed to provide content!') unless $data;
```
tmooney: If we're going to use `validate_pos`, I think it makes sense to consolidate the other parameter validation into it.

susannasiebert: Using `validate_pos` I can definitely remove lines 25-26 and line 16. Do you suggest to also consolidate the remaining tests in the `parse` subroutine using `validate` instead of `validate_pos`?

tmooney: I think `validate_pos` would still work, e.g.:

```perl
validate_pos(@_, 1, { callbacks => { 'file has size' => sub { -s $_[0]; }}});
```

susannasiebert: Thanks for the pointer. I just pushed a commit that consolidates those checks into `validate_pos`.

susannasiebert:
@davidlmorton @ebelter I just pushed another commit to add a QC config lister. Could you please review the commit to make sure that your +1 still stands?

ebelter:
+1, save my sort position comment.

susannasiebert:
Good point. I'm not even sure that `sub_command_sort_position` does anything. I removed it from this module.

tmooney:
The required migration for this PR has been applied; it'll need to be reconciled with the current master.

susannasiebert:
The previous sqitch changes failed to set permissions correctly. PR https://github.com/genome/genome-sqitch/pull/4 needs to be migrated before this PR can be merged.

tmooney:
It would be most delightful would the CI test this as it sees fit!

tmooney:
+1.

It might be nice to standardize the capitalisation in the documentation; I see all of `QC`, `qc`, and `Qc`.  (`qC` is right out :smiley:)

PR #848, opened 2015-07-02
Typofix: aliger => aligner

tmooney:
Do we still support/use the `Star` alignment module?


jasonwalker80:
The Star alignment module should still be supported.  Ideally we would switch to Star instead of TopHat for RNA-seq alignment.  However, there are some compatibility issues to address with existing RNA-seq QC and there were noticeable, unexplained changes in Cufflinks run times.  The Griffith lab is working on a project called regtools that would replace the existing RNA-seq splice junction QC and generalize it for any input BAM file.

ebelter:
+2 for the simple update.

PR #849, opened 2015-07-06
Remove unwanted points from LOH plot

gatoravi:
This pull-request removes individual LOH points from the plots which are a bit confusing and don't add a lot of information. The tumor VAF for these points are still plotted, along with the identified segments which should provide the information we need.


ebelter:
+1

mkiwala:
+1

PR #850, opened 2015-07-06
Extract DV2 dispatcher resource requirements to a configuration variable.

tmooney:


ebelter:
+1

lib/perl/Genome/Model/Tools/DetectVariants2/Dispatcher.pm
```diff
@@ -102,6 +102,9 @@ class Genome::Model::Tools::DetectVariants2::Dispatcher {
         lsf_queue => {
             default_value => Genome::Config::get('lsf_queue_dv2_workflow'),
         },
+        lsf_resource => {
+            default => Genome::Config::get('lsf_resource_dv2_dispatcher'),
```
acoffman: What is the difference between the `default` here and the `default_value` immediately above it? Anything?

ebelter: These are synonyms: default, value and default_value. The default and value both normalize to default_value.

acoffman: Gotcha. I might advocate for keeping the label consistent within the same class header, but I don't feel strongly either way.

acoffman:
+1

lib/perl/Genome/Model/Tools/DetectVariants2/Dispatcher.pm
tmooney: I've changed the line that said `default_value` to match the other five that are easily seen in this PR. :smile:

PR #851, opened 2015-07-06
Library Duplication Rates from Reference Alignment builds

jasonwalker80:
Initial version of a new command to dump library duplication rates from the MarkDuplicates metrics file for ReferenceAlignment builds.


lib/perl/Genome/Model/ReferenceAlignment/Command/LibraryDuplicationRates.pm
```diff
@@ -0,0 +1,44 @@
...
+        my @instrument_data = $build->instrument_data;
+        for my $library_name (keys %{$library_metrics}) {
+            my $data = $library_metrics->{$library_name};
+            my @library_instrument_data = grep {$_->library_name eq $library_name} @instrument_data;
+            $data->{BUILD_ID} = $build->id;
+            $data->{MODEL_ID} = $build->model->id;
+            $data->{SUBJECT_NAME} = $build->model->subject->name;
+            $data->{COUNT_INSTRUMENT_DATA} = scalar(@library_instrument_data);
+            unless ($writer) {
+                my @headers = keys %{$data};
```
tmooney: This will not always be the same order across executions.

lib/perl/Genome/Model/ReferenceAlignment/Command/LibraryDuplicationRates.pm
```diff
@@ -0,0 +1,44 @@
...
+                $writer = Genome::Utility::IO::SeparatedValueWriter->create(
+                    separator => "\t",
+                    headers => \@headers,
+                );
+            }
+            $writer->write_one($data);
+        }
+    }
+    $writer->output->close;
+    exit;
```
tmooney: This `exit` probably doesn't belong.

lib/perl/Genome/Model/ReferenceAlignment/Command/LibraryDuplicationRates.pm
```diff
@@ -0,0 +1,44 @@
...
+}
```
tmooney: Modules customarily end with a `1;` on the very last line.

tmooney:
Does this command eliminate the need for `Genome::Model::Tools::TechD::PicardDuplicationRates` to exist?

jasonwalker80:
exit; removed, oops

1; added, oops

help updated to describe what this command is intended to do.  I did not anticipate needing a sort order. I figure most users will be dumping this to a spreadsheet where they will sort by a desired field.

Also, I removed the old command that had similar functionality.

jasonwalker80:
Basic sort applied to output headers.

lib/perl/Genome/Model/ReferenceAlignment/Command/LibraryDuplicationRates.pm
```diff
@@ -0,0 +1,63 @@
...
+    return <<"EOS"
+All reference alignment builds currently use Picard MarkDuplicates to flag PCR duplicates after merging the individual instrument data alignments.
+One output from Picard MarkDuplicates is a metrics file that provides a per library breakdown of PCR and optical duplicate reads.  This tool simply parses the metrics file and dumps those metrics along with basic build/model information in tab-delimited form.
+EOS
+}
+
+sub execute {
+    my $self = shift;
+
+    my $writer;
```
ebelter: Would it be better to store a hash for each build, the at the end, uniq the keys of the data for the headers, open the writer, and write the entries? This eliminates the need to check if the writer exists, and would prevent the output method from possibly being called on the undefined writer below.

```perl
# @data = { BUILD_ID => 1, ... } ...
if ( not @data ) { # necessary?
    $self->warning_message('No mark dup found...');
    return 1; # command ran OK
}
my @headers = List::MoreUtils::uniq(sort map { keys %$_ } @data);
my $writer = Genome::Utility::IO::SeparatedValueWriter->create(
    separator => "\t",
    headers => \@headers,
);
for ( @data ) { $writer->write_one($_); }
```

lib/perl/Genome/Model/ReferenceAlignment/Command/LibraryDuplicationRates.pm
```diff
@@ -0,0 +1,63 @@
...
+    for my $build ($self->builds) {
+        my $library_metrics = $build->mark_duplicates_library_metrics_hash_ref;
+        my @instrument_data = $build->instrument_data;
+        for my $library_name (keys %{$library_metrics}) {
+            my $data = $library_metrics->{$library_name};
+            my @library_instrument_data = grep {$_->library_name eq $library_name} @instrument_data;
+            $data->{BUILD_ID} = $build->id;
+            $data->{MODEL_ID} = $build->model->id;
+            $data->{SUBJECT_NAME} = $build->model->subject->name;
+            $data->{COUNT_INSTRUMENT_DATA} = scalar(@library_instrument_data);
```
ebelter: The scalar function is unnecessary. Calling an array in scalar context returns the number.

lib/perl/Genome/Model/ReferenceAlignment/Command/LibraryDuplicationRates.pm
```diff
@@ -0,0 +1,63 @@
...
+One output from Picard MarkDuplicates is a metrics file that provides a per library breakdown of PCR and optical duplicate reads.  This tool simply parses the metrics file and dumps those metrics along with basic build/model information in tab-delimited form.
+EOS
+}
+
+sub execute {
+    my $self = shift;
+
+    my $writer;
+    for my $build ($self->builds) {
+        my $library_metrics = $build->mark_duplicates_library_metrics_hash_ref;
```
ebelter: Just an FYI - Calling this function causes the metrics to be set on the build unless they were set previously. It also does not check the metrics first, where it could return the values in the DB. If the mark dup files do not exist, the values could still be in the database, but would not be returned from this function.

lib/perl/Genome/Model/ReferenceAlignment/Command/LibraryDuplicationRates.pm
```diff
@@ -0,0 +1,63 @@
...
+        my @instrument_data = $build->instrument_data;
+        for my $library_name (keys %{$library_metrics}) {
```
ebelter: What if the mark dup files do not exist, and $library_metrics is undef?

ebelter:
This looks a lot like https://github.com/genome/genome/pull/853.

jasonwalker80:
I pushed a new version that uses DB metrics.

jasonwalker80:
@ebelter and @tmooney , I have updated this command to use the DB metrics rather than parsing the files.  Any comments or suggestions?

lib/perl/Genome/Model/ReferenceAlignment/Command/LibraryDuplicationRates.pm
```diff
@@ -0,0 +1,86 @@
...
+    'A command to print Picard MarkDuplicates duplication metrics for a reference alignment build.'
+}
+
+sub help_synopsis {
+    'Dump reference alignment build duplication metrics to standard out in tab-delimited format.'
+}
+
+sub help_detail{
+    return <<"EOS"
+All reference alignment builds currently use Picard MarkDuplicates to flag PCR duplicates after merging the individual instrument data alignments.
```
tmooney: While in practice most reference alignment builds do this, I do not believe MarkDuplicates is required by the pipeline.

lib/perl/Genome/Model/ReferenceAlignment/Command/LibraryDuplicationRates.pm
```diff
@@ -0,0 +1,86 @@
...
+One output from Picard MarkDuplicates is a metrics file that provides a per library breakdown of PCR and optical duplicate reads.  This tool simply parses the metrics file and dumps those metrics along with basic build/model information in tab-delimited form.
```
tmooney: Does this tool ever fall back on parsing the file, or does it now rely solely on the database-stored metrics?

lib/perl/Genome/Model/ReferenceAlignment/Command/LibraryDuplicationRates.pm
```diff
@@ -0,0 +1,86 @@
...
+        my @instrument_data = $build->instrument_data;
+        my @library_names = List::MoreUtils::uniq(sort map { $_->library->name } @instrument_data);
+
+        for my $library_name (@library_names) {
+            my @library_instrument_data = grep {$_->library_name eq $library_name} @instrument_data;
+            my $data = {
+                'BUILD_ID' => $build->id,
+                'MODEL_ID' => $build->model->id,
+                'SUBJECT_NAME' => $build->model->subject->name,
+                'COUNT_INSTRUMENT_DATA' => scalar(@library_instrument_data),
```
tmooney: If `@library_instrument_data` is used only to get this count, we could do something like this:
```perl
my %library_counts;
for my $instrument_data (@instrument_data) {
  $library_counts{$instrument_data->library_name}++;
}
```
and then not need to `grep` through the instrument data again each time through the library name loop.

lib/perl/Genome/Model/ReferenceAlignment/Command/LibraryDuplicationRates.pm
jasonwalker80: Now it solely relies on the db metrics.  Are you suggesting it should have support for both?  If the db metrics do NOT exist, call another subroutine that calls the old hash method?

lib/perl/Genome/Model/ReferenceAlignment/Command/LibraryDuplicationRates.pm
jasonwalker80: If duplicate marking was performed, it was by Picard MarkDuplicates.  Samblaster and SpeedSeq are alternatives for duplicate marking, but are only currently available in SomVal pipelines.

lib/perl/Genome/Model/ReferenceAlignment/Command/LibraryDuplicationRates.pm
tmooney: Nope, I'm not suggesting it should have support for both.  The documentation claims this tool `simply parses the metrics file`, so I was asking if that happens.  Since it doesn't, I'm suggesting that the text be changed.

lib/perl/Genome/Model/ReferenceAlignment/Command/LibraryDuplicationRates.pm
tmooney: True, if duplicate marking was performed, it was by Picard MarkDuplicates.  But not "[a]ll reference alignment builds currently use Picard MarkDuplicates to flag PCR duplicates".  Some instead do not use anything to flag PCR duplicates.

lib/perl/Genome/Model/ReferenceAlignment/Command/LibraryDuplicationRates.pm
ebelter: With the refcov metrics, if the refcov step is shortcut, the metrics are not in the DB. I am not sure if that is how the dup metrics work. The hashref method on the build should return the DB metrics (if the exist), then fall back to the file.

lib/perl/Genome/Model/ReferenceAlignment/Command/LibraryDuplicationRates.pm
```diff
@@ -0,0 +1,86 @@
...
+                $library_metric->name =~ /^$regex/;
+                my $key = $1;
+                my $value = $library_metric->value;
+                $data->{$key} = $value;
+            }
+            push @data, $data;
+        }
+    }
+    unless (@data) {
+        $self->warning_message('MarkDuplicates metric files were not found for builds!');
```
tmooney: This warning message also refers to files rather than the metrics themselves.

jasonwalker80:
I see your point now.  The db metrics are not stored on all builds.  I added both methods for resolving the metrics, file-based and build db based.  I also added your suggestion to pre-calculate the count of instrument data per library.

lib/perl/Genome/Model/ReferenceAlignment/Command/LibraryDuplicationRates.pm
jasonwalker80: I made a few updates.

For now, the command will first look for db metrics and parse the files if needed using `mark_duplicates_library_metrics_hash_ref`.  It would be great if `mark_duplicates_library_metrics_hash_ref` returned the same result regardless of source, db or file based.  That seems like a warranted improvement, but this command should continue to work as-is.  If that is implemented, this command could be simplified to use the common interface for retrieving duplication metrics.

jasonwalker80:
@tmooney  or @ebelter additional comments or review?

ebelter:
+1 @jasonwalker80  How do you feel about merging this with the ref cov metrics? Maybe add a metrics sub command?

tmooney:
+1 :coffee:

jasonwalker80:
@ebelter I would fully support refactoring to a metrics sub command.  I wonder how this overlaps with the need to dump QC metrics.  I know @apregier was asking if a tool exists to dump QC metrics to a tsv file.  I will merge this PR and assume future work would make it in a followup PR.

PR #852, opened 2015-07-06
Don't worry about design_set reference in SomaticValidation

tmooney:
The design_set isn't used in the pipeline--it is the source of the target_region_set (whose reference we already ignore).


ebelter:
+1

acoffman:
+1

PR #853, opened 2015-07-06
RefCov Metrics

ebelter:
This is a command to get metrics from ref-align models. It gets the metrics directly from the merged result, so model metrics do not need to exist in the database. It also allows for metric gathering for large model groups when the view crashes.


lib/perl/Genome/Model/ReferenceAlignment/Command/RefCovMetrics.pm
```diff
@@ -0,0 +1,124 @@
...
+            default_value => '-',
+            doc => 'Path to send output.',
+        },
+    },
+};
+
+sub help_brief { 'get ref cov metrics from models' }
+sub help_detail {
+    return <<HELP;
+This command will get reference coverage from a model's succeeded builds. It goes directly to the refcof software result, and does not look at the model/build metrics.
```
tmooney: typo: `refcof`

lib/perl/Genome/Model/ReferenceAlignment/Command/RefCovMetrics.pm
```diff
@@ -0,0 +1,124 @@
...
+
+use Genome;
+
+require List::MoreUtils;
+
+class Genome::Model::ReferenceAlignment::Command::RefCovMetrics {
+    is => 'Command::V2',
+    has_input => {
+        models => {
+            is => 'Genome::Model::ReferenceAlignment',
```
tmooney: Should we make this operate on any `Genome::Model`, go through `$build->results` to find ref-cov results (of which there may be two for `SomaticValidation`, and call the accessors on the results directly?

ebelter: @tmooney  That is a possibility. To do so:
* move the command to a different namespace (genome instdata alignment-result merged?)
* pick a different row identifier for each result (it is currently model_name)

Additionally we could allow different 'get' methods: instdata, models, result id). What do yo think?

tmooney: That's even more generic than I was thinking of making this, but that'd be fine by me.  The object that was used to do the query could be listed in the row, but the result's ID would have to be the primary identifier.

It's enough to me if the command works on both reference alignments and somatic validation, since those are the primary two model types used to look at coverage.  (Or, alternatively, there's a second command in the `genome model somatic-validation` namespace that does the same thing as this for those models.)

ebelter:
@tmooney  I added the ability to use som-val models, plus specify what 'row ids' should be included. Please review.

tmooney:
+1, except for the previously noted typo!

acoffman:
+1

PR #856, opened 2015-07-07
GP Eukaryotic: Make fasta_files is_many in SplitFasta

tmooney:
`WorkflowBuilder` cares about the difference between an `ARRAY` and an `is_many` property as the input to a `parallelBy` workflow.  (This is the second and hopefully final part of fixing gene-prediction eukaryotic.)


ebelter:
+1

tmooney:
I rebased the commit to address a conflict between the `fasta_file` input and the `fasta_files` output.  (This makes the two properties a lot less confusing, anyway!)

acoffman:
+1

PR #857, opened 2015-07-08
Allow QC framework to work on merged alignment results

susannasiebert:
This change is needed to support running verifyBamId on the merged alignment result. The QC framework continues to support running QC tools on the per-lane alignment results.

### DIFF
--- a/lib/perl/Genome/Qc/Tool.pm
+++ b/lib/perl/Genome/Qc/Tool.pm
@@ -65,8 +65,12 @@ sub sample {
     my $self = shift;
 
     my @samples = uniq map {$_->sample} $self->alignment_result->instrument_data;
-    die "More than one sample" if scalar(@samples) > 1;
-    return $samples[0];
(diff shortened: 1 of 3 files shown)


acoffman:
+1

lib/perl/Genome/Qc/Tool.pm
```diff
@@ -65,8 +65,12 @@ sub sample {
     my $self = shift;
 
     my @samples = uniq map {$_->sample} $self->alignment_result->instrument_data;
-    die "More than one sample" if scalar(@samples) > 1;
-    return $samples[0];
+    if (scalar(@samples) > 1) {
+        die $self->error_message("More than one sample: %s", join(', ', map {$_->name} @samples));
```
tmooney: Would it be useful for this error message to include the Alignment Result ID as well?

tmooney:
+1 :grapes:

PR #859, opened 2015-07-08
Project purge command

brummett:
This adds a command for purging/expunging software results given one or more analysis projects.

It's based on jwalker's basic script, converted to a genome model command, and added locking around each anp to prevent multiple instances from stomping on each other.

### DIFF
--- a/lib/perl/Genome/Model/Command/Admin/PurgeSoftwareResultsFromAnalysisProject.pm
+++ b/lib/perl/Genome/Model/Command/Admin/PurgeSoftwareResultsFromAnalysisProject.pm
@@ -136,11 +136,6 @@ sub purge_one_analysis_project {
 
     my $unlocker = $self->_get_lock_for_analysis_project($anp);
 
-    my $sth = $dbh->prepare($sql);
-    die $self->error_message("preparing SQL failed : $DBI::errstr") unless $sth;
-
-    $sth->execute($anp->id);
-
     my $total_kb_purged = 0;
     my $software_result_count = 0;
 
@@ -150,6 +145,11 @@ sub purge_one_analysis_project {
                                 int($total_kb_purged / KB_IN_ONE_GB),
                                 $software_result_count);
     });
+
+    my $sth = $dbh->prepare($sql);
+    die $self->error_message("preparing SQL failed : $DBI::errstr") unless $sth;
+
+    $sth->execute($anp->id);
 
     while (my $data = $sth->fetchrow_hashref()) {
         my $sr = Genome::SoftwareResult->get($data->{result_id});
--- a/lib/perl/Genome/Model/Command/Admin/PurgeSoftwareResultsFromAnalysisProject.pm
+++ b/lib/perl/Genome/Model/Command/Admin/PurgeSoftwareResultsFromAnalysisProject.pm
@@ -30,7 +30,8 @@ class Genome::Model::Command::Admin::PurgeSoftwareResultsFromAnalysisProject {
 };
 
 my $sql = qq(
-    SELECT * FROM (
+    SELECT candidates.result_id, candidates.profile_item_id, candidates.anp_name, candidates.anp_id, candidates.kilobytes_requested
+    FROM (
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Model/Command/Admin/PurgeSoftwareResultsFromAnalysisProject.pm
```diff
@@ -0,0 +1,171 @@
...
+
+        $software_result_count++;
+        $total_kb_purged += $data->{kilobytes_requested};
+
+        if ($self->dry_run) {
+            $self->warning_message('Dry run, not removing software result '.$sr->id);
+        } else {
+            $self->status_message($reason);
+            $sr->expunge($reason);
+            UR::Context->commit();
```
davidlmorton: Do we want some error-checking around this `Context->commit`?

lib/perl/Genome/Model/Command/Admin/PurgeSoftwareResultsFromAnalysisProject.pm
```diff
@@ -0,0 +1,171 @@
...
+        }
+    }
+}
```
ebelter: Need a '1;'

lib/perl/Genome/Model/Command/Admin/PurgeSoftwareResultsFromAnalysisProject.pm
```diff
@@ -0,0 +1,171 @@
...
+    my $unlock_and_print_report = Scope::Guard->new(sub {
+        $unlocker->();
+        $self->status_message('Removed %d GB from %d software results',
+                                int($total_kb_purged / KB_IN_ONE_GB),
+                                $software_result_count);
+    });
+
+    my $sth = $dbh->prepare($sql);
+    die $self->error_message("preparing SQL failed : $DBI::errstr") unless $sth;
+
```
ebelter: Maybe prepare the SQL statement handle outside the purging AnP method? And then reuse the $sth to execute?

lib/perl/Genome/Model/Command/Admin/PurgeSoftwareResultsFromAnalysisProject.pm
ebelter: Maybe each AnP purge should be in a transaction?

lib/perl/Genome/Model/Command/Admin/PurgeSoftwareResultsFromAnalysisProject.pm
```diff
@@ -0,0 +1,171 @@
...
+        my $sr = Genome::SoftwareResult->get($data->{result_id});
+        die $self->error_message('Cannot get software result '. $data->{result_id}) unless $sr;
+
+        my $reason = 'Expunge software result uniquely used by model from disabled config item ('. $data->{profile_item_id} .') for analysis project \''. $data->{anp_name} .'\' ('. $data->{anp_id} .')';
+
+        $software_result_count++;
+        $total_kb_purged += $data->{kilobytes_requested};
+
+        if ($self->dry_run) {
+            $self->warning_message('Dry run, not removing software result '.$sr->id);
```
ebelter: What about having the same message with prepended with 'Would remove' and Removing', like 'git clean'? That would you can see what would happen on a 'wet_run' :smile_cat:

ebelter: I suppose it isn't as simple to just set UR_DBI_NO_COMMIT=1 for the dry_run.

brummett: No, ```expunge()``` (and the things it calls) don't respect UR_DBI_NO_COMMIT before removing files.

brummett:
Added new commits to die if the commit() fails, and unifying the messages printed while running

brummett:
Added new commits to extract the SQL prepare, and to only purge data if --dry-run is off

ebelter:
+1

jeldred:
+1 :shipit:

PR #861, opened 2015-07-09
Command to Deprecate Analysis Projects

jasonwalker80:
New analysis project status Depcrecated and a new Deprecate command. Support for Deprecated status added to existing status update commands.

### DIFF
--- a/lib/perl/Genome/Config/AnalysisProject/Command/Deprecate.pm
+++ b/lib/perl/Genome/Config/AnalysisProject/Command/Deprecate.pm
@@ -1,0 +1,55 @@
+package Genome::Config::AnalysisProject::Command::Deprecate;
+
+use strict;
+use warnings;
+
+use Genome;
+
+class Genome::Config::AnalysisProject::Command::Deprecate {
+    is => 'Command::V2',
+    has_input => [
+       analysis_projects  => {
+            is                  => 'Genome::Config::AnalysisProject',
+            doc                 => 'the analysis projects to deprecate',
+            is_many             => 1,
+            shell_args_position => 1,
+        }
+    ],
+};
+
+sub help_brief {
+    return 'deprecate the analysis project';
+}
+
+sub help_synopsis {
+    return "genome config analysis-project deprecate <analysis-projects>";
+}
+
+sub help_detail {
+    return <<"EOS"
+Given some analysis projects, this will deprecate the analysis project.  This includes setting all config profile items to disabled and updating the project status to Deprecated.
+EOS
+}
+
+sub execute {
+    my $self = shift;
+
+    for my $anp ($self->analysis_projects) {
+        my @active_config_profile_items = grep {$_->status eq 'active'} $anp->config_items;
+        for my $config_profile_item (@active_config_profile_items) {
+            my $cmd = Genome::Config::AnalysisProject::Command::DisableConfigFile->create(
+                profile_item => $config_profile_item,
+            );
+            unless ( $cmd->execute() ) {
+                $self->error_message('Failed to disable config profile item: '. $config_profile_item->id);
+                die($self->error_message);
+            }
+        }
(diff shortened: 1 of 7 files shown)


lib/perl/Genome/Config/AnalysisProject/Command/Deprecate.pm
```diff
@@ -0,0 +1,55 @@
...
+    return 'deprecate the analysis project';
+}
+
+sub help_synopsis {
+    return "genome config analysis-project deprecate <analysis-projects>";
+}
+
+sub help_detail {
+    return <<"EOS"
+Given some analysis projects, this will deprecate the analysis project.  This includes setting all config profile items to disabled and updating the project status to Deprecated.
```
tmooney: There is a disagreement in the number of analysis projects in the first sentence.

ebelter:
+1

lib/perl/Genome/Config/AnalysisProject/Command/Release.pm
```diff
@@ -46,7 +46,7 @@ sub __errors__ {
     my $self = shift;
     my @errors = $self->SUPER::__errors__(@_);
     for my $status (map{$_->status} $self->analysis_projects){
-        unless(grep{$_ eq $status} ("Pending", "Hold", "In Progress")){
+        unless(grep{$_ eq $status} ("Pending", "Hold", "In Progress","Deprecated")){
```
tmooney: To verify: we want to **allow** setting the status back to `Hold` or `In Progress` from `Deprecated`?

Either way, we may want to put a similar status check on the new command to prevent deprecating AnPs that are in the various "final" statuses (like `Purged` or `Template`).

jasonwalker80:
Good catch Tom, I removed Deprecated from the list of allowed statuses for Hold and Release.  Deprecated should be a final status along with Archived, Purged, Template, etc.

susannasiebert:
The capitalization of the `valid_values` for `status` is inconsistent between `Genome::Config::AnalysisProject` and `Genome::Config::Profile::Item`. It seems like in general we tend to use all-lowercase in our code base. +1 either way.

tmooney:
I'd still like to see this command check the status of the AnPs it's operating on to make sure they're not already `Purged` or other statuses that wouldn't make sense to deprecate.

jasonwalker80:
Updated version.  I modified the command to use Base class to reuse the status validation logic.

@susannasiebert you are correct about the inconsistency in capitalization.

lib/perl/Genome/Config/AnalysisProject/Command/Deprecate.t
```diff
@@ -0,0 +1,40 @@
...
+
+my $profile_item = Genome::Config::Profile::Item->create(
+    analysis_project => $ap,
+    status => 'active',
+);  
+
+my $cmd = $class->create(analysis_project => $ap);
+ok($cmd->execute(), 'succesful command execution');
+
+is($ap->status, 'Deprecated', 'it should set the status to hold');
```
tmooney: Perhaps `to Deprecated` instead of `to hold`.

tmooney:
Since #807 has been merged, we'll want to add `Deprecated` to the excluded list for `Genome::Config::AnalysisProject->is_current`.  (I can do this as a separate PR if you like.)

+1 aside from that and my nitpicking of test descriptions!

lib/perl/Genome/Config/AnalysisProject/Command/Deprecate.t
jasonwalker80: Correction made.  copy/paste error

jasonwalker80:
I can't seem to find the code for `Genome::Config::AnalysisProject->is_current`.  That will have to make it in as a separate PR.  Merging after tests pass.

PR #865, opened 2015-07-14
Deduplicate Aligner and Annotation Index Classes

tmooney:
AnnotationIndex was copy+pasted from AlignerIndex and the two had started to drift.  This reconciles most of the differences and eliminates most of the duplication between them.

### DIFF
--- a/lib/perl/Genome/Model/Build/ReferenceSequence/AnnotationIndex.pm
+++ b/lib/perl/Genome/Model/Build/ReferenceSequence/AnnotationIndex.pm
@@ -125,29 +125,11 @@ sub create {
 sub generate_dependencies_as_needed {
     my $self = shift;
 
-    my %params = (
-        aligner_name => $self->aligner_name,
-        aligner_params => $self->aligner_params,
-        aligner_version => $self->aligner_version,
-    );
-
     # if the reference is a compound reference
     if ($self->reference_build->append_to and $self->_supports_multiple_reference) {
         die('Compound references are not currently supported in '. __PACKAGE__);
-        for my $b ($self->reference_build->append_to) { # (append_to is_many)
-            $params{reference_build} = $b;
-            $self->debug_message("Creating AlignmentIndex for build dependency " . $b->name);
-            my $result = Genome::Model::Build::ReferenceSequence::AlignerIndex->get_or_create(%params);
-            unless($result) {
-                $self->error_message("Failed to create AlignmentIndex for dependency " . $b->name);
-                return;
-            }
-            unless ($result->generate_dependencies_as_needed()) {
-                $self->error_message("Failed while checking dependencies of " . $b->name);
-                return;
-            }
-        }
     }
+
     return 1;
 }
 
--- a/lib/perl/Genome/Model/Build/ReferenceSequence/AnnotationIndex.pm
+++ b/lib/perl/Genome/Model/Build/ReferenceSequence/AnnotationIndex.pm
@@ -41,50 +41,9 @@ sub __display_name__ {
                );
 }
 
-sub get {
-    my $class = shift;
-
-    my @objects;
-    if (@_ % 2 == 0) {
-        my %p = @_;
-        unless ($p{test_name}) {
-            $p{test_name} = (Genome::Config::get('aligner_index_test_name') || undef);
-        }
-        if (exists $p{aligner_name} && $class->aligner_requires_param_masking($p{aligner_name})) {
-            $p{aligner_params} = undef;
-        }
-        @objects = $class->SUPER::get(%p);
-    } else {
-        @objects = $class->SUPER::get(@_);
-    }
-
-    return unless @objects;
-
-    for my $obj (@objects) {
-        next unless ref($obj); # sometimes UR gives us back the package name when deleting?
-
-        unless ($obj->generate_dependencies_as_needed()) {
-            $obj->error_message("Failed to get AnnotationIndex objects for dependencies of " . $obj->__display_name__);
-            return;
-        }
-    }
-
-    if (@objects > 1) {
-        return @objects if wantarray;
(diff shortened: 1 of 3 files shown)


ebelter:
+1

lib/perl/Genome/Model/Build/ReferenceSequence/IndexBase.pm
```diff
@@ -0,0 +1,277 @@
...
+            is_optional => 1
+        },
+        aligner => {
+            calculate_from => [qw/aligner_name aligner_version aligner_params/],
+            calculate => q|no warnings; "$aligner_name $aligner_version $aligner_params"|
+        },
+    ],
+    has_input => [
+        reference_build_id  => {
+            is => 'Number',
```
apregier: Probably should be Text

lib/perl/Genome/Model/Build/ReferenceSequence/IndexBase.pm
```diff
@@ -0,0 +1,277 @@
...
+        $self->error_message("Failed to prepare reference index!");
+        return;
+    }
+
+    unless ($self->_promote_prepared_index) {
+        $self->error_message("Failed to save reference index to output location!");
+        return;
+    }
+
+    unless ($self->generate_dependencies_as_needed($self->_user_data_for_nested_results)) {
```
apregier: Should this be added as an abstract method with a Carp::confess if it is not defined in the subclass?

lib/perl/Genome/Model/Build/ReferenceSequence/AnnotationIndex.pm
```diff
@@ -79,227 +41,33 @@ sub __display_name__ {
...
-    unless ($self->check_dependencies()) {
-        $self->error_message("Failed to create AnnotationIndex objects for dependencies");
-        return;
-    }
-
-    return $self;
-}
-
-sub check_dependencies {
+sub generate_dependencies_as_needed {
```
apregier: So the AnnotationIndex doesn't actually need to generate any dependencies?

lib/perl/Genome/Model/Build/ReferenceSequence/AnnotationIndex.pm
```diff
@@ -79,227 +41,33 @@ sub __display_name__ {
...
-        $reference_fasta_file = $self->reference_build->full_consensus_path('fa');
-    }
-
-    unless (-s $reference_fasta_file) {
-        $self->error_message(sprintf("Reference fasta file %s does not exist", $reference_fasta_file));
-        return;
-    }
-
-    $self->debug_message(sprintf("Confirmed non-zero reference fasta file is %s", $reference_fasta_file));
+    my $reference_fasta_file = shift;
```
apregier: $reference_fasta_file is not actually used in this method, although it is part of the interface.  Is it better to shift it explicitly since it is part of the interface, or ignore it since it isn't used?

lib/perl/Genome/Model/Build/ReferenceSequence/IndexBase.pm
```diff
@@ -0,0 +1,277 @@
...
+        $p->{params}{test_name} = (Genome::Config::get('aligner_index_test_name') || undef);
+    }
+    if (exists $p->{params}{aligner_name} && $class->aligner_requires_param_masking($p->{params}{aligner_name})) {
+        $p->{params}{aligner_params} = undef;
+    }
+
+    return $p;
+}
+
+sub _modify_params_for_lookup_hash {
```
apregier: This wasn't in AnnotationIndex before, but it seems more correct to have it (just checking!)

lib/perl/Genome/Model/Build/ReferenceSequence/IndexBase.pm
```diff
@@ -0,0 +1,277 @@
...
+        return @objects if wantarray;
+        my @ids = map { $_->id } @objects;
+        die "Multiple matches for $class but get or create was called in scalar context! Found ids: @ids";
+    }
+    else {
+        return $objects[0];
+    }
+}
+
+sub _gather_params_for_get_or_create {
```
apregier: This wasn't in AnnotationIndex before, but it seems ok for it to be in there now (just checking again!)

lib/perl/Genome/Model/Build/ReferenceSequence/AnnotationIndex.pm
tmooney: The "dependency" thing was added for some particular aligner index, and none of our annotation indices work with it.  Until we have one that does, there's nothing to do...except maybe crash! :smile:   Perhaps this version of `generate_dependencies_as_needed` should go in the base class (in lieu of your `Carp::confess` suggestion)?

lib/perl/Genome/Model/Build/ReferenceSequence/IndexBase.pm
tmooney: Aye--updated!

lib/perl/Genome/Model/Build/ReferenceSequence/AnnotationIndex.pm
tmooney: I think I prefer listing it.

Maybe we don't need to bother doing the `resolve_reference_fasta_file` in the base class?  It preserves the old behaviour of checking it, but it is a bit silly if we're not using it.

lib/perl/Genome/Model/Build/ReferenceSequence/IndexBase.pm
tmooney: Yep, these are important!

This is the better way to accomplish what was previously being done in `get` and `create`.  The aligner index used to do it that way as well, but was changed previously.

apregier:
+1

PR #867, opened 2015-07-15
Inactive analysis project config profile item status

jasonwalker80:
Add inactive status for analysis project config profile items.  The inactive status is used for the store-only option when adding config.


tmooney:
A few suggested additions to this PR are in jasonwalker80/genome#1.

acoffman:
+1

brummett:
+1

PR #869, opened 2015-07-17
Speedseq sv

Origamigirl11:
This is the wrapper and test for Speedseq Sv detection command.


apipe-tester:

apipe-tester:

apregier:
I'm going to take a look at this first - I'll update again when we're ready for wider review!

lib/perl/Genome/Model/Tools/Speedseq/Sv.pm
```diff
@@ -0,0 +1,136 @@
...
+			is_optional => 1,
+                        tool_param_name => 'm',
+                },
+		trim_threshold => {
+                        is => 'Float',
+                        doc => 'trim threshold [0]',
+                        is_optional => 1,
+                        tool_param_name => 'r',
+                },
+		temp_directory => {
```
apregier: By default, we should create a temp directory using Genome::Sys->create_temp_directory()
I think we can do this in a calculated property.

lib/perl/Genome/Model/Tools/Speedseq/Sv.t
```diff
@@ -0,0 +1,80 @@
...
+   genotype_svtyper => 1,
+   discordant_read_bam_file => $discordant_bam,
+);
+isa_ok($sv_cmd3,$pkg);
+ok($sv_cmd3->execute,'execute command '. $pkg);
+
+#VCF Difference
+my $output_file3 = "$output_prefix3.sv.vcf.gz";
+
+my $expected_output_file3 = ('/gscmnt/gc2801/analytics/mfulton/genome2/lib/perl/Genome/Model/Tools/Speedseq/Sv.t.out/Sv.t.out3.sv.vcf.gz');
```
apregier: We'll need to remove this hard-coded path - I think you can use $expected_output_dir

lib/perl/Genome/Model/Tools/Speedseq/Sv.t
```diff
@@ -0,0 +1,80 @@
...
+compare_ok("$output_file.tbi", "$expected_output_file.tbi");
+
+#~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#~~~~~~~Third Test CNVnator Check~~~~~~~
+#~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+my $output_prefix3 = Genome::Sys->create_temp_directory() .'/example.CNV';
+#my $output_prefix3 = ('/gscmnt/gc2801/analytics/mfulton/genome/lib/perl/Genome/Model/Tools/Speedseq/Sv.t.out/Sv.t.out3');
+# Do not use the same temp directory for output.  speedseq cleans up the temp directory.
+my $expected_output_prefix3 = ('/gscmnt/gc2801/analytics/mfulton/genome2/lib/perl/Genome/Model/Tools/Speedseq/Sv.t.out/Sv.t.out3');
```
apregier: Another hard-coded path that needs to be removed

lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.pm
```diff
@@ -0,0 +1,132 @@
...
+    ]
+};
+
+	my $temp_directory = Genome::Sys->create_temp_directory();
+	my $pkg = 'Genome::Model::Tools::Speedseq::Sv';
+
+# For the Parameters what I need to do is make two hashes.
+# The first one will have the name of the command in the SV class and the letter calling it.  
+# The second Hash will have the class string name and the value wil be either a boolean or a file location.
+	my %ref_library = (
```
apregier: I think we may be able to get these from the __meta__ objects of the command class, since the parameters have the property "tool_param_name"

lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.pm
```diff
@@ -0,0 +1,132 @@
...
+sub find_split{
+        my $self = shift;
+        my @bam_dir = @_;
+	my @final = ();
+        
+	foreach(@bam_dir){
+		my $split;
+        	my @dir_split = split('/',$_);
+        	my $editor = pop(@dir_split);
+        	$editor =~ s/aligned/splitters/g;
```
apregier: The file name pattern that we expect doesn't actually have the string "aligned" in the bam - that was just some test data that we made.  So, we'll need to adjust this regular expression and the file names of the test data.  What we actually expect is something like:
47633c8db598491b83171ea874957bb4.bam
47633c8db598491b83171ea874957bb4.discordants.bam
47633c8db598491b83171ea874957bb4.splitters.bam

lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.pm
```diff
@@ -0,0 +1,132 @@
...
+};
+
+sub find_split{
+        my $self = shift;
+        my @bam_dir = @_;
+	my @final = ();
+        
+	foreach(@bam_dir){
+		my $split;
+        	my @dir_split = split('/',$_);
```
apregier: Instead of split, use the package File::Spec  It has dirname() and basename() methods

lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.pm
```diff
@@ -0,0 +1,132 @@
...
+        	my $editor = pop(@dir_split);
+        	$editor =~ s/aligned/splitters/g;
+        	push (@dir_split, $editor);
+        	$split = join('/', @dir_split);
+        	push (@final, $split);
+	}
+	my $combined_splits = join (',',@final);
+        return (
```
apregier: We probably only want to return this if at least one of the split bam files actually exist.

lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.pm
```diff
@@ -0,0 +1,132 @@
...
+		split_read_bam_file => $combined_splits,
+	);
+};
+
+sub find_discord{
```
apregier: There is a lot of duplication between the find_split subroutine and this subroutine - can you refactor to reduce the code duplication?

lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.pm
```diff
@@ -0,0 +1,132 @@
...
+
+};
+
+sub split_params_to_letter {
+        my $self = shift;
+        my $parms = $self->params;
+        my %params_hash = ();
+	my @params = split(',',$parms);
+	foreach (@params){
+		if ($_ =~ /:/){
```
apregier: Rather than using substr, this might be clearer and more flexible if you implement it by splitting on ':' and then checking for the length of the resulting array - if it is 1 then you can interpret it as a boolean, and if it is 2 you can interpret it as a key-value parameter.

lib/perl/Genome/Model/Tools/Speedseq/Sv.t
```diff
@@ -0,0 +1,80 @@
...
+   split_read_bam_file => $split_bam,
+   discordant_read_bam_file => $discordant_bam,
+);
+isa_ok($sv_cmd,$pkg);
+ok($sv_cmd->execute,'execute command '. $pkg);
+
+#VCF Difference
+my $output_file = "$output_prefix.sv.vcf.gz";
+
+my $expected_output_file = ('/gscmnt/gc2801/analytics/mfulton/genome2/lib/perl/Genome/Model/Tools/Speedseq/Sv.t.out/Sv.t.out.sv.vcf.gz');
```
apregier: Another hard-coded path

lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.pm
```diff
@@ -0,0 +1,132 @@
...
+	my %final_cmd = (
+   		output_prefix => $self->_sv_staging_output,
+   		full_bam_file => $aligned_bams,
+		$self->find_split(@fullBam),
+		$self->find_discord(@fullBam),
+	);	
+
+	my %list_params = $self->split_params_to_letter();
+ 	
+	while (my ($key, $value) = each(%ref_library)) {$final_cmd{$value} = $list_params{$key} if exists $list_params{$key};}
```
apregier: This will look clearer if you split it up onto multiple lines with proper indentation

apregier: We probably need to die with an error message if the key doesn't exist in the hash

lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.pm
```diff
@@ -0,0 +1,132 @@
...
+
+use Genome;
+use File::Basename;
+use IPC::System::Simple;
+use Genome::File::Tsv;
+
+class Genome::Model::Tools::DetectVariants2::SpeedseqSv {
+    is => 'Genome::Model::Tools::DetectVariants2::Detector',
+    has_optional => [
+        _legend_output => {
```
apregier: Is this just copy and pasted from somewhere?  I don't see it used.

lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.pm
```diff
@@ -0,0 +1,132 @@
...
+		g => 'genotype_svtyper',
+		d => 'CNVnator_read_depth',
+		A => 'annotate_vcf',
+		k => 'keep_temp_files',
+		#End of Boolean files#
+		t => 'threads',
+		x => 'exclude_bed_file',
+		m => 'min_sample_weight',
+		r => 'trim_threshold',
+		T => 'temp_directory',
```
apregier: I don't think we should expose the temp_directory parameter at this level - we should just make one using Genome::Sys->create_temp_directory inside of the _detect_variants subroutine and pass that in

lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.pm
```diff
@@ -0,0 +1,132 @@
...
+    is => 'Genome::Model::Tools::DetectVariants2::Detector',
+    has_optional => [
+        _legend_output => {
+            calculate_from => ['_temp_staging_directory'],
+            calculate => q{ File::Spec->join($_temp_staging_directory, 'legend.tsv'); },
+        },
+    ]
+};
+
+	my $temp_directory = Genome::Sys->create_temp_directory();
```
apregier: I think this should be moved into the _detect_variants subroutine

lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.pm
```diff
@@ -0,0 +1,132 @@
...
+
+
+package Genome::Model::Tools::DetectVariants2::SpeedseqSv;
+use warnings;
+use strict;
+
+use Genome;
+use File::Basename;
+use IPC::System::Simple;
+use Genome::File::Tsv;
```
apregier: Check and see whether all of these packages are actually used, and remove ones that aren't

apregier:
It looks like the file Genome/Model/Tools/DetectVariants2/Genome::Model::Tools::DetectVariants2::Detector, (with a comma) somehow got created and added to a commit - just use the command `git rm` to remove it again

apregier:
CI, test this

tmooney:
CI, test this.

apregier:
CI, test this.

apregier:
We are ready for wider review on this

lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.pm
```diff
@@ -0,0 +1,109 @@
+#SpeedseqSv.pm
```
tmooney: This comment can be removed.

lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.pm
```diff
@@ -0,0 +1,109 @@
...
+use File::Basename qw(basename dirname);
+use File::Spec;
+use IPC::System::Simple;
+use Genome::File::Tsv;
+
+class Genome::Model::Tools::DetectVariants2::SpeedseqSv {
+    is => 'Genome::Model::Tools::DetectVariants2::Detector',
+};
+
+	my $temp_directory = Genome::Sys->create_temp_directory();
```
tmooney: Having this here will create a temporary directory at "compile time".

If this needs to be a single temporary directory shared classwide, the variable should be wrapped by a subroutine that initializes it the first time it's called.  Otherwise we should use that subroutine to wrap a property on the instance (or use a calculated property).

lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.pm
```diff
@@ -0,0 +1,109 @@
...
+	my $pkg = 'Genome::Model::Tools::Speedseq::Sv';
```
tmooney: This might benefit from being a constant classwide property, like:
```perl
has => [
  gmt_package => {
    is => 'Text',
    is_classwide => 1,
    is_constant => 1,
    value => 'Genome::Model::Tools::Speedseq::Sv',
  },
],
```
(There are likely better name choices than `gmt_package` :smile: )

lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.pm
```diff
@@ -0,0 +1,109 @@
...
+	my $self = shift;
+	
+	my @fullBam = ($self->aligned_reads_input,$self->control_aligned_reads_input);
+
+	my $aligned_bams = join(',',@fullBam);
+
+	my %library = $self->get_parameter_hash();
+
+	while (my ($key, $value) = each(%library)){
+		print "$key => $value\n";
```
tmooney: Is this for debugging, or part of the output?
If it's for debugging, it could be `$self->debug_message("$key => $value")`.
if it's part of the output, it might be good to sort the keys so that the output is consistent across runs.

lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.pm
```diff
@@ -0,0 +1,109 @@
...
+	my $disc = shift;
+	my $value = shift;
+	my @bam_dir = @_;
+	my @final = ();
+	
+	foreach (@bam_dir){
+		my $editor = basename($_);
+		my $dir = dirname($_);
+		$editor =~ s/\.bam/\.$disc\.bam/g;
+		my $discord = "$dir/$editor";
```
tmooney: `File::Basename` has another function, `fileparse`, that may help here:

```perl
my ($editor, $dir, $suffix) = fileparse($_, '.bam');
my $discord = "$dir$editor.$disc$suffix";
```

lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.pm
```diff
@@ -0,0 +1,109 @@
...
+sub split_params_to_letter {
+        my $self = shift;
+        my $parms = $self->params;
+        my %params_hash = ();
+	my @params = split(',',$parms);
+	foreach (@params){
+		if ($_ =~ /:/){
+			my @options = split(':',$_);
+			my $value = pop(@options);
+			my $num = shift(@options);
```
tmooney: If you wanted, you could assign to `$num` and `$value` directly:
```perl
my ($num, $value) = split(':', $_, 2);
```
(The `2` at the end is useful if the value might itself contain a `:`.)

lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.pm
tmooney: (Since it's only used once, it could also be taken out, and `Genome::Model::Tools::Speedseq::Sv->create` could be called directly.  We usually only use this `$pkg` shortcut in tests.)

lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.t
```diff
@@ -0,0 +1,81 @@
...
+	my $pkg2 = 'Genome::Model::Tools::DetectVariants2::SpeedseqSv';
+	
+	my $refbuild_id = 101947881;
+	my $result_users = Genome::Test::Factory::SoftwareResult::User->setup_user_hash(reference_sequence_build_id => $refbuild_id,);
+
+	my $params = "-R:$reference_fasta,-g,-d,-o:Hello";
+
+	print "Here is the base name of the bam variable: \t";
+	print basename($bam);
+	print "\n\n";
```
tmooney: The best way to print debug information in tests is with `diag`.  This is guaranteed to print the messages in a way that won't break the test output.

lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.pm
```diff
@@ -0,0 +1,109 @@
...
+
+package Genome::Model::Tools::DetectVariants2::SpeedseqSv;
+use warnings;
+use strict;
+
+use Genome;
+use File::Basename qw(basename dirname);
+use File::Spec;
+use IPC::System::Simple;
+use Genome::File::Tsv;
```
tmooney: Are `IPC::System::Simple` and `Genome::File::Tsv` needed here?

lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.t
```diff
@@ -0,0 +1,81 @@
...
+	result_users => $result_users,	
+	aligned_reads_input => $bam,
+	params => $params,
+	control_aligned_reads_input => $bam2,
+);
+
+ok($command2->execute, 'Executed `gmt detect-variants2 Speedseq` command');
+
+
+$DB::single=1;
```
tmooney: A breakpoint!  (Probably should be removed.)

lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.t.d/svs.hq.sv.NA12878.20slice.30X.aligned.bam.readdepth.bed
```diff
@@ -0,0 +1,2 @@
+20_slice	51001	53900	0.559811
+20_slice	260701	263200	0.504321
```
tmooney: It'd be good to make sure there's a newline at the end of these BED files.  Some of our tools don't work well when it's missing.

lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.t.d/svs.hq.sv.NA12878.20slice.30X2.aligned.bam.readdepth.bed
```diff
@@ -0,0 +1,2 @@
+20_slice	51001	53900	0.559811
+20_slice	260701	263200	0.504321
```
tmooney: Are these `30X2` files being used anywhere?  At first glance they're identical to the `30X` versions.

lib/perl/Genome/Model/Tools/Speedseq/Sv.pm
```diff
@@ -0,0 +1,137 @@
...
+                        doc => 'discordant reads BAM file(s) (comma separated list) (Auto Generated if absent)',
+                        example_values => ['input1.bam'],
+                        tool_param_name => 'D',
+                        is_optional => 1,
+                        tool_input_file => 1,
+                },
+		reference_fasta => {
+			is => 'Text',
+			doc => 'fasta file (indexed with bwa)',
+			example_values => ['refernce.fa'],
```
tmooney: `reference` in the example value is missing an `e`.

lib/perl/Genome/Model/Tools/Speedseq/Sv.pm
```diff
@@ -0,0 +1,137 @@
...
+        	output_prefix => {
+        		is => 'Text',
+        		doc => 'output prefix [in1.bam]',
+        		is_optional => 1,
+        		example_values => ['in1.fq','in.realign'],
+        		tool_param_name => 'o',
+        	},
+		threads => {
+        		is => 'Integer',
+        		doc => 'threads [1]',
```
tmooney: It's not necessary for the `doc` to contain the `default_value`.  The output for `--help` will list it in a structured way.

lib/perl/Genome/Model/Tools/Speedseq/Sv.pm
```diff
@@ -0,0 +1,137 @@
...
+       			default_value => 1,
+        		is_optional => 1,
+        		tool_param_name => 't',
+        	},
+		exclude_bed_file => {
+                        is => 'Text',
+                        doc => 'BED file to exlude',
```
tmooney: In the doc, `exclude` is missing a `c`.

lib/perl/Genome/Model/Tools/Speedseq/Sv.pm
```diff
@@ -0,0 +1,137 @@
...
+                },
+		trim_threshold => {
+                        is => 'Float',
+                        doc => 'trim threshold [0]',
+                        is_optional => 1,
+                        tool_param_name => 'r',
+                },
+		temp_directory => {
+            		is => 'Text',
+			default_value => Genome::Sys->create_temp_directory(),
```
tmooney: Putting this in a `default_value` section creates it at "compile time" much like putting it in a `my` variable at the top level does.  (Whatever solution is used for `Genome::Model::Tools::DetectVariants2::SpeedseqSv` would work here, too!)

lib/perl/Genome/Model/Tools/Speedseq/Sv.pm
```diff
@@ -0,0 +1,137 @@
...
+                        tool_param_name => 'k',
+                },
+		
+
+	],
+};
+
+
+sub path_for_version {
+	return '/gscmnt/gc2719/halllab/users/cchiang/bin/speedseq';
```
tmooney: This ought to defer to the `path_for_version` defined in the base class.  (But if there's something in that version that's not in any of our "frozen" versions yet, it can wait...)

apregier:
It looks like tests are failing because some of the test data is missing - did some of it not get committed?

apregier:
CI, test this.

Origamigirl11:
Its ready to be looked at again. @tmooney t

tmooney:
Looking pretty good! A few thoughts:

* I see some of the "30X2" files have been removed, but others have appeared (e.g. `lib/perl/Genome/Test/Data.pm.d/NA12878/NA12878.20slice.30X2.splitters.bam`).  Are they needed?
* Some of the files use a mix of tabs and spaces for indentation.  It'd be better if each file stuck to one or the other.
* It looks like a couple more defaults are still listed in `doc` lines.  (In one case, that might be okay since the default is only implied--though it could be moved to a `default_value` instead.)
* The names of the internal variables to `find_file` suggest it is designed to run on the discordant read BAM, but it also runs on the split read BAM.  It might be nice if they had more neutral names.

But notwithstanding all that, if it works I say +1.

apregier:
CI, test this.

lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.t
```diff
@@ -0,0 +1,78 @@
...
+	my $test_dir = __FILE__.".d";
+	
+	my $version = '0.0.3a-gms';
+
+	my $output = Genome::Sys->create_temp_directory();
+        
+	my $reference_fasta = get_test_file('NA12878', 'human_g1k_v37_20_42220611-42542245.fasta');
+	my $bam = get_test_file('NA12878', 'NA12878.20slice.30X.aligned.bam');
+	my $bam2 = get_test_file('NA12878', 'NA12878.20slice.30X.aligned.bam');
+	my $split_bam = get_test_file('NA12878','NA12878.20slice.30X.splitters.bam');
```
apregier: $split_bam and $discordant_bam aren't used anywhere, right?

lib/perl/Genome/Model/Tools/Speedseq/Sv.t
```diff
@@ -0,0 +1,86 @@
...
+my $pkg = 'Genome::Model::Tools::Speedseq::Sv';
+use_ok($pkg);
+
+my $speedseq_version = '0.0.3a-gms';
+
+my $expected_output_dir = __FILE__.".out";
+
+my $reference_fasta = get_test_file('NA12878', 'human_g1k_v37_20_42220611-42542245.fasta');
+my $bam = get_test_file('NA12878', 'NA12878.20slice.30X.aligned.bam');
+my $split_bam = get_test_file('NA12878','NA12878.20slice.30X.splitters.bam');
```
apregier: let's use the 30X.aligned.* versions of these files and remove the other ones that you added - they are identical except for the file name.

apregier:
CI, test this.

PR #871, opened 2015-07-20
It'll now work with aggrs containing any number of digits.

brummett:
The regex was failing to match /vol/aggr3/ which only contained one digit


ebelter:
+2

tmooney:
+1

PR #872, opened 2015-07-20
Filter out thread count, max memory, and verbosity from lookup hash.

tmooney:
Threads are overwritten inside the tool.  I heartell it should be safe to ignore the other two as well.

(Existing processing profile mostly have no parameters specified at all, so this shouldn't have much effect on results we already have.)


apregier:
+1

acoffman:
Is BWA non-deterministic depending on the number of threads selected? If that's the case, would we want the thread count to be included in the lookup hash? Or do we not care?

+1 if we don't care!

jasonwalker80:
+1, we would not want to invalidate our existing results because we change the number of threads to optimize performance on our hardware.

PR #873, opened 2015-07-21
Add status to AnalysisMenu Items

tmooney:
This helps reduce the list shown by default to the commonly used options.

The two statuses are `active` and `inactive` since in neither case does the menu item cease to function (i.e. `disabled`).

The changed commands won't work until the database is updated to add this column.


tmooney:
It would be most helpful if we could have the CI test this now that the migration has been applied!

acoffman:
+2

jeldred:
+1 :anchor:

PR #874, opened 2015-07-21
Use error_message for SendToPap's Workflow Errors

tmooney:
These were originally `status_message` and were then changed to `debug_message`.  Since they're errors, it seems like they should be `error_message`s instead.


brummett:
+1

ebelter:
+2

PR #875, opened 2015-07-21
Model define.t simplification

ebelter:


tmooney:
+1, it might make sense for `Define.t` to use an auto-generated define command (either an existing class like `Genome::Model::Command::Define::SomaticValidation` or a custom-defined one in the test).  This test could then be `Define/ReferenceAlignment.t`.

ebelter:
@tmooney  I agree. The whole space is a mess. I wanted to get this done first, then look at creating an actual define test.

PR #876, opened 2015-07-21
CLE "variant-reporting wrappers gold-evaluation" addons

indraniel:
The `genome variant-reporting wrappers gold-evaluation` command needed some additional work to get it to run in practice. This set of commits are those patches.  They should be applied to both the mainline master branch, and in theory, the CLE snapshot, `genome-3590`.  The details about these patches and its "testing" are described in [BIO-1491](https://jira.gsc.wustl.edu/browse/BIO-1491).

@susannasiebert, @davidlmorton, and @jasonwalker80 provided the inspiration and recommendations for these changes.


lib/perl/Genome/VariantReporting/Command/Wrappers/GoldEvaluation.pm
```diff
@@ -86,5 +88,89 @@ sub is_single_bam {
...
+}
+
+sub normal_sample {
+    my $self = shift;
+    return $self->model->normal_sample;
+}
+
+sub discovery_sample {
+    my $self = shift;
+    return $self->model->tumor_sample;
```
tmooney: These could be properties, like:
```perl
discovery_sample => {
  is => 'Genome::Sample',
  via => 'model',
  to => 'tumor_sample',
},
discovery_sample_name => {
  is => 'Text',
  via => 'discovery_sample',
  to => 'name',
},
```
There is no requirement that the current inputs to this model match those on the `last_succeeded_build`.  If this command is operating on the data from that succeeded build, it may be better to rely on that build for all information instead of mixing between the two.  (Alternatively, if `last_succeeded_build->is_current` the build believes no consequential changes to the model have occurred.)

lib/perl/Genome/VariantReporting/Command/Wrappers/GoldEvaluation.t
```diff
@@ -0,0 +1,70 @@
...
+subtest "sample and library labels" => sub {
+
+    my $cmd = $pkg->create(
+        model => $build1->model,
+        gold_sample_name => "GOLD",
+        snvs_input_vcf => Genome::Sys->create_temp_file_path,
+        indels_input_vcf => Genome::Sys->create_temp_file_path,
+    );
+
+    $DB::single = 1;
```
tmooney: There's a breakpoint here.

lib/perl/Genome/VariantReporting/Command/Wrappers/GoldEvaluation.pm
```diff
@@ -86,5 +88,89 @@ sub is_single_bam {
     return (!defined($model->normal_sample));
 }
 
+sub get_common_translations {
```
apregier: This is only used in the test - did you intend to use it at lines 39 and 48?  I actually don't see much point in it, since both those use cases already have the is_single_bam conditional

lib/perl/Genome/VariantReporting/Command/Wrappers/GoldEvaluation.pm
```diff
@@ -86,5 +88,89 @@ sub is_single_bam {
...
+    return $self->model->normal_sample;
+}
+
+sub discovery_sample {
+    my $self = shift;
+    return $self->model->tumor_sample;
+}
+
+my %counters;
+sub get_library_name_labels {
```
apregier: This is identical to subroutines in ModelPairFactory and ModelReport - maybe with a third instance it is time to refactor so the code isn't duplicated?

indraniel: Perhaps.  I'm still not too familiar with this code base.  

Part of my "fixing" involved copying the logic from ModelReport and making "adjustments" from there.

Some of the nomenclature is a bit confusing to me.  For example, when we are supposed to be using the term "tumor" versus "discovery"?  To my initial eyes, the "discovery" sample should be really called the "tumor" sample.

lib/perl/Genome/VariantReporting/Command/Wrappers/GoldEvaluation.pm
indraniel: This is again an artifact from my refactoring to get the module to simply"work".  I initially copied the execute and some supporting logic from `Genome::VariantReporting::Command::Wrappers::ModelReport`.  `get_common_translations` was one of the support methods.  The test was initially created to test the new `get_common_translations` code changes I placed into this module.  Along the path to getting things working,  the `common_translations` method got refactored into the broken down `get_somatic_translations` and `get_germline_translations`.  

I left the `get_common_translations` method around, to keep the "framework" symmetry with the other variant reporting wrapper modules.

indraniel:
Is it okay for me to merge the code in its latest state?

apregier:
I understand 'just get it working' (especially since it wasn't working in part because of my initial laziness!) Just a few small refactors will make it way easier to 'keep it working'.  Could we pair on this sometime?

indraniel:
> I understand 'just get it working' (especially since it wasn't working in part because of my initial laziness!) Just a few small refactors will make it way easier to 'keep it working'. Could we pair on this sometime?

Most definitely!  We can talk about it later today.

apregier:
+1 pending tests passing!

indraniel:
Looks like the tests are passing.  @tmooney, do things pass your review?

tmooney:
+1 :space_invader:

PR #878, opened 2015-07-22
Model: remove unused _trigger_downstream_builds

ebelter:


tmooney:
We should also remove the [call to this in `Genome::Model::Build`](https://github.com/ebelter/genome/blob/d6c983f43b9cef0534df2919f96db8c45d92fc6f/lib/perl/Genome/Model/Build.pm#L1600).

tmooney:
+1 :pencil2:

apregier:
+1 Maybe a follow-on pr can clean up auto_build and downstream_models properties, which seem like they are only used here

PR #879, opened 2015-07-22
Disconnect Database Handles before potentially long-running actions

tmooney:
The specific motivation for this change: The database "pausing" feature will disconnect handles, but not if the pause is both set and cleared before the program has a chance to notice.  It may then be left holding a stale handle when it does return.


brummett:
+1

apregier:
In Genome::WorkflowBuilder::DAG, why do we need to disconnect the db handles in _execute_with_workflow but not _execute_with_ptero?

tmooney:
@apregier Good question!  I've added a disconnect before we wait on PTero now as well.

acoffman:
+1

apregier:
+1

PR #880, opened 2015-07-22
Dump flagstat data when comparison fails.

tmooney:
Since the `$temp_flagstat` gets cleaned up on failure, this provides a way to see how it was different from expected.


ebelter:
+2

PR #882, opened 2015-07-23
Parse MEMLIMIT with units, normalize to kb as expected

rfugina:


rfugina:
Either I don't see the failed tests, or they're all failed.  If someone could point out to me where to look...

davidlmorton:
CI test this.

ebelter:
Ptero to JSON tests are failing. Click the details link, then the 'Test Result' or 'Show all failed tests' links.

rfugina:
CI test this

davidlmorton:
+1

davidlmorton:
This code worked fine while running the somatic-validation test build on lsf9.

lib/perl/Genome/Site/TGI/Extension/Sys.pm
```diff
@@ -630,7 +630,23 @@ sub get_mem_limit_from_bjobs {
...
+        }
+        elsif ($bjobs_mem_limit_units eq 'K') {
+            $bjobs_mem_limit_kb = int($bjobs_mem_limit);
+        }
+        else {
+            # failed to parse
+        }
         $mem_limit = $bjobs_mem_limit_kb if ($bjobs_mem_limit_kb);
     }
     return $mem_limit;
```
brummett: Looks like ```$bjobs_mem_limit_kb``` isn't necessary.  The if/else blocks can just assign to ```$mem_limit```.

Also, is it OK for this function to return undef if it couldn't parse, or does it need to die?

rfugina: The sub that calls it, mem_limit_kb, and the unit tests for that sub, allow for the possibility of undef being returned.  It seems there is the possibility that the job has no mem limit, and mem_limit_kb falls back to reporting on physical memory.

As for the extra variable...  Absolutely right.

rfugina:
CI test this

rfugina:
So I ran a couple of jobs with additional limits -- stack size, processes -- just to see what the output of bjobs would look like.  Take a look at job 5927089 on LSF8, or 45903 on LSF9 to see.  I'm going to put more thought into it...

davidlmorton:
Can an LSF job really have no MEMLIMIT? I thought if it wasn't specified it was set to 4 GB.  Regardless, I'm on board with Rob's new plan of putting the LSF job information into env variables instead of parsing bjobs output in the future, which will probably make this PR moot.

jeldred:
closing until " I'm going to put more thought into it..." time

davidlmorton:
CI test this.

davidlmorton:
CI test this (Ran out of time last attempt... changes to Genome::Sys trigger just about every test to run)

davidlmorton:
+1 I'm suggesting we merge this since the environment variable solution does not now seem possible.  This is needed to run on lsf9.

tmooney:
+1, seems to work!

PR #885, opened 2015-07-23
Use shellcmd instead of IPC::Run.

tmooney:
The motivation for this change is to make sure the database handle is disconnected before running.  Alternatively we could add an explicit disconnect before the call here.


brummett:
It's not using any of the fancy IPC::Run features, so this looks like the right change

+1

apregier:
+1

PR #886, opened 2015-07-23
GM Concordance

ebelter:
add available versions for picard
add latest and available versions to bedtools
add available versions for bedtools
add default value for bedtools version
use bedtools version

### DIFF
--- a/lib/perl/Genome/Model/ReferenceAlignment/Command/GenotypeMicroarrayConcordance.pm
+++ b/lib/perl/Genome/Model/ReferenceAlignment/Command/GenotypeMicroarrayConcordance.pm
@@ -21,8 +21,7 @@ class Genome::Model::ReferenceAlignment::Command::GenotypeMicroarrayConcordance 
         },
         bedtools_version => {
             is => 'Text',
-            doc => 'Version of bed tools intersect to use.',
-            default_value => Genome::Model::Tools::BedTools->latest_bedtools_version,
+            doc => 'Version of bed tools intersect to use. Recommended version: '.Genome::Model::Tools::BedTools->latest_bedtools_version,
             valid_values => [ Genome::Model::Tools::BedTools->available_bedtools_versions ],
         },
         picard_version => {
--- a/lib/perl/Genome/Model/ReferenceAlignment/Command/GenotypeMicroarrayConcordance.pm
+++ b/lib/perl/Genome/Model/ReferenceAlignment/Command/GenotypeMicroarrayConcordance.pm
@@ -22,7 +22,7 @@ class Genome::Model::ReferenceAlignment::Command::GenotypeMicroarrayConcordance 
         bedtools_version => {
             is => 'Text',
             doc => 'Version of bed tools intersect to use.',
-            default_value => Genome::Model::Tools::BedTools->latest_bedtools_Version,
+            default_value => Genome::Model::Tools::BedTools->latest_bedtools_version,
             valid_values => [ Genome::Model::Tools::BedTools->available_bedtools_versions ],
         },
         picard_version => {
--- a/lib/perl/Genome/Model/ReferenceAlignment/Command/GenotypeMicroarrayConcordance.pm
+++ b/lib/perl/Genome/Model/ReferenceAlignment/Command/GenotypeMicroarrayConcordance.pm
@@ -19,6 +19,12 @@ class Genome::Model::ReferenceAlignment::Command::GenotypeMicroarrayConcordance 
             doc => 'The minimum depth required to consider concordant.',
             example_values => ['4'],
         },
+        bedtools_version => {
+            is => 'Text',
+            doc => 'Version of bed tools intersect to use.',
+            default_value => Genome::Model::Tools::BedTools->latest_bedtools_Version,
+            valid_values => [ Genome::Model::Tools::BedTools->available_bedtools_versions ],
+        },
         picard_version => {
             is => 'Text',
             doc => 'The version of picard to use.',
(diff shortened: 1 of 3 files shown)


lib/perl/Genome/Model/Tools/BedTools.pm
```diff
@@ -49,6 +49,19 @@ my %BEDTOOLS_VERSIONS = (
...
+sub available_bedtools_versions {
+    my $sort = sub{ 
+        my @a = split(/\./, $_[0]);
+        my @b = split(/\./, $_[1]);
+        return $a[0] <=> $b[0] || $a[1] <=> $b[1] || $a[2] <=> $b[2];
+    }; # could use Sort::Versions it has a lucid pkg
+    return sort { $sort->($b, $a) } keys %BEDTOOLS_VERSIONS;
+}
+
+sub latest_bedtools_Version {
```
tmooney: This should probably not have a capital `V`.

ebelter:
@tmooney Any more input for this PR?

lib/perl/Genome/Model/ReferenceAlignment/Command/GenotypeMicroarrayConcordance.pm
```diff
@@ -19,10 +19,16 @@ class Genome::Model::ReferenceAlignment::Command::GenotypeMicroarrayConcordance
...
+            is => 'Text',
+            doc => 'Version of bed tools intersect to use.',
+            default_value => Genome::Model::Tools::BedTools->latest_bedtools_version,
+            valid_values => [ Genome::Model::Tools::BedTools->available_bedtools_versions ],
+        },
         picard_version => {
             is => 'Text',
             doc => 'The version of picard to use.',
-            example_values => ['1.123'],
+            valid_values => [ Genome::Model::Tools::Picard->available_picard_versions ],
```
tmooney: I believe this tool was added to Picard in `1.122`, so many of these won't be valid versions for this tool.

lib/perl/Genome/Model/Tools/BedTools.pm
```diff
@@ -49,6 +49,19 @@ my %BEDTOOLS_VERSIONS = (
     '2.3.2' => Genome::Config::get('sw') . '/bedtools/BEDTools-2.3.2',
 );
 
+sub available_bedtools_versions {
```
acoffman: Would using something like `Perl::Version` (which I think we already have installed) make this sort a little more robust?

lib/perl/Genome/Model/ReferenceAlignment/Command/GenotypeMicroarrayConcordance.pm
```diff
@@ -19,10 +19,16 @@ class Genome::Model::ReferenceAlignment::Command::GenotypeMicroarrayConcordance
             doc => 'The minimum depth required to consider concordant.',
             example_values => ['4'],
         },
+        bedtools_version => {
+            is => 'Text',
+            doc => 'Version of bed tools intersect to use.',
+            default_value => Genome::Model::Tools::BedTools->latest_bedtools_version,
```
acoffman: Do we want a default value here? I know in general we try to avoid them.

lib/perl/Genome/Model/Tools/BedTools.pm
brummett: The "version" module from the Perl standard library can do this, too.

lib/perl/Genome/Model/ReferenceAlignment/Command/GenotypeMicroarrayConcordance.pm
ebelter: @acoffman I agree - although, an old buggy version is the reason for this PR. Also, This command is run by hand, and not in a pipeline. Ideas?

lib/perl/Genome/Model/Tools/BedTools.pm
ebelter: @acoffman @brummett Perl::Version and Sort::Versions are installed for 5.8.7, but not 5.10.1. version does not correctly sort the maintenance release number:
```
$ sort { $b cmp $a } map { version->parse($_)->normal } (qw/ v2.2.1 2.2.2 v2.2.3 2.2.11 2.3.0 2.1.0 /)
v2.3.0
v2.2.3
v2.2.2
v2.2.11
v2.2.1
v2.1.0
```

brummett: You're supposed to sort the version objects directly
```
map { print "$_\n" }
sort { $b <=> $a }
map { version ->parse($_) }
(qw/ v2.2.1 2.2.2 v2.2.3 2.2.11 2.3.0 2.1.0 /);

2.3.0
2.2.11
v2.2.3
2.2.2
v2.2.1
2.1.0
```

lib/perl/Genome/Model/ReferenceAlignment/Command/GenotypeMicroarrayConcordance.pm
ebelter: Thanks! I will have a PR to address this, only returning the available versions based on the sub command.

lib/perl/Genome/Model/ReferenceAlignment/Command/GenotypeMicroarrayConcordance.pm
acoffman: If its run by hand, I think a default value is fine for now with the caveat that we should probably remove it should it become part of an automated process!

ebelter:
@acoffman I moved the default to a recommendation in the doc.
@tmooney @brummett I replaced custom sort w/ perl's version handler

PR #887, opened 2015-07-24
Add the ability to specify a minimum number of hits required

apregier:
In some cases, we want to require more than one hit to count an sv as a 'true positive'.  We also want to save the actual svs in json for further processing.

### DIFF
--- a/lib/perl/Genome/Model/Tools/Bedpe/EvaluateBedpe.pm
+++ b/lib/perl/Genome/Model/Tools/Bedpe/EvaluateBedpe.pm
@@ -94,7 +94,7 @@ sub _get_true_positives {
 
 sub _sv_with_min_support_count {
     my ($self, $file, $original_file, $output) = @_;
-    my $f = new IO::File($file, "r");
+    my $f = Genome::Sys->open_file_for_reading($file);
     my %results;
 
     my $id_index = _count_cols($original_file) + 6;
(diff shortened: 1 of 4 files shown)


lib/perl/Genome/Model/Tools/Bedpe/EvaluateBedpe.pm
```diff
@@ -67,10 +77,84 @@ sub _get_stat {
...
+        my $sv_name = $f[6];
+        unless (defined $sv_name) {
+            die "Sv name not defined on line:\n$line";
+        }
+        next if exists $seen{$sv_name};
+        $seen{$sv_name} = 1;
+
+        next unless exists $results{$sv_name} && ($results{$sv_name}{count} >= $min_hit_support);
+        $count++;
+        if ($output) {
```
tmooney: It doesn't make a big difference, but would it be better for this to check `if ($out)`?

lib/perl/Genome/Model/Tools/Bedpe/EvaluateBedpes.pm
```diff
@@ -50,13 +58,24 @@ sub execute {
     return 1;
 }
 
+sub _read_tps {
+    my $file = shift;
+    my @tps = Genome::Sys->read_file($file);
+    for my $tp (@tps) {
+        chomp $tp;
```
tmooney: This could be `chomp @tps;` instead of the loop.

apregier:
Thanks for the suggestions, I've pushed commits to address them.

tmooney:
+1 :athletic_shoe: 

If we're going to do a lot of parsing of these files, it might be nice to have a constant for the `6` that appears in a few places, a subroutine for extracting the `sv_name` from the line, or somesuch similar.

ebelter:
+1

PR #888, opened 2015-07-24
PTero: Control flow feature

davidlmorton:
The PTero system now understands links between tasks that are not data-flow related.  This PR updates the building of PTero workflows to utilize this new feature.

I've also added a couple of commands related to running a Genome::Process::Test::Process that are helpful for manually testing the ability to submit to Processes to PTero without using the more computationally intensive and complex trio variant-report.

Since libptero-perl 0.1.7 has been deployed before this PR, the unit-tests in Genome::Ptero will fail until this PR is merged.  That is because the libptero-perl library changed how it renders JSON representations of workflows to be compatible with the PTero workflow service.


davidlmorton:
Merging this without additional review because this will allow the quick-tests to pass for other people's PRs.

PR #891, opened 2015-07-28
Only schedule annotation if we have variants to annotate.

tmooney:
A common misconfiguration of a project seems to be to assign an annotation build input to a reference alignment that is not going to run variant detection.  This currently results in a build failure.

With this change, the build should succeed and not worry about the (then unused) annotation input.  (The alternative change would be to make such a build "Unstartable" so it fails fast.)


acoffman:
I'm generally fine with this change - my only question/concern is: are there places where we're filtering/querying or otherwise using the absence of an annotation build to indicate which RefAlign builds have do variant detection and which dont? It seems like we should be using the PP to determine that, but you never know :)

tmooney:
@acoffman If something is using the annotation build to decide whether or not variant detection occurred, I am not remorseful about breaking that logic!

(A quick `grep` of the tree doesn't not point out anything immediately suspicious. Tools looking at the property are generally doing something that would not change based on whether these models have the input is set.)

acoffman:
I generally agree, just wanted to be sure! :)

+1

apregier:
+1

PR #892, opened 2015-07-28
LDAP Synch: Fix counting bug and cleanup

tmooney:
* The tool has a threshold to avoid doing too much work, but it was failing to count deletes towards that limit.
* If the limit is exceeded it prints a message, but that message had no newline (so it would potentially be buffered for a long while).


brummett:
+1

acoffman:
+1

PR #893, opened 2015-07-29
Don't try to run LOH without a normal to compare to.

tmooney:
This lets a `SomaticVariation` build proceed even though the combination of inputs and processing profile doesn't work.  (This is similar to #891.)  I could see a stronger argument for this one leaning the `Unstartable` way.  On the other hand, we should stop using `SomaticVariation` altogether!

### DIFF
--- a/lib/perl/Genome/Model/SomaticVariation/Command/Loh.pm
+++ b/lib/perl/Genome/Model/SomaticVariation/Command/Loh.pm
@@ -130,9 +130,9 @@ sub should_skip_run {
         return 1;
     }
 
-    my $normal_build = $build->normal_build;
-    unless($normal_build->can('snv_detection_strategy') and $normal_build->snv_detection_strategy) {
-        $self->debug_message("No SNV Detection Strategy on normal build, skipping LOH.");
+    my $normal_model = $build->normal_model;
+    unless($normal_model->can('snv_detection_strategy') and $normal_model->snv_detection_strategy) {
+        $self->debug_message("No SNV Detection Strategy on normal model, skipping LOH.");
         return 1;
     }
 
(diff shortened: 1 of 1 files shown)


brummett:
I don't understand how this changes anything.  It looks like the only effective change is that it might print a debug_message().

Does retrieving the normal build or calling ```$normal_build->snv_detection_strategy()``` have some kind of side-effect?

tmooney:
There's a simple explanation: I failed at `git`, and pushed a version that was missing the `return 1;`.  (Rebased!)

brummett:
+1

lib/perl/Genome/Model/SomaticVariation/Command/Loh.pm
```diff
@@ -130,6 +130,12 @@ sub should_skip_run {
         return 1;
     }
 
+    my $normal_build = $build->normal_build;
+    unless($normal_build->can('snv_detection_strategy') and $normal_build->snv_detection_strategy) {
```
dufeiyu: This probably won't work since normal_build is ref-align build that doesn't have snv_detection_strategy as property (unlike som-var build).  Instead, $normal_model should work.

tmooney: Good point! I've change this to use the model instead.

lib/perl/Genome/Model/SomaticVariation/Command/Loh.pm
```diff
@@ -130,6 +130,12 @@ sub should_skip_run {
         return 1;
     }
 
+    my $normal_model = $build->normal_model;
+    unless($normal_model->can('snv_detection_strategy') and $normal_model->snv_detection_strategy) {
+        $self->debug_message("No SNV Detection Strategy on normal build, skipping LOH.");
```
dufeiyu: trivial, model here.

dufeiyu:
+1

PR #894, opened 2015-07-29
Allow Programmatic Access to ModelSummary information

tmooney:
A monitoring tool is being developed to help track the progress of data through sequencing and analysis.  Right now it uses `$model->status`, but it would benefit from being able to grab the full summary (without needing to capture `STDOUT` and parsing it).


ebelter:
+1

brummett:
+1

PR #895, opened 2015-07-29
Ignore the Y chromosome for LOH calling.

gatoravi:
Add another chromosome to ignore for this tool.


tmooney:
+1 if it's okay that this skips any chromosome that contains 'Y' rather than one named specifically 'Y'.

PR #897, opened 2015-07-31
Don't require builds to be from the same model to diff

apregier:
We have a use case for CLE regression testing where
we are making new models in a testdb and want to diff
the resulting builds with builds from old models.  They
should be equivalent, but not the same model.
I don't see any reason why this check adds to the
integrity of the diff

### DIFF
--- a/lib/perl/Genome/Model/Build.pm
+++ b/lib/perl/Genome/Model/Build.pm
@@ -2135,9 +2135,6 @@ sub compare_output {
     my $other_build = Genome::Model::Build->get($other_build_id);
     confess "Could not get build $other_build_id!" unless $other_build;
 
-    unless ($self->model_id eq $other_build->model_id) {
-        confess "Builds $build_id and $other_build_id are not from the same model!";
-    }
     unless ($self->class eq $other_build->class) {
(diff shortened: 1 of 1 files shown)


acoffman:
+1 pending tests!

apregier:
CI, test this.

ebelter:
Seems the model check was just for sanity. Should we check that the model type, sample or processing profile or inputs are the same?

apregier:
It is checking model type - because otherwise the comparison really wouldn't make sense!  I am hesitant to add comparisons of specific inputs/sample/etc. because we might want to compare 'equivalent' things, but that have different ids.  But I could be persuaded otherwise if you feel strongly.  The main objective of the diff (in my opinion) is to look at the outputs - files and metrics.

ebelter:
Odd that it would check the model id AND type. Seems checkont the id wouldTo me, checkojng that it was Looking at the outputs and metrics. I see it as 'did these inputs with this processing produce the same results on a different code base.' The model id check sort of does that, although inputs and processing profile could be different. Anywho, as long as we are aware of it, +1

PR #898, opened 2015-07-31
Utility Test data_dir: add doc, use validate_pos

ebelter:


lib/perl/Genome/Utility/Test.pm
```diff
@@ -532,6 +535,34 @@ Like command_execute_ok(), but this test fails if the command's execution return
...
+
+  Synopsis
+  Given a package name and optional version, locate the directory for test data.
+
+  Params
+  $package => package name
+  $version => optionally, the version of the test data
+
+  Returns
+  $dir     => test data diretor for package/version
```
tmooney: `directory` is missing a `c` and a `y`.

lib/perl/Genome/Utility/Test.pm
```diff
@@ -532,6 +535,34 @@ Like command_execute_ok(), but this test fails if the command's execution return
...
+
+=item data_dir_ok
+
+  my $dir = Genome::Utility::Test->data_dir_ok($package, $version);
+
+  Synopsis
+  Params and return are the same as data_dir(), but tests (via ok) if the directory exists.
+
+  Returns
+  $dir     => test data diretor for package/version regardless if the test for existance passes
```
tmooney: `directory` is missing a `c` and a `y`. `existence` has an `a` where an `e` goes.

PR #899, opened 2015-07-31
Picard

ebelter:
add minimumrequired versions to commands that need a min version
move enforcing min version to exectue from validate params
restrict available versions if minimum required version set
use GC vesions in ref-align command

did NOT try to set available versions as valid values for use_version

### DIFF
--- a/lib/perl/Genome/Model/Tools/Picard.t
+++ b/lib/perl/Genome/Model/Tools/Picard.t
@@ -129,7 +129,7 @@ subtest 'Available picard versions' => sub{
 
     $obj->minimum_version_required($versions[2]);
     my @versions_with_miniumum_required = $obj->available_picard_versions;
-    is(@versions_with_miniumum_required, 3, '3 available verions with minimum required version set');
+    is(@versions_with_miniumum_required, 3, '3 available versions with minimum required version set');
     cmp_ok(
         @versions, '>', @versions_with_miniumum_required, 
         'less versions available with minimum set',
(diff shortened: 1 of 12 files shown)


lib/perl/Genome/Model/Tools/Picard.t
```diff
@@ -109,15 +117,24 @@ subtest "Enforce minimum version" => sub {
...
+subtest 'Available picard versions' => sub{
+    plan tests => 3;
+
+    $obj->minimum_version_required(undef);
+    my @versions = $obj->available_picard_versions;
+    ok(@versions, 'all available picard versions');
+
+    $obj->minimum_version_required($versions[2]);
+    my @versions_with_miniumum_required = $obj->available_picard_versions;
+    is(@versions_with_miniumum_required, 3, '3 available verions with minimum required version set');
```
tmooney: `versions` is missing an `s`.

apregier:
+1

lib/perl/Genome/Model/Tools/Picard/GenotypeConcordance.t
```diff
@@ -66,8 +66,8 @@ ok(-e $output_detailed_metrics, 'detailed metrics exist');
 my $filters = [
    qr(^# .*$),
 ];
-compare_ok($expected_detailed_metrics,$output_detailed_metrics, filters => $filters );
-compare_ok($expected_summary_metrics,$output_summary_metrics, filters => $filters );
+compare_ok($expected_detailed_metrics,$output_detailed_metrics, name => 'detailed metrics files mactch', filters => $filters );
```
tmooney: `match` has an extra `c`.

PR #900, opened 2015-08-03
Don't disconnect if we're not going to wait on the lock.

tmooney:
If we're not waiting this will return quickly and so we don't need to disconnect.


ebelter:
+1

brummett:
+1

PR #901, opened 2015-08-03
Add cle test

apregier:
This will be used by the jenkins script: https://github.com/genome/wugc-hudson/pull/26

### DIFF
--- a/lib/perl/Genome/Interfaces/Comparable/Command/Diff.pm
+++ b/lib/perl/Genome/Interfaces/Comparable/Command/Diff.pm
@@ -64,5 +64,10 @@ sub diffs_message {
     }
     return $diff_string;
 }
+
+sub has_diffs {
+    my $self = shift;
+    return (defined ($self->_diffs) && scalar(keys %{self->_diffs})) || 0;
+}
 1;
 
--- a/lib/perl/Genome/Interfaces/Comparable/Command/Diff.pm
+++ b/lib/perl/Genome/Interfaces/Comparable/Command/Diff.pm
@@ -67,7 +67,7 @@ sub diffs_message {
 
 sub has_diffs {
     my $self = shift;
-    return (defined ($self->_diffs) && scalar(keys %{self->_diffs})) || 0;
+    return (defined ($self->_diffs) && scalar(keys %{$self->_diffs})) || 0;
 }
 1;
 
--- a/lib/perl/Genome/Site/TGI/CleTest.pm
+++ b/lib/perl/Genome/Site/TGI/CleTest.pm
@@ -1,0 +1,201 @@
+package Genome::Site::TGI::CleTest;
+
+use strict;
+use warnings;
+use Genome;
+
+class Genome::Site::TGI::CleTest {
+    has_transient_optional => {
+        analysis_project => {
+            is => 'Genome::Config::AnalysisProject',
+        },
+    },
+};
+
+sub get_builds {
+    my $self = shift;
+    my $cmd0 = Genome::Config::AnalysisProject::Command::Create->create(
+        environment => "ad-hoc",
+        name => "CLE Test",
+        no_config => 1
+    );
+    my $analysis_project = $cmd0->execute;
+    $self->analysis_project($analysis_project);
+    my $menu_item1 = Genome::Config::AnalysisMenu::Item->get("9ab6e28f832a428393b87b171d444401");
+    my $menu_item2 = Genome::Config::AnalysisMenu::Item->get("3770b8510d5a459f9c0bb01fabf56337");
+    my $discovery_tag = Genome::Config::Tag->get(name => 'discovery');
+    my $followup_tag = Genome::Config::Tag->get(name => 'followup');
+    my $germline_tag = Genome::Config::Tag->get(name => 'germline');
+    my $cmd1 = Genome::Config::AnalysisProject::Command::AddMenuItem->create(
+        analysis_menu_items => $menu_item1,
+        tags => $discovery_tag,
+        analysis_project => $analysis_project,
+    );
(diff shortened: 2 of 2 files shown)


apregier:
Good `catch`, thanks!

lib/perl/Genome/Site/TGI/CleTest.pm
```diff
@@ -0,0 +1,203 @@
...
+    my $post_cqid_model_count = scalar @post_cqid_models;
+    my $expected_number_of_models = 7;
+    my $actual_number_of_models = $post_cqid_model_count - $pre_cqid_model_count;
+    unless ($actual_number_of_models == $expected_number_of_models) {
+        die $self->error_message("We expected CQID to create %s models and it actually created %s\n",
+                $expected_number_of_models,
+                $actual_number_of_models);
+    }
+
+    my @models = Genome::Model->get(analysis_projects => [$analysis_project]);
```
tmooney: The `analysis_projects` property is deprecated in favour of the singular `analysis_project`.

lib/perl/Genome/Site/TGI/CleTest.pm
```diff
@@ -0,0 +1,203 @@
...
+    my $analysis_project = $self->analysis_project;
+    unless (defined $analysis_project) {
+        die $self->error_message("Must run get_builds first");
+    }
+
+    my $discovery_subject = $self->get_sample_from_subject_mapping('discovery');
+    my $followup_subject = $self->get_sample_from_subject_mapping('followup');
+    my $germline_subject = $self->get_sample_from_subject_mapping('germline');
+
+    my @models_for_process = Genome::Model::SomaticValidation->get(analysis_projects => [$analysis_project],
```
tmooney: (Ibid.)

lib/perl/Genome/Site/TGI/CleTest.pm
```diff
@@ -0,0 +1,203 @@
...
+    );
+    unless ($diff_cmd->execute) {
+        die $self->error_message("Diff command failed to execute!\n");
+    }
+
+    if ($diff_cmd->has_diffs) {
+        return $diff_cmd;
+    }
+    return;
+}
```
tmooney: Now that this is a module, a `1;` should be added to the end.

lib/perl/Genome/Site/TGI/CleTest.pm
```diff
@@ -0,0 +1,204 @@
...
+        my $instrument_data = Genome::InstrumentData->get($id);
+        Genome::Config::AnalysisProject::InstrumentDataBridge->create(
+            analysis_project => $analysis_project,
+            instrument_data => $instrument_data,
+            status => 'new',
+        );
+    }
+
+    my $cmd5 = Genome::Config::AnalysisProject::Command::Release->create(
+        analysis_project => $analysis_project,
```
tmooney: This one is not a `Genome::Model` and will need to stay plural.

lib/perl/Genome/Interfaces/Comparable/Command/Diff.pm
```diff
@@ -64,5 +64,10 @@ sub diffs_message {
     }
     return $diff_string;
 }
+
+sub has_diffs {
+    my $self = shift;
+    return (defined ($self->_diffs) && scalar(keys %{self->_diffs})) || 0;
```
davidlmorton: missing $ on 'self'

lib/perl/Genome/Site/TGI/CleTest.pm
```diff
@@ -0,0 +1,204 @@
...
+        name => "CLE Test",
+        no_config => 1
+    );
+    my $analysis_project = $cmd0->execute;
+    $self->analysis_project($analysis_project);
+    my $menu_item1 = Genome::Config::AnalysisMenu::Item->get("9ab6e28f832a428393b87b171d444401");
+    my $menu_item2 = Genome::Config::AnalysisMenu::Item->get("3770b8510d5a459f9c0bb01fabf56337");
+    my $discovery_tag = Genome::Config::Tag->get(name => 'discovery');
+    my $followup_tag = Genome::Config::Tag->get(name => 'followup');
+    my $germline_tag = Genome::Config::Tag->get(name => 'germline');
```
davidlmorton: I would suggest looping over the names in a for loop here.

davidlmorton:
I think it would be clearer if all the wugc helper functions and such were in the genome repo, even if they are located somewhere in Site/TGI. That way all the logic in the wugc script about waiting on builds and processes could just be in an execute method on the test command.  It seems odd to make an object with a bunch of functions that are then called in some sequence in a script, I would prefer that the object have a method that called them itself (better encapsulation).

apregier:
Thanks, @davidlmorton I agree that encapsulation would be way better.  I am getting stuck on how many things from the `Library` in `wugc-hudson` are being called - making it a non-trivial refactor.  I think I'd like to see how it runs for a while before making that commitment, because if we end up changing it to run in separate jenkins builds like we talked about today, then we would want it split up anyway and may not need a lot of that stuff.  What do you think?

davidlmorton:
Rather than move the wugc-hudson code into genome, it might make more sense to move the functions in this class over there.  In either case, they ought to be together (the functions in this class, and the script that calls them).  This class isn't much of a class after all, its more of a collection of functions that want to be strung together in a sequence.  If the code that does that is in another repo, it just makes maintenance more cumbersome (imagine renames or changing the call signature of the functions).

ebelter:
I agree with @davidlmorton that this should live inside wugc-hudson. The CLE test is MGI specific, and would be better to maintain it where we maintain the other testing/CI code.

apregier:
Replaced by #908

PR #902, opened 2015-08-03
Account for the existence of other non-disabled statuses.

tmooney:
Without this change, the script might delete results that are also used by `inactive` config items.


ebelter:
+1

brummett:
+1

PR #903, opened 2015-08-03
Genotype Microarray: Remove instrument type param

ebelter:


tmooney:
+1 :statue_of_liberty:

PR #904, opened 2015-08-04
Only match on QC models that are "active".

tmooney:
Ideally this would key on the QC tag rather than going through model names, but this is better than accidentally finding ancient QC models.

### DIFF
--- a/lib/perl/Genome/Model/ReferenceAlignment.pm
+++ b/lib/perl/Genome/Model/ReferenceAlignment.pm
@@ -459,7 +459,12 @@ sub get_lane_qc_models {
         my $lane_qc_model_name = $self->default_lane_qc_model_name_for_instrument_data($instrument_data);
 
         my $existing_model = Genome::Model->get(name => $lane_qc_model_name);
-        if ($existing_model) {
+        my $has_current_qc =
+            $existing_model &&
+            $existing_model->analysis_project_bridges &&
+            $existing_model->analysis_project_bridges->config_profile_item &&
+            $existing_model->analysis_project_bridges->config_profile_item->is_current;
+        if ($has_current_qc) {
             $self->debug_message("Default lane QC model " . $existing_model->__display_name__ . " already exists.");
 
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Model/ReferenceAlignment.pm
```diff
@@ -459,7 +459,12 @@ sub get_lane_qc_models {
...
 
         my $existing_model = Genome::Model->get(name => $lane_qc_model_name);
-        if ($existing_model) {
+        my $has_current_qc =
+            $existing_model &&
+            $existing_model->analysis_project_bridges &&
+            $existing_model->analysis_project_bridges->config_profile_item &&
+            $existing_model->analysis_project_bridges->config_profile_item->is_current;
+        if ($has_current_qc) {
             $self->debug_message("Default lane QC model " . $existing_model->__display_name__ . " already exists.");
```
ebelter: Would a function on Model that returns if it is active be a more concise and reusable way to make this check?

tmooney: This should definitely be simplified!  For the moment it mirrors [a similar check in GenotypeMicroarray](https://github.com/genome/genome/blob/a4f871bd95af2f81dc6ccd509a26826e4bb29e38/lib/perl/Genome/Model/GenotypeMicroarray.pm#L177-L181).

I've got a branch that depends on this PR and #894 that will add a new status to models and will reduce the cascade of arrows here!

ebelter: :smile: I did +1 that PR!

tmooney:
I've included this commit inside #912 for expediency.  I'll close this PR.

@ebelter In the other PR Model has `is_disabled`. That's not *quite* what I want here for the moment, because I  want to exclude models that are not linked to Analysis Projects as well.  Let me know if you think some sort of `is_enabled` would be a good complement to this to further reduce the `->` count.

PR #907, opened 2015-08-05
Save a copy of the users hash.

tmooney:
This fixes a bug where someone else with the reference was manipulating the hash between passing it and us using it, making our users disappear.


ebelter:
+1

apregier:
+1 Wow this is subtle - I wonder where else this might be a problem but not necessarily cause a crash?

PR #908, opened 2015-08-05
Add cle test2

apregier:
This replaces #901 and goes along with https://github.com/genome/wugc-hudson/pull/26
This is admittedly a very simplistic config, and it's structure is highly coupled to the script in the jenkins pr.  But, it should at least make more sense to, say, update blessed builds in the genome repo rather than the jenkins repo.
When we need to put in another cle test (for example, for another protocol) we may want to revisit the structure of the config.


ebelter:
+1 otherwise!

tmooney:
+1 :fork_and_knife:

PR #909, opened 2015-08-05
Add ability to skip exome CNV calling in ClinSeq.

gatoravi:
Sometimes it doesn't make a lot of sense to do copy-number calling for e.g when the data is from targeted capture of a handful of genes etc and we'd like to turn off this step. This pull-request uses a parameter in the clin-seq processing profile to handle this.

### DIFF
--- a/lib/perl/Genome/Model/Build/ClinSeq.pm
+++ b/lib/perl/Genome/Model/Build/ClinSeq.pm
@@ -4,9 +4,10 @@ use warnings;
 
 use Genome;
 class Genome::Model::Build::ClinSeq {
-  is => ['Genome::Model::Build',
-    'Genome::Model::Build::ClinSeq::FileAccessors',
-    'Genome::Model::Build::ClinSeq::InputBuilds'],
+    is => ['Genome::Model::Build',
+           'Genome::Model::Build::ClinSeq::FileAccessors',
+           'Genome::Model::Build::ClinSeq::InputBuilds'
+          ],
 };
 
 sub special_compare_functions {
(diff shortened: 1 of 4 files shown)


tmooney:
Should all existing processing profiles have this enabled?  Either way, we can backfill after it's merged and the extra parameter will be ready once it's in a snapshot.

gatoravi:
Yes, all existing processing-profiles should have this parameter, I did backfill all of them already (there are only 4.) I was trying to test a model-group that required this code-change and also to make sure that this doesn't change existing results.

tmooney:
+1.

If the logic that's currently `$build->exome_build and $self->exome_cnv` becomes much more complicated, it might be useful to have a helper like `$self->should_run_exome_cnv` so it won't have to be kept in synch in so many places.

gatoravi:
Hi @tmooney, I'd like to do this but I'm unsure how to pull out the `$exome_build` from `$self` in the new method. Looks like `$exome_build` is obtained from `$build` which is passed to `_resolve_workflow_for_build`.

 Should I pass `$exome_build` as a parameter to  `should_run_exome_cnv`?

tmooney:
Yeah, it could be something like:
```perl
sub should_run_exome_cnv {
  my ($self, $build) = @_;
  return ($build->exome_build and $self->exome_cnv);
}
```
Alternatively it could be defined on the `$build` itself:
```perl
sub should_run_exome_cnv {
  my $self = shift;
  return ($self->exome_build and $self->processing_profile->exome_cnv);
}
```
Some other build classes are set up to generate accessors for processing profile parameters, which would simplify the above.

gatoravi:
Thanks for the idea @tmooney I've added this property to the `$build` object.

PR #910, opened 2015-08-05
Fix typo in method call in PP help generation

tmooney:
I'm not sure this condition is ever met in production, anyway.

### DIFF
--- a/lib/perl/Genome/ProcessingProfile/Command/Create/Base.pm
+++ b/lib/perl/Genome/ProcessingProfile/Command/Create/Base.pm
@@ -52,7 +52,7 @@ sub _get_help {
         return $help;
     }
     elsif ($self->_target_class_name->can($help_fn_name)) {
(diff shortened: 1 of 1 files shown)


brummett:
+1 for this change.

But it looks like ```_target_class_name()``` is something used/defined by Command::SubCommandFactory.  Does that mean this module's class needs to be a subclass of SubCommandFactory instead of Command::V2?

tmooney:
`Command::SubCommandFactory` [adds `_target_class_name`  to the commands](https://github.com/genome/UR/blob/b40400caea6deab59d7ed5923fc720b95e35fd96/lib/Command/SubCommandFactory.pm#L99) it generates.  Any commands generated by `Genome::ProcessingProfile::Command::Create` will thus have it, but any manually created inheritors would not.

Maybe it's better to remove this `elsif` block completely?

acoffman:
+1

PR #911, opened 2015-08-06
adding example file, changing 'reference-genome' param to 'reference-fasta',

chrisamiller:
and cleaning up extraneous comments


tmooney:
+1 :poultry_leg:

PR #912, opened 2015-08-06
Disabled configs imply disabled models

tmooney:
If these models are truly not wanted, running new builds for them is merely a waste of resources!


jeldred:
+1 coffman says it weird but ok

ebelter:
+1

PR #913, opened 2015-08-06
Remove FileSummary creation when allocations are purged

acoffman:
This will require a follow-up database migration to remove the file_summaries table.


tmooney:
+1 :tulip:

brummett:
+1

PR #914, opened 2015-08-06
Don't auto-retry MEMLIMIT errors and orphaned allocations.

tmooney:
Without manual intervention these builds will eventually fail again, possibly after wasting hours or days of CPU.


brummett:
+1

jweible:
+1

PR #915, opened 2015-08-06
Remove old `gmt somatic run-reports`.

tmooney:
It refers to some other modules that were removed in commit d08d6b84701442346ccfd9076c1be997c608732b in 2013, so it doesn't work, anyway!


brummett:
+1

acoffman:
+1

PR #916, opened 2015-08-06
Add CRC as a valid tumor common-name

gatoravi:


obigriffith:
+1 This is needed so that we can analyze some new OSCC tumors that come in the form of CRCs (conditionally reprogrammed cells).

tmooney:
+1 :+1:

gatoravi:
Merging, since this is very minor. :pager:

PR #917, opened 2015-08-06
QC framework build metrics

jasonwalker80:
Initial revision of a command to dump build QC metrics.


ernfrid:
Would be better if this had tests of the code itself.

+1

lib/perl/Genome/Qc/Command/BuildMetrics.pm
```diff
@@ -0,0 +1,70 @@
...
+    my $self = shift;
+
+    my @metrics;
+    for my $build ($self->builds) {
+        my @qc_results = grep {$_->isa('Genome::Qc::Result')} $build->results;
+        for my $qc_result (@qc_results) {
+            my $as = $qc_result->alignment_result;
+            my @instrument_data = $as->instrument_data;
+            if (@instrument_data > 1) {
+                $self->error_message('Please add support for merged alignment results and multiple instrument data QC results!');
```
apregier: You could easily support this by putting the instrument data ids in a comma-separated list (for example) - I think everything else is generic enough that you wouldn't need to change anything.

apregier:
+1 Code looks good to me, although I agree a test will help prevent future problems.  You may need to write a quick Genome::Test::Factory module for a qc result

lib/perl/Genome/Qc/Command/BuildMetrics.pm
jasonwalker80: I expect the metric names to be very different for results with multiple instrument data.  The generic separated value writer can handle missing values.  Another option is they become a separate list of `@merged_metrics` with a different writer.

I was waiting to know more once we have a QC config that defines both per read group and merged result QC in the same build.

jasonwalker80:
I started looking at making a qc result test factory and that escalated quickly.  I would need an AlignedBamResult, SoftwareResult and other objects to make it work.  Perhaps I need a nudge on how to make a Genome::Test::Factory::Qc::Result.  There seem to be many dependencies to set one up correctly.

jasonwalker80:
@susannasiebert is helping me through the Genome::Test::Factory::Qc::Result.  There are several changes to other test factories that are needed for this to work.

jasonwalker80:
@ernfrid , @apregier and @susannasiebert I pushed a few commits.  The first one is the necessary changes to make a test factory qc result.  An additional commit improves the AR factory to allow the use of the same build for all results.

apregier:
Would it work to simply call `$test_alignment_result->add_user`
instead of the rigmarole of passing the build around?  Or is it mostly an optimization so that only one test build is created?

lib/perl/Genome/Qc/Command/BuildMetrics.t
```diff
@@ -0,0 +1,68 @@
...
+
+my $pkg = 'Genome::Qc::Command::BuildMetrics';
+use_ok($pkg);
+
+my $cmd = Genome::Qc::Command::BuildMetrics->create(
+   builds => $test_build,
+);
+isa_ok($cmd,'Genome::Qc::Command::BuildMetrics');       
+ok($cmd->execute,'execute build-metrics command');
+
```
apregier: If you want to capture the output in a file and compare it, you could use compare_ok with the replace option to fill in the instrument data and build ids, which will change from run to run:
```
use Genome::Utility::Test qw(compare_ok);
my $replace = {
    INSTRUMENT_DATA_ID => $test_inst_data_1->id,
    BUILD_ID => $test_build->id
};
compare_ok($output, $expected_output, replace => $replace, 
    name => "Build metrics printed correctly");
```
Your expected output file would look like:
```
build_id\tinstrument_data_id\tmetric_A\tmetric_B\tmetric_C\tmetric_D
BUILD_ID\tINSTRUMENT_DATA_ID\t1\t2\t3\t4
```

jasonwalker80:
Passing the build was an optimization to avoid the unnecessary creation of additional models and builds. Your solution might work for this test, but it also seemed like a feature other tests might need.

jasonwalker80:
Update to command to add output file option.  Also added test to compare to expected output file.

apregier:
:+1:

lib/perl/Genome/Qc/Command/BuildMetrics.t
```diff
@@ -0,0 +1,93 @@
...
+use Test::More;
+use Genome::Utility::Test qw(compare_ok);
+
+use Genome::Test::Factory::InstrumentData::Solexa;
+use Genome::Test::Factory::InstrumentData::AlignmentResult;
+use Genome::Test::Factory::Model::SomaticValidation;
+use Genome::Test::Factory::Build;
+use Genome::Test::Factory::Qc::Result;
+
+use Sub::Install qw (reinstall_sub);
```
ernfrid: nitpick - space after qw breaks style from lines above.

lib/perl/Genome/Test/Factory/InstrumentData/AlignmentResult.pm
```diff
@@ -9,9 +9,17 @@ use Genome::Test::Factory::SoftwareResult::User;
...
 sub generate_obj {
     my $self = shift;
+    my %params = @_;
 
-    my $result_users = Genome::Test::Factory::SoftwareResult::User->setup_user_hash();
-    return Genome::InstrumentData::AlignmentResult::Bwa->__define__(@_, _user_data_for_nested_results => $result_users);
+    my $build = delete($params{'build'});
+    
+    my $result_users;
+    unless ($build) {
```
ernfrid: I find the ```unless ($build)``` construct to be touch difficult to interpret vs a more standard ```if ($build)``` (with proper rearrangement below of course).

ernfrid:
+1 one from me (I think my nitpicks are optional) with the caveat that I'm familiar with the Test::Factory paradigm.

PR #919, opened 2015-08-07
Inactivate AnP Config Files

jasonwalker80:
Initial revision of a command to inactivate config files.


tmooney:
+1

It'd be nice to have a `.t` to go along with this.  (The test for `DisableConfigFile.pm` should work with minimal changes.)

brummett:
I agree with needing a test.  It'll help in the future if this command gets additional functionality.

jasonwalker80:
I agree with the need for a test case.  I'll look at the DisableConfigFile.t and see where I get.

jasonwalker80:
Test added.

brummett:
+1

PR #923, opened 2015-08-11
Add script to prime testdb for cle test

apregier:
Not sure if this is the best place for this, but it needs to be in vc somewhere - open to suggestions!


tmooney:
How about `Genome/Site/TGI/CleTest/prime-db.pl`?  If more `CleTest`-related things are coming, it'd be nice to group them together.

ernfrid:
I agree with Tom. That sounds like a good idea.

____

PR #906, opened 2015-08-05
Rnaseq spawned workflows

davidlmorton:
This PR makes it so that all workflows spawned from an rnaseq model are run with WorkflowBuilder.  WorkflowBuilder is where we honor the environment variable that specifies the workflow be run using PTero.

This PR also adds the ability to view spawned workflows using 'genome process view' or 'genome model build view' commands.  Here is an example of the rnaseq test build, run using PTero:

```
Using libraries at /home/archive/dmorton/genome/lib/perl
'build' may require verification...
Resolving parameter 'build' from command argument 'a0781aaead9541d5966764191830a441'... found 1
=== Build ===
Build ID: a0781aaead9541d5966764191830a441      Build Status: Succeeded
Model ID: 13ba093990d34294b8219ee013f62c4a      Model Name: apipe-test-rnaseq
Run by: apipe-tester                            Processing Profile ID: 2762d4c6c0ed4590bd3424b6c213ed7d
Build Scheduled: 2015-08-03 14:12:09            Build Completed: 2015-08-03 15:24:19

Build Class: Genome::Model::Build::RnaSeq
Software Revision: /gscuser/dmorton/deployments/rnaseq_spawned_workflows/genome/lib/perl/Genome/Site/TGI/SiteLib:/gscuser/dmorton/deployments/rnaseq_spawned_workflows/genome/lib/perl:/gsc/scripts/opt/genome/snapshots/genome-3625/lib/perl:/etc/perl:/usr/local/lib/site_perl
Software Result Test Name(s): "dmorton_ptero_test_wb_2" (11), undef (4)
Data Directory: /gscmnt/gc7005/info/model_data/13ba093990d34294b8219ee013f62c4a/builda0781aaead9541d5966764191830a441

=== Workflow ===
   STAGE     STATUS              STARTED       DURATION  P-INDEX  NAME
          succeeded  2015-08-03T19:11:24       01:11:32           a0781aaead9541d5966764191830a441
 execute  succeeded  2015-08-03T19:11:37       00:16:27           . RnaSeq Alignment
   spawn  succeeded  2015-08-03T19:11:51       00:15:32           . . 7CpxtmkHS2uMxOHcpDZwkw
shortcut  succeeded  2015-08-03T19:11:58       00:00:08           . . . per-lane-tophat index #1
          succeeded  2015-08-03T19:12:08       00:00:00           . . . block
          succeeded  2015-08-03T19:12:09       00:13:50           . . . Alignment Dispatcher for 2891578700 (segment A)
[...]

PR #924, opened 2015-08-11
Alignment Dispatcher uses WorkflowBuilder

tmooney:
* A step on our quest to stop relying on `Workflow`!
* The old version had some UR optimizations built into it for the benefit of `PhenotypeCorrelation`.  I took those out in this conversion, but they could potentially be re-added if needed.
* This adds programmatic support for declaring an input `is_optional` in `Genome::WorkflowBuilder::DAG`s.  (The changes backing this were added in #906.)

### DIFF
--- a/lib/perl/Genome/InstrumentData/Composite/Workflow.pm
+++ b/lib/perl/Genome/InstrumentData/Composite/Workflow.pm
@@ -57,11 +57,7 @@ sub execute {
         $workflow->log_dir($self->log_directory);
     }
 
-    my @errors = $workflow->validate;
-    if (@errors) {
-        $self->error_message($_) for @errors;
-        die 'Errors validating workflow';
-    }
+    $workflow->validate;
 
     my @results = $self->_run_workflow($workflow, $inputs);
     $self->_result_ids(\@results);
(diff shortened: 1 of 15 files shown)


mkiwala:
@tmooney -- I plan to review this later today.  I'll see if I can get @davidlmorton to look at it too.

lib/perl/Genome/InstrumentData/Composite/Workflow.pm
```diff
@@ -110,9 +106,8 @@ sub _run_workflow {
     Genome::Sys->disconnect_default_handles;
 
     $self->debug_message('Running workflow...');
-    my $dag = Genome::WorkflowBuilder::DAG->from_xml($workflow->save_to_xml());
     my %inputs_hash = (@$inputs); # For some reason $inputs is an array-ref
```
davidlmorton: I kind of feel like this was a good opportunity to fix this.  It may be scope creep though.

davidlmorton:
+1 Sorry this took so long for me to review.  I still don't fully understand some of the template business, but that is carry-over from the old version of the code so I'm not too concerned about it.

lib/perl/Genome/InstrumentData/Composite/Workflow.pm
tmooney: Good point; I'll make a follow-up PR once this one is done.  I think it'll be easier to follow as a fresh request rather than tacked on here.

susannasiebert:
+1

tmooney:
I've rebased to resolve the merge conflicts.  If this passes I'll see about merging it!

PR #925, opened 2015-08-11
Remove unused variables and the logic to set them.

tmooney:
These are controlled by PP params in SomaticValidation.


brummett:
+1

dufeiyu:
+1

PR #926, opened 2015-08-11
Lend functionality to "is_output" build properties.

tmooney:
They will be mapped via `Genome::SoftwareResult::User` the way inputs are mapped via `Genome::Model::Input`.

This is intended to make it possible to simplify accessors like [this one](https://github.com/genome/genome/blob/045d09fa691caa7ff99f115485f98f978f5d24ac/lib/perl/Genome/Model/Build/SomaticValidation.pm#L94) and those immediately following it.

### DIFF
--- a/lib/perl/Genome/Model/Build.pm
+++ b/lib/perl/Genome/Model/Build.pm
@@ -2509,6 +2509,14 @@ sub _preprocess_subclass_description {
             $prop_desc->{'is_delegated'} = 1;
         }
 
+        if (exists $prop_desc->{'is_output'} and $prop_desc->{'is_output'}) {
+            $prop_desc->{'via'} = 'result_users';
+            $prop_desc->{'to'} = 'software_result';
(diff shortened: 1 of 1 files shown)


brummett:
+1 Looks ok on its own.

But this change would require some of the property metadata in the linked class description to be changed.  This change sets ```where => [ label => $prop_name]``` and the linked property has the label linked to something similar to, but not the same as the property name.

Also, I don't think you need to use ```exists``` before checking the truthiness of the hash value.  The key won't get autovivified for a check only one level deep.

tmooney:
Indeed--I held off changing these properties for now because they don't match their labels.  I was adding this for use in adding some new properties (to a different build subclass).

As to the autovivification--good point!  I've removed it from all the cases here.

acoffman:
+1

Based on the fact that it will resolve to a SoftwareResult, I might prefer it be called `is_result` rather than `is_output`, but on the whole `is_output` gives nicer symmetry with other things in our codebase. I'm good either way!

PR #927, opened 2015-08-12
Disk Volume Sync Command Updates

ebelter:


lib/perl/Genome/Disk/Command/Volume/SyncUsage.pm
```diff
@@ -27,44 +25,30 @@ class Genome::Disk::Command::Volume::SyncUsage {
...
-                $self->debug('Sync total kB...');
-                $volume->sync_total_kb(%args);
-            }
-            if ($self->unallocated_kb) {
-                $self->debug('Sync unallocated kB...');
-                $volume->sync_unallocated_kb(%args);
-            }
-            $transaction->commit or die 'Failed to commit volume! '.$volume->mount_path;
+        $self->status_message('Syncing volume: %s', $volume->mount_path);
+        for my $type (qw/ total unallocated /) {
```
tmooney: Could this loop be replaced with a call to `$volume->sync_usage`?

(Maybe the `status_message` would then go inside there?)

ebelter: Sure. The previous version had status messages in the command and volume.pm I hesitated using this function because then the types of things to sync (total./unallocated) would need to be passed in. That would create 2 places where the params live and wouldneed to be maintined. One in the command and the other the function in volume.pm.

tmooney: Makes sense.  I wasn't considering that the command has the option to skip one type of thing while still updating the other.  I'd rather take that out than pass the options through :smile:

ebelter: Does it make sense to sync one and not the other?

tmooney: It might make sense if you think `df` is expensive and you trust your volumes not to change sizes.  I don't trust our volumes not to change size, though! 

(It's fine by me if you want to leave it as-is, or if you want to take those options out.)

ebelter: I will leave it as is.

PR #929, opened 2015-08-12
Make GATK Base more like Picard Base

tmooney:
* GATK Base will now build the command lines and execute them, so logic to do this need not be duplicated in each subcommand.
* A few legacy tools (`GermlineIndel`, `GermlineIndelUnifiedGenotyper`, and `SomaticIndel`) are not updated in this PR.  They don't work with recent versions of GATK, anyway.


brummett:
+1

This is a pretty slick way of managing the command-line params

ebelter:
The tests checking the command have been removed, and there is no testing of command building added to Base.t I like what is done here, but would like some testing around these methods in the base class.

tmooney:
@ebelter I've added a test for the Base class!

ebelter:
+1

PR #930, opened 2015-08-13
verifyBamId: Make the --ignoreRG param optional

susannasiebert:
For production QC we want to be able to run verifyBamId by read group.


tmooney:
+1 :deciduous_tree:

jasonwalker80:
+1

PR #931, opened 2015-08-13
Ptero runs workflows with optional inputproperties

davidlmorton:


davidlmorton:
Integration tests do pass when run by hand as apipe-tester:

```
$ PTERO_WORKFLOW_SUBMIT_URL=http://workflow.apipe-deis.gsc.wustl.edu/v1/workflows /usr/bin/prove -v Ptero/IntegrationTest.pl
```

These tests are run automatically by jenkins at https://apipe-ci.gsc.wustl.edu/view/Ptero/job/ptero_genome_integration_tests_1/ but not until the perl tests have passed, hence after the PR has been merged.

lib/perl/Genome/WorkflowBuilder/Detail/Operation.pm
```diff
@@ -131,6 +131,18 @@ sub get_xml {
...
+sub remove_links_from_unspecified_input {
+    return;
+}
+
+sub remove_links_to_unused_input_properties {
+    return;
+}
+
+sub recurse_do {
+    return;
```
tmooney: I think this would make more sense as `return $self->$method_name;`, though practically this currently saves us a method call.

tmooney:
+1 :strawberry:

PR #932, opened 2015-08-13
Remove allocations in test db

brummett:
This command inspects the current test-db and it's template to determine which allocations were made since the test-db was created.  These allocations are then deleted.


lib/perl/Genome/Model/Command/Admin/RemoveDiskAllocationsFromTestdb.pm
```diff
@@ -0,0 +1,307 @@
...
+    my @allocations = $self->collect_newly_created_allocations();
+    $self->report_allocations_to_delete(@allocations);
+    $self->delete_allocations(@allocations);
+    return 1;
+}
+
+sub is_running_in_test_env {
+    my $self = shift;
+
+    return $ENV{XGENOME_TESTING};
```
tmooney: Should this be examining `Genome::Config::get('testing')` instead of the environment variable?

lib/perl/Genome/Model/Command/Admin/RemoveDiskAllocationsFromTestdb.t
```diff
@@ -0,0 +1,253 @@
...
+#
+#    no warnings 'redefine';
+#
+#    local *Genome::Model::Command::Admin::RemoveDiskAllocationsFromTestdb::_make_iterator_for_template_allocations = sub {
+#        make_allocation_iterator_from_list_with_kb_requested(b => 1, c => 2);
+#    };
+#    local *Genome::Model::Command::Admin::RemoveDiskAllocationsFromTestdb::_make_iterator_for_database_allocations = sub {
+#        make_allocation_iterator_from_list_with_kb_requested(a => $three_gb_in_kb, b => 1, c => 2, d => $one_gb_in_kb);
+#    };
+#};
```
tmooney: Is this commented-out code still needed?

lib/perl/Genome/Model/Command/Admin/RemoveDiskAllocationsFromTestdb.pm
brummett: Maybe? yes?  Should all these env variable access go through Genome::Config?

tmooney: I vote yes, unless there's a reason to go outside the configuration system.

brummett:
Added new commits to use Genome::Config and remove cruft

lib/perl/Genome/Model/Command/Admin/RemoveDiskAllocationsFromTestdb.pm
```diff
@@ -0,0 +1,308 @@
...
+    }
+    return \%connection;
+}
+
+sub execute {
+    my $self = shift;
+
+    unless ($self->is_running_in_test_env) {
+        die $self->error_message('Must be run within a test environment created with genome-test-env');
+    }
```
apregier: I think these will running under genome-env, not genome-test-env, because we want to be committing.  Or maybe I am unclear on the consequences of genome-test-env vs. genome-env?

lib/perl/Genome/Model/Command/Admin/RemoveDiskAllocationsFromTestdb.pm
```diff
@@ -0,0 +1,308 @@
...
+}
+
+# Genome::Process objects delete their own allocations
+sub _delete_processes {
+    my($self, $processes) = @_;
+
+    foreach my $process ( @$processes ) {
+        my $allocation = $process->disk_allocation;
+        $self->_print_debug_message_about_deleting_allocation($allocation, $process);
+        $process->delete();
```
apregier: I don't understand why this is necessary - the point of the script is to remove the files in the allocation - the db objects are going to get blown away anyway, so we don't really need to worry about dependencies?

lib/perl/Genome/Model/Command/Admin/RemoveDiskAllocationsFromTestdb.pm
brummett: I suppose that's true.  genome-test-env does everything genome-env does, plus sets no-commit to true, right?

As far as I can tell, the only way to tell if you're running in any kind of genome-env is to look at the XGENOME_DS_GMSCHEMA_SERVER env var directly.

lib/perl/Genome/Model/Command/Admin/RemoveDiskAllocationsFromTestdb.pm
```diff
@@ -0,0 +1,308 @@
...
+
+    return Genome::Config::get('testing');
+}
+
+sub collect_newly_created_allocations {
+    my $self = shift;
+
+    my $tmpl_allocations = $self->_make_iterator_for_template_allocations();
+    my $db_allocations = $self->_make_iterator_for_database_allocations();
+
```
apregier: Holy cats we are going to have to be careful with this.  By comparing the test db to the template rather than 'production', one danger would be that someone (e.g. me) would have gone into the test db and possible mucked around, maybe using fill_db to load some extra things from production, etc. etc. This would delete those things.  Maybe the answer to this is "Allison don't muck around with test_filldb!"  Just something to keep in mind.

lib/perl/Genome/Model/Command/Admin/RemoveDiskAllocationsFromTestdb.pm
```diff
@@ -0,0 +1,308 @@
...
+        }
+        unless (defined $next_db_allocation) {
+            $next_db_allocation = $db_allocations->();
+        }
+
+        last unless defined($next_db_allocation);
+
+        if (!defined($next_tmpl_allocation)
+            or
+            $next_tmpl_allocation->id gt $next_db_allocation->id
```
apregier: just checking that gt will always use the same ordering as ORDER BY ?

lib/perl/Genome/Model/Command/Admin/RemoveDiskAllocationsFromTestdb.pm
brummett: It's not strictly necessary, but doing it this way avoids an ugly message about a constraint violation if the Process's allocation gets deleted without deleting the Process.

lib/perl/Genome/Model/Command/Admin/RemoveDiskAllocationsFromTestdb.pm
brummett: The use-case here is that a test DB is made from a template, stuff is done storing the changes in the test DB.  Any rows in the test DB that aren't in the template must have been created during testing and should be cleaned up.

Checking whether it exists in the production DB seems prudent.  I'll need to figure out how to get the DB connection info for production out of Genome::Config even when the env var is set to override it.

lib/perl/Genome/Model/Command/Admin/RemoveDiskAllocationsFromTestdb.pm
brummett: Short answer: yes.

These UUID IDs are stored in the DB as text, so Postgres sorts them as text. This is the same comparison mechanism that UR's object loader uses when comparing non-numeric IDs for in-memory objects to objects loaded from the DB.

lib/perl/Genome/Model/Command/Admin/RemoveDiskAllocationsFromTestdb.t
```diff
@@ -0,0 +1,242 @@
...
+sub make_allocation_iterator_from_list {
+    my @list = @_;
+    return sub {
+        my $alloc_id = shift @list;
+        return unless defined $alloc_id;
+        return Genome::Disk::StrippedDownAllocation->new(id => $alloc_id, kilobytes_requested => 1);
+    };
+}
+
+sub make_allocation_iterator_from_list_with_kb_requested {
```
apregier: this helper method isn't used -maybe you intended to?

lib/perl/Genome/Model/Command/Admin/RemoveDiskAllocationsFromTestdb.pm
apregier: Yeah, the concept of 'production db' is a little bit strange - that there is one canonical db and then a bunch of separate but equal test dbs.  Maybe the connection information could be an optional input to the command, and if specified it will do a sanity check before deleting anything?

brummett:
Added new commits to change the genome-env detection and check for allocation in the production DB before deleting them

lib/perl/Genome/Model/Command/Admin/RemoveDiskAllocationsFromTestdb.pm
```diff
@@ -0,0 +1,348 @@
...
+    my @allocations = $self->collect_newly_created_allocations();
+    $self->report_allocations_to_delete(@allocations);
+    $self->delete_allocations(@allocations);
+    return 1;
+}
+
+sub is_running_with_test_db {
+    my $self = shift;
+
+    return (! ! $self->get_template_name_for_database_name($self->database_name));
```
apregier: The double !! just casts it into a boolean?

apregier:
+1 I feel much safer about using this

lib/perl/Genome/Model/Command/Admin/RemoveDiskAllocationsFromTestdb.pm
brummett: yes

lib/perl/Genome/Model/Command/Admin/RemoveDiskAllocationsFromTestdb.pm
```diff
@@ -0,0 +1,348 @@
...
+                                           { AutoCommit => 0, RaiseError => 1, PrintError => 0 });
+    }
+    return $self->{$cache_key};
+}
+
+my @ENV_VARS_FOR_TEST_DB = qw( XGENOME_DS_GMSCHEMA_LOGIN XGENOME_DS_GMSCHEMA_AUTH XGENOME_DS_GMSCHEMA_SERVER );
+sub _dbh_for_production_db {
+    my $self = shift;
+
+    delete local @ENV{ @ENV_VARS_FOR_TEST_DB };
```
tmooney: `Genome::Config::spec($key)` returns a `Genome::ConfigSpec` object which has `->env` to provide the environment variable name.  There is also `Genome::Config::default_value($key)` that will return the default value directly.

brummett:
Added a new commit to use ```Genome::Config::spec()``` to convert the config key to its env var.

I'd rather have used ```default_value()```, but it returned undef for all three of these config keys

tmooney:
+1.  These probably do not have defaults defined, but are relying on the values in our site configuration instead.

PR #933, opened 2015-08-14
Ignore strelka subdirectory

apregier:
The important strelka results will be diffed in the vcf files.
The remaining files in this directory are intermediate files
and contain a lot of non-diffable items like times and file
paths.

Our current model test for somatic validation doesn't run strelka, but the ones we want to use for the cle regression test do.

### DIFF
--- a/lib/perl/Genome/Model/Build/SomaticValidation.pm
+++ b/lib/perl/Genome/Model/Build/SomaticValidation.pm
@@ -247,6 +247,8 @@ sub dirs_ignored_by_diff {
         variants/sv/union
         validation/small_indel/realigned_bams
         /validation/small_indel
+        variants/snv/strelka
+        variants/indel/strelka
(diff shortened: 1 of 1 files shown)


ebelter:
+1

dufeiyu:
Can we at least diff_compare snvs.hq.bed and indels.hq.bed ?

ernfrid:
@dufeiyu - I'm not as familiar with the diff process as I'd like to be yet, but don't the BED files exist outside of the "output" directory and thus would be diffed?

dufeiyu:
@ernfrid those bed files are combined bed from different callers. Anyway, Alliosn's solution should work by ignoring the whole strelka/output dir. Those .vcf.gz and .hq diff will be handled properly by diff_vcf_gz and diff_hq methods in special_compare_functions.

PR #934, opened 2015-08-14
Deduplicate between config and prime-db script

apregier:
Some of the items from config are also needed in the prime db script
We might consider putting all of the items needed in the prime db script
in the config, but that might also be confusing, I'm not sure

### DIFF
--- a/lib/perl/Genome/Site/TGI/CleTest/prime-db.pl
+++ b/lib/perl/Genome/Site/TGI/CleTest/prime-db.pl
@@ -10,6 +10,7 @@
 # 5. use `test-db template create` to create a new template
 # 6. Use the new template in the jenkins configuration
 use Genome;
+use Genome::Site::TGI::CleTest;
 use List::MoreUtils qw(uniq);
 
 #things we need to load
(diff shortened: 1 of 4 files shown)


ernfrid:
FWIW, I'd think moving all of it to the config would be preferable. At least then it is ALL in one place in terms of database ids or other entities that we may need. If you're worried it will be confusing in the config what is what, I'd just clearly separate our the prime-specific configuration and throw a comment in there. That would be clear enough for me at least.

+1

lib/perl/Genome/Site/TGI/CleTest/prime-db.pl
```diff
@@ -143,38 +144,14 @@ sub load_users_by_username {
...
-        2893815448
-        2893815449
-        2893815451
-        2893815484
-        2893815485
-        2893815489
-        2893815492
-    );
+
+    use Genome::Site::TGI::CleTest;
```
tmooney: This could be located up top with the other `use` statements, since all occur at "compile time".

tmooney:
+1, yay deduplication!

PR #936, opened 2015-08-17
Attempt to fix MutationSpectrumSequenceContext 

gatoravi:
This is an attempt to fix a sporadic diff in the test-builds when running this tool. Instead of writing to an existing file, we now write to a temporary file and copy that over at the end.


ebelter:
+1

tmooney:
+1 for this change.

`Genome::Model::ClinSeq::Command::CreateMutationSpectrum` should have an LSF resource string added to reserve temp space when this is run in the build.

gatoravi:
Ok, the param has been added in #937

PR #937, opened 2015-08-17
Add lsf param to CreateMutationSpectrum

gatoravi:
See #936

### DIFF
--- a/lib/perl/Genome/Model/ClinSeq/Command/CreateMutationSpectrum.pm
+++ b/lib/perl/Genome/Model/ClinSeq/Command/CreateMutationSpectrum.pm
@@ -61,7 +61,7 @@ class Genome::Model::ClinSeq::Command::CreateMutationSpectrum {
     ],
     has_param => [
         lsf_resource => {
-            default => "-R 'select[tmp>1000] span[hosts=1] rusage[tmp=1000]' -M 16000000",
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Model/ClinSeq/Command/CreateMutationSpectrum.pm
```diff
@@ -59,6 +59,11 @@ class Genome::Model::ClinSeq::Command::CreateMutationSpectrum {
             default => 0,
         },
     ],
+    has_param => [
+        lsf_resource => {
+            default => "-R 'select[tmp>1000] span[hosts=1] rusage[tmp=1000]' -M 16000000",
```
tmooney: Does this step need 16GB of RAM?  If so, it should also indicate `mem=16000` in the select and rusage.  If not, the `-M` parameter can be removed.

ebelter: The rusage should use gtmp (MGI inhouse resource] instead of tmp, and is indicated in Gb. This would reserve 1 Gb of tmp space:
```
rusage[gtmp=1]
```

gatoravi:
You're right 16 GB is overkill! I've changed this and the param string now uses `gtmp`.

ebelter:
+1

tmooney:
+1 :+1:

PR #938, opened 2015-08-17
Don't try to queue "Buildless" models.

tmooney:
We don't usually run this on unbuilt models, so this hasn't been a big issue so far.  But it's better to not queue models that will inevitably fail!


ebelter:
+1

acoffman:
+1

PR #939, opened 2015-08-17
Handle insert_centric Option for Bam Readcount

ebelter:
The bam_readcount insert_centric option was not introduced until version 0.5. This PR adds a check if the bam readcount version supports  the option before using it in the GMT Analysis Coverage BamReadcount class and asserts it is supported in the GMT Sam ReadCount command.

### DIFF
--- a/lib/perl/Genome/Model/Tools/Analysis/Coverage/BamReadcount.pm
+++ b/lib/perl/Genome/Model/Tools/Analysis/Coverage/BamReadcount.pm
@@ -476,10 +476,8 @@ sub execute {
             %params,
             output_file       => "$tempdir/readcounts_indel",
             region_list       => "$tempdir/indelpos",
+            insertion_centric => 1,
         );
-        if ( $readcount_cmd->version_has_insertion_centric( $self->bam_readcount_version ) ) {
-            $readcount_cmd->insertion_centric(1);
-        }
         unless($readcount_cmd->execute) {
             $self->error_message("Failed to execute.");
             die $self->error_message;
--- a/lib/perl/Genome/Model/Tools/Analysis/Coverage/BamReadcount.pm
+++ b/lib/perl/Genome/Model/Tools/Analysis/Coverage/BamReadcount.pm
@@ -476,8 +476,10 @@ sub execute {
             %params,
             output_file       => "$tempdir/readcounts_indel",
             region_list       => "$tempdir/indelpos",
-            insertion_centric => 1,
         );
+        if ( $readcount_cmd->version_has_insertion_centric( $self->bam_readcount_version ) ) {
+            $readcount_cmd->insertion_centric(1);
+        }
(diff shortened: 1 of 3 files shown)


tmooney:
+1 :dolls:

ernfrid:
This all looks good, but the results will not be correct with the current code and version 0.4. There were significant chunks of code revolving around the counting of indels that I removed when I released 0.5, foolishly thinking this wasn't run in pipelines anywhere (or maybe it wasn't at the time).

These need to be added back in if we're going to support 0.4. I'm open to recommendations on if that's a good idea and how best to do it.

jasonwalker80:
Perhaps we should NOT support calling 0.4 with the -i option.  Builds should fail and we should handle it by updating the config for the AnP or PP used by the analyst.

ebelter:
@jasonwalker80  @ernfrid I am up for whatever you gentlemen decide. Not supporting 0.4 seems the easiest.

jasonwalker80:
@ernfrid please confirm.  The -i option is passed to version 0.4 when running the false-positive indel filter that we all agree should no longer be used.  Therefore, if 0.4 is called with the -i option, it should fail.

If we were to support indel centric calling with 0.4, it would require adding some old code back to genome.  We prefer not to do this.  Also, please confirm that the old Genome code was only specific to indels.

Other false-positive SNV filters require 0.4 because the output bam-readcount format has changed.  Also testing with 0.5 or later did not produce satisfactory results.

Summary:  Support processing profiles with bam-readcount 0.4 for SNV filters only.  Indel filters should be removed from the processing profile and other bam-readcount versions in the PP should be updated to the latest version.

ebelter:
@jasonwalker80 @ernfrid I believe reverting: https://github.com/ebelter/genome/commit/551627d2f81c17d56ecd77981fb09dcb30cc8c60
to blindly indicate insertion centric to and then let Genome::Model::Tools::Sam::Readcount throw the error about 0.4 and -i wehn buildign the bam-readcount command. Previously, the command was just assembled, and the stderr was sent to /dev/null. This made it difficult to debug this problem.

ernfrid:
@ebelter - That seems smart to me. I like that idea as a good solution to the current complexity.

ernfrid:
Here is what I think is going on:

1) We still need to use 0.4 with the false-indel filter. It shouldn't 
use -i if used with this software. If we wanted to support a newer 
bam-readcount version and use the false-indel filter (but we don't) then 
the false-indel filter would need to be updated. I think we are in 
agreement that this filter should ultimately be dropped.
2) The false-positive filter (for SNVs) should be ok with newer versions 
of bam-readcount, but I am not sure. Still need to check with @cmiller.
3) I can confirm that the code I dropped was only to do with indels.

Dave

____

ebelter:
@jasonwalker80 @ernfrid Commit reverted.

PR #940, opened 2015-08-18
updating default version to 0.4 to fix ensembl transcript issues

chrisamiller:


tmooney:
+1, this tool doesn't seem to be used by any other modules.  (`ClinSeq` is calling the `pairoscope` executable by name.)

PR #942, opened 2015-08-18
Reconcile Genotype: Make a lookup hash for the attributes.

tmooney:
This avoids repeatedly searching through a list with tens of thousands of items!


brummett:
It has a purpose; it can stay.

+1

ebelter:
+1

PR #943, opened 2015-08-19
Add required users to Ensembl import runner.

tmooney:
I guess we haven't tried to run this since they became required!


jasonwalker80:
+2

PR #944, opened 2015-08-19
Rewrite get_merged_bam_to_revivify_per_lane_bam method to meet the need of removing superseded merged bams

dufeiyu:
The original get_merged_bam_to_revivify_per_lane_bam will get smallest unarchived merged bam to revivify per lane bam. It will fail after we remove superseded merged bam. This rewrite will get whatever bam that is available.

### DIFF
--- a/lib/perl/Genome/InstrumentData/AlignmentResult.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult.pm
@@ -1821,10 +1821,6 @@ sub revivified_alignment_bam_file_path {
     my $revivified_bam = File::Spec->join($temp_allocation->absolute_path, 'all_sequences.bam');
     my $merged_bam = $self->get_merged_bam_to_revivify_per_lane_bam;
 
-    unless ($merged_bam and -s $merged_bam) {
-        die $self->error_message('Failed to get valid merged bam to recreate per lane bam '.$self->id);
-    }
-
     my $bam_header_path = $self->create_bam_header;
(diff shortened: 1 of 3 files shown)


lib/perl/Genome/InstrumentData/AlignmentResult.pm
```diff
@@ -1901,19 +1897,32 @@ sub _get_uuid_string {
...
 sub get_merged_bam_to_revivify_per_lane_bam {
     my $self = shift;
-    my $merged_result = $self->get_smallest_merged_alignment_result($self->get_unarchived_merged_alignment_results);
+    my $merged_bam;
 
-    unless ($merged_result) {
-        $merged_result = $self->get_smallest_merged_alignment_result($self->get_merged_alignment_results);
-        unless ($merged_result) {
-            die $self->error_message('Failed to get archived merged result for per lane alignment '.$self->id);
+    for my $mr ($self->get_unarchived_merged_alignment_results) {
```
tmooney: This could be changed to `get_active_merged_alignment_results`.  If it filters on `$_->status eq 'active'` instead of `$_->is_archived`, it will exclude both archived and purged results.

dufeiyu: good point. I will make the change.

ebelter:
+1 except the else block comment, which isn't imperative.

tmooney:
+1 :rocket:

PR #947, opened 2015-08-20
Trio process fixes

apregier:
These changes are needed to get consistent process outputs from the cle regression test process.

### DIFF
--- a/lib/perl/Genome/Interfaces/Comparable/Command/Diff.pm
+++ b/lib/perl/Genome/Interfaces/Comparable/Command/Diff.pm
@@ -58,6 +58,7 @@ sub diffs_message {
     my $self = shift;
     my $diffs = $self->_diffs;
     my $diff_string = "DIFFERENCES FOUND:\n";
+    $diff_string .= sprintf(" Comparing new object %s to blessed object %s\n", $self->new_object->id, $self->blessed_object->id);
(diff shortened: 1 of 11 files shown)


lib/perl/Genome/Interfaces/Comparable/Command/Diff.pm
```diff
@@ -58,6 +58,7 @@ sub diffs_message {
     my $self = shift;
     my $diffs = $self->_diffs;
     my $diff_string = "DIFFERENCES FOUND:\n";
+    $diff_string .= sprintf(" Comparing new object %s to blessed object %s\n", $self->new_object->id, $self->blessed_object->id);
```
tmooney: It's a little strange that executing this command will print this line twice if there are diffs.

apregier: Yeah, the main thing is that this one will make it into the e-mail, whereas the other one is just a status message and will only be in the jenkins console log

tmooney: Fair enough!  It is nice to know the info in both places :smile:

tmooney:
+1.

It looks like the `AlignmentStatsSummary` and `CoverageStatsSummary` commands might be ripe for deduplication!

PR #948, opened 2015-08-20
Handle error rate

apregier:
This pr is a little strange - I am fixing how the alignment stats summary prints the error rate when it is not available (it should print 'NA'), while at the same time causing the process diff to ignore that column.  The reason for ignoring it is that we don't have a way to make a blessed process for the cle regression test in production that would have 'NA' in this column - because the value is available.  However, in our clean test db that is used for the test, this value will not be available and should be NA (this is actually what is expected for most cle data - the error rate will not be available in this report).
If/when we update the cle processing profile to produce error rate information, and update the alignment stats report to access the error rate information from a place that will actually be produced in the cle, we can remove this exception to the diff.


lib/perl/Genome/Model/SomaticValidation/Command/AlignmentStatsSummary.pm
```diff
@@ -144,7 +144,8 @@ sub _alignment_metrics_from_result {
         $total_bases += $lane->total_base_count;
         $total_mapped_bases += $lane->total_aligned_base_count;
         my $filt_error_rate_avg;
-        if ($mismatches ne "NA" and $lane->instrument_data->can("filt_error_rate_avg")) {
+        if ($mismatches ne "NA" and $lane->instrument_data->can("filt_error_rate_avg")
+            and defined($lane->instrument_data->can("filt_error_rate_avg"))) {
```
tmooney: This last one should call `filt_error_rate_avg` instead of `can`.

apregier: Thanks for the catch - awesome to be introducing bugs and removing the test that would potentially catch them at the same time :-\

acoffman:
+1

tmooney:
+1 :camel:

ernfrid:
@apregier - should we rethink the strategy here on simply ignoring these columns in the alignment summary? Your comment about removing tests make me much less comfortable with this approach than I was this morning.

apregier:
The right thing to do may be to write a unit test - then we are not depending on the cle test for this particular detail.

apregier:
@ernfrid I just pushed a test case that failed before the change (it showed an error rate of 0.00) and now passes, showing 'NA'.  Does that make you feel better?

apregier:
fwiw it makes _me_ feel better

ernfrid:
Yay! Makes me feel better too.

PR #949, opened 2015-08-20
Extract metrics_for_build subroutine for programmatic convenience.

tmooney:
I heartell it may be useful to call this command from within Perl.  In that case it's more convenient to get the hashes back directly instead of printing and re-parsing the output.


brummett:
+1

acoffman:
+1

PR #950, opened 2015-08-20
Add path for 0.1.0 freeze of speedseq gms branch

acoffman:
The new version of SpeedSeq (v 0.1.0) from the GMS branch is frozen at `/gscmnt/sata849/info/speedseq_freeze/v3/speedseq`

It was built with `make align realign sv` and appears to be working. I also believe that I have updated the config file appropriately (though I would appreciate someone double checking). This PR adds it as an available version for use in the GMS.


tmooney:
+1 The new version exists!

`speedseq.config` looks to be faithfully changed from `v2` to `v3`.  (It seems like most of the lines should use `$SPEEDSEQ_HOME` instead of repeating the full path.)

susannasiebert:
We should update the speedseq versions in the various Speedseq tests. This will ensure that the config is indeed setup correctly for speedseq to run. Better yet, maybe these tests should run on all speedseq versions but that might be outside of the scope of this PR.

acoffman:
Good call!

jasonwalker80:
+1

PR #952, opened 2015-08-21
Get test passing for new version of GATK

brummett:
When GATK 3.4 was installed, this test needed to be updated to match it


dufeiyu:
+1  this need a early merge so we can have a possible snapshot promotion Monday morning.

ebelter:
+2

brummett:
Merging early to get tests passing

PR #953, opened 2015-08-21
DV2: Use config to get lsf resources

ebelter:
This is in response to many failed builds for the Cyclin-E project. This will allow users to run builds without a custom snapshot.


tmooney:
We should declare environment variables for these (like [for the dispatcher](https://github.com/genome/genome/blob/04a6b72d7dfdafd51ab6b602621527849ebfbc19/etc/genome/spec/lsf_resource_dv2_dispatcher.yaml#L2)) so they're easy to specify when launching a build.

(Once we stop needing to rely on old snapshots hopefully we can take the "X"s out of the variable names again!)

ebelter:
@tmooney ENVs added.

tmooney:
+1 :circus_tent:

brummett:
+1

PR #951, opened 2015-08-21
translated_unprocessed_pseudogene added to AnnotationStructures

jasonwalker80:
Add translated_unprocessed_pseudogene to the list of coding_pseudogene biotype classes

PR #954, opened 2015-08-21
More annotation structure updates to biotype map.

jasonwalker80:
Several new gene biotypes added to the AnnotationStructures biotype class map.


jasonwalker80:
New commit to include changes from #951 which where accidentally merged and reverted in #955 .  All to avoid merge conflicts.

tmooney:
+1, as best I can tell these match [the docs](http://www.gencodegenes.org/gencode_biotypes.html)!

The existing `ncrna` list is alphabetized (albeit with capitals sorted before minuscules).  It might be preferable to keep them ordered.

jasonwalker80:
alphabetical order added.  need one more +1 and tests to pass.

apregier:
+1

PR #956, opened 2015-08-24
Defer reporting "Build Needed" if not enough instrument data assigned.

tmooney:
A build without instrument data assigned for each sample on the model will fail, so report it as "Buildless" rather than "Build Needed". [Note that if there are older builds, it will report their status rather than "Buildless" or "Build Needed".]

This will have the effect of making CQID wait to launch builds until data for both samples is present.

We currently allow starting the build in this situation since it may be desired to expedite the alignments for one of the samples, but this is written to be used as-is for a start validation later if desired.


jweible:
+1

PR #957, opened 2015-08-25
Microarray cnv update

gatoravi:
This module currently only runs on tumor-normal pairs. In cases where the normal is not available or when we are only interested in a single sample, this tool module now call CNVs based off the logR values for that sample alone.


ebelter:
Instead of merging, maybe rebase would be better to catch your branch your with master.

gatoravi:
Hi @ebelter I think I've addressed most of your comments except for the 2 listed below, thanks for the great feedback.

1. Looks like `compare_ok` works on single files and not directories, would it be better to loop through every file in the directory and run `compare_ok` on them.
2. I would really split that commits into two separate commits but I really suck at git and git-surgery might be bloody, but I'll try to get this PR rebased soon.(curious - What would be the advantage of doing that in this case?)

ebelter:
Yes - I missed that the diff was on the entire directory.  We do not have a compare_dir_ok yet!

If rebase was used instead of merging origin/master into this branch, there would be a cleaner project history. You would not end up with the two 'Merge remote-tracking...' commits. This is speaking specifically for developing locally on your new feature branch.

gatoravi:
Ok great, I'll remember rebase for the next time.

ebelter:
+1 NP! We are always available!

lib/perl/Genome/Model/ClinSeq/Command/MicroarrayCnv.pm
```diff
@@ -321,14 +412,21 @@ sub run_cnview {
...
+            my $cnview_cmd = Genome::Model::Tools::CopyNumber::CnView->create(annotation_build => $self->annotation_build_id, cnv_file => $cnv_hq_file, segments_file => $cnv_hmm_file, output_dir => $symbol_outdir, name => $symbol, cancer_annotation_db => $cancer_annotation_db, window_size => 0, verbose => 1, chr => $plot_chr);
             $cnview_cmd->execute();
         } else {
             my $gene_targets_file = "$gene_symbol_dir/$symbol" . ".txt";
-            my $cnview_cmd = Genome::Model::Tools::CopyNumber::CnView->create(annotation_build =>  $self->annotation_build_id, cnv_file => $cnv_hq_file, segments_file => $cnv_hmm_file, output_dir => $symbol_outdir, gene_targets_file => $gene_targets_file, name => $symbol, cancer_annotation_db => $cancer_annotation_db, window_size => 0, verbose => 1);
+            my $cnview_cmd = Genome::Model::Tools::CopyNumber::CnView->create(annotation_build =>  $self->annotation_build_id, cnv_file => $cnv_hq_file, segments_file => $cnv_hmm_file, output_dir => $symbol_outdir, gene_targets_file => $gene_targets_file, name => $symbol, cancer_annotation_db => $cancer_annotation_db, window_size => 0, verbose => 1, chr => $plot_chr);
```
tmooney: If these two differ only in `gene_targets_file`, the duplication of these params listings could be removed using something like:

```perl
my %params = (annotation_build => ..., chr => $plot_chr);
if ($symbol ne 'All') {
  $params{gene_targets_file} = "$gene_symbold_dir/$symbol" . ".txt";
}
my $cnview_cmd = Genome::Model::Tools::CopyNumber::CnView->create(%params);
$cnview_cmd->execute;
```

lib/perl/Genome/Model/ClinSeq/Command/MicroarrayCnv.pm
```diff
@@ -310,7 +396,12 @@ sub run_cnview {
...
+    my $make_hmmfile_cmd;
+    if($self->microarray_model_single) {
+        $make_hmmfile_cmd = 'awk \'{ size = $3-$2; nmarkers=$4; cn_diff = $5; event = "NA"; if(cn_diff>0) { event = "Gain" } else if(cn_diff<0) { event = "Loss" } cn1 = $5; cn2 = 0; if((event == "Gain" || event == "Loss") && (cn_diff > ' . $min_cnv_diff . ' || cn_diff < -1 * ' . $min_cnv_diff . ' ) && $4 >=5) print $1"\t"$2"\t"$3"\t"size"\t"nmarkers"\t"cn1"\t"cn1"\t"cn2"\t"cn2"\tNA\t"event; } \' ' . $cbs_op . ' > ' .  $cnv_hmm_file;
+    } else {
+        $make_hmmfile_cmd = 'awk \'{ size = $3-$2; nmarkers=$4; cn_diff = 2^$5*2-2; event = "NA"; if(cn_diff>0) { event = "Gain" } else if(cn_diff<0) { event = "Loss" } cn1 = cn_diff + 2; cn2 = 2; if((event == "Gain" || event == "Loss") && (cn_diff > ' . $min_cnv_diff . ' || cn_diff < -1 * ' . $min_cnv_diff . ' ) && $4 >=5) print $1"\t"$2"\t"$3"\t"size"\t"nmarkers"\t"cn1"\t"cn1"\t"cn2"\t"cn2"\tNA\t"event; } \' ' . $cbs_op . ' > ' .  $cnv_hmm_file;
```
tmooney: It might be clearer to use Perl variables for the values of `cn_diff`, `cn1`, and `cn2` rather than repeating the `awk` command in each of these two cases.

tmooney:
+1, the deduplication might be a nice touch, but it could also wait until the next time these need to be changed.

gatoravi:
Thanks for the suggestions @tmooney , I've pushed these changes up and I'll merge this soon.

lib/perl/Genome/Model/ClinSeq/Command/MicroarrayCnv.pm
```diff
@@ -310,7 +396,18 @@ sub run_cnview {
...
+    if($self->microarray_model_single) {
+        my $cn_diff = '$5';
+        my $cn1 = '$5';
+        my $cn2 = '0';
+        $make_hmmfile_cmd = 'awk \'{ size = $3-$2; nmarkers=$4; cn_diff = ' . $cn_diff . '; event = "NA"; if(cn_diff>0) { event = "Gain" } else if(cn_diff<0) { event = "Loss" } cn1 = ' . $cn1 . '; cn2 = ' . $cn2 . '; if((event == "Gain" || event == "Loss") && (cn_diff > ' . $min_cnv_diff . ' || cn_diff < -1 * ' . $min_cnv_diff . ' ) && $4 >=5) print $1"\t"$2"\t"$3"\t"size"\t"nmarkers"\t"cn1"\t"cn1"\t"cn2"\t"cn2"\tNA\t"event; } \' ' . $cbs_op . ' > ' .  $cnv_hmm_file;
+    } else {
+        my $cn_diff = '(2 ^ $5 * 2) - 2';
+        my $cn1 = 'cn_diff + 2';
+        my $cn2 = '2';
+        $make_hmmfile_cmd = 'awk \'{ size = $3-$2; nmarkers=$4; cn_diff = ' . $cn_diff . '; event = "NA"; if(cn_diff>0) { event = "Gain" } else if(cn_diff<0) { event = "Loss" } cn1 = ' . $cn1 . '; cn2 = ' . $cn2 . '; if((event == "Gain" || event == "Loss") && (cn_diff > ' . $min_cnv_diff . ' || cn_diff < -1 * ' . $min_cnv_diff . ' ) && $4 >=5) print $1"\t"$2"\t"$3"\t"size"\t"nmarkers"\t"cn1"\t"cn1"\t"cn2"\t"cn2"\tNA\t"event; } \' ' . $cbs_op . ' > ' .  $cnv_hmm_file;
```
tmooney: If `$cn_diff` etc. were declared outside the conditional (the way `$make_hmmfile_cmd` is now), then the repetition of the `awk` could be eliminated and its declaration could follow the conditional instead.

PR #958, opened 2015-08-25
Skip running the small indel workflow if there aren't any small indels.

tmooney:


chrisamiller:
+1 - looks good - thanks!

lib/perl/Genome/Model/Tools/Varscan/PullOneTwoBpIndels.pm
```diff
@@ -417,7 +417,11 @@ sub execute {
         push(@cmds,"$bsub -N -u $user\@$email_domain -J varscan_process_validation_reltum -w \'ended(JOB8)\' \'gmt varscan process-validation-indels --validation-indel-file $output_indel.reltum --validation-snp-file $output_snp.reltum --variants-file $small_indel_list_nobed --output-file $final_output_file.reltum\'");
     }
     else{
-        # This was added so things are testable for now
+        if (-z $small_indel_list) {
+            $self->status_message('Skipping small indel validation because small indel list <%s> is empty.', $small_indel_list);
+            return 1;
+        }
+
```
brummett: -z will return false if the file doesn't exist.  Is that an issue here?

tmooney: I believe getting to this point without the file existing constitutes an error, since the file should have been created by this step.  As is, it will fail inside `ValidateSmallIndels`.

brummett:
+1

PR #961, opened 2015-08-26
Add configurable LSF queues to many pipelined commands

tmooney:


ebelter:
+1 Although it might be worth it to have an base/role class such as CommandWithLsfQueue or the like to hold this option.

brummett:
+1

PR #962, opened 2015-08-26
Optional microarray build for Picard GenotypeConcordance

jasonwalker80:
Add options to over-ride the microarray data used.  
Limit the concordance for Exome analysis to the ROI. 
Utilize gzipped and indexed VCF whenever possible.
Add bedtools 2.24.0 as an available version.
Use the bedtools -header option when available.

### DIFF
--- a/lib/perl/Genome/Model/ReferenceAlignment/Command/GenotypeMicroarrayConcordance.pm
+++ b/lib/perl/Genome/Model/ReferenceAlignment/Command/GenotypeMicroarrayConcordance.pm
@@ -43,7 +43,7 @@ class Genome::Model::ReferenceAlignment::Command::GenotypeMicroarrayConcordance 
             is => 'Genome::Model::Build::GenotypeMicroarray',
             doc => 'Use this microarray build as the truth VCF regardless of default genotype data for the sample.',
         },
-        lane_qc_builds => {
+        lookup_lane_qc_builds => {
             is => 'Boolean',
             doc => 'Lookup the lane QC build for each instrument data assigned to the input reference alignment build.',
             default_value => 1,
@@ -296,7 +296,7 @@ sub resolve_builds {
     my $self = shift;
 
     my @builds;
-    if ($self->lane_qc_builds) {
+    if ($self->lookup_lane_qc_builds) {
         my @instrument_data = $self->build->instrument_data;
         unless (@instrument_data) {
             $self->error_message('Found no instrument data assigned to build: '. $self->build->__display_name__);
--- a/lib/perl/Genome/Model/ReferenceAlignment/Command/GenotypeMicroarrayConcordance.pm
+++ b/lib/perl/Genome/Model/ReferenceAlignment/Command/GenotypeMicroarrayConcordance.pm
@@ -38,6 +38,19 @@ class Genome::Model::ReferenceAlignment::Command::GenotypeMicroarrayConcordance 
         },
     ],
     has_optional =>[
+        microarray_build => {
+            is => 'Genome::Model::Build::GenotypeMicroarray',
+            doc => 'Use this microarray build as the truth VCF regardless of default genotype data for the sample.',
+        },
+        lane_qc_builds => {
+            is => 'Boolean',
+            doc => 'Lookup the lane QC build for each instrument data assigned to the input reference alignment build.',
+            default_value => 1,
+        },
+        output_dir => {
+            is => 'Text',
+            doc => 'A directory to output the raw Picard files to.',
+        },
         _seqdict => {},
         _genotype_samples_and_sorted_vcfs => { is => 'HASH', default_value => {}, },
     ],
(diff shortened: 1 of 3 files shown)


lib/perl/Genome/Model/ReferenceAlignment/Command/GenotypeMicroarrayConcordance.pm
```diff
@@ -167,48 +178,35 @@ sub summary_headers {
...
     
-    my $build_vcf = $qc_build->get_detailed_snvs_vcf;
+    my $build_vcf = $build->get_detailed_snvs_vcf;
     unless (-e $build_vcf) {
-        $self->error_message('Unable to find the snvs VCF for build : '. $qc_build->__display_name__);
+        $self->error_message('Unable to find the snvs VCF for build : '. $build->__display_name__);
         die($self->error_message);
     }
+    #TEST
+    return $build_vcf;
```
tmooney: This looks to be left over from some debugging.

lib/perl/Genome/Model/ReferenceAlignment/Command/GenotypeMicroarrayConcordance.pm
```diff
@@ -252,28 +279,100 @@ sub resolve_genotype_microarray_vcf_and_sample {
...
+
+sub resolve_roi_intervals_from_build {
+    my $self = shift;
+    my $build = shift;
+    
+    my $roi_bed = $self->create_temp_file_path($build->id .'.bed');
+    $build->region_of_interest_set_bed_file($roi_bed);
+
+    my $roi_intervals = $self->create_temp_file_path($build->id .'.intervals');
+    # TODO: Cache this file or store it's location for later use.  Same ROI is likely used for all QC builds.
```
tmooney: There is a stray apostrophe in `its`.

lib/perl/Genome/Model/ReferenceAlignment/Command/GenotypeMicroarrayConcordance.pm
```diff
@@ -54,53 +67,47 @@ sub execute {
...
+    }
+
+    my @data;
+    for my $build ($self->resolve_builds) {
+        my $build_vcf = $self->resolve_snvs_vcf_for_build($build);
+        unless ($microarray_build) {
+            $microarray_build = $build->genotype_microarray_build;
+            unless ($microarray_build) {
+                $self->error_message('Failed to find a Microarray build.  None exists as input to SNVs build.');
+                die($self->error_message);
```
ebelter: Should we also check it's reference as above when provided the GM build?

lib/perl/Genome/Model/ReferenceAlignment/Command/GenotypeMicroarrayConcordance.pm
```diff
@@ -252,28 +279,100 @@ sub resolve_genotype_microarray_vcf_and_sample {
...
-    );
+sub create_temp_file_path {
+    my $self = shift;
+    my $basename = shift;
 
-    # Check if we have seen this sample, and return the sorted vcf file name
-    my $sorted_microarray_vcf = $self->_genotype_samples_and_sorted_vcfs->{$genotype_sample->id};
-    return $sorted_microarray_vcf if $sorted_microarray_vcf;
+    if ($self->output_dir) {
+        return $self->output_dir .'/'. $basename;
```
ebelter: File::Spec->join is the prefered way to create file names.

lib/perl/Genome/Model/ReferenceAlignment/Command/GenotypeMicroarrayConcordance.pm
```diff
@@ -54,53 +67,47 @@ sub execute {
...
-        unless ($intersect_cmd->execute) {
-            $self->error_message('Failed to execute bedtools intersect!');
-            die($self->error_message);
+    }
+
+    my @data;
+    for my $build ($self->resolve_builds) {
+        my $build_vcf = $self->resolve_snvs_vcf_for_build($build);
+        unless ($microarray_build) {
+            $microarray_build = $build->genotype_microarray_build;
```
ebelter: If $self->microarray_build is not specified to the command, this would cause the $microarray_build from the first build to be set and used for all builds. Is that desired?

ebelter:
This would have been easier to review if specific commits were used. Fo large PRs, I step through the commits to understand the changes. Additionally, I am concerned no tests were added for the new functionality.

lib/perl/Genome/Model/ReferenceAlignment/Command/GenotypeMicroarrayConcordance.pm
jasonwalker80: Since the microarray model is an input to the reference alignment model, I assumed compatibility checks would happen at that time.  If that's not the case, then perhaps we do want to check the reference sequence build for consistency.

jasonwalker80:
New commits pushed.  I agree this could benefit from a test case.  I'm concerned about the time it would take to run on complete inputs. If there are examples of small mocked microarray data sets, that would be a great start.

lib/perl/Genome/Model/ReferenceAlignment/Command/GenotypeMicroarrayConcordance.pm
```diff
@@ -38,6 +39,19 @@ class Genome::Model::ReferenceAlignment::Command::GenotypeMicroarrayConcordance
         },
     ],
     has_optional =>[
+        microarray_build => {
+            is => 'Genome::Model::Build::GenotypeMicroarray',
+            doc => 'Use this microarray build as the truth VCF regardless of default genotype data for the sample.',
+        },
+        lane_qc_builds => {
```
davidlmorton: This might be better named 'lookup_lane_qc_builds' to indicate that it is not itself a list of the builds but is a boolean option.

lib/perl/Genome/Model/ReferenceAlignment/Command/GenotypeMicroarrayConcordance.pm
```diff
@@ -252,28 +275,99 @@ sub resolve_genotype_microarray_vcf_and_sample {
...
+}
+
+sub resolve_roi_intervals_from_build {
+    my $self = shift;
+    my $build = shift;
+    
+    my $roi_bed = $self->create_temp_file_path($build->id .'.bed');
+    $build->region_of_interest_set_bed_file($roi_bed);
+
+    my $roi_intervals = $self->create_temp_file_path($build->id .'.intervals');
```
davidlmorton: I wish we would create tools that, if they want a filename to write a file to, would default to creating you a tempfile and then all you'd have to do is look at the command's output property for the path.  I imagine it would be easy to achieve this either with a UR property type or as a mixin class.  This pattern (create temp file, create command, execute command, return temp file's path) is used a lot inside of Genome.

lib/perl/Genome/Model/ReferenceAlignment/Command/GenotypeMicroarrayConcordance.pm
jasonwalker80: Thanks.  New commit pushed.

lib/perl/Genome/Model/ReferenceAlignment/Command/GenotypeMicroarrayConcordance.pm
```diff
@@ -252,28 +275,99 @@ sub resolve_genotype_microarray_vcf_and_sample {
...
 
-    # Create and store the sorted VCF file name for this sample
-    $sorted_microarray_vcf = Genome::Sys->create_temp_file_path(
-        Genome::Utility::Text::sanitize_string_for_filesystem($genotype_sample->name).'_microarray_sorted.vcf'
+sub resolve_builds {
+    my $self = shift;
+
+    my @builds;
+    if ($self->lane_qc_builds) {
+        my @instrument_data = $self->build->instrument_data;
```
davidlmorton: This lookup logic better belongs on the build's class, that way any other commands that would like to look up lane_qc_builds associated with this build could do so without replicating this function.  Its nearly impossible to find this kind of functionality if its in these commands, so it gets replicated over and over throughout the system.

lib/perl/Genome/Model/ReferenceAlignment/Command/GenotypeMicroarrayConcordance.pm
jasonwalker80: I agree, that pattern is repeated all over Genome.  If there is a better way, I'm all ears.

lib/perl/Genome/Model/ReferenceAlignment/Command/GenotypeMicroarrayConcordance.pm
jasonwalker80: The lookup of lane qc builds is on it's way out, hence this PR to make it optional.  There isn't much additional functionality except iterating over each instrument data and returning the results of that instrument data property.  It might be best to consolidate as a property of a build, but given the short lifespan of lane qc to begin with, I'm hesitant to make those changes at this time.

ebelter:
+1

PR #964, opened 2015-08-26
Ref Cov Metrics Command

ebelter:
add param to inlcude non succeeded builds sort by date_scheduled


tmooney:
+1.  

We often order by `created_at` instead of `date_scheduled` these days.  (In particular `Unstartable` and `New` builds do not have a `date_scheduled`, but they aren't too useful for this command anyway.)

brummett:
+1

PR #965, opened 2015-08-27
Update CLE Test

apregier:


tmooney:
+1 inasmuch as the new ID does in fact identify a `Genome::Process` in our database!

apregier:
This is an update to the CLE Test and must be approved by the Medical Director of the CLE

apregier:
Changes:
-the sort order of the entries in the coverage stats file has changed
-the library labels are now numbered starting at 1.  This will be consistent in future runs.  This changes most of the report.txt files and the legend files.
-the docm reports contain additional annotations
-the vcf reports have additional info fields that specify the values of each individual 'soft' filter

ernfrid:
Allison — We have approval. Please merge and move forwards!!!

PR #966, opened 2015-08-27
Only unlock SoftwareResults once.

tmooney:
* Subsequent commits of SoftwareResults generally don't need to unlock again, as the object already exists and was unlocked.
* `Genome::Db::Ensembl::AnnotationStructures` commits several times as it is working.  To avoid unlocking early, add an explicit unlock after the final commit.
* Minor fix for test names in the aforementioned class is included.


ebelter:
+1

brummett:
+1

PR #967, opened 2015-08-27
Script to set up the fill_db environment in which you can pull in obj…

apregier:
…ects from the production db into a test db

### DIFF
--- a/bin/fill_db
+++ b/bin/fill_db
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+DS_GMSCHEMA_SERVER=`genome config get ds_gmschema_server`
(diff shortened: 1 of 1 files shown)


tmooney:
+1.  It'd be nice if there were a better way to get back to the defaults than mixing `genome config` commands with environment variables, but I'm not sure what that would be.

apregier:
Maybe if there were a 'genome config set-default' or 'genome config unset' command?  I'm not sure if either of those would make sense.  I got the idea for doing it this way from Tony's new command to clean up test-db allocations on the filesystem, so if we come up with something it could be used there as well.

apregier:
Thanks, good catch - updated

brummett:
+1

tmooney:
We could make a `genome config get-default` command (using `Genome::Config::default_value`) whose output can be passed to `genome config set-env`.  A `genome config unset-env` command would work well, too.

PR #969, opened 2015-08-27
ClinSeq::LOH - Write results to temp dir first.

gatoravi:
This PR modifies this module to write the results
to a temp dir and copy the results over at the end for
better performance.

### DIFF
--- a/lib/perl/Genome/Model/ClinSeq/Command/IdentifyLoh.pm
+++ b/lib/perl/Genome/Model/ClinSeq/Command/IdentifyLoh.pm
@@ -191,7 +191,6 @@ sub filter_snvs {
     $self->get_tumor_normal_bam($somvar_build);
     my $refbuild = $self->reference_build($somvar_build);
     my $ref_fa = $refbuild->full_consensus_path('fa');
-    my $temp_op_dir = Genome::Sys->create_temp_directory;
     my $filter = Genome::Model::Tools::Varscan::SomaticFilterWorkflow->
     create(
         outdir => $outdir,
--- a/lib/perl/Genome/Model/ClinSeq/Command/IdentifyLoh.pm
+++ b/lib/perl/Genome/Model/ClinSeq/Command/IdentifyLoh.pm
@@ -3,6 +3,7 @@ package Genome::Model::ClinSeq::Command::IdentifyLoh;
 use strict;
 use warnings;
 use File::Spec;
+use File::Copy;
 use Genome;
 
 class Genome::Model::ClinSeq::Command::IdentifyLoh {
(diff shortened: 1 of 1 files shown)


brummett:
As far as I can tell, all the I/O is writes to the output directory, whether it's to the final ```outdir``` or the temp directory.  Are there reads from that directory I'm missing?

If there's no reads, and it's writing the same amount of data either way, how is it faster?

gatoravi:
Sorry, that was a stray variable. `$temp_outdir` is the one that is actually used. I've removed it now. Writes now happen on `tmp` and results copied over at the end.

gatoravi:
Wouldn't writing to local temp space on a blade and then copying over to disk be faster if we were constrained by disk bandwidth?

brummett:
The same disk contention is going to still exist whether it's in the old code with the tools writing directly to the output directory, or in the new code copying to the output directory.

One case where it's better to write to /tmp locally first is if it writes some data to a file and reads it back in during the same step.  Then, the reads are taking place locally on the same machine rather than going over the network.

malachig:
Maybe it was my imagination but it seemed like there were files there during processing that did not exist in the final result.  If these files are not part of the final result then wouldn't there be an advantage to writing to temp and them moving the final result over (that way the intermediate files are never going over the network)?

lib/perl/Genome/Model/ClinSeq/Command/IdentifyLoh.pm
```diff
@@ -255,11 +262,12 @@ sub filter_loh {
 #Get rid of intermediate files
 sub cleanup {
     my $self = shift;
+    my $outdir = shift;
     my @patterns = qw(*.Somatic* *.readcounts *.hc *.LOH
```
malachig: Do we actually need all of these files in the final result under non debugging conditions?

gatoravi:
Sorry, I see your point now @brummett .
One point is we do a cleanup on the temp dir before the copy so some chunky readcount files are removed before the copy.

lib/perl/Genome/Model/ClinSeq/Command/IdentifyLoh.pm
gatoravi: I don't think so and hence I got rid of them. What do you think?

brummett:
I get it now.  If it's not copying all the files - that makes sense.

+1

lib/perl/Genome/Model/ClinSeq/Command/IdentifyLoh.pm
malachig: Right I see now that these are the files being removed and not kept.  Duh.  This makes sense!

malachig:
+1

PR #970, opened 2015-08-28
Feature List: do not sub off chr in names

ebelter:


ebelter:
CI, test this.

ebelter:
@susannasiebert I updated the test data featture list file to remove the chr in the chromosome names. This PR stops subbing off the leading 'chr', which was an artifact from the original install of ref-cov. Does the update to the feature list file seem to be OK?

tmooney:
+1.  This will probably break some builds that previously might have succeeded!  But it's also more correct :smile:

susannasiebert:
+1. Looks good to me.

Just out of curiosity, are we fixing all the feature lists in our database to remove any leading 'chr'?

ebelter:
@susannasiebert @tmooney I was not planning on removing leading chr in existing feature lists. Should we push this change out if we don't?

tmooney:
I don't think we should remove the `chr`s from existing lists.  I'm more inclined to say we should update the reference set on the FLs if it is wrong.  When it's set properly the `chr`s already get removed by the refseq converter.

ebelter:
@tmooney Good point. Should we do this proactively, or wait for failures?

apregier:
The processed_bed_file method in feature list is used in other places (e.g. variant reporting), e.g https://github.com/genome/genome/blob/master/lib/perl/Genome/VariantReporting/Suite/Vep/RunResult.pm#L211

jeldred:
nobody want's this to lead to SILENT failures....does anyone have ideas how we can catch those?

apregier:
I think it may be sufficient to look at the handful of places where FeatureList->processed_bed_file (and maybe processed_bed_file_content?) is used and insert an alternate way to remove 'chr' or at least detect where a mismatch is occurring and fail loudly.  Otherwise I am concerned that tools using the feature list will simply fail to match coordinates because the chromosome names don't match.
At the very least we could verify that test cases for these uses cover the scenario where the feature list contains chr and the other thing doesn't.

ebelter:
Closing for now. May revisit!

PR #972, opened 2015-08-28
Update GetBamReadCountsMatrix.t with the new som-var build

dufeiyu:
This test has been failing for several days. The failure is caused by the merged bams of normal/tumor ref-align builds of used som-var builds are both removed during the recent removal of superseded merged bam event. A new som-var build is used along with the new test data ouput directory. This PR need early merge to make jenkins test pass.


gatoravi:
+1 The new results look right!

brummett:
+1

PR #973, opened 2015-08-28
Move prepare-submodules script from jenkins repo into main repo

apregier:
I want to move this script from the jenkins repo into this repo because sometimes we might want to prepare the jenkins submodule, and we can't without this script ;-)
Once this commit is merged, I will update the configurations in jenkins to reference it, and then merge a commit in wugc-hudson removing the script from that repo.


apregier:
I don't really need to do this just now.

PR #974, opened 2015-08-28
Basic refalign QC summary.

jasonwalker80:
Remove the ADSP QC command and make it applicable to all projects.


jasonwalker80:
Yeah, duplication_rate still is still there.  Number of unique bases would have been calculated from that.  Duplication rate seems to be more commonly used across all projects.

ernfrid:
Yep, very ADSP specific.

____

tmooney:
+1.  It'd be nice to have a test that runs the tool.  (That might be easier to create if it wrote a TSV instead of an XLS file.)

jasonwalker80:
The transition from XLS to TSV might be a worthwhile follow up PR to enable tests.  I'm not sure anyone requires the XLS functionality and TSV will open as a spreadsheet.

PR #975, opened 2015-08-31
Revert "ClinSeq::LOH - Write results to temp dir first."

tmooney:
Reverts genome/genome#969

We can't use a local temporary directory as the output location for a workflow.  The LSF jobs that are spawned almost always won't be able to read that directory.  (Since this causes jobs to fail in pre-exec in Flow, they sit in PENDing state forever.)


gatoravi:
I'm totally ok with this revert. We'll just have to come up with a different way of doing this if we still want to. +1 to help get past failing test builds.

PR #976, opened 2015-08-31
Update how ptero workflows are constructed

davidlmorton:
Ptero::Builder::ShellCommand has been generalized to
Ptero::Builder::Job, but it requires a service_url to be specified.


davidlmorton:
This will not pass tests until the PR in to genome_snapshot_deps has been merged and blades have installed the libptero-perl v0.1.10

lib/perl/Genome/WorkflowBuilder/Command.pm
```diff
@@ -84,13 +85,15 @@ sub _get_ptero_shortcut_method {
...
+    require Ptero::Builder::Job;
+
 
     my $self = shift;
     my $log_dir = shift;
-    # XXX This should use the LSF service, or be configuration based
-    return Ptero::Builder::ShellCommand->new(
+    return Ptero::Builder::Job->new(
         name => 'execute',
+        # XXX This should use the LSF service, or be configuration based
```
tmooney: What is this comment suggesting still needs to be done?

davidlmorton: There will be a different environment variable that points to the lsf service url and the resource string (if it exists) on the command should be included in the parameters.

tmooney:
Merging to fix tests!

PR #977, opened 2015-08-31
Add required-users for fast-tier bitmask generation.

tmooney:
Like #945, this adds user data to the SR construction.  It also makes the `build` parameter non-optional, since it is used to get this user data.


ebelter:
+2

ebelter:
+1

PR #978, opened 2015-09-01
Iterate through ldapsearch output

dufeiyu:
Currently 1000 is used as response size limit for ldapsearch.  Modify this so more info can be obtained.

### DIFF
--- a/lib/perl/Genome/Site/TGI/Command/SyncSysUserWithLdap.pm
+++ b/lib/perl/Genome/Site/TGI/Command/SyncSysUserWithLdap.pm
@@ -4,6 +4,7 @@ use strict;
 use warnings;
 
 use Genome;
+use IO::File;
 
 class Genome::Site::TGI::Command::SyncSysUserWithLdap{
     is => 'Command::V2',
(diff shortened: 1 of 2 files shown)


brummett:
Since this may return a considerable amount of data, it may be worth switching to an iterator-based read from ldapsearch rather than slurping everything up with ```my @c = `ldapsearch -z 0 -x`;```

dufeiyu:
Good point. I will make the change.

ebelter:
@brummett @tmooney Has anyone used Net::LDAP->search ?

brummett:
I haven't used Net::LDAP.  It seems to have a decent set of tests and one of the contributors (gbarr) has a few modules in the Perl standard library.  I'd say it's worth trying out, though the debian package we have available is pretty old.

brummett:
I can't figure out how to use Net::LDAP.  trying to find all the user records with ```$ldap->search('(objectClass=person)')``` or entries with email addresses (as this function seems to be doing) with ```$ldap->search('(email=*)')``` always return 0 records.  I guess it's either not working or I don't know what I'm doing...  Though Net::LDAP would be better than the janky parsing this function is using.

+1, this change improves the existing function.

ebelter:
I believe we are switching to FreeIPA. Not sure how that will affect this.

ebelter:
Talking to Rob F about using Net::LDAP.

dufeiyu:
@ebelter Eddie agrees to push this for now to fix current problem. Future PR can be created for the new change.

ebelter:
I will have a PR using et::LDAP from an example @rfugina provided.

PR #979, opened 2015-09-01
Add entry for speedseq version to bwa version mapping

acoffman:
SpeedSeq bundles its own version of BWA. This mapping identifies the version of BWA packaged with a particular version of SpeedSeq so that we can pass it along to other results. The 0.1.0 release of SpeedSeq is already out, but I missed updating this at the time.


susannasiebert:
+2 once tests pass.

acoffman:
@susannasiebert @jasonwalker80 In addition to the new version mapping, I've updated this PR with a test that should ensure future Speedseq version are added to this hash as they're made available.

PR #980, opened 2015-09-01
Update test to account for new DGIdb release.

tmooney:
There was a new DGIdb release recently, which caused this test to diff.  This adds updated expected data.

The changes can be viewed with:
```
diff $GENOME_TEST_INPUTS/Genome-Model-ClinSeq-Command-AnnotateGenesByDgidb/{v3,v4}/all_interactions.tsv
```


dufeiyu:
+1

acoffman:
+1

PR #981, opened 2015-09-01
Increase speedseq LSF memory requested 

susannasiebert:
36GB is not enough for some large instrument data. Ultimately this well get calculated based on the size of the instrument data once we have run more samples and have a better idea what the formula should be (similar to gtmp requested).


jasonwalker80:
+2

dufeiyu:
+1

tmooney:
+:100:

PR #983, opened 2015-09-01
Commit actions

brummett:
This branch implements an alternative API for registering callbacks to run then the base Context commits or rolls-back intended to have fewer gotchas than using Context observers:
* They only ever fire when the base Process context is saving to data sources, never for software transactions
* They only ever fire once (except during a failure in sync_database - discussed in the docs)
* commit/rollback callbacks may by grouped together and only one of them will fire

### DIFF
--- a/lib/perl/Genome/Process.pm
+++ b/lib/perl/Genome/Process.pm
@@ -620,9 +620,8 @@ sub delete {
         $input->delete;
     }
 
-    my $observer;
-    $observer = Genome::Sys::CommitAction->create(
-        on_commit => $self->_disk_allocation_cleanup_closure(\$observer)
+    my $observer = Genome::Sys::CommitAction->create(
+        on_commit => $self->_disk_allocation_cleanup_closure()
     );
     if ($observer) {
         $self->status_message("Registered observer to delete disk allocation " .
--- a/lib/perl/Genome/Process.pm
+++ b/lib/perl/Genome/Process.pm
@@ -620,7 +620,8 @@ sub delete {
         $input->delete;
     }
 
-    my $observer = Genome::Sys::CommitAction->create(
+    my $observer;
+    $observer = Genome::Sys::CommitAction->create(
         on_commit => $self->_disk_allocation_cleanup_closure(\$observer)
     );
     if ($observer) {
--- a/lib/perl/Genome/Process.pm
+++ b/lib/perl/Genome/Process.pm
@@ -620,9 +620,9 @@ sub delete {
         $input->delete;
     }
 
-    my $observer;
-    $observer = UR::Context->current->add_observer(aspect=>'commit',
-        callback=>$self->_disk_allocation_cleanup_closure(\$observer));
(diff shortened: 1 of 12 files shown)


lib/perl/Genome/Process.pm
```diff
@@ -621,8 +621,9 @@ sub delete {
     }
 
     my $observer;
-    $observer = UR::Context->current->add_observer(aspect=>'commit',
-        callback=>$self->_disk_allocation_cleanup_closure(\$observer));
+    $observer = Genome::Sys::CommitAction->create(
+        on_commit => $self->_disk_allocation_cleanup_closure(\$observer)
```
tmooney: I think we can stop passing `\$observer` into this call.  It looks like it was trying to accomplish the same thing as `once => 1`.

lib/perl/Genome/SoftwareResult.pm
```diff
@@ -788,7 +788,7 @@ sub delete {
     };
 
     #hook our anonymous sub into the commit callback
-    $observer = $class_name->ghost_class->add_observer(aspect=>'commit', callback=>$upon_delete_callback);
```
tmooney: `$observer` can be removed from this subroutine entirely (in addition to on this line).

tmooney:
Will these `on_commit` actions fire after all other data sources have committed?

brummett:
These on_commit actions are guaranteed to fire after RDBMS-type data has been saved/committed.  UR commit 10b27695 made it so UR::DataSource::Default always saves and commits last.

tmooney:
Cool, +1.

lib/perl/Genome/Sys/CommitAction.pm
```diff
@@ -0,0 +1,128 @@
...
+throws an exception.  CommitActions are then considered live again, and
+all their C<con_sync> callbacks may run again, if the original Context commit
+exception is caught and it tries to commit again.  Note that C<on_sync_fail>
+callbacks are _not_ run when the context is rolled back with
+UR::Context->rollback.
+
+The callbacks are passed a single argument, whatever is stored in the
+CommitAction's 'data' attribute.
+
+Also note that if the C<con_sync> or C<on_commit> callbacks change the program
```
tmooney: `C<con_sync>` should be `C<on_sync>`.

lib/perl/Genome/Sys/CommitAction.pm
```diff
@@ -0,0 +1,128 @@
...
+If any data fails to save during sync_database, all data sources are asked
+to rollback.
+
+For these CommitAction objects, their C<on_sync> callback is run during
+sync_database, C<on_commit> is run during commit.  C<on_rollback> is run when
+the Process Context is rolled back.
+
+The C<on_sync_fail> callback is only run when a subsequent C<on_sync> callback
+throws an exception.  CommitActions are then considered live again, and
+all their C<con_sync> callbacks may run again, if the original Context commit
```
tmooney: `C<con_sync>` should be `C<on_sync>`.

PR #984, opened 2015-09-01
Remove newly created disk allocations when exiting genome-env

brummett:
Before the test database is deleted, remove any allocations created during
this invocation of genome-env


apregier:
+1

tmooney:
+1, yay less leaky allocations!

PR #987, opened 2015-09-02
Ignore DGIdb interaction files in the clinseq build-test

gatoravi:


tmooney:
+1 :green_apple:

ebelter:
+1 Would it be safer to add dgidb to the regex like so:

*.dgidb/all_interactions.tsv

ebelter:
Merging early to get model tests passing.

PR #990, opened 2015-09-08
new options for epitope binding filter

chrisamiller:


chrisamiller:
Good suggestion - I added code that I think solves this, but can't for the life of me figure out what this error even means:

Code is at:
/gscmnt/gc6134/cancer-genomics/medseq/git/genome/lib/perl/Genome/Model/Tools/EpitopePrediction/BindingFilter.pm

Test data is here: (feel free to play around in this dir)
/gscmnt/sata921/info/medseq/cmiller/pancVacc/H_TG-567-03-03-A1-M1500207/tmp

Command and output:
$ /usr/bin/perl -I ~cmiller/git/genome/lib/perl `which gmt` epitope-prediction binding-filter --fof-file=epi.fof --output-file=epitopes.combined --minimum-fold-change 1 --variant-file snvs.indels.annotated.ponpass.evs.filt.3llr.hq

ERROR: Can't locate object method "validate_file_for_writing" via package "epitopes.combined.varanno" at /gsc/scripts/opt/genome/current/user/bin/gmt line 25

ernfrid:
`open_file_for_writing` expects to be called by the package or the class, not to be referenced directly. I changed it to `Genome::Sys->open_file_for_writing` instead of what you had and it works now. I also added a check on whether or not you are able to create the SeparatedValueWriter as that was giving odd errors as well.

chrisamiller:
Just pushed some code that seems to do the trick - thanks for the assist, Dave.

brummett:
+1

PR #991, opened 2015-09-09
Use the correct key for passing users.

tmooney:
This will work better if we use the key the code expects to see. :smile:


ebelter:
+1

jasonwalker80:
+1, this change worked in a local git repo when both of the nested repos have the 'users' param passed on.

PR #992, opened 2015-09-10
Add another users hash to the annotation result get_or_create

jasonwalker80:
Additional users hash to complement #991


tmooney:
+1 :bust_in_silhouette:

ebelter:
+1 :koko:

PR #993, opened 2015-09-10
CQID processes data batched by sample

tmooney:
In order to reduce the creation of builds that are nearly immediately superseded, this processes all available data for a sample in a single CQID run.

* This may run over the `limit` parameter, since otherwise it's possible to have a sample that will never be processed.  This version is pretty tolerant of going over, but it could be made more strict (e.g. only accepting a larger sample batch if nothing else is going to be processed) if that seems better.
* There's SQL added that is not covered by the `.t`.  It looked like it was working when I tried it manually, but it probably warrants extra attention in review!

### DIFF
--- a/lib/perl/Genome/Config/Command/ConfigureQueuedInstrumentData.pm
+++ b/lib/perl/Genome/Config/Command/ConfigureQueuedInstrumentData.pm
@@ -346,50 +346,22 @@ sub _get_items_to_process {
 sub _ordered_sample_id_iterator {
     my $self = shift;
 
-    if(!defined UR::Context->query_underlying_context or UR::Context->query_underlying_context) {
-        my $dbh = Genome::Sample->__meta__->data_source->get_default_handle;
-        my $sth = $dbh->prepare(q{
-            SELECT sample.subject_id
-            FROM subject.subject sample
-            INNER JOIN instrument.fragment_library library ON library.sample_id = sample.subject_id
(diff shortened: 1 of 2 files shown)


brummett:
Is the SQL in there for performance reasons?  Would the code be more clear if it used a UR Set query instead of a bifurcated SQL/UR query?

You can get the same query (sans ordering) by adding a property on Genome::InstrumentData
```
new_and_failed_analysis_project_bridges => {
    is => 'Genome::Config::AnalysisProject::InstrumentDataBridge',
    reverse_as => 'instrument_data',
    is_many => 1,
    where => [ 'status in' => ['new','failed'] ],
},
```
Then the iterator construction becomes something like this
```
my $set = Genome::Sample->define_set(
    'instrument_data.new_and_failed_analysis_project_bridges.analysis_project.status' => 'In Progress',
);
my @subsets = $set->group_by('subject_id');
return sub {
    my $subset = shift @subsets;
    return $subset && $subset->subject_id;
};
````
Both the SQL and this query are returning 0 items at the moment, so I'm not sure it's doing the right thing

tmooney:
I agree it's better to not have a giant SQL query in there.  I've pushed a different approach that still examines `fail_count`, but doesn't try to aggregate it for each sample.

ebelter:
+1

brummett:
+1

PR #994, opened 2015-09-10
Return analysis ID but join on seq ID.

tmooney:
The "analysis ID" is only used to map to the Genome object for Illumina data.  Everywhere else the "seq ID" is used to refer to the object within the LIMS schema.


ebelter:
+1

brummett:
+1

PR #995, opened 2015-09-10
Config List: sort configs for easier viewing

ebelter:


tmooney:
The file `Genome/Config/Command/30000]` should be removed.

PR #996, opened 2015-09-11
transcript-info use toplevel Ensembl coord system

jasonwalker80:
Iterate over all slices and use the toplevel coord system rather than chromosome and supercontig slices.

@apregier please take a look if you have the time.


apregier:
+1 I assume this worked when you ran it on build 38, since it doesn't have unit tests?

jasonwalker80:
@apregier this did work on Ensembl 79_38 for human. I'm still testing/running mouse annotation.  There are additional problems I encountered outside of the scope of this PR or this tool for the mouse annotation.

PR #999, opened 2015-09-14
ModelSummary: "abandon" action should not be overridden by "rebuild" action

tmooney:
(The rebuilds would fail to queue, anyway, since the models are disabled.)


ebelter:
+1

jweible:
+1

PR #1000, opened 2015-09-14
add option to run on non-succeeded build

chrisamiller:
This is useful when a build crashes in the late stages, preventing create reports from running. Gives us an easy way to at least pull out SNVs and indels without waiting for a rebuild.


tmooney:
With the name `--force-use-latest-build`, I expected it to use the latest build even if there is an earlier build that Succeeded, while this implementation appears to prefer a Succeeded build.  Maybe it should be something like `--fallback-to-latest-build`?

chrisamiller:
Good suggestion, made that change.

tmooney:
+1 :japanese_castle:

brummett:
+1

PR #1001, opened 2015-09-14
Create Gtf::Cat command before executing.

jasonwalker80:
First create command before executing.  Calling execute twice results fatal error.


brummett:
+1

tmooney:
+1 :cat2:

PR #1003, opened 2015-09-15
Use should_skip_run from Loh.pm to decide whether loh runs or not

dufeiyu:
Currently $build->loh_version is used to decide whether loh runs or not, which is not accurate. This PR is to fix the bug.


tmooney:
+1.

It's a little strange to call `$class->build` since in that case `$class` is definitely a concrete object.  It might be better to pass `$build` when calling within the `LOH` command as well.

dufeiyu:
Agree. make the change.

jweible:
+1

PR #1004, opened 2015-09-16
Tweaks to output of number of builds abandoned by ModelSummary

tmooney:
* Now with matching parentheses!
* Don't bother printing out zeroes.


brummett:
+1

jweible:
+1

PR #1005, opened 2015-09-17
Sync Users: use Net::LDAP and add tests

ebelter:


lib/perl/Genome/Site/TGI/Command/SyncSysUserWithLdap.t
```diff
@@ -5,16 +5,49 @@ use warnings;
...
 
-my $ldap_users = $class->get_ldap_users;
-my $users = Set::Scalar->new(keys %$ldap_users);
-my $apipe_users = Set::Scalar->new(map{$_.'@'.Genome::Config::get('email_domain')}qw(apipe apipe-builder apipe-tester));
+my @apipe_uids = (qw/ apipe apipe-builder apipe-tester /);
+my $apipe_users = Set::Scalar->new( map{$_.'@'.Genome::Config::get('email_domain')} @apipe_uids );
+my @apipe_db_users = map { my $o = Test::MockObject->new; $o->set_always('email', $_); $o; } $apipe_users->members;
 
-ok($users->is_proper_superset($apipe_users), 'Found apipe users in ldap users');
+subtest 'lpad_users' => sub {
```
tmooney: Should `lpad_users` be either `load_users` or `ldap_users`?

ebelter:
@dufeiyu Can you review? [That rhymes btw!]

lib/perl/Genome/Site/TGI/Command/SyncSysUserWithLdap.pm
```diff
@@ -52,40 +54,36 @@ sub execute {
...
+    $mesg->code && die $mesg->error;
+
+    $mesg = $ldap->bind;
+    $mesg->code && die Dumper($mesg);
+
+    $mesg = $ldap->search(
+        base => "dc=gsc,dc=wustl,dc=edu",
+        filter => "(objectClass=Person)",
+        #filter => "(&(objectClass=Person)(uid=$username))",
+    );
```
brummett: The server used by ```Net::LDAP->new()``` should probably be a Genome::Config item.  The params used in the search also look like they should be config items, too.

ebelter: I asked Tom what he thought about it before committing. Since it is in 'site' we decided that it was not worth worrying about. Something like this wis generally useful, and col live in Genome::Sys. Whatcha think @tmooney ?

lib/perl/Genome/Site/TGI/Command/SyncSysUserWithLdap.pm
```diff
@@ -52,40 +54,36 @@ sub execute {
...
-        } else {
-            my ($key, $value) = split(/\:\s+/,$c);
-            $user->{$key} = $value;
-        }
+    my $ldap = Net::LDAP->new('ipa1.gsc.wustl.edu', version => 3);
+    my $mesg = $ldap->start_tls(verify => 'none');
+    $mesg->code && die $mesg->error;
+
+    $mesg = $ldap->bind;
+    $mesg->code && die Dumper($mesg);
```
dufeiyu: Need to use Data::Dumper ?

lib/perl/Genome/Site/TGI/Command/SyncSysUserWithLdap.pm
brummett: Didn't see this was under Site/; that makes sense.  Other installations will have their own implementations of user syncing, so I'd leave this as-is.

brummett:
+1

ebelter:
I also confirmed the command works! :small_blue_diamond:

PR #1006, opened 2015-09-17
Remove picard path

jasonwalker80:
remove picard 1.123 path to jwalker directory.  It's also installed system-wide in /usr/share/java/


lib/perl/Genome/Model/Tools/Picard.pm
```diff
@@ -145,7 +145,6 @@ EOS
 # Please put the most recent first.
 my $sw_legacy_java = Genome::Config::get('sw_legacy_java');
 my @PICARD_VERSIONS = (
-    '1.123' => '/gscmnt/sata132/techd/solexa/jwalker/lib/picard-tools-1.123',
```
dufeiyu: Don't we need version 1.123 here ? just thought it should be $sw_legacy_java . '/samtools/picard-tools-1.123'

jasonwalker80: It was debian packaged by Ian a while back.  The hash version should not be necessary.

dufeiyu: You are correct. After second look on path_for_picard_version, it shows if no version is found in legacy hash, it will try to find one (like 1.123) from /usr/share/java/picard-tools*, which is kind of odd and might need refactoring later but is out of scope of this PR.  

+1

tmooney:
+1 :anchor:

PR #1007, opened 2015-09-17
A superseding alignment's samples must match the samples of the alignment it supersedes.

tmooney:
We encountered a situation wherein a BAM was superseded by another that was a combination of data for 16 samples.  This led to the per-sample BAMs being removed.


chrisamiller:
+1

dufeiyu:
Can this be made optional ? Sometimes there might be a need to get full list of superseding results, regardless of multiple or single sample.

tmooney:
Perhaps the the superseding BAM should have the same set of samples as the current BAM?

dufeiyu:
OK, agree. It makes more sense.

tmooney:
I've updated this PR with the requirement that the sample(s) be the same in both the original and the superseding alignment.

dufeiyu:
+1  if tests pass

PR #1008, opened 2015-09-18
Create a valid vcf

apregier:
The header doesn't have and entry for the INFO field TYPE


tmooney:
+1 :information_desk_person:

ebelter:
+1

PR #1010, opened 2015-09-20
add converter to remove extra contigs

ernfrid:
This is a proposal to try to help address RT 108134.

This attempts to add a converter between two references that would hopefully drop contigs we don't care to have coverage calculated on in the destination. There is no test yet. 

I'd like a little feedback on whether or not this strategy will work. If the imported reference has chr's (for example) AND more chromosomes than the destination reference then is that addressed in the current system? How? Order of operations could be important here if multiple converters need to be applied.


tmooney:
As implemented this converter would work for mapping between one reference and another that is a proper subset of it.  The practical situation we're needing to address also must incorporate the existing `convert_chrXX_contigs_to_GL` process on the chromosome name before evaluating whether to keep it.  (Otherwise all lines from the file will be dropped!)

ernfrid:
Ok. That was my concern if there was no way to apply multiple converters. Will ponder this.

ernfrid:
I would probably lean towards just making a version of this that also does the conversion and calling it something new. I would probably leave in the current implementation as well. @tmooney — how does that sound good or dumb? Is there a better way to do it that you'd advise?

tmooney:
One option is to make `drop_extra_contigs` take an optional per-position algorithm name in the `resource` property.  If provided, run it on the `$chrom, $start, $stop` and then re`join` those results, something like:
```perl
my $per_position_algorithm = $self->resource;

#later, inside the loop
if($per_position_algorithm) {
  $line = join("\t", $self->$per_position_algorithm($chrom, $start, $stop), @extra);
}
```

ernfrid:
I've tried to make some changes to implement your design @tmooney. It does need tests still, but I thought I'd ask for feedback now.

tmooney:
Looks reasonable, though I hope that we don't start adding a bunch of `$site_test`s to this in the future!

lib/perl/Genome/Model/Build/ReferenceSequence/Converter.pm
```diff
@@ -184,4 +167,60 @@ sub no_op {
...
+        unless($self->validate_algorithm_name($algorithm) && $self->is_per_position_algorithm($algorithm)) {
+            die "Per-position algorithm from resource property is not a valid converter\n";
+        }
+    }
+
+    my $site_test = sub {
+        my ($chrom) = @_;
+        return $chromosome_set->has($chrom);
+    };
+    $self->parse_and_write_bed($source_bed, $source_reference, $algorithm, $site_test);
```
tmooney: `parse_and_wirte_bed` takes `$destination_bed` rather than `$source_reference`.

ernfrid:
Ok. Changes are rebased, including that last one.

ernfrid:
Found one more bug and added simple test. Let me know how things look now.

ernfrid:
Ok. I did rebase after finding some bugs and adding a test. @tmooney - let me know what you think. Testing code is pretty low-bar and copy-paste.

lib/perl/Genome/Model/Build/ReferenceSequence/Converter.t
```diff
@@ -45,6 +48,33 @@ my $diff = Genome::Sys->diff_file_vs_text($converted_bed_file, $expected_result)
...
+    my $dc_converter = Genome::Model::Build::ReferenceSequence::Converter->create(
+        source_reference_build => $ref_seq_a,
+        destination_reference_build => $ref_seq_c,
+        algorithm => 'drop_extra_contigs',
+        resource => 'convert_chrXX_contigs_to_GL',
+    );
+    my $dc_converted_bed_file = $feature_list->converted_bed_file(reference => $ref_seq_c);
+    ok($dc_converted_bed_file, 'got converted BED file');
+
+    Memoize::flush_cache('Genome::FeatureList::processed_bed_file');
```
tmooney: This line doesn't affect the test in its present location.  It could be moved before the call to `->converted_bed_file` or removed.

lib/perl/Genome/Model/Build/ReferenceSequence/Converter.t
```diff
@@ -45,6 +48,33 @@ my $diff = Genome::Sys->diff_file_vs_text($converted_bed_file, $expected_result)
 ok(!$diff, 'converter produced expected result')
     or diag('diff: ' . $diff);
 
+{
```
tmooney: Since we're introducing a block here, we could make this into a `subtest`.

tmooney:
+1 :droplet:

PR #1011, opened 2015-09-21
Remove early/duplicate variable declaration.

tmooney:
This variable is declared when it is initialized later (on line 409).


brummett:
+2

PR #1012, opened 2015-09-22
fixed broken regex to appropriately display RNA expression

kkrysiak:


tmooney:
I tried running both the old and new regular expressions on the `genes.fpkm.expsort.top1percent.tsv` file from the latest build of the `apipe-test-clinseq-wer` model. I don't believe either regular expression properly parses the file.

I recommend `split`ting on tabs and collecting/parsing each field individually as needed. Using `Genome::Utility::IO::SeparatedValueReader` may make this easier.

kkrysiak:
The problem with the test build is that it has both normal and tumor rna-seq which should not invoke that regular expression at all. It should only be used when you have tumor rna-seq only. I used it on such a sample using the command below. You can see the output directory as well.

genome model clin-seq make-circos-plot --build=ecaa7532589c4b9b9aa36b7b0786998f --output-directory=/gscmnt/gc2547/mardiswilsonlab/kkrysiak/lymphoma/LYM175_circos/

tmooney:
The new version truncates the decimal from the FPKM values (e.g., it reports `65227` instead of `65227.9`).  Is this expected and desired?

kkrysiak:
It's rather moot for the visualization as that amount of difference will not be seen in the plots generated by this tool. However, it's a good catch so I fixed it. I tested it on the same build. 

--output-directory=/gscmnt/gc2547/mardiswilsonlab/kkrysiak/lymphoma/LYM175_circos_2/

tmooney:
The `apipe-clinseq-test-wer` model does not have a `normal_rnaseq_model` specified.  The difference is that there are coverage scores available in the file in the test, whereas the other file contains only `NA` in that column.

Here's what it could look like with the `SeparatedValueReader`:
```perl
my $reader = Genome::Utility::IO::SeparatedValueReader->create(separator => "\t", input => "$output_directory/raw/genes.fpkm.expsort.top1percent.tsv");
my $expression_fh = Genome::Sys->open_file_for_writing("$output_directory/data/genes.fpkm.expsort.top1percent.tsv");
while(my $data = $reader->next) {
    my ($chr, $start, $stop) = split(/[:-]/, $data->{locus});
    $$genes_noAmpDel{$data->{mapped_gene_name}} = join("\t", "hs$chr", $start, $stop);
    $$genes_AmpDel{$data->{mapped_gene_name}} = join("\t", "hs$chr", $start, $stop);
    print $expression_fh ("hs$chr $start $stop " . log($data->{FPKM})/log(2)."\n");
}
```

kkrysiak:
My apologies. I had it in my head that the blessed build was still the TST1/HCC1395. I'll take a crack at it with the separator. Thanks for the help.

kkrysiak:
Thanks for the help @tmooney , now that I understand how to use SeparatedValueReader, I'll try doing so more in the future. When I tested the new code on the blessed build, it appears to be functioning as I would expect.

genome model clin-seq make-circos-plot --build=d64a5bb2f808434fa30f252c418c1d2c --output-directory=/gscmnt/gc2547/mardiswilsonlab/kkrysiak/lymphoma/Blessed_build_test/

tmooney:
+1, you're welcome!

jasonwalker80:
+1

PR #1014, opened 2015-09-23
binding threshold exposed as parameter

chrisamiller:


tmooney:
+1 :boom:

lib/perl/Genome/Model/Tools/EpitopePrediction/BindingFilter.pm
```diff
@@ -27,6 +27,11 @@ class Genome::Model::Tools::EpitopePrediction::BindingFilter {
             doc => 'Minimum fold change between mutant binding score and wild-type score. The default is 0, which filters no results, but 1 is often a sensible default (requiring that binding is better to the MT than WT)',
             default => 0,
         },
+        binding_threshold => {
+            is => 'Numeric',
```
dufeiyu: Numeric or Number ?

chrisamiller: minimum_fold_change, right above it is "Numeric". Is one preferred over the other? Is there a difference?

dufeiyu: Both should be Number. There is no UR::Value::Numeric. UR treats that unrecognized data type as SloppyPrimitive.

chrisamiller:
Both sites updated to "Number". Thanks for the catch, Feiyu!

dufeiyu:
+1

PR #1015, opened 2015-09-23
Deprecated AnP status Timeline event.

jasonwalker80:
Add valid_value for Deprecated AnP status.

### DIFF
--- a/lib/perl/Genome/Timeline/Event/AnalysisProject.pm
+++ b/lib/perl/Genome/Timeline/Event/AnalysisProject.pm
@@ -30,7 +30,7 @@ class Genome::Timeline::Event::AnalysisProject {
             is => 'Text',
             len => 255,
             default_value => 'Pending',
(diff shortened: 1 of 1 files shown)


tmooney:
+1.  Should `Template` be added as well?

jasonwalker80:
I'm not sure.  The timeline events are for when the status is changed.  With Template, it's either created with --store-only or it's not.  I don't think changing status too or from Template would be desired.  I'm open to other ideas or opinions though.

tmooney:
I believe it will try to store the status when noting changes to `is_cle` or `run_as` as well. (None of these are likely to be changed for `Template`s either.)

Maybe instead of adding it explicitly we could use `Genome::Config::AnalysisProject->__meta__->property(property_name => 'status')->valid_values` so the lists always match.

Or this is fine as is!

jasonwalker80:
I like the suggestion @tmooney.  New commit to use all valid_values from AnPs to avoid duplication.

acoffman:
+1

PR #1016, opened 2015-09-23
Make read group accumulation level not the default at this time

susannasiebert:
- Adding the read group accumulation level drastically increases runtime
- Picard uses the RG PU field as the read group value which in our BAMs is not unique and, thus, not useful
- It's not needed for CCDG at this time

Once we have the config management system deployed this option can be turned on as needed.

### DIFF
--- a/lib/perl/Genome/Qc/Config.pm
+++ b/lib/perl/Genome/Qc/Config.pm
@@ -42,7 +42,7 @@ sub get_commands_for_alignment_result {
                 input_file => 'bam_file',
                 refseq_file => 'reference_sequence',
                 use_version => 1.123,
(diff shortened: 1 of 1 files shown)


tmooney:
+1 :recycle:

dufeiyu:
+1

susannasiebert:
After talking to Jason we also decided to remove the SAMPLE metric_accumulation level from the current hard-coded config since it adds a lot of runtime and in our BAMs all the reads belong to the same sample anyway so it didn't add any further information.

tmooney:
+1² :black_small_square:

jasonwalker80:
+2

PR #1017, opened 2015-09-23
Fixes to Metric Parsing in ClinSeq SummarizeBuilds

tmooney:
This is a follow-up to changes in #998 to update a few cases that aren't covered by unit tests but were caught by the `ClinSeq` model test.


susannasiebert:
+2

PR #1018, opened 2015-09-24
Remove more category names from keys.

tmooney:
Hot on the heels of #1017 here are more metric changes to benefit ClinSeq.


susannasiebert:
+2

PR #1022, opened 2015-09-28
Allow evaluation to be re-parsed

apregier:
If the evaluation has already been run and we just want to
get the stats, we don't need to re-run everything, which takes
a while.  Detect this situation if the compare.txt file exists with
size.
The disadvantage here is that if you actually do want to rerun (e.g.
something has changed), it may not do what you expect (it does print
a debug message that results are just being parsed).
Previously, it would crash when trying to create a symlink that
already existed.

### DIFF
--- a/lib/perl/Genome/Model/Tools/Vcf/EvaluateVcf.pm
+++ b/lib/perl/Genome/Model/Tools/Vcf/EvaluateVcf.pm
@@ -136,41 +136,45 @@ sub execute {
     my $fp_roi_file = $output_dir->file("fp_in_roi.vcf")->stringify;
     my $compare_file = $output_dir->file("compare.txt")->stringify;
 
-    Genome::Sys->create_directory($variants_dir);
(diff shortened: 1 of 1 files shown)


indraniel:
Could a `--force` boolean parameter to be added to forcefully re-run everything if the `compare.txt` file already exists?  That way one could use the option to purposely override things without having to delete the file before re-running evaluate-vcf.

This would be useful if we noticed a corrupt `compare.txt` file.

ernfrid:
I like @indraniel's idea too. We would need to add clean up to the script then though or we will just fail out. I think that's probably outside the scope of this request.

tmooney:
I'd suggest choosing a different name than `--force`, though.  In the past we've run into trouble with it being unclear exactly what would or wouldn't happen when it's specified!

apregier:
I think I'll just merge as-is and next time we run this and need a force-like option we can add it.

PR #1024, opened 2015-09-28
Add an additional dir to ignored paths.

gatoravi:


tmooney:
+1, but will grep still match the desired lines if the initial `\\w+/\\w+/(info|test)` is removed from the regex entirely?  As-is, this may be a problem in other environments (e.g. `cle`) if ClinSeq is used in them in the future.

gatoravi:
I guess we were just trying to be a bit more stringent with the match:) Would other environments including `cle` use a separate testing framework? I can definitely remove the check it if so.

dufeiyu:
How about use  grep -vP '^file'  here so we don't need to deal with different build path.

dufeiyu:
CI, test this

gatoravi:
Nice idea, @dufeiyu, updated this.

dufeiyu:
+1  please merge this asap.

PR #1025, opened 2015-09-29
Disk Allocation Archive command accepts paths

tmooney:
The unarchive command already accepts paths; this makes them more symmetric.


brummett:
+1

acoffman:
+1

PR #1026, opened 2015-09-30
Show more info on error message to help troubleshooting

dufeiyu:
Instrument data id and merged result id are added into error message.

### DIFF
--- a/lib/perl/Genome/InstrumentData/AlignmentResult/Merged.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult/Merged.pm
@@ -370,7 +370,8 @@ sub collect_individual_alignments {
                 my $lookup_hash = Genome::InstrumentData::AlignmentResult->calculate_lookup_hash_from_arguments(%all_params); 
                 my @alignments  = Genome::InstrumentData::AlignmentResult->get(lookup_hash => $lookup_hash);
                 unless (@alignments and @alignments == 1) {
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/InstrumentData/AlignmentResult/Merged.pm
```diff
@@ -370,7 +370,8 @@ sub collect_individual_alignments {
                 my $lookup_hash = Genome::InstrumentData::AlignmentResult->calculate_lookup_hash_from_arguments(%all_params); 
                 my @alignments  = Genome::InstrumentData::AlignmentResult->get(lookup_hash => $lookup_hash);
                 unless (@alignments and @alignments == 1) {
-                    die $self->error_message('Only 1 alignment is expected, but got '.scalar @alignments);
+                    die $self->error_message(sprintf("Only 1 alignment is expected, but got %d for instrument data %s of merged result %s",
```
acoffman: `$self->error_message()` will actually do `sprintf` style substitution for you, so I think this can be simplified down to just:

````perl
 die $self->error_message("Only 1 alignment is expected, but got %d for instrument data %s of merged result %s", scalar @alignments, $i->id, $self->id)
````

acoffman:
+2 with the `sprintf` removed :+1:

dufeiyu:
good point, unnecessary sprintf removed.

PR #1027, opened 2015-09-30
Genome ptero lsf integration

davidlmorton:
These tests will not pass until the perlSDK package is updated, they pass when PERL5LIB points at local perl-sdk deployment.  I'll have 'ci test this' when the perlSDK package has been updated.

### DIFF
--- a/lib/perl/Genome/Ptero/ToJSON.t
+++ b/lib/perl/Genome/Ptero/ToJSON.t
@@ -22,8 +22,8 @@ for my $test_directory (glob test_data_directory('*')) {
 
     my $json_filename = do {
         local %ENV; # otherwise the JSON contains the current user's environment.
-        $ENV{XGENOME_PTERO_SHELL_COMMAND_SERVICE_URL} = 'http://example.com/v1';
-        $ENV{XGENOME_PTERO_LSF_SERVICE_URL} = 'http://lsf.example.com/v1';
+        Genome::Config::set_env(ptero_shell_command_service_url => 'http://example.com/v1');
+        Genome::Config::set_env(ptero_lsf_service_url => 'http://lsf.example.com/v1');
         my $ptero_builder = $workflow->get_ptero_builder();
 
         my $json_filename = Genome::Sys->create_temp_file_path();
--- a/lib/perl/Genome/Ptero/ToJSON.t
+++ b/lib/perl/Genome/Ptero/ToJSON.t
@@ -23,6 +23,7 @@ for my $test_directory (glob test_data_directory('*')) {
     my $json_filename = do {
         local %ENV; # otherwise the JSON contains the current user's environment.
         $ENV{XGENOME_PTERO_SHELL_COMMAND_SERVICE_URL} = 'http://example.com/v1';
+        $ENV{XGENOME_PTERO_LSF_SERVICE_URL} = 'http://lsf.example.com/v1';
         my $ptero_builder = $workflow->get_ptero_builder();
 
         my $json_filename = Genome::Sys->create_temp_file_path();
@@ -41,6 +42,7 @@ for my $test_directory (glob test_data_directory('*')) {
         filters => [
             qr/^.*"workingDirectory" :.*$/,
             qr/^.*"user" :.*$/,
+            qr/^.*"cwd" :.*$/,
         ],
     );
 }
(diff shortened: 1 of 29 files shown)


davidlmorton:
CI test this

davidlmorton:
+1

davidlmorton:
These changes allowed an rna-seq model test build to complete: 

```
genome model build view 4cd42b2df0d3465e83fffb07c14e7cb9
```

lib/perl/Genome/Ptero/ToJSON.t
```diff
@@ -23,6 +23,7 @@ for my $test_directory (glob test_data_directory('*')) {
     my $json_filename = do {
         local %ENV; # otherwise the JSON contains the current user's environment.
         $ENV{XGENOME_PTERO_SHELL_COMMAND_SERVICE_URL} = 'http://example.com/v1';
+        $ENV{XGENOME_PTERO_LSF_SERVICE_URL} = 'http://lsf.example.com/v1';
```
tmooney: This could use `Genome::Config::set_env`.

lib/perl/Genome/Ptero/workflow_tests/parallel-cmd/ptero_workflow_for_process.json
```diff
@@ -85,26 +85,23 @@
...
+                                    "resReq" : "select[gtmp>1 && mem>3000] span[hosts=1] rusage[gtmp=1 && mem=3000]"
+                                 },
+                                 "rLimits" : {
+                                    "RSS" : "3000000"
+                                 },
+                                 "user" : "dmorton"
                               },
                               "service" : "job",
-                              "serviceUrl" : "http://example.com/v1"
+                              "serviceUrl" : "http://lsf.apipe-deis.gsc.wustl.edu/v1"
```
tmooney: It's a little strange to me that some of the domains in these `ptero_workflow_for_process.json` files are `example.com` and some are `wustl.edu`.

lib/perl/Genome/Utility/LSFResourceParser.pm
```diff
@@ -0,0 +1,221 @@
+package Genome::Utility::LSFResourceParser;
```
tmooney: Should this go with the other modules under `Genome::Sys::LSF`?

lib/perl/Genome/WorkflowBuilder/Command.pm
```diff
@@ -90,22 +91,42 @@ sub _get_ptero_execute_method {
 
     my $self = shift;
     my $log_dir = shift;
+    my $ptero_lsf_parameters = $self->_get_ptero_lsf_parameters();
+    $ptero_lsf_parameters->{command} = sprintf(
+        'genome ptero wrapper --command-class %s '
+        .'--method execute --log-directory %s',
+        $self->command, $log_dir);
```
tmooney: Is there a reason for the shift from the `commandLine` list to this string version, `command`?

davidlmorton: This is in line with how the lsf c-api expects commands.  In the shell-command service, we use a different method to launch the job that has a different api, one that expects a list of strings.  We could homogenize these interfaces though, it is worth discussing (perhaps outside this PR).

mkiwala: I initially shared @tmooney's concern, but I eventually decided that I'm ok with these api's being different.

The api's of the job services are different because the underlying api's which the different job services expose as web services are different.  Shell-command exposes execvp(2) via Pythons subprocess.Popen().  The lsf service exposes lsb_submit(3).  The workflow service makes few demands as possible on the api's that the job services expose.  This allows us to keep the definition of what a job service can do as a broad as possible.

If we have stricter requirements on the api which a job service must expose, the we limit the kinds of job services which we could eventually offer.

tmooney: Does this still work if the desired log directory contains spaces or other characters that a shell might treat specially?  That's my main concern in switching away from an array version.

mkiwala: Some layer above PTero would have to be responsible for crafting the appropriate command for submission to LSF, especially if a log directory contained spaces or other characters that required quoting.

mkiwala: +1

tmooney: Maybe all we need is to call `Genome::Sys->quote_for_shell` on the `log_directory`?  (This code is some layer above PTero!)

PR #1028, opened 2015-10-01
Make the volume lister work

brummett:
Disk Groups validate that their disk_group_name belongs to a short list
and dies otherwise to ensure Genome stuff only messes with analysis
allocations.  For a basic lister, we don't want want this safety net.

Getting the free space of an unmounted volume also throws an exception
if the volume is unmounted to avoid mounting tons of volumes.  For this
simple lister, we just print out "<unmounted>" for such volumes instead of
dying.

### DIFF
--- a/lib/perl/Genome/Disk/Command/Volume/List.pm
+++ b/lib/perl/Genome/Disk/Command/Volume/List.pm
@@ -19,5 +19,34 @@ class Genome::Disk::Command::Volume::List {
     doc => 'Lists Genome::Disk::Volume objects',
 };
 
+sub execute {
+    my $self = shift;
+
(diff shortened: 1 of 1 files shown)


tmooney:
Despite its name, `is_mounted` will trigger the automounter for each volume queried.  If we want to avoid mounting all possible volumes we'll need a different approach.

brummett:
New commits added to use /proc/mounts instead of Genome::Disk::Volume::is_mounted to determine mounted-ness, and added an option to attempt mounting if you really want the sizes for all displayed volumes (defaults to false)

tmooney:
+1 :floppy_disk:

PR #1029, opened 2015-10-01
Use generic config variable for alignment queue.

tmooney:
This is the only place left referring to this config instead of the regular one.  After this change is in all active snapshots we can remove the other variable from our deployed `config.yaml` and then from config altogether.


brummett:
+1

jasonwalker80:
+1

PR #1030, opened 2015-10-01
In the UsageLog, don't call getpwuid.

tmooney:
In my tests it reduces the runtime of "use Genome;" by about half a second–from ~1.2s to ~0.7s).

This potentially trades accuracy for speed, but, If we're using this as an audit log, we're going about it all wrong!


brummett:
This seems ok as long as no one is trying any funny business.  Al alternative would be for record_usage() to store the user ID instead of name, and the reporting tool can do the translation.

tmooney:
That's a better idea!  It still doesn't prevent funny business if someone's motivated, but it captures the same info as the old version.  Updated.

brummett:
+1

acoffman:
+1

PR #1031, opened 2015-10-02
Move variant sequence fasta generation into variant reporting framework and handle frameshifte and in frame indels

susannasiebert:
This is step one in adding frameshift mutations and inframe indels to the epitope prediction pipeline. This PR does the following:
- Move the creation of the variant sequence fasta file to the variant reporting framework
- Use VEP for annotation
- Make a new VEP plugin to get the wildtype protein sequence
- Handle frame shifts and in frame indels in addition to missense variants which were already handled in the original pipeline


susannasiebert:
@jhundal If you could have a look at this PR, that would be great. I would especially appreciate another set of eyes on the test cases in Genome/VariantReporting/Suite/Vep/EpitopeVariantSequenceInterpreter.t. The corresponding test vcf files are at Genome/VariantReporting/Suite/Vep/EpitopeVariantSequenceInterpreter.t.d.

lib/perl/Genome/Model/Tools/Annotate/TranscriptVariants.pm
```diff
@@ -768,11 +768,8 @@ sub file_has_header {
 
     my @expected_column_headers = $class->variant_attributes();
     my $expected_header = join("\t", @expected_column_headers);
-    if ($first_line =~ m/^$expected_header/) {
-        return 1;
-    }
-    else {
-        return 0;
+    unless ($first_line =~ m/^$expected_header/) {
```
tmooney: [This isn't something you changed, but] since `$expected_header` is a constant this could be written as `unless (index($first_line, $expected_header) == 0) {`.

chrisamiller:
+1

lib/perl/Genome/Model/Tools/Annotate/TranscriptVariants.pm
susannasiebert: Fixed.

susannasiebert:
While talking to @jhundal about this PR she pointed out that the ".p" in the fasta keys is an artifact from the the genome annotator and is not needed. I change the interpreter not not add the ".p" and all the appropriate epitope prediction gmt to work with the new fasta key format.

tmooney:
+1 :fried_shrimp:

susannasiebert:
@jhundal @chrisamiller I will merge this pull request. If we uncover any bugs we can make new pull requests with the fixes at that time.

PR #1032, opened 2015-10-02
Adds sanity checks to header and required cols

chrisamiller:


chrisamiller:
Bad things happen if the header is malformed - this checks for that. Some whitespace cleanup too - append "?w=1" to the URL to ignore it.

ernfrid:
Might be somewhat cleaner to create a method which checks for the appropriately named header and then throws the error immediately. This would give a more specific error message and potentially simplify the main body there.

Might also be better to use die("") instead of 

```
$self->error_message("$your_error_msg"); 
return(undef);
````

But I defer to someone in APIPE. Error handling is all over the board in the current code base and I don't know what current best practice is.

**Comment by tmooney** [comment]
I'd generally say `return;` is better than `return undef;`.  In a pipeline I'd recommend `die`, but in a command returning false is fine.  (Preview: genome/UR#101 adds `$self->fatal_message` to both print an error and die.)

File: lib/perl/Genome/Model/Tools/Validation/IdentifyOutliers.pm [hunk 1]
```diff
@@ -105,27 +105,38 @@ sub execute {
...
-        if($line =~ /chromosome_name/){
-            $normal_ref_col = List::MoreUtils::first_index {$_ =~ /Normal_ref_count/ } @F;
-            $normal_var_col = List::MoreUtils::first_index {$_ =~ /Normal_var_count/ } @F;
-            $tumor_ref_col = List::MoreUtils::first_index {$_ =~ /Tumor_ref_count/ } @F;
-            $tumor_var_col = List::MoreUtils::first_index {$_ =~ /Tumor_var_count/ } @F;
-
-            print OUTFILE $line . "\tFilterCall\tProb\tLLR\n";
-            if(defined($self->lq_output_file)){
-                print LQFILE $line . "\tFilterCall\tProb\tLLR\n";
+        if($linecount == 0){
```
tmooney: It might be nice to pull this part out to before the loop, since we know it'll happen exactly once.

lib/perl/Genome/Model/Tools/Validation/IdentifyOutliers.pm
```diff
@@ -105,49 +105,60 @@ sub execute {
     my $normal_var_col;
     my $tumor_ref_col;
     my $tumor_var_col;
+    my $linecount = 0;
```
tmooney: (`$linecount` is now unused.)

chrisamiller: linecount removed

tmooney:
+1 :apple:

brummett:
+1

PR #1033, opened 2015-10-05
Better User Handling with IntermediateAlignmentResults

tmooney:
This fixes an issue where the IAR could not be deleted when custom users were supplied in the result user hash.  It also removes a spurious error about linking alignments to the build in ReferenceAlignment.


ebelter:
+2

PR #1034, opened 2015-10-05
Fix the Search Queue test

brummett:
This test uses a nonsensical pattern:
    * create a software transaction
    * commit the lowest-level Context (inside the dedup functionality)
    * roll back the software transaction

Since the guts of Genome::Search::Queue::dedup() make a low-level commit,
the whole test needs to operate on that level.

An impending UR PR will make the old behavior throw an exception


tmooney:
+1

ebelter:
+1

PR #1035, opened 2015-10-06
Wait 10s before checking status of LSF jobs again.

tmooney:
Better solutions than this one have been proposed, but this is quick and should ameliorate the problem!


acoffman:
+1

chrisamiller:
+1

PR #1036, opened 2015-10-06
Don't really start builds in this test.

tmooney:
The integration (model) tests will try running a build all the way through.  This just needs to make sure Start works.  (This was testing the "inline" build process, anyway, which is not how builds are usually run in production.)


brummett:
+1

acoffman:
+1

tmooney:
Merging early to make progress towards a snapshot!

PR #1038, opened 2015-10-07
Addition of an option, cleanup of edge cases and tests

chrisamiller:


ernfrid:
+1

lib/perl/Genome/Model/Tools/Analysis/Concordance.pm
```diff
@@ -209,28 +215,53 @@ sub execute {
...
+        }
+    }
     close $IN2;
 
+
+    #output a file with all genotypes, if specified
+    if($self->output_genotypes){
+        my $outfile = Genome::Sys->open_file_for_writing($self->output_genotypes);
+        $outfile->print(join("\t",("Chr","Pos","ReferenceBase","Sample1","Sample2")) . "\n");
+        for my $pos (keys(%genotypes)){
```
tmooney: The order from `keys` is not guaranteed to be consistent.  You may want to sort these before writing them out.

chrisamiller: Good point - now sorts the keys before output

PR #1040, opened 2015-10-08
More efficient unarchive wait for lsf jobs

brummett:
Adds a new method to Genome::Sys called bsub_and_wait_completion() that can schedule multiple LSF jobs in parallel, and wait on them without running bjobs over and over.  It sets a post-exec command for each job to send a message back to the waiter over a TCP socket.  The waiter knows all the jobs are done when it gets the message from all the jobs it's waiting on.

Then, Genome::Model::Build::Command::Unarchive is changed to use this new method.

### DIFF
--- a/lib/perl/Genome/Site/TGI/Extension/Sys.pm
+++ b/lib/perl/Genome/Site/TGI/Extension/Sys.pm
@@ -74,8 +74,9 @@ sub get_lsf_job_status {
 sub bsub_and_wait_for_completion {
     my($class, %bsub_args) = @_;
 
-    my $commands;
-    ($commands, %bsub_args) = _bsub_and_wait_for_completion__validate_args(%bsub_args);
+    my $special_args;
+    ($special_args, %bsub_args) = _bsub_and_wait_for_completion__validate_args(%bsub_args);
+    my($commands, $on_submit_cb, $on_complete_cb) = @$special_args{'commands','on_submit','on_complete'};
 
     my $hostname = hostname;
     my $waiting_socket = IO::Socket::INET->new(Listen => 5, Proto => 'tcp', LocalHost => $hostname);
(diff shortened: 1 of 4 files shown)


lib/perl/Genome/Site/TGI/Extension/Sys.pm
```diff
@@ -69,6 +71,82 @@ sub get_lsf_job_status {
     return $status;
 }
 
+sub bsub_and_wait_for_completion {
+    my($class, %bsub_args) = @_;
+
+    my $special_args;
+    ($special_args, %bsub_args) = _bsub_and_wait_for_completion__validate_args(%bsub_args);
+    my($commands, $on_submit_cb, $on_complete_cb) = @$special_args{'commands','on_submit','on_complete'};
+
```
ebelter: It would be clearer if these were all separate key value pairs passed in as params. Maybe they were too many places to update to the new functionality?

ebelter:
+1 Exciting to have this in production!

lib/perl/Genome/Site/TGI/Extension/Sys.pm
brummett: I did it this way to make calling it simpler by having the regular bsub args (like ```queue``` and ```job_group```) passed together with the args for this method (like ```cmds``` and ```on_submit```).  Otherwise, you'd end up with a more complicated function call like this
```
Genome::Sys->bsub_and_wait_completion(
    { queue => 'short',
      job_group => 'blah',
    },
    {  cmds => [ '/bin/true', 'bin/false'],
      on_submit => sub { ... },
    }
);

ebelter: Yep. I think it is fine as implemented.

chrisamiller:
+1

PR #1041, opened 2015-10-08
New test-db snapshot

acoffman:
This commit updates the test database template to a new version which contains the `config.qc` table. I believe that merging this in is a prerequisite to #846 passing tests!


tmooney:
Would the CI test this please?

tmooney:
+1.

I don't think this is blocking #846 since we have 669522d, but it seems fine!

brummett:
+1

PR #1042, opened 2015-10-08
Inst Data Import: use read group id as instdata id

ebelter:
When processing files for import, generate a uuid for each read group and type (paired, singleton). Then use that red group id as the instrument data id.


ebelter:
Sry for all the commits...turned out to be more complicated than expected. A lot of commit s revolve around updating test data, as every bam is now different because of the newly added read group/instdata id.

ebelter:
@brummett Thanks for the review. Commits pushed to address your concerns.

brummett:
+1

PR #1043, opened 2015-10-08
Disable index creation in `gmt mutect`.

tmooney:
When run in the pipeline it uses a hardcoded ROD that should already be indexed.  (See also
[this GATK forum thread](http://gatkforums.broadinstitute.org/discussion/1252/unifiedgenotyper-gets-hung-waiting-for-file-lock).)


dufeiyu:
+1  but probably need ok from @ernfrid

ernfrid:
Looks fine in principal. Has anyone tested the option exists in our current install and run with it on real data?

tmooney:
I created this PR in response to a few RT tickets (e.g. 108455), so it was successful in creating succeeded builds where previously there were failures!

ernfrid:
Ok by me then! I saw the tickets and the successes.

+1

PR #1044, opened 2015-10-12
New SR to create "buckets" of chromosomes for a reference.

tmooney:
This PR is part of adding the HaplotypeCaller to the pipeline.  The buckets will be used to parallelize the task into roughly equal-sized tasks rather than by chromosomes whose sizes vary significantly.

### DIFF
--- a/lib/perl/Genome/Model/Build/ReferenceSequence/Buckets.pm
+++ b/lib/perl/Genome/Model/Build/ReferenceSequence/Buckets.pm
@@ -58,7 +58,6 @@ sub _run {
         $bucketizer->add_item( @$chr_with_length );
     }
 
-    #skip optimization to preserve order
     $bucketizer->optimize(maxrounds => 100_000);
 
     my @buckets = $bucketizer->buckets();
--- a/lib/perl/Genome/Model/Build/ReferenceSequence/Buckets.pm
+++ b/lib/perl/Genome/Model/Build/ReferenceSequence/Buckets.pm
@@ -50,7 +50,7 @@ sub bucket {
 sub _run {
     my $self = shift;
 
-    my $chr_lengths = $self->_chromosomes_with_lengths;
+    my $chr_lengths = $self->reference_sequence_build->chromosomes_with_lengths;
     my $max_length = max( map $_->[1], @$chr_lengths );
 
     my $bucketizer = Algorithm::Bucketizer->new(bucketsize => $max_length);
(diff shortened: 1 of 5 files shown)


lib/perl/Genome/Model/Build/ReferenceSequence/Buckets.pm
```diff
@@ -0,0 +1,125 @@
...
+    $self->count(scalar(@buckets));
+
+    for my $bucket (@buckets) {
+        $self->_write_bucket($bucket);
+    }
+
+    return 1;
+}
+
+sub _chromosomes_with_lengths {
```
jasonwalker80: There is some overlap between this subroutine and `Genome::Model::Build::ReferenceSequence->get_or_create_genome_file`.

tmooney: I've deduplicated the two parsing algorithms.

chrisamiller:
From skimming this code, it doesn't look like this subsets the chrs at all - it just bins them together?  So this doesn't actually reduce wallclock time? Is batching them to take similar amounts of time advantageous in some way that I'm not immediately recognizing?

tmooney:
Correct, this does not subset the chromosomes.

It still reduces wallclock time compared to not running in parallel, and no dataset to process will be larger than the largest chromosome, so wallclock time should be similar.  More homogeneous jobs are easier to request resources for and schedule in LSF.

@jasonwalker80 has been developing some more finely tweaked buckets based on experimentally-determined runtimes, which we can drop in for specific references, but it's infeasible to obtain that for all references in our system.

jasonwalker80:
Dividing work by chromosome is an easy strategy to implement, but it results in a wide array of sizes and corresponding run time, compute resources as Tom mentioned.  In addition, GRCh38 with alternate haplotypes consists of more than 400 chromosomes/contigs.  

The binning algorithm by chr size is not perfect but should yield consistent results with more predictable resource allocations.  Also as Tom mentioned, I have been evaluating GATK HaplotypeCaller on a handful of samples to see if the load can be balanced based on prior run time better than nucleotide length alone.  This approach is also not without flaws, but the rate-limiting run time is closer to the maximum run time when split by chromosome, eg. chr1.

I'm sure there are more efficient approaches that might result in splitting chromosomes by gaps and centromeres.  The bin by chr size approach seems like the right balance between complexity and performance.

lib/perl/Genome/Model/Build/ReferenceSequence/Buckets.pm
```diff
@@ -0,0 +1,84 @@
...
+    my $chr_lengths = $self->reference_sequence_build->chromosomes_with_lengths;
+    my $max_length = max( map $_->[1], @$chr_lengths );
+
+    my $bucketizer = Algorithm::Bucketizer->new(bucketsize => $max_length);
+    for my $chr_with_length (@$chr_lengths) {
+        $bucketizer->add_item( @$chr_with_length );
+    }
+
+    #skip optimization to preserve order
+    $bucketizer->optimize(maxrounds => 100_000);
```
dufeiyu: just wanna make sure that 100k here means skip.

tmooney: That comment is poorly worded.  It means that, in a case where order must be preserved, this line should be commented out.  100k here is the maximum number of attempts to make at optimizing the distribution.

Since the comment isn't terribly useful, I'll remove it.

lib/perl/Genome/Model/Build/ReferenceSequence/Buckets.pm
```diff
@@ -0,0 +1,83 @@
...
+
+    my $chr_lengths = $self->reference_sequence_build->chromosomes_with_lengths;
+    my $max_length = max( map $_->[1], @$chr_lengths );
+
+    my $bucketizer = Algorithm::Bucketizer->new(bucketsize => $max_length);
+    for my $chr_with_length (@$chr_lengths) {
+        $bucketizer->add_item( @$chr_with_length );
+    }
+
+    $bucketizer->optimize(maxrounds => 100_000);
```
dufeiyu: How about setting maxrounds as software result property with 100k as default value ?

tmooney: The algorithm is non-deterministic, so the important thing is to make it high enough that we're likely to get close to the optimal solution.  If we re-run `$bucketizer->optimize` and find a better solution, we should replace the result with that better one rather than let the results proliferate.

dufeiyu: It makes sense.

dufeiyu:
+1

susannasiebert:
+1

PR #1045, opened 2015-10-12
Changes to handle either a Workflow::Model or a G::WorkflowBuilder::DAG.

tmooney:
This allows using `WorkflowBuilder` to construct the workflow for a `Build`.

### DIFF
--- a/lib/perl/Genome/Model/Build.pm
+++ b/lib/perl/Genome/Model/Build.pm
@@ -1035,7 +1035,7 @@ sub start {
         }
 
         # Launches the workflow (in a pend state, it's resumed by a commit hook)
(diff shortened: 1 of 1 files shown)


brummett:
What about changing the method name in one class (or adding a function alias) to match the other?

tmooney:
I've opened genome/tgi-workflow#6 to add `get_xml` to `Workflow::Model`.  My eventual intent is to switch all model types to using `Genome::WorkflowBuilder`, so that'll be the new interface going forward.

I'll update this PR once that is merged.

brummett:
+1

acoffman:
+1

PR #1046, opened 2015-10-12
Set queue to apipe-pd and request 4 slots instead of 6 for QC runs

susannasiebert:


tmooney:
* While we're making changes here, I would consider creating a new config variable to hold the LSF resource string as in 1a28db02729b961076938697f4841e1693b3ad5e.
* The `lsf_queue` could be its own `param` like [here](https://github.com/genome/genome/blob/d0f92bf2798ebf081e10931919fc14afc9d73608/lib/perl/Genome/Model/RnaSeq/Command/DetectFusions/Chimerascan/DetectorBase.pm#L42-L50) if you want to keep them separate.

susannasiebert:
@tmooney I just pushed two commits to implement your suggestions.

tmooney:
+1 :first_quarter_moon:

jasonwalker80:
+1

PR #1047, opened 2015-10-12
Create a Test Factory for speedseq results.

tmooney:
I'll be adding another test in a subsequent PR that also uses this Factory.  (Splitting this out to reduce the size of that PR.)


susannasiebert:
+1 assuming tests pass.

dufeiyu:
+1 looks better now

PR #1048, opened 2015-10-12
Add intervals parameter for GATK HaplotypeCaller

tmooney:
Once the HaplotypeCaller is added to the pipeline, we'll be using the buckets from #1044 to fill in the `intervals` parameter added here.


jasonwalker80:
+1

brummett:
+1

PR #1051, opened 2015-10-13
Move AlignAndMerge LSF Resource generation to the command.

tmooney:
This change facilitates running the `AlignAndMerge` command outside of the alignment dispatcher.


susannasiebert:
+1 assuming tests pass.

acoffman:
+1

PR #1052, opened 2015-10-13
Build with results from https://github.com/genome/genome/pull/1012

gatoravi:
#1012 switched to separatedvaluereader from an incorrect regex.
The updated blessed-build has the correct results.


tmooney:
+1, build ID does point to a Succeeded clinseq-wer build!

ebelter:
+1 It does indeed! Merging to get model tests working.

PR #1053, opened 2015-10-15
Add run_as to test factory model.

tmooney:
This way `Genome::SoftwareResult::User->user_hash_for_build` can find a sponsor.

(I don't understand how `Genome/Model/Build/MetagenomicComposition16s/ProcessAndMergeInstrumentData.t` has been passing before, but it currently fails for me even on the deployed snapshot.  This change fixes it.)


brummett:
+1

acoffman:
+1

tmooney:
Merging early to fix the test.

PR #1054, opened 2015-10-15
Use array form of shellcmd.

tmooney:
This was failing on a BAM with parentheses in its name.


chrisamiller:
+1

We'll test it on our specific case out of a checkout once it's merged, but this looks good.

brummett:
+1

PR #1055, opened 2015-10-15
Remove old db_skip_postgres configuration variable.

tmooney:
This variable could be used to skip writing to postgres.  Since support for writing to multiple databases is long gone, this is not useful.


brummett:
+2

PR #1056, opened 2015-10-19
Use Filesys::Df to get available KB

tmooney:
(This is similar to how it is done in `Genome::Disk::Volume`.)


ebelter:
+1

jasonwalker80:
+1, once tests are passing

PR #1057, opened 2015-10-19
Make Java 1.7 the default for running GATK tools.

tmooney:
The system default is currently at 1.6, but various tools in the "refine-reads" step have been segfaulting on 1.6, and using 1.7 resolves this.  Additionally newer versions of GATK require 1.7 or higher (such as for the `HaplotypeCaller`).


brummett:
+1

jasonwalker80:
+1

PR #1058, opened 2015-10-19
Disk Alloc Move: do not allow specifying target group and volume

ebelter:


lib/perl/Genome/Disk/Detail/Allocation/Mover.pm
```diff
@@ -30,6 +30,22 @@ class Genome::Disk::Detail::Allocation::Mover {
...
+    my $self = shift;
+
+    my @errors = $self->SUPER::__errors__;
+    return @errors if @errors;
+
+    if ( $self->disk_group_name and $self->target_mount_path ) {
+        push @errors, UR::Object::Tag->create(
+            type => 'error',
+            properties => [qw/ group_name target_mount_path /],
+            desc => 'Can not specify both target group and volume!',
```
tmooney: Should we let both be specified as long as they don't conflict?

ebelter: @tmooney I wanted to do that, but the whole interface is strings, and not objects. In order to check them, the objects would need to be retrieved. I think that would be better, but it seemed that it should be done higher up, and I wanted to keep the PR small. It would be much trouble to update the commands and allocation to pass objects around. What do you think?

tmooney: Your reasoning is sound!  I wouldn't mess with changing everything just to allow both.

lib/perl/Genome/Disk/Command/Allocation/Move.pm
```diff
@@ -47,23 +47,31 @@ sub execute {
...
+}
 
+sub _resolve_move_params_for_allocation {
+    my ($self, $allocation) = @_;
+
+    my %params;
+    if ($self->target_volume) {
+        $params{target_mount_path} = $self->target_volume->mount_path;
+    }
+    else {
```
tmooney: If the user specifies both a `--target-volume` and a `--target-group` in this command should it check to make sure they match before ignoring the latter parameter?  (In this case we do have both the objects!)

ebelter:
@tmooney At your convenience, please review if this volume/group check looks good!

tmooney:
The check looks good to me!  I'd add a test where both volume and group are passed to the command to show that it properly only sends along one to the mover itself.

tmooney:
+1 :movie_camera:

brummett:
+1

PR #1059, opened 2015-10-19
Disk Group Validate: add info_archive and update test

ebelter:


lib/perl/Genome/Site/TGI/SiteLib/Genome/Disk/Group/Validate/GenomeDiskGroups.t
```diff
@@ -22,12 +24,12 @@ do {
         unix_uid => 0,
         unix_gid => 0,
     );
-    ok( ! exception { $DB::single = 1; $group->validate },
-        'no exception thrown when disk_group_name is GENOME_DISK_GROUP_MODELS'
-    );
+    ok(!$group->validate, 'no exception thrown when disk_group_name is GENOME_DISK_GROUP_MODELS');
```
tmooney: This could use `lives_ok`.  That way we don't have to care about the return value from `validate`.

tmooney:
+1.

Do we need to care if someone might now try to specify the archive disk group when creating/moving/etc. allocations?

ebelter:
@tmooney I can add the check to not allow that to be a group. If it becomes necessary, we can allow it!

ebelter:
@tmooney Pushed some commits for your suggestions. Please take a look when you can.

brummett:
+1

tmooney:
+1 again :smiley:

PR #1060, opened 2015-10-20
Stop testing that Genome classes work under older Perl.

tmooney:
LIMS is no longer calling these directly.

(This "fixes" the tests for #1056.)


brummett:
+1

ebelter:
+! Sweet! :purse:

PR #1061, opened 2015-10-20
Operations QC WGS Coverage

jasonwalker80:
Add operations qc WGS coverage calculation.

This is a work-in-progress.  The GENOME_TERRITORY metric is calculated by CollectWgsMetrics.  I am unaware of an alternative GMS source for the number of non-N bases for each reference sequence.  

A software result for each RefSeq could be made that runs a tool or grep that counts the number on non-N bases and it could have it's own metric.  This software result would be generated once per RefSeq.  In that case, CollectWgsMetrics, the longest running QC tool, could be turned off based on the current CCDG/HiSeqX10 operations definition of WGS coverage.

@tmooney, @acoffman, and @susannasiebert feedback, suggestions or alternative approach?


lib/perl/Genome/Qc/Command/BuildMetrics.t
```diff
@@ -49,7 +55,7 @@ my $pkg = 'Genome::Qc::Command::BuildMetrics';
 use_ok($pkg);
 
 my $test_dir = __FILE__.'.d';
-my $expected_output_file = $test_dir .'/expected_build_metrics.yml';
+my $expected_output_file = $test_dir .'/expected_build_metrics_v2.yml';
```
tmooney: Since this is in the repo, we can edit the original `expected_build_metrics.yml` if nothing else is using it.

tmooney:
If the result of running this tool is only the creation of a metric, it could be stored as a build metric on the refseq build itself rather than creating a new result.  We could make a command like `Genome::Model::ReferenceSequence::Command::CalculateGenomeTerritory` that could either run by hand or be added to the refseq build process to set this metric.

It might be possible to run it in the else block where the error message is now, but we do have to worry about concurrency then (with which a SoftwareResult could help!).

lib/perl/Genome/Qc/Command/BuildMetrics.t
jasonwalker80: commits pushed

jasonwalker80:
A new set of commits to generate the GENOME_TERRITORY metric if necessary.

tmooney:
+1 :metro:

acoffman:
+1

If you wanted, you could use `tee` in your shellcmd so that you only have to read through the file one time, but I wouldn't let that hold up this PR.

PR #1063, opened 2015-10-21
Disk Volume Under Allocated

ebelter:
This changes the return value of the disk volume under allocated command to be undef if under allocated volumes are found and removess the built in alert system. Instead of alerts being sent from the command, the return value can be monitored. Also, update it to be a true Command::V2 and not split/join disk group names on commas.


tmooney:
+1 :alien:

brummett:
+1

PR #1064, opened 2015-10-21
Build View Status XML: remove unused load jobs from lsf file methods

ebelter:
properties


tmooney:
+1. It looks like there's some related commented-out code in the Event view that can also be removed.

brummett:
+1

PR #1066, opened 2015-10-22
Model create delete

brummett:
Fixes APIPE-3017 by using a rolled-back transaction to clean up after a problem in Genome::Model::create(), rather than calling delete() on a partially constructed object.


lib/perl/Genome/Model.pm
```diff
@@ -471,6 +471,8 @@ sub __display_name__ {
 sub create {
     my $class = shift;
 
+    no warnings 'redefine';
+    local *__errors__ = sub { () };
```
tmooney: What does suppressing errors here do?

brummett: When the software transaction commits at the end of create(), it checks whether all the objects changed within are valid (same as when the main Context commits).  The Model object it's created might not yet be completely valid, but it's valid enough to pass back to the caller.

ebelter:
+1

lib/perl/Genome/Model.pm
tmooney: If we're pretending to the caller that there's not a transaction here, would it make sense to override `UR::Context::Transaction->changes_can_be_saved` instead of picking and choosing which errors to ignore?  (Or, would it be overkill to introduce an option for software transactions to not check validity of objects before promoting them to the next context?)

brummett: Overriding ```UR::Context::Transaction->changes_can_be_saved``` would be a gigantic break of encapsulation.  Adding an option to check validity isn't too onerous; I'll go with that.

tmooney:
+1 :dolphin:

PR #1067, opened 2015-10-22
Remove obsolete postgres logging

brummett:
These logs were used during the transition from Oracle to Postgres to judge whether our Postgres implementation was ready.  Now that that job is done, and we have no intention of going back to Oracle, this code can be removed.


tmooney:
I think we can also remove the `$ENV{LIMS_PERL}` exception from this and `Genome/Model/Tools/Lims.pm`.  It looks like they're up to a recent enough version of `DBD::Pg`:

```
$ lims-perl -MDBD::Pg -e 'print $DBD::Pg::VERSION,"\n"'
2.19.3
```

ebelter:
+1

acoffman:
+1

PR #1068, opened 2015-10-22
Add Genome Site Command

ebelter:
Add directory for site commands. Consolidate commands into this directory.


ebelter:
Command is in https://github.com/genome/wugc-hudson/pull/35

brummett:
+1

lib/perl/Genome/Site/TGI/Command/LimsAndGenome.pm
```diff
@@ -1,4 +1,4 @@
-package Genome::Site::TGI::Synchronize::SyncLimsAndGenome;
+package Genome::Site::TGI::Command::LimsAndGenome;
```
tmooney: It might be better to keep a verb in this command name.

lib/perl/Genome/Site/TGI/Command/SyncCaptureSets.pm
```diff
@@ -1,4 +1,4 @@
-package Genome::Site::TGI::CaptureSet::Command::Synchronize;
+package Genome::Site::TGI::Command::SyncCaptureSets;
```
tmooney: I think it's time to retire this script (and its Jenkins job)!  These are now created using `genome feature-list import` directly.

ebelter:
@tmooney Suggestions applied!

tmooney:
+1 :sake:

PR #1069, opened 2015-10-26
Remove old "Msi" and "ReferenceVariationSanger" definers (and build class)

tmooney:
The processing profile and implementation of these pipelines were removed in 2012.  No instances of these model or build types exist in our database.  Thus the define commands are useless (They would fail if run.), and the `Msi` build class does not need to remain to load objects.


ebelter:
+1

brummett:
+1

PR #1070, opened 2015-10-26
New Very Specific Pipeline for Alignment + HaplotypeCaller

tmooney:
Here's an initial implementation of the ~~`ReferenceVariation`~~ `SingleSampleGenotype` pipeline.

A diagram of its steps:
* Alignment
  * QC
  * HaplotypeCaller

The Alignment Dispatcher is not used so those last two steps can happen in parallel.

There are two user commands to generate:
* AlignerIndex
* ReferenceSequence "Buckets"

These are not run within the pipeline, but must be run once per aligner/reference and once per reference.  Particularly in the case of the buckets, we may want to manually tweak the result that is used in this pipeline.  The idea here was to make the actual pipeline simpler and swifter by not scheduling separate steps to check for and possibly generate these, but the existing commands could be added as steps if that'd be preferable.


susannasiebert:
+1

acoffman:
+1

lib/perl/Genome/Model/Build/SingleSampleGenotype.pm
```diff
@@ -0,0 +1,74 @@
...
+        buckets_result => {
+            is => 'Genome::Model::Build::ReferenceSequence::Buckets',
+            doc => 'the buckets used for parallelization',
+        },
+    ],
+};
+
+sub workflow_name {
+    my $self = shift;
+    return $self->build_id . ' Reference Variation Pipeline';
```
dufeiyu: Isn't the name singlesamplegenotype ?

lib/perl/Genome/Model/SingleSampleGenotype.t
```diff
@@ -0,0 +1,33 @@
...
+my $pkg = 'Genome::Model::SingleSampleGenotype';
+use_ok($pkg);
+
+my $pp = Genome::Test::Factory::ProcessingProfile::SingleSampleGenotype->setup_object();
+my $reference = Genome::Test::Factory::Model::ReferenceSequence->setup_reference_sequence_build();
+my $sample = Genome::Test::Factory::Sample->setup_object;
+
+my $model = $pkg->create(
+    processing_profile => $pp,
+    name => 'Model/RefereVariation.t test',
```
susannasiebert: Another straggler

lib/perl/Genome/Model/SingleSampleGenotype/Command/HaplotypeCaller.pm
```diff
@@ -0,0 +1,61 @@
...
+    my $build = $self->build;
+
+    my $result_users = Genome::SoftwareResult::User->user_hash_for_build($build);
+
+    my @params = (
+        alignment_result => $build->merged_alignment_result,
+        result_users => $result_users,
+        haplotype_caller_version => $build->model->haplotype_caller_version,
+        emit_reference_confidence => 'GVCF',
+        gvcf_gq_bands => [5,20,60],
```
dufeiyu: Could these values be changed ? Furthermore, maybe "haplotype_caller_params" that combines emit_reference_confidence, gvcf_gq_bands and intervals can be added into model has_param so users can change these parameters via pp.

jasonwalker80: This is a good suggestion but I would wait on implementing these as processing profile param(s) until absolutely necessary.  This pipeline is intended to be very specific.  I think it's ok to wait on flexible features until the need arises.

jasonwalker80:
+1

PR #1071, opened 2015-10-27
Add formatters to ResourceParser

davidlmorton:
This ensures that later, when we prepare to submit a workflow to PTero, that the JSON formatter treats some values as numbers instead of strings.

### DIFF
--- a/lib/perl/Genome/Sys/LSF/ResourceParser.pm
+++ b/lib/perl/Genome/Sys/LSF/ResourceParser.pm
@@ -81,9 +81,9 @@ sub _create_option_spec {
     return ("$option=s" => sub { my ($option_name, $value, )});
 }
 
-sub null_formatter {
+sub string_formatter {
     my $value = shift;
-    return $value;
+    return "$value";
 }
 
 sub number_formatter {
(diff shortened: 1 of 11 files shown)


lib/perl/Genome/Sys/LSF/ResourceParser.pm
```diff
@@ -79,6 +81,42 @@ sub _create_option_spec {
...
+    my $value = shift;
+    return $value;
+}
+
+sub number_formatter {
+    my $value = shift;
+    return $value + 0;
+}
+
+my %formatters = (
```
tmooney: If you want to encapsulate this lookup hash, it could be made a [`state`](http://perldoc.perl.org/functions/state.html) variable inside of `_formatters`.

lib/perl/Genome/Sys/LSF/ResourceParser.pm
```diff
@@ -79,6 +81,42 @@ sub _create_option_spec {
     return ("$option=s" => sub { my ($option_name, $value, )});
 }
 
+sub null_formatter {
+    my $value = shift;
+    return $value;
```
tmooney: Is it ever important that the value definitely be a string and not a number (i.e. `return "$value";`)?

tmooney:
+1 once the testcases are updated!

PR #1072, opened 2015-10-27
Use submit_url option when submitting PTero workflows

davidlmorton:
Instead of setting the PTERO_WORKFLOW_SUBMIT_URL env variable.  This uses features made available here: https://github.com/genome/ptero-perl-sdk/pull/96


tmooney:
+1 :clapper:

acoffman:
+1

PR #1073, opened 2015-10-27
Lock before setting build metrics.

tmooney:
The changes introduced in #1061 mean that multiple processes could be trying to set the same metric for a build.  This change is to stop them from accidentally trying to commit multiple copies of the same metric.

This is similar to a `Genome::Utility::ObjectWithLockedConstruction`, but in this case the new value is kept rather than the old one.


acoffman:
+1

jasonwalker80:
+1

PR #1074, opened 2015-10-27
Add wrappers to run the epitope prediction pipeline with the variant reporting components

susannasiebert:
This is step two in in adding frameshift mutations and inframe indels to the epitope prediction pipeline. In step one (https://github.com/genome/genome/pull/1031) the fasta generation was moved to the variant reporting system. This PR now adds various wrappers:
- A command to run the variant reporting system to create the fasta file
- A command that runs a workflow to execute the remaining steps in the epitope prediction pipeline with the fasta file from the first command as an input.
- A wrapper to run these two workflows in succession.


lib/perl/Genome/Model/Tools/EpitopePrediction/NetmhcPipeline.pm
```diff
@@ -0,0 +1,371 @@
...
+    my $workflow = $self->_construct_workflow();
+
+    $self->debug_message("Getting Workflow Inputs...");
+    my $inputs = $self->_get_workflow_inputs();
+
+    $self->debug_message("Running Workflow...");
+    my $result = $workflow->execute(inputs => $inputs);
+
+    unless($result){
+        $self->error_message( join("\n", map($_->name . ': ' . $_->error, @Workflow::Simple::ERROR)) );
```
tmooney: `Genome::WorkflowBuilder::DAG->execute()` takes care of reporting errors, so we shouldn't need to `use Workflow::Simple` any more when creating new workflows.

lib/perl/Genome/Model/Tools/EpitopePrediction/NetmhcPipeline.pm
```diff
@@ -0,0 +1,371 @@
...
+        die $self->error_message("Workflow did not return correctly.");
```
tmooney: This would be a prime opportunity to use the new `$self->fatal_message`!

lib/perl/Genome/Model/Tools/EpitopePrediction/NetmhcPipeline.pm
```diff
@@ -0,0 +1,371 @@
...
+            die $self->error_message("Sample name must be defined if no somatic variation build is given")
+        }
+    }
+
+    unless (-s $self->input_fasta_file) {
+        die $self->error_message("Input fasta file %s does not exist or has no size", $self->input_fasta_file);
+    }
+
+    unless (Genome::Sys->create_directory($self->output_directory)) {
+        die $self->error_message("Coult not create directory (%s)", $self->output_directory);
```
tmooney: The last part of `Could` has become unvoiced.

lib/perl/Genome/Model/Tools/EpitopePrediction/NewPipeline.pm
```diff
@@ -0,0 +1,192 @@
+package Genome::Model::Tools::EpitopePrediction::NewPipeline;
```
tmooney: Will this still be the `NewPipeline` in five years?  (Not that I have better names to suggest!)

lib/perl/Genome/Model/Tools/EpitopePrediction/NewPipeline.pm
```diff
@@ -0,0 +1,192 @@
...
+            is => 'Text',
+            doc =>
+                'Type of epitopes to report in the final output - select \'top\' to report the top epitopes in terms of fold changes,  \'all\' to report all predictions ',
+            valid_values => ['top', 'all'],
+        },
+    ],
+    has_calculated_optional => [
+        build => {
+            calculate_from => [qw(model)],
+            calculate => q/$model->last_succeeded_build/,
```
tmooney: Would it make sense to let the user specify a build directly?  If they gave a model ID they'd get the interactive prompt asking them to choose which build to use.

lib/perl/Genome/Model/Tools/EpitopePrediction/NewPipeline.pm
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    return;
+}
+
+sub translations_file {
+    my $self = shift;
+    my $translations_file = Genome::Sys->create_temp_file_path;
+    my %translations;
+    $translations{reference_fasta} = $self->build->tumor_build->reference_sequence_build->full_consensus_path("fa");
```
tmooney: `Genome::Model::Build::SomaticVariation` has a `reference_sequence_build` method defined.  (It goes through the normal, but the build will be `Unstartable` if the two don't match.)

lib/perl/Genome/Model/Tools/EpitopePrediction/NewPipeline.pm
```diff
@@ -0,0 +1,192 @@
...
+
+sub plan_file {
+    my ($self, $variant_type) = @_;
+    my $base_dir = dirname(dirname(dirname(dirname(__FILE__))));
+    return File::Spec->join($base_dir, 'VariantReporting', 'plan_files', "epitope_prediction_$variant_type.yaml");
+}
+
+sub input_vcf {
+    my ($self, $variant_type) = @_;
+    my $accessor = "get_detailed_$variant_type"."_vcf";
```
tmooney: This looks like it wants to be `"get_detailed_${variant_type}_vcf"`.

lib/perl/Genome/VariantReporting/Command/Wrappers/EpitopeBindingPredictionFasta.pm
```diff
@@ -0,0 +1,78 @@
...
+    my $self = shift;
+    $self->generate_translations_file;
+    $self->run_reports;
+}
+
+
+sub generate_translations_file {
+    my $self = shift;
+    my %translations;
+    $translations{reference_fasta} = $self->build->tumor_build->reference_sequence_build->full_consensus_path("fa");
```
tmooney: This looks familiar.  How much do this and the GMT have in common?

susannasiebert: Yeah, there is some overlap, mostly in the helper functions which are actually used in other places as well. Time for a ReportHelpers class of sorts!

lib/perl/Genome/Model/Tools/EpitopePrediction/NewPipeline.pm
```diff
@@ -0,0 +1,192 @@
...
+    my $translations_file = Genome::Sys->create_temp_file_path;
+    my %translations;
+    $translations{reference_fasta} = $self->build->tumor_build->reference_sequence_build->full_consensus_path("fa");
+    YAML::DumpFile(File::Spec->join($translations_file), \%translations);
+    return $translations_file;
+}
+
+sub plan_file {
+    my ($self, $variant_type) = @_;
+    my $base_dir = dirname(dirname(dirname(dirname(__FILE__))));
```
tmooney: Is this the equivalent of `Genome->base_dir`?

lib/perl/Genome/Model/Tools/EpitopePrediction/NetmhcPipeline.pm
susannasiebert: Oh, good to know. So this line can just go away? There is a similar line on the Pipeline.pm as well...

lib/perl/Genome/Model/Tools/EpitopePrediction/NewPipeline.pm
susannasiebert: The new system doesn't support bed files which is currently still a feature used by some analysts. Ultimately we want to move to this piepleine exclusively but it might require adding on some pre-processing of bed to files to create vcf files that can be used in the variant reporting system. Once that is hall hashed out the old pipeline should go away and this can be renamed.

lib/perl/Genome/Model/Tools/EpitopePrediction/NewPipeline.pm
susannasiebert: Yes. I was planning on changing that before making this PR but it fell through the cracks.

lib/perl/Genome/Model/Tools/EpitopePrediction/NewPipeline.pm
susannasiebert: It might be which would be awesome! Serves me right for just copy and pasting code from different places.

lib/perl/Genome/Model/Tools/EpitopePrediction/NetmhcPipeline.pm
tmooney: Looks like the one in `Pipeline.pm` is a holdover from 22ec93038d9b973e9a2339cfdc3f7722e736f558 before it was changed to use `WorkflowBuilder`.

susannasiebert:
@tmooney I just pushed a few commits that I believe address all of your comments.

lib/perl/Genome/VariantReporting/Command/Wrappers/EpitopeBindingPredictionFasta.pm
```diff
@@ -0,0 +1,87 @@
+package Genome::VariantReporting::Command::Wrappers::EpitopeBindingPredictionFasta;
+
+use strict;
+use warnings;
+use Genome;
+use File::Basename qw(dirname);
```
tmooney: `dirname` is no longer used here.

lib/perl/Genome/Model/Tools/EpitopePrediction/NewPipeline.pm
```diff
@@ -0,0 +1,161 @@
+package Genome::Model::Tools::EpitopePrediction::NewPipeline;
+
+use strict;
+use warnings;
+use Genome;
+use File::Basename qw(dirname);
+use YAML;
```
tmooney: Neither `File::Basename` nor `YAML` are still used here.

lib/perl/Genome/Model/Tools/EpitopePrediction/NewPipeline.pm
```diff
@@ -0,0 +1,161 @@
...
+class Genome::Model::Tools::EpitopePrediction::NewPipeline {
+    is => 'Command::V2',
+    has_input => [
+        output_directory => {
+            is => 'Text',
+            doc => 'the directory where you want results stored',
+        },
+        build => {
+            is => "Genome::Model::Build::SomaticVariation",
+        },
```
tmooney: This parameter stands out as not having any `doc`umentation.

lib/perl/Genome/Model/Tools/EpitopePrediction/NewPipeline.pm
```diff
@@ -0,0 +1,161 @@
...
+}
+
+sub validate_inputs {
+    my $self = shift;
+
+    if (-d $self->output_directory) {
+        $self->fatal_message("Output directory %s already exists", $self->output_directory);
+    }
+
+    Genome::Sys::make_path($self->output_directory);
```
tmooney: This should use `Genome::Sys->create_directory` to make sure permissions are set appropriately.

tmooney:
+1 :shirt:

acoffman:
+1 :shipit:

PR #1076, opened 2015-10-28
Don't Generate RefAlign Summary Report Data Twice

tmooney:
The failing testcase in #1073 pointed out that we've been doing the work twice in this report.  This saves the result from the first time to use the second time.


ebelter:
+1

brummett:
+1

PR #1077, opened 2015-10-28
Look for output from new UR fatal_message().

tmooney:
We now need to be able to find messages like these in the logs:
```
$ genome-perl -e 'use UR; UR::Object->fatal_message("Example!");'
FATAL: Example! at -e line 1
```


brummett:
+1

I was going to do the same thing, but didn't think anyone was using it yet.

ebelter:
+1

PR #1078, opened 2015-10-29
Add adjacent normal to the case of is_not_tumor

dufeiyu:
Adjacent normal is not tumor. Refer to https://rt.gsc.wustl.edu/Ticket/Display.html?id=108687. This PR needs early merge.


tmooney:
+1 :tulip:

brummett:
+1

PR #1079, opened 2015-10-30
Refactor of epitope prediction NetMHC parser

susannasiebert:
This PR is step three in in adding frameshift mutations and inframe indels to the epitope prediction pipeline. This PR does a refactor of the module which parses output from the NetMHC software. It now is able to correctly read in files that contain frameshifts and inframe indels. However, the actual parsing right now filters out all frameshifts since the wildtype peptide sequence has no overlap with the downstream peptide sequence "mutant positions". This will need further consideration from @chrisamiller  and @jhundal


chrisamiller:
Hi Susanna, this looks great.  

>the actual parsing right now filters out all frameshifts since the wildtype peptide sequence has no overlap with the downstream peptide sequence "mutant positions".

I thought we had already outlined a strategy for this, namely that when there is no corresponding wildtype sequence, there will be no WT prediction, and that field should just contain "NA".  Is that a quick fix that you can add?

jhundal:
Yes I think we had talked about adding  separate parsers based on type of mutation. And as Chris mentioned, we can just add NAs for frameshift WT counterpart.

susannasiebert:
Yes it is a quick fix to add that. There is no need to make a separate parser.

I just need some clarification on which fields you would like to display as NA. These are currently the fields in the output:
* 'Gene Name'
* 'Point Mutation'
* 'Sub-peptide Position'
* 'MT score'
* 'WT score'
* 'MT epitope seq'
* 'WT epitope seq'
* 'Fold change'

jhundal:
WT score, WT epitope seq and Fold change.
This will only be in cases where there is no WT counterpart. If there is an insertion or deletion of 2-3 amino acids, then we would like to retain these fields and report wildtype values where possible.

chrisamiller:
1) It also affects the "Sub-peptide position" column. Let's introduce a special value there: "novel_sequence", to be used for frameshifts past the end of the protein

2) I'm having trouble finding the code that states what goes into the 'Point Mutation' column for indels.

3) We should probably change that column header to just "Mutation", as 'point mutation' implies a 1bp event.

jhundal:
@chrisamiller  : As for #1, if the new (MT) amino acid sequence resulting from frameshift sequence is, say 20bp, we will need to the Sub-peptide position since we are only interested in 8-10mer epitopes, and this field indicates starting position of the 8-10mer in the entire novel sequence.

susannasiebert:
>WT score, WT epitope seq and Fold change.
Ok
>This will only be in cases where there is no WT counterpart.
Just so we are on the same page with out terminology: this applies to frameshifts only?
>If there is an insertion or deletion of 2-3 amino acids, then we would like to retain these fields and report wildtype values where possible.
And this would be the proposed handling of inframe indels? Or am I missing a case where there could be no WT counterpart and we are having an inframe indel?

I believe inframe indels already get handled correctly but it would be great if you guys could double check the test data: `/gsc/var/cache/testsuite/data/Genome-Model-Tools-EpitopePrediction-ParseNetmhcOutput/3/netmhc_file.3.4.indels.xls` Entry_5 and Entry_6 are an insertion. The parsed file is at `/gsc/var/cache/testsuite/data/Genome-Model-Tools-EpitopePrediction-ParseNetmhcOutput/3/parsed_file.3.4.indels.all`

For frameshifts (i.e., no WT counterpart) would the "all" option still output all of the subpeptides of the mutant? And for the "top" option is the current handling of picking the one with the highest mt score correct?

susannasiebert:
> 1) It also affects the "Sub-peptide position" column. Let's introduce a special value there: "novel_sequence", to be used for frameshifts past the end of the protein

Can you be more specific as to what is considered "frameshifts past the end of the protein"? Any mutant subpeptide that doesn't have any overlap with the wildtype sequence?

> 2) I'm having trouble finding the code that states what goes into the 'Point Mutation' column for indels.

It is the last part of the protein identifier (second column in the key file), e.g. FS.215 for the idenitifier >WT.UBR5.FS.215 (frameshift) and G733GK for >WT.NFX1.G733GK (insertion).

3) We should probably change that column header to just "Mutation", as 'point mutation' implies a 1bp event.

OK

susannasiebert:
I think it would be best if you guys could come up with some test data (netmhc output file and the corresponding parsed entries) even if it would be totally made up. This would be really helpful.

chrisamiller:
@susannasiebert :
- NAs will also be necessary in large insertions. I believe the metric we decided on was any time that the WT sequence differed from the mutant sequence by more than 3 AAs (due to large in-frame or frameshift events), the wildtype comparison was meaningless and should just be NA.  (correct me if I'm misremembering the exact number - it may have been 2AAs)

@jhundal :
> As for #1, if the new (MT) amino acid sequence resulting from frameshift sequence is, say 20bp, we will need to the Sub-peptide position since we are only interested in 8-10mer epitopes, and this field indicates starting position of the 8-10mer in the entire novel sequence.

In the case of SNVs, that position shows where the SNV is in the MT sequence, right?  In the case of frameshift, once you're into the new sequence, the entire thing is novel and unrelated to the MT sequence. Thinking about it farther, though, I guess you'd just record the position at which the new sequence begins.  So, two cases, where W is wildtype and M is mutant AAs:

    WWWWMWWW  #pos is 5
    WWWWMMMM  #pos is 5
    MMMMMMMM #pos is 1(no matter how far into the novel frameshift we are)

Does that make sense, or am I missing something?

jhundal:
@chrisamiller  : No, the sub-peptide position is assigned by NetMHC to specify the starting position of the 8,9,or 10-mer epitope in the entire n-mer fasta.
It is know as "pos" in the netmhc output : 
http://www.cbs.dtu.dk/services/NetMHC/output.php

susannasiebert:
The original parsing code only compared the wildtype and the mutant at the same sub-peptide position; which makes sense since we only have a 1bp replacement so the sequences at the same sub-peptide positions will "match up".

Now that we are encountering inframe indels do we also need to start comparing the mutants at different sub-petite positions depending on the length of the indel? A given wildtype sub-peptide sequence will have multiple mutant sub-peptide sequences that can be regarded as a "match" depending on whether you look at the wildtype sequence before or after the indel, correct?

Example full sequence
>ABCDE
ABCMDE C->CM insertion

current 4-mer comparisons:
>ABCD - ABCM - positon 0
BCDE - BCMD - position 1

addition 4-mer comparison to match after the indel?
>ABCD - BCMD
BCDE - CMDE

susannasiebert:
@chrisamiller @jhundal I pushed a few commits to implement the handling of frameshifts and large indels as we had discussed during our last meeting. I also implemented the suggestions you gave in this PR.

chrisamiller:
I've looked through this and can't find anything obviously wrong.  +1

chrisamiller:
Why don't you go ahead and merge this?  We'll then test it out on a few caes and then can give additional feedback. Thanks!

PR #1080, opened 2015-10-30
Ncallers fix

apregier:
The reports were including variants that were filtered out upstream by one or more of the individual callers, but those callers were being counted toward the 'n-callers' filter cutoff, which is not desirable.  I also did a small refactor so that the n-callers filter and variant-callers interpreter now use the same bit of code to decide which filters to count/include


susannasiebert:
+1

tmooney:
+1 :blowfish:

apregier:
We are holding off on merging this until another cle fix is ready, as we will likely need to make new blessed processes.

apregier:
@ernfrid this latest commit should fix it - the example I looked at used '.' instead of PASS in the FT fields, but our process has 'PASS', so it was thinking that all of the callers had been filtered.

PR #1081, opened 2015-11-02
removing comma that makes text harder to select

chrisamiller:


brummett:
+1

tmooney:
+1 :speech_balloon:

lib/perl/Genome/ModelGroup/Command/Create.pm
```diff
@@ -55,7 +55,7 @@ sub execute {
         return;
     }
     $self->status_message('Created model group:');
-    $self->status_message('ID: ' . $model_group->id . ', NAME: ' . $model_group->name);
+    $self->status_message('ID: ' . $model_group->id . '  NAME: ' . $model_group->name);
```
dufeiyu: It would be better if format print is used:
$self->status_message("ID: %s    NAME: %s", $model_group->id, $model_group->name);

dufeiyu:
+1  with or without that change

PR #1084, opened 2015-11-02
Exclude index results from CLE verification.

tmooney:
These results do not depend on any of the data being analyzed, and as such are not CLE-specific.


apregier:
+1

davidlmorton:
+1

PR #1085, opened 2015-11-03
Lsf cputime

mkiwala:


davidlmorton:
+1 It sucks that the regenerate test data thing adds so much spam to the commits.  That makes it harder to evaluate the PR than it should be.  One could use 'git commit -p' or make the regeneration logic better, but it shouldn't stop this PR.

mkiwala:
I had considered using `git commit -p` to patch-in just the changes that were important.  I could have also just made the modifications that I wanted without setting `GENERATE_TEST_DATA=1`.

tmooney:
+1, my sentiment is the same as @davidlmorton's.

PR #1088, opened 2015-11-03
missed some defines

ernfrid:
Missed a few additional places where this data should have been tested using defined rather than truth.

### DIFF
--- a/lib/perl/Genome/Model/Build/ReferenceSequence/Converter.pm
+++ b/lib/perl/Genome/Model/Build/ReferenceSequence/Converter.pm
@@ -81,7 +81,7 @@ sub is_per_position_algorithm {
     my $self = shift;
     my ($algorithm) = @_;
 
-    return !grep($_ eq $algorithm, 'liftOver', 'no_op', 'drop_extra_contigs');
+    return !grep($_ eq $algorithm, 'lift_over', 'no_op', 'drop_extra_contigs');
 }
 
 sub convert_position {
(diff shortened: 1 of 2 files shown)


brummett:
Is it really valid for ```$chrom``` or ```$new_chrom``` to be 0 or an empty string?

ernfrid:
Probably not. Same for the stop, it should be an integer.

ernfrid:
Better now @brummett ?

gatoravi:
Unrelated to your changes, but might be a bug, should https://github.com/ernfrid/genome/blob/master/lib/perl/Genome/Model/Build/ReferenceSequence/Converter.pm#L84 check for "lift_over" instead of "liftOver"?

ernfrid:
I think you are correct and have made the change as well as harmonized capitalization on an error message.

brummett:
As you have it now, ```0``` would be an acceptable chromosome; that seems wrong.
Should it be ok if both ```$start``` and ```$stop``` were 0 to denote an insertion?
Is this the place to check that ```$stop``` is freater or equal to ```$start```?

ernfrid:
I can't find any documentation of the BED format that suggests that 0 shouldn't be an allowable string name for a chromosome (though I agree it would be bad form to do so).

I think technically, `$stop` is supposed to be one-based, but I think you are correct in terms of how we utilize this internally so I have changed that.

brummett:
+1

apregier:
+1

PR #1089, opened 2015-11-03
Run the HaplotypeCaller serially on each contig in a bucket.

tmooney:
It has been decided that it is preferable to produce per-chromosome GVCFs rather than have collections of chromosomes bundled together. In order to still have consistently-sized jobs (and avoid splitting hundreds of ways, which will pound the source BAM file), run the HaplotypeCaller on each entry in the bucket in turn.


jasonwalker80:
+1

ebelter:
+1

PR #1090, opened 2015-11-03
Export/Copy Allocations to 'ad hoc' disk

jasonwalker80:
Initial revision of copy/export commands for allocations.

### DIFF
--- a/lib/perl/Genome/Disk/Command/Allocation/Export.pm
+++ b/lib/perl/Genome/Disk/Command/Allocation/Export.pm
@@ -37,7 +37,7 @@ sub execute {
 
     my $existing_allocation = Genome::Disk::Allocation->get_allocation_for_path($self->output_dir);
     if ( $existing_allocation && ($existing_allocation->absolute_path eq $self->output_dir) ) {
-        die('Can not export allocation to allocated disk direcotory: '. $self->output_dir);
+        die('Can not export allocation to allocated disk directory: '. $self->output_dir);
     }
 
     for my $allocation ($self->allocations) {
--- a/lib/perl/Genome/Disk/Command/Allocation/Export.pm
+++ b/lib/perl/Genome/Disk/Command/Allocation/Export.pm
@@ -41,10 +41,10 @@ sub execute {
     }
 
     for my $allocation ($self->allocations) {
-        $allocation->copy( output_dir => $self->output_dir );
+        $allocation->copy( output_dir => File::Spec->join($self->output_dir, $allocation->id) );
     }
 
-    $self->status_message("Successfully copied allocations!");
+    $self->status_message("Successfully exported allocations!");
     return 1;
 }
 
--- a/lib/perl/Genome/Disk/Command/Allocation/Export.pm
+++ b/lib/perl/Genome/Disk/Command/Allocation/Export.pm
@@ -35,6 +35,11 @@ sub help_brief {
 sub execute {
     my $self = shift;
 
+    my $existing_allocation = Genome::Disk::Allocation->get_allocation_for_path($self->output_dir);
+    if ( $existing_allocation && ($existing_allocation->absolute_path eq $self->output_dir) ) {
+        die('Can not export allocation to allocated disk direcotory: '. $self->output_dir);
+    }
+
     for my $allocation ($self->allocations) {
         $allocation->copy( output_dir => $self->output_dir );
     }
--- a/lib/perl/Genome/Disk/Command/Allocation/Export.pm
+++ b/lib/perl/Genome/Disk/Command/Allocation/Export.pm
@@ -1,0 +1,47 @@
+package Genome::Disk::Command::Allocation::Export;
+
+use strict;
+use warnings;
+use Genome;
+
+class Genome::Disk::Command::Allocation::Export {
+    is => 'Command::V2',
+    has => [
+        allocations => {
+            is => 'Genome::Disk::Allocation',
+            is_many => 1,
+            shell_args_position => 1,
+            doc => 'Allocations to export',
+        },
+        output_dir => {
+            is => 'Text',
+            doc => 'The output directory to copy allocation contents to.',
+        },
+    ],
+    doc => 'copy allocations from allocated disk to another network filesystem path.',
+};
+
+sub help_detail {
+    return <<EOS
+Copies allocations from allocated disk to a specified output directory. 
+
+EOS
+}
+
+sub help_brief {
+    return 'copy allocations from allocated disk to a specified output directory';
+}
+
+sub execute {
+    my $self = shift;
+
+    for my $allocation ($self->allocations) {
+        $allocation->copy( output_dir => $self->output_dir );
+    }
+
+    $self->status_message("Successfully copied allocations!");
+    return 1;
+}
+
+1;
+
--- a/lib/perl/Genome/Disk/Detail/Allocation/Copier.pm
+++ b/lib/perl/Genome/Disk/Detail/Allocation/Copier.pm
@@ -45,20 +45,21 @@ sub copy {
         target_directory => $output_dir,
     );
     if ($allocation_object->is_archived) {
-        my $tar_path = $allocation_object->tar_path;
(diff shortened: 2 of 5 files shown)


jasonwalker80:
New commits.  The command works on a test allocation, both archived and active.  This may beyond WIP at this time.

jasonwalker80:
@brummett would you mind having a look at this PR?

lib/perl/Genome/Disk/Detail/Allocation/Copier.pm
```diff
@@ -0,0 +1,168 @@
...
+                # $allocation_object, and $archive_path).
+                my @signals = qw/ INT TERM /;
+                for my $signal (@signals) {
+                    $SIG{$signal} = sub {
+                        print STDERR "Cleanup activated within allocation, ",
+                            "cleaning up LSF jobs\n";
+                        eval { Genome::Sys->kill_lsf_job($job_id) } if $job_id;
+                        die "Received signal, exiting.";
+                    }
+                }
```
brummett: Something like this might be cleaner:
```
my $sig_handler = sub {
    print ...
    eval ...
    die ...
};
local @SIG{'INT','TERM'} = ($sig_handler, $sig_handler);
```
Then you can get rid of the ```for``` loop below that deletes the keys out of ```%SIG```, and it's safer in case this block is exited by other means and those deletes don't get called.

brummett: It actually looks like this whole section was copied from Genome::Disk::Detail::Allocation::Unarchiver.  This code should be using Unarchiver or the common code should be extracted somewhere else.

jasonwalker80: It all started in Archiver.  It appears it was copy/paste to Unarchiver.  The Copier version is slightly different because there is no need to cleanup the archive directory.  That would be bad.

Now that you point out Unarchiver, it appears an unsuccessful unarchive will cleanup the archive_path as well.  That seems bad.

lib/perl/Genome/Disk/Detail/Allocation/Copier.pm
```diff
@@ -0,0 +1,168 @@
...
+
+sub copy {
+    my $self = shift;
+
+    my $id = $self->allocation_id;
+
+    my $output_dir = $self->output_dir;
+
+    my ($allocation_object, $allocation_lock) =
+        Genome::Disk::Allocation->get_with_lock($id);
```
brummett: I think it makes sense to use a Scope::Guard here to do the unlocking
```
my $unlocker = Scope::Guard->new(sub { $allocation_lock->unlock if $allocation_lock });
```

jasonwalker80:
@brummett  I think this is ready for another look.  I believe comments have been addressed.

jasonwalker80:
@tmooney is the above commit something like you had in mind for checking allocation paths.

ebelter:
+1

lib/perl/Genome/Disk/Detail/Allocation/Unarchiver.pm
```diff
@@ -177,4 +133,49 @@ sub _swap_shadow_allocation {
     return 1;
 }
 
+sub _do_unarchive_cmd {
+    my $class = shift;
+    my $id = shift;
+    my $cmd = shift;
```
brummett: It would be better if this sub took an allocation object and a destination path.  Then you can pull out construction of the ```tar``` command from the other two places into this function.

lib/perl/Genome/Disk/Detail/Allocation/Unarchiver.pm
```diff
@@ -177,4 +133,49 @@ sub _swap_shadow_allocation {
...
+
+        # Entire path must be wrapped in quotes because older allocation
+        # IDs contain spaces If the command isn't wrapped in quotes, the
+        # '&&' is misinterpreted by bash (rather than being "bsub '1 && 2'
+        # it is looked at as 'bsub 1' && '2')
+        ($job_id, $status) = Genome::Sys->bsub_and_wait(
+            queue => Genome::Config::get('archive_lsf_queue'),
+            job_group => '/unarchive',
+            log_file => "/tmp/$id",
+            cmd => $cmd,
```
brummett: If you pass the command as an arrayref, then you don't have to worry about special characters messing up the shell
```cmd => ['tar', '-C', $target_path, '-xf', $tar_path]```

File: lib/perl/Genome/Disk/Command/Allocation/Export.pm [hunk 5]
```diff
@@ -0,0 +1,52 @@
...
+sub help_brief {
+    return 'copy allocations from allocated disk to a specified output directory';
+}
+
+sub execute {
+    my $self = shift;
+
+    my $existing_allocation = Genome::Disk::Allocation->get_allocation_for_path($self->output_dir);
+    if ( $existing_allocation && ($existing_allocation->absolute_path eq $self->output_dir) ) {
+        die('Can not export allocation to allocated disk direcotory: '. $self->output_dir);
```
tmooney: `directory` has gained an `o`.

tmooney:
Yes, that's what I had in mind for making sure this is really an export!

jasonwalker80:
I believe I have addressed all concerns and improved the sig handling that kills the LSF job.  Please review.

brummett:
+1

lib/perl/Genome/Disk/Command/Allocation/Export.pm
```diff
@@ -0,0 +1,52 @@
...
+    my $existing_allocation = Genome::Disk::Allocation->get_allocation_for_path($self->output_dir);
+    if ( $existing_allocation && ($existing_allocation->absolute_path eq $self->output_dir) ) {
+        die('Can not export allocation to allocated disk directory: '. $self->output_dir);
+    }
+
+    for my $allocation ($self->allocations) {
+        $allocation->copy( output_dir => $self->output_dir );
+    }
+
+    $self->status_message("Successfully copied allocations!");
```
tmooney: Should this status message say `exported` instead of `copied`?

lib/perl/Genome/Disk/Command/Allocation/Export.pm
```diff
@@ -0,0 +1,52 @@
...
+sub execute {
+    my $self = shift;
+
+    my $existing_allocation = Genome::Disk::Allocation->get_allocation_for_path($self->output_dir);
+    if ( $existing_allocation && ($existing_allocation->absolute_path eq $self->output_dir) ) {
+        die('Can not export allocation to allocated disk directory: '. $self->output_dir);
+    }
+
+    for my $allocation ($self->allocations) {
+        $allocation->copy( output_dir => $self->output_dir );
```
tmooney: Depending what these allocations are, it seems likely there will be collisions among the filenames of the contents.  Would it make sense to pass in `FIle::Spec->join($self->output_dir, $allocation->id)` as the `output_dir` here?  Or should the command fail if there's a collision?

jasonwalker80:
@tmooney I believe I've addressed your comments with the latest commit.

tmooney:
+1, I see that `copied` has been added to the database:

```
$ ur list objects --subject-class 'Genome::Timeline::Event::AllocationEventType' --show id
ID
--
archived
copied
created
finalized
invalidated
moved
preserved
purged
reallocated
strengthened
unarchived
unpreserved
weakened
```

lib/perl/Genome/Disk/Command/Allocation/Export.pm
```diff
@@ -0,0 +1,52 @@
...
+
+sub help_brief {
+    return 'copy allocations from allocated disk to a specified output directory';
+}
+
+sub execute {
+    my $self = shift;
+
+    my $existing_allocation = Genome::Disk::Allocation->get_allocation_for_path($self->output_dir);
+    if ( $existing_allocation && ($existing_allocation->absolute_path eq $self->output_dir) ) {
```
tmooney: This might also want to check that the `output_dir` is not inside the allocation's directory.

PR #1091, opened 2015-11-03
Unarchiving now requires an AnP instead of a lab.

tmooney:
Unarchiving now requires an analysis project to be specified instead of a lab.  This analysis project will be linked to any software results that are unarchived.  Additionaly, unarchiving a build now
requires that its model belong to an analysis project.

Arbitrary unarchives will be supported via the new export command in #1090.

### DIFF
--- a/lib/perl/Genome/Disk/Command/Allocation/Unarchive.t
+++ b/lib/perl/Genome/Disk/Command/Allocation/Unarchive.t
@@ -12,6 +12,10 @@ use above "Genome";
 use Test::More; #skip_all => 'unarchiving not fully implemented yet';
 use File::Temp 'tempdir';
 use Filesys::Df qw();
+use Test::Exception;
+
+use Genome::Test::Factory::Model::SingleSampleGenotype;
+use Genome::Test::Factory::Build;
 
 use_ok('Genome::Disk::Allocation') or die;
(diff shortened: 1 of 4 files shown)


lib/perl/Genome/Disk/Command/Allocation/Unarchive.t
```diff
@@ -124,9 +133,18 @@ unlink join('/', $allocation->absolute_path, 'a.out');
...
+$allocation->owner_id($build->id);
+$allocation->owner_class_name($build->class);
+$cmd = Genome::Disk::Command::Allocation::Unarchive->create(
+    allocations => [$allocation],
+    analysis_project => $analysis_project,
+);
+
 ok($cmd->execute, 'successfully executed unarchive command');
 is($allocation->volume->id, $volume->id, 'allocation moved to active volume');
 ok($allocation->is_archived == 0, 'allocation is not archived');
```
ebelter: Should we test that the AnP association was created?

tmooney: Sure, it'd be a nice thing to test!  I'll have to make a testcase with an SR as the owner, since we're not making extra associations for models.

ebelter:
+1

acoffman:
+1

PR #1094, opened 2015-11-05
Use config to get lsf_resource for chimerascan

dufeiyu:
So users do not need run builds with changed lsf limit from custom snapshot or checkout.


dufeiyu:
@ebelter  Please review

ebelter:
+1 Maybe we should add rnaseq to the config name to set a namespace? Like: lsf_resource_rnaseq_chimerascan.

dufeiyu:
Good point, I will make the change.

davidlmorton:
+1 If I were being nit-picky I'd say you should just have combined the first two commits together and the last three commits together.  There is no utility to having but not using the configuration, nor is there any benefit to separate commits to fix the name of the config-spec and env-variable.  I like the idea though, good way to reduce the number of custom checkouts!

PR #1095, opened 2015-11-05
Added 'genome test run-model-test' command

davidlmorton:
This command runs a model test but does not:
  - alter your environment in (i.e. setting a software result test name)
  - determine what version of perl to run
  - email anybody


davidlmorton:
Merging early since I want to start using this.  I cannot imagine bad things happening because it exists.

PR #1096, opened 2015-11-05
Readcount annotation fix

ernfrid:
Indel alleles are not being comprehensively handled in variant reporting currently. This is a work in progress to try to remedy that. I am still running some live tests on the new code to verify we've addressed this everywhere, but would greatly appreciate some review.


lib/perl/Genome/File/Vcf/BamReadcountUtilities.t
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/env perl
```
tmooney: We tend to use `#!/usr/bin/env genome-perl` in our tests.

ernfrid:
Fixed. Thanks Tom.

ernfrid:
I just added a bit of additional code to address some bam-readcount bugs. I will need to re-run my out-of-band testing, but I think this is ready for any additional review/feedback.

ernfrid:
The VAF code is not going to work due to upstream complexity in how bam-readcount is run. I've reverted.

ernfrid:
I think this is ready for serious review now.

mkiwala:
+1 :koala:

apregier:
+1

PR #1098, opened 2015-11-06
Attempt to fix Trio Reporting IGV labels

ernfrid:
This attempts to prevent overwriting of the IGV session labels to mistakenly indicate that the Followup Sample is the Discovery and vice versa.


apregier:
+1

susannasiebert:
+1

PR #1099, opened 2015-11-06
Make 'instrument-data ignore' accept a list

gatoravi:
instrument-data ignore currently only accepts one ID as a parameter, this PR attempts to make it so that it can accept multiple ID's. This will be useful when ignoring a large number of Instrument-data.


ebelter:
The command to update ignored is auto generated and exists as

genome instrument-data update ignored --h

This command accepts multiple instdata. We should close this PR. I will create a new PR to remove this extra ignored command.

ebelter:
This command is removed in https://github.com/genome/genome/pull/1101. Please advise if the auto generated update ignored command does not act as expected.

PR #1100, opened 2015-11-06
InstData Import: Avoid shell errors associated with odd charcters

ebelter:
use arrayref shellcmd sanitize sample name before using it in  afile name
problems

### DIFF
--- a/lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/FastqsToBam.pm
+++ b/lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/FastqsToBam.pm
@@ -8,6 +8,7 @@ use Genome;
 use Data::Dumper 'Dumper';
 require File::Basename;
 require File::Spec;
+require Genome::Utility::Text;
 require List::Util;
 use Try::Tiny;
 
--- a/lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/FastqsToBam.pm
+++ b/lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/FastqsToBam.pm
@@ -7,6 +7,7 @@ use Genome;
 
 use Data::Dumper 'Dumper';
 require File::Basename;
+require File::Spec;
 require List::Util;
 use Try::Tiny;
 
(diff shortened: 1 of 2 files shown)


lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/SortBam.pm
```diff
@@ -64,8 +64,8 @@ sub _sort_bam {
     my $output_bam_path = $self->output_bam_path;
     $self->debug_message("Sorted bam path: $output_bam_path");
 
-    my $cmd = "samtools sort -m 3000000000 -n $bam_path $sorted_bam_prefix";
-    my $rv = eval{ Genome::Sys->shellcmd(cmd => $cmd); };
+    my @cmd = (qw/ samtools  sort -m 3000000000 -n /, $bam_path, $sorted_bam_prefix );
```
tmooney: Is the second space between `samtools` and `sort` intentional?

lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/FastqsToBam.pm
```diff
@@ -35,7 +36,7 @@ class Genome::InstrumentData::Command::Import::WorkFlow::FastqsToBam {
         output_bam_path => {
             is => 'Text',
             calculate_from => [qw/ working_directory sample_name /],
-            calculate => q( return $working_directory.'/'.$sample_name.'.bam'; ),
+            calculate => q| return File::Spec->join($working_directory, Genome::Utility::Text::sanitize_string_for_filesystem($sample_name).'.bam'); |,
```
tmooney: This module should `use Genome::Utility::Text` in order to prevent a `Use of inherited AUTOLOAD for non-method ... is deprecated` warning (or, moreover, to not potentially rely on a deprecated feature :smile:).

ebelter: Yes, auto loading has made me lazy!

tmooney:
+1 :zap:

brummett:
+1

PR #1102, opened 2015-11-09
Use primary queue to submit SomVal AlignReads job.

tmooney:
Some jobs scheduled by the dispatcher (e.g. the new QC framework job) will end up in the "alt" queue, so this job needs to be in the lower-priority queue.

(In our current configuration, the "alt" queue is `apipe` and the other is `long` (with corresponding `pd` queues for builder).)


dufeiyu:
+1

ebelter:
+1

PR #1103, opened 2015-11-09
Sync Users: use status_message

ebelter:
Use status_message instead of custom STDOUT redirect status to allow printing
of messages during cron runs.


brummett:
+1

tmooney:
+1 :surfer:

PR #1104, opened 2015-11-10
Shortcut Several Skipped SomVal Steps

tmooney:
I noticed several builds that were done save for PENDing jobs for things that weren't going to run.  This allows these two steps to shortcut so we don't wait for slots just to find there's nothing to do.


brummett:
+1

ebelter:
+1

PR #1108, opened 2015-11-12
Convert Ptero/IntegrationTest.pl into a Command

davidlmorton:
This makes it so that it can easily be run like so:

$ genome test run-ptero-integration-tests --log-directory ...

It can even run without a local checkout of genome since it will use the
system-wide installation by default.


lib/perl/Genome/Test/Command/RunPteroIntegrationTests.pm
```diff
@@ -0,0 +1,102 @@
...
+sub get_test_outputs {
+    my $name = shift;
+    my $file = File::Spec->join(test_data_directory($name), 'outputs.json');
+    die "Cannot locate test outputs for workflow_test: $name" unless -e $file;
+    return from_json(Genome::Sys->read_file($file));
+}
+
+sub test_data_directory {
+    my $name = shift;
+    my $genome_dir = dirname(dirname(dirname(__FILE__)));
```
tmooney: This could use `Genome->base_dir`.

ebelter:
+1

lib/perl/Genome/Test/Command/RunPteroIntegrationTests.pm
mkiwala: I agree with Tom's comment here.

PR #1110, opened 2015-11-12
Fix list evaluation in scalar context.

tmooney:
This is an `is_many` property, so be ready for multiple values!


jasonwalker80:
+1

brummett:
Is the consumer of this data also prepared for a listref instead of a simple scalar?

tmooney:
@brummett Yes.  This was broken in #1089--the old version always received an array reference directly.  When `intervals` was changed from a subroutine to an `is_many` property, it began returning a list instead of the reference.

brummett:
+1

PR #1112, opened 2015-11-13
InstData Import Inputs

ebelter:
This PR fixes issues in serialization of instdata workflow inputs. Inputs has a 'get' method in order to be retrieved from  the process import file. The inputs were reorganized into a separate directory where creation is delegated to a factory. The factory contains the logic of the csv parser, and returns an input object instead of a hash.

It also addresses a couple of issues running in flow vs legacy workflow that casused the imports to fail.


lib/perl/Genome/InstrumentData/Command/Import/Basic.pm
```diff
@@ -87,12 +87,15 @@ sub execute {
...
     my $self = shift;
 
-    return Genome::InstrumentData::Command::Import::WorkFlow::Inputs->create(
-        analysis_project => $self->analysis_project,
-        library => $self->library,
-        source_files => [ $self->source_files ],
-        instrument_data_properties => $self->_resolve_instrument_data_properties,
-    );
+    my $factory = Genome::InstrumentData::Command::Import::Inputs::Factory->get;
+    $factory->analysis_project($self->analysis_project);
```
tmooney: Relying on the code to update the global `analysis_project` property of this Singleton appropriately seems fragile.  If it's only used in `from_params` perhaps it should be passed in as a parameter?

tmooney: I think I'm not understanding the idea behind switching this Factory to a Singleton in general.

lib/perl/Genome/InstrumentData/Command/Import/Inputs.pm
```diff
@@ -0,0 +1,110 @@
...
+    my $self = shift;
+    return substr(
+        Genome::Sys->md5sum_data( join(' ', $self->library_name, $self->source_paths) ), 
+        0, 6,
+    );
+}
+
+sub analysis_project {
+    my $self = shift;
+    if ( $self->process ) {
```
tmooney: What would it mean for there to not be a process if `process_id` is part of the `id` of this object?

lib/perl/Genome/InstrumentData/Command/Import/Inputs.pm
```diff
@@ -0,0 +1,110 @@
...
+        %params = ( name => $self->entity_params->{library}->{name} );
+    }
+    else {
+        $self->fatal_message('No library id or name given to inputs!');
+    }
+    return Genome::Library->get(%params);
+}
+
+sub process {
+    return Genome::InstrumentData::Command::Import::Process->get(id => $_[0]->process_id);
```
tmooney: This could be turned into an `id_by => 'process_id'` property in the class definition.

lib/perl/Genome/InstrumentData/Command/Import/Inputs.pm
ebelter: @tmooney  Currently, there are 2 entry points to import instdata. Manual (or single), by invoking the 'basic' command. and one a little more automated (multiple) through the 'launch' command. The basic command does not create a process. And a 'fake' process id needs to be used. I hope to eliminate the basic importer to have just one import command. As a tie over, a process could be created in the basic command. Since I hope to eliminate it, I did not add that bit of code. Do you have an opinion?

lib/perl/Genome/InstrumentData/Command/Import/Inputs.pm
ebelter: @tmooney That is the eventual plan. See comment above.

lib/perl/Genome/InstrumentData/Command/Import/Basic.pm
ebelter: @tmooney I'm SURE I had a reason :smile: Let me review it and see if it is necessary.

lib/perl/Genome/InstrumentData/Command/Import/Inputs.pm
tmooney: Cool; I don't think we need to do a lot of extra work if `Basic` is going away soon.  It seems like that'll simplify a good bit of this.

ebelter:
@tmooney  I reverted the inputs factory to a UR::Object. Take a look if you'd like.

tmooney:
+1, I look forward to the later simplifications to this!

PR #1113, opened 2015-11-13
Dont fail ptero builds

davidlmorton:
Because PTero run builds don't have an LSF job related to the workflow, the build scan thinks they're bad.


ebelter:
+1 :ring:

mkiwala:
+1

PR #1115, opened 2015-11-16
Updates to HaplotypeCaller file naming in SSG

tmooney:
* Include the name of the contig (if there's only one).
* Use `.gz` extension so the file is compressed.
* Add an accessor so the VCF files can be listed more easily.


mkiwala:
+1

acoffman:
+1

PR #1117, opened 2015-11-16
Several improvements for TCGA submission

dufeiyu:
Refer to https://jira.gsc.wustl.edu/browse/APIPE-2742 and https://jira.gsc.wustl.edu/browse/APIPE-2747. Alternative ways are added to get TCGA names and patient ids from sample attributes, and get sample is_tumor or not from sample type code of TCGA names.


dufeiyu:
@apregier  can you review this ? thanks

apregier:
Seems ok to me +1

lib/perl/Genome/Sample.pm
```diff
@@ -267,11 +268,60 @@ sub sample_name_to_name_in_vcf {
...
     my $self = shift;
     my $sample_tcga_name = $self->extraction_label;
-    if ($sample_tcga_name and $sample_tcga_name =~ /^TCGA\-/) {
-        $self->debug_message("Found TCGA name: $sample_tcga_name for sample: %s", $self->name);
-        return $sample_tcga_name;
+
+    unless ($sample_tcga_name and $sample_tcga_name =~ /^TCGA\-/) {
+        $sample_tcga_name = $self->name;
+        my @tcga_names = $self->get_tcga_names;
+        if (@tcga_names and @tcga_names == 1) {
```
tmooney: This is logically equivalent to `if (@tcga_names == 1)`.

Should we be concerned if there is more than one?

dufeiyu: There are cases that TCGA sample has multiple TCGA barcodes (from different sources) stored as external_name in sample attributes. It is ok.  If code reaches this step (no TCGA names as extraction_label and  @tcga_names != 1), the only way to find correct TCGA names is to parse TCGA submission data inventory .dat files. See comments in Genome::Info::TCGASpecialSampleNamingConversion.

tmooney:
+1, though it seems like the mapping should also be stored as an attribute on the samples instead of in code.

dufeiyu:
@tmooney  I agree. Currently there is no good way on LIMS to find that unique TCGA names for some LAML samples (caTissue as nomenclature, they are TCGA samples but can't be stored with TCGA- names) from Ley lab. For now, I place the name mapping under Genome so that at least there is a proper place to hold this name mapping information and prevent parsing those badly structured .dat files over and over. We can always revisit this space once the better solution for this is in place on LIMS (if still needed).

PR #1118, opened 2015-11-17
Ptero submit after process committed

mkiwala:
These changes are necessary for starting builds using PTero as the workflow backend.


davidlmorton:
+1 This was the reason PTero builds were not being committed during 'genome model build start'

PR #1119, opened 2015-11-17
Add extra parameters when gzipped GVCF requested in version 3.4 of HaplotypeCaller.

tmooney:
We switched to requesting gzipped GVCF in #1115.  However, GATK 3.4 contains a bug that requires two additional (deprecated) parameters to be supplied when a gzipped file is requested.  (This is fixed in 3.4-46, but that version is not tagged in their public repository.)

### DIFF
--- a/lib/perl/Genome/Model/Tools/Gatk/HaplotypeCaller.pm
+++ b/lib/perl/Genome/Model/Tools/Gatk/HaplotypeCaller.pm
@@ -95,4 +95,19 @@ sub _shellcmd_extra_params {
     );
 }
 
+sub _cmdline_args {
+    my $self = shift;
+
+    my @args = $self->SUPER::_cmdline_args(@_);
+
+    if ($self->version eq '3.4' and
+            $self->output_vcf =~ /g\.vcf\.gz$/ and
+            $self->emit_reference_confidence eq 'GVCF') {
+        #Handle a bug in GATK 3.4 by including these deprecated parameters.
+        push @args, qw(-variant_index_type LINEAR -variant_index_parameter 128000);
+    }
+
+    return @args;
+}
+
 1;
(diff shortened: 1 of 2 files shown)


lib/perl/Genome/Model/Tools/Gatk/HaplotypeCaller.pm
```diff
@@ -95,4 +95,19 @@ sub _shellcmd_extra_params {
...
+sub _cmdline_args {
+    my $self = shift;
+
+    my @args = $self->SUPER::_cmdline_args(@_);
+
+    if ($self->version eq '3.4' and
+            $self->output_vcf =~ /g\.vcf\.gz$/ and
+            $self->emit_reference_confidence eq 'GVCF') {
+        #Handle a bug in GATK 3.4 by including these deprecated parameters.
+        push @args, qw(-variant_index_type LINEAR -variant_index_parameter 128000);
```
acoffman: Is there ever any need to vary these values? Or is it fine for them to be hardcoded?

tmooney: They don't need to vary.  Here's the error message if they're omitted; it specifies these exact parameters and values:
```
##### ERROR ------------------------------------------------------------------------------------------
##### ERROR A USER ERROR has occurred (version debian/3.4-2-0-g9765327): 
##### ERROR
##### ERROR This means that one or more arguments or inputs in your command are incorrect.
##### ERROR The error message below tells you what is the problem.
##### ERROR
##### ERROR If the problem is an invalid argument, please check the online documentation guide
##### ERROR (or rerun your command with --help) to view allowable command-line arguments for this tool.
##### ERROR
##### ERROR Visit our website and forum for extensive documentation and answers to 
##### ERROR commonly asked questions http://www.broadinstitute.org/gatk
##### ERROR
##### ERROR Please do NOT post this error to the GATK forum unless you have really tried to fix it yourself.
##### ERROR
##### ERROR MESSAGE: GVCF output requires a specific indexing strategy.  Please re-run including the arguments -variant_index_type LINEAR -variant_index_parameter 128000.
##### ERROR ------------------------------------------------------------------------------------------
```

jeldred:
+ π

acoffman:
+1

lib/perl/Genome/Model/Tools/Gatk/HaplotypeCaller.pm
acoffman: Good enough for me!

PR #1120, opened 2015-11-18
New classes to store custom files as a software result

susannasiebert:
This PR is in preparation for allowing analyst to produce custom, project-specific VCFs to be used with VerifyBamID (https://jira.gsc.wustl.edu/browse/APIPE-3318).


lib/perl/Genome/SoftwareResult/ImportedFile.pm
```diff
@@ -0,0 +1,26 @@
...
+
+use strict;
+use warnings;
+use Genome;
+
+class Genome::SoftwareResult::ImportedFile {
+    is => 'Genome::SoftwareResult::StageableSimple::SingleFile',
+    has_input => [
+        source_file_path => {
+            is => 'Path',
```
tmooney: When we "import" files into the system it's better to use an MD5 of the file to uniquely identify it rather than the path from which it came.

`Genome::Model::Tools::DetectVariants2::Result::Manual` is one example of doing this.  The InstrumentData import process is another.

lib/perl/Genome/SoftwareResult/ImportedFile.pm
```diff
@@ -0,0 +1,26 @@
...
+        }
+    ],
+};
+
+sub _run{
+    my $self = shift;
+    Genome::Sys->copy_file($self->source_file_path, $self->_temp_staging_file_path);
```
tmooney: Speaking of `Genome::Model::Tools::DetectVariants2::Result::Manual`, it uses the `_needs_symlinks_followed_when_syncing` feature of `SoftwareResult::Stageable` to avoid making two copies of the file.  That might be useful here, too.

lib/perl/Genome/SoftwareResult/StageableSimple/SingleFile.pm
```diff
@@ -0,0 +1,27 @@
+package Genome::SoftwareResult::StageableSimple::SingleFile;
+
+use strict;
+use warnings;
+use Genome;
+use File::Basename;
+
+class Genome::SoftwareResult::StageableSimple::SingleFile {
+    is => 'Genome::SoftwareResult::StageableSimple',
+};
```
tmooney: This should probably be `is_abstract => 1`.

susannasiebert:
I pushed a few commit to address the comments as well as add a new command to import a genotype vcf file for QC.

susannasiebert:
@tmooney @ebelter @acoffman Is there anything else you would like to see done with this PR before giving your seal of approval?

acoffman:
+1 :cocktail:

lib/perl/Genome/SoftwareResult/StageableSimple/SingleFile.pm
```diff
@@ -0,0 +1,28 @@
...
+
+sub _temp_staging_file_path {
+    my $self = shift;
+    return File::Spec->join($self->temp_staging_directory, $self->_file_name);
+}
+
+sub _file_name {
+    my $self = shift;
+    $self->fatal_message("Abstract method (_file_name). Needs to be overwritten in your child class");
```
tmooney: `Carp::confess` is handy for cases of programmer error, so it's easier to track down where this is getting called on an incorrectly implemented class.

tmooney:
+1 :airplane:

PR #1122, opened 2015-11-18
apipe-builder user and the apipe queue

brummett:
This should prevent apipe-builder from submitting jobs to the apipe queue.  Then we'll be able to track them down and move the jobs to a different queue.


brummett:
I'm closing this branch in favor of another that just notifies me when it happens and doesn't kill the build

PR #1123, opened 2015-11-19
Remove clean_bam from workflow inputs

davidlmorton:
It is not used in the workflow and WorkflowBuilder, when launching a
PTero workflow disallows unused inputs.


mkiwala:
+1

tmooney:
+2 :goberserk:

PR #1124, opened 2015-11-19
Notify me when apipe-builder uses the apipe queue

brummett:
Then I can track down the sources and move them to different queues


ebelter:
+1 Title says tester, but code says builder. I presumed you meant buiider.

brummett:
Yes, builder.  I've fixed the title.

jeldred:
+1

PR #1125, opened 2015-11-19
Let UR re-call with the correct subclass if StageableSimple called abstractly.

tmooney:
I don't think this fails in practice, because it is far more common to call on either the specific subclass or on `Genome::SoftwareResult` itself.  Just in case, though :smile:


brummett:
You're right.  I was looking at the wrong file.

+1

acoffman:
+1

PR #1126, opened 2015-11-19
Set a default resource request for the BucketIterator.

tmooney:
Thus far the jobs have fit within the standard 4GB memory reservation.  The ones I looked at showed `/tmp` usage around 100-150MB; the reservation of 1 GB should ensure they don't land on a completely full blade.

This will then also be ready if some cases have stronger requirements.


brummett:
+1

dufeiyu:
+1  though I like lsf_resource_ssg_gatk_haplotype_caller a little bit more.

PR #1127, opened 2015-11-20
Don't try to format body_text when displaying notes.

davidlmorton:
The clean_up_timestamp was throwing away everything after the first '.'
which isn't desired here.


susannasiebert:
+1

acoffman:
+1

PR #1128, opened 2015-11-23
Add more TCGA name mapping

dufeiyu:
Refer to https://jira.gsc.wustl.edu/browse/APIPE-2742. More TCGA name mapping (for LAML samples) are added. This needs early merge.


jasonwalker80:
+2

PR #1129, opened 2015-11-24
Set queue and projectName when getting ptero lsf params

mkiwala:
This commit makes getting lsf parameters for use with PTero work more like other cases by using the existing `operation_type_attributes()` method to decide whether the lsf resource string should come from the Genome command class or from `Genome::WorkflowBuilder::Command`.  It also gets the queue name and project name from the `operation_type_attributes()`.

Since the queue and project name could be specified in the resource string, the workflow builder command, and the genome command class, we should consider the order of precedence.  Clearly, the queue and project specified on the workflow builder command takes precedence over a value supplied on the Genome command class.  It is not clear to me whether a queue or project specified in the lsf resource string should take precedence over the value specified on the workflow builder command.

### DIFF
--- a/lib/perl/Genome/WorkflowBuilder/Command.pm
+++ b/lib/perl/Genome/WorkflowBuilder/Command.pm
@@ -106,32 +106,22 @@ sub _get_ptero_execute_method {
         parameters => $ptero_lsf_parameters);
 }
 
-sub _get_lsf_resources_from_command {
-    my $self = shift;
-    my $prop = $self->command->__meta__->property_meta_for_name('lsf_resource');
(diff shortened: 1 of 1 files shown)


mkiwala:
@davidlmorton points to [this](https://github.com/genome/tgi-workflow/blob/master/lib/Workflow/Server/Hub.pm#L515-L516) code which suggests that the queue specified in the lsf_resource string should take precedence over the queue specified in the lsf_queue option.

Given this new information, I'll update my PR.

tmooney:
+1 :guardsman:

PR #1130, opened 2015-11-24
Make QC genotype VCF file an input to the SSG pipeline and use in verifyBamId

susannasiebert:


lib/perl/Genome/Model/SingleSampleGenotype/Command/QualityControl.pm
```diff
@@ -35,6 +35,7 @@ sub _params_for_command {
     my @params = (
         alignment_result => $build->merged_alignment_result,
         config_name => $build->model->qc_config,
+        qc_genotype_vcf_file => $build->model->qc_genotype_vcf_file,
```
tmooney: Inputs should be obtained from the `$build` directly rather than going back to the model.

susannasiebert: Fixed.

acoffman:
Is there any backfilling of results that we'll need to do for things that were previously run?

tmooney:
+1 :dragon:

susannasiebert:
@acoffman I don't think so. `qc_genotype_vcf_file` is optional so it's ok if it wasn't set previously. And the way to use it, is to put `vcf: genotype_vcf_file` in your config which will call the `genotype_vcf_file` subroutine that is in `Genome::Qc::Tool`. Since the hard-coded config doesn't use that right now it shouldn't be a problem. Once we can get #846 merged, we can make configs that use that subroutine.

acoffman:
Cool, sounds good!

+1

PR #1131, opened 2015-11-25
PTero LSF Stuff

davidlmorton:
This PR enables you to configure the polling interval that the PTero lsf service uses to track jobs.  Additionally, this PR ensures that lsf jobs associated with a build run with PTero all have a projectName set.

### DIFF
--- a/etc/genome/spec/ptero_lsf_polling_interval.yaml
+++ b/etc/genome/spec/ptero_lsf_polling_interval.yaml
@@ -1,3 +1,5 @@
 ---
 env: XGENOME_PTERO_LSF_POLLING_INTERVAL
 default_value: 300
+validators:
+  - Numeric
--- a/etc/genome/spec/ptero_lsf_polling_interval.yaml
+++ b/etc/genome/spec/ptero_lsf_polling_interval.yaml
@@ -1,0 +1,3 @@
+---
+env: XGENOME_PTERO_LSF_POLLING_INTERVAL
+default_value: 300
--- a/lib/perl/Genome/Model/Build.pm
+++ b/lib/perl/Genome/Model/Build.pm
@@ -1396,9 +1396,9 @@ sub _launch {
     local $ENV{UR_COMMAND_DUMP_STATUS_MESSAGES} = 1;
 
     # we don't use 'local' because we want this to persist through
-    # spawned workflows
-    $ENV{Genome::Config::spec('lsf_project_name')->env} =
-        sprintf("build/%s", $self->id);
+    # spawned workflows, and since the build is launched in a commit
+    # observer the 'local' designations ensure the ENV isn't still around then
+    Genome::Config::set_env('lsf_project_name', sprintf("build/%s", $self->id));
 
     my $build_id_guard = set_build_id($self->id);
 
--- a/lib/perl/Genome/Model/Build.pm
+++ b/lib/perl/Genome/Model/Build.pm
@@ -1395,6 +1395,11 @@ sub _launch {
     local $ENV{UR_DUMP_STATUS_MESSAGES} = 1;
     local $ENV{UR_COMMAND_DUMP_STATUS_MESSAGES} = 1;
 
+    # we don't use 'local' because we want this to persist through
+    # spawned workflows
+    $ENV{Genome::Config::spec('lsf_project_name')->env} =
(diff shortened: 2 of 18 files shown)


lib/perl/Genome/Model/Build.pm
```diff
@@ -1395,6 +1395,11 @@ sub _launch {
     local $ENV{UR_DUMP_STATUS_MESSAGES} = 1;
     local $ENV{UR_COMMAND_DUMP_STATUS_MESSAGES} = 1;
 
+    # we don't use 'local' because we want this to persist through
+    # spawned workflows
+    $ENV{Genome::Config::spec('lsf_project_name')->env} =
+        sprintf("build/%s", $self->id);
```
tmooney: This looks like what `Genome::Config::set_env` is designed for.

etc/genome/spec/ptero_lsf_polling_interval.yaml
```diff
@@ -0,0 +1,3 @@
+---
+env: XGENOME_PTERO_LSF_POLLING_INTERVAL
+default_value: 300
```
tmooney: It'd be good to add validation to this spec:
```yaml
validators:
  - Numeric
```

lib/perl/Genome/Model/Build.pm
tmooney: I don't understand the comment about not using `local`.  How will spawned workflows that jobs for this build produces know whether or not the environment variable was reset when this `_launch` completes?

It seems like this may cause the ID of nested builds to improperly be associated with jobs spawned by the parent build.

lib/perl/Genome/WorkflowBuilder/Command.pm
```diff
@@ -99,6 +99,13 @@ sub _get_ptero_execute_method {
     $ptero_lsf_parameters->{environment} = $self->_get_sanitized_env();
     $ptero_lsf_parameters->{user} = Genome::Sys->username;
     $ptero_lsf_parameters->{cwd} = Cwd::getcwd;
+    $ptero_lsf_parameters->{pollingInterval} =
+        Genome::Config::get('ptero_lsf_polling_interval') + 0;
+
+    my $project_name = Genome::Config::get('lsf_project_name');
+    if ($project_name) {
+        $ptero_lsf_parameters->{options}{projectName} = $project_name;
```
tmooney: Properly handling the project name seems like it depends on #1129, which gets the project name from the Workflow XML.  Deciding between that XML and this environment variable might then go inside `_get_ptero_lsf_parameters`.

lib/perl/Genome/Model/Build.pm
davidlmorton: I've tested this out, and because of the scope-guard that `Genome::Config::set_env` the value does not make it into the environment of the workflow steps.  The same is true of the `local` modifier.  As for nested builds (something that we should eliminate if they exist imo) if they are started by calling `Genome::Model::Build->start()` or something that invokes that then they will behave as expected and have the `projectName` associated with their build and not their parent's build.

lib/perl/Genome/WorkflowBuilder/Command.pm
davidlmorton: Michael and I have discussed this briefly, and it appears that almost nobody is using the lsf_project property and maybe it's best to just deprecate it.

lib/perl/Genome/Model/Build.pm
tmooney: The `Scope::Guard` is only created if you ask for a return value from `set_env`, so I believe it should be used regardless.

My previous comment was referring to jobs that do belong to the parent build becoming incorrectly associated with the child thanks to this not being localized.  I agree nested builds are not great--unfortunately they're in `SomaticValidation` of all places...

I would guess the problem with localization is because the workflow is finally submitted in a commit observer.  The project variable needs to be set at that time to work properly.

mkiwala: > I would guess the problem with localization is because the workflow is finally submitted in a commit observer. The project variable needs to be set at that time to work properly.

That makes perfect sense.  Thanks for pointing that out.

mkiwala:
+1 :koala:

tmooney:
+1 :godmode:

PR #1133, opened 2015-11-25
Add optional property to VerifyBamId QC tool to set default genotype vcf file

susannasiebert:
This file would then be used if no qc_genotype_vcf_file is specified in the build

### DIFF
--- a/lib/perl/Genome/Qc/Tool/VerifyBamId.pm
+++ b/lib/perl/Genome/Qc/Tool/VerifyBamId.pm
@@ -7,6 +7,16 @@ use List::MoreUtils qw(any);
 
 class Genome::Qc::Tool::VerifyBamId {
     is => 'Genome::Qc::Tool::WithVariationListVcf',
+    has => [
+        default_genotype_vcf_file_id => {
+            is => 'Text',
+            is_optional => 1,
+        },
+        default_genotype_vcf_file => {
+            is => 'Genome::SoftwareResult::ImportedFile',
+            id_by => 'default_genotype_vcf_file_id',
+        },
+    ],
 };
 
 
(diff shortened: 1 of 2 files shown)


lib/perl/Genome/Qc/Tool/VerifyBamId.pm
```diff
@@ -7,6 +7,16 @@ use List::MoreUtils qw(any);
 
 class Genome::Qc::Tool::VerifyBamId {
     is => 'Genome::Qc::Tool::WithVariationListVcf',
+    has => [
+        default_genotype_vcf_file_id => {
+            is => 'Text',
+            is_optional => 1,
+        },
+        default_genotype_vcf_file => {
+            is => 'Genome::SoftwareResult::ImportedFile',
```
tmooney: Would it makes sense for this to be a `Genome::SoftwareResult::StageableSimple::SingleFile` so it could be either an imported file or one of our Genotype VCFs?  It doesn't seem likely as a use case, but should we rule it out?

susannasiebert: I could see a case where they might want to do that. I changed it to be a `SingleFile`

lib/perl/Genome/Qc/Tool/VerifyBamId.t
```diff
@@ -94,6 +98,13 @@ my @expected_cmd_line = (
 );
 is_deeply([$tool->cmd_line], [@expected_cmd_line], 'Command line list as expected');
 
+my $command_without_qc_genotype_vcf_file = Genome::Qc::Run->create(
+    config_name => 'testing-qc-run',
+    alignment_result => $alignment_result,
+    %{Genome::Test::Factory::SoftwareResult::User->setup_user_hash},
+);
+ok($command_without_qc_genotype_vcf_file->execute, "Command with qc_genotype_vcf_file executes ok");
```
tmooney: The variable says `without_qc` but the test description says `with qc`.

tmooney:
+1 when the with/without is reconciled!

acoffman:
+1 :snowman:

PR #1134, opened 2015-11-25
Increase memory limit of lsf_resource_dv2_dispatcher

dufeiyu:
Recently there are quite a few of som-var builds failed at Detect_Variants because memory usages are over 600 MB limit.


dufeiyu:
@tmooney  Can you check this ?

tmooney:
+1 seems fine to me.  @jasonwalker80 had originally requested this be lowered from the default 4GB so he might be interested in reviewing.

jasonwalker80:
+1 - no need to risk a failure over 400MB.  4GB was entirely too much though.  If this requires more than 1GB, we may want to determine the cause.  Perhaps it's related to a specific type of data or possibly bad data we don't really want to process.

dufeiyu:
@tmooney  @jasonwalker80  Recently I get failed builds including a few of TCGA ones because of this. It seems memory used in those failed builds range from 620 to 950 MB. So 1 GB should be ok for now.

PR #1138, opened 2015-12-01
updating path to bam-window cmd in help

chrisamiller:


tmooney:
+1, I see that the old command under `copy-number` was removed in #204.

acoffman:
+1

PR #1139, opened 2015-12-01
Add running build limit in "list-build-queue"

tmooney:
As we move to PTero, we can't rely on the size of the workflow LSF queues to limit our build starts.  

It is not ideal that this hard-codes a multiplier for apipe-builder; several things about the build process seem to have this "feature".


mkiwala:
+1

acoffman:
+1

PR #1143, opened 2015-12-02
RemoveDiskAllocationsFromTestDb remembers which TestDB was used.

tmooney:
`delete local` did not only remove the key from the hash locally. Thus, when the allocation shelled out to perform the deletion, it connected to the production database and couldn't find the allocation.

Splitting this to two lines seems to have the intended effect of removing the localized version.


brummett:
+1

acoffman:
+1

PR #1144, opened 2015-12-03
TCGA Priorities: Add alternative way to get cghub id

dufeiyu:
For some TCGA builds, current cghub id retrieve ways do not work. For example, mount path of merged bam could be changed after unarchiving and build subject extraction_label do not always return proper TCGA names. This PR adds a lookup hash by searching for unique bam base_name and uses name_in_vcf to replace extraction_label. This PR needs early merge.

### DIFF
--- a/lib/perl/Genome/Model/Tools/Tcga/Sdrf.pm
+++ b/lib/perl/Genome/Model/Tools/Tcga/Sdrf.pm
@@ -220,7 +220,7 @@ sub resolve_cghub_id {
         unless (defined $id) {
             $id = $id_by_bam_base->{basename $merged_bam};
             unless (defined $id) {
-                $self->fatal_message("CGHub id could not be resolved for build %s with bam file: %s", $build->id, $build->whole_rmdup_bam_file);
+                $self->fatal_message("CGHub id could not be resolved for build %s with bam file: %s", $build->id, $merged_bam);
             }
         }
     }
--- a/lib/perl/Genome/Model/Tools/Tcga/Sdrf.pm
+++ b/lib/perl/Genome/Model/Tools/Tcga/Sdrf.pm
@@ -216,9 +216,9 @@ sub resolve_cghub_id {
     my $merged_bam = $build->whole_rmdup_bam_file;
     my $id = $CGHUB_INFO->{$merged_bam};
     unless (defined $id) {
-        $id = $id_by_bam_base->{basename $merged_bam};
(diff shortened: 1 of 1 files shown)


dufeiyu:
@apregier  can you review this ? thanks.

lib/perl/Genome/Model/Tools/Tcga/Sdrf.pm
```diff
@@ -197,21 +198,30 @@ sub fill_in_common_fields {
...
-    my $build = shift;
+    my ($self, $build) = @_;
 
     unless (defined $CGHUB_INFO) {
         $CGHUB_INFO = $self->load_cghub_info("BAM_path");
     }
 
+    my $id_by_bam_base;
+    while (my($bam_path, $cghub_id) = each %$CGHUB_INFO) {
+        $id_by_bam_base->{basename $bam_path} = $cghub_id;
```
apregier: does this need to be `basename $bam_path`?

apregier: oh, never mind, I see the library use above

lib/perl/Genome/Model/Tools/Tcga/Sdrf.pm
```diff
@@ -197,21 +198,30 @@ sub fill_in_common_fields {
...
-    my $id = $CGHUB_INFO->{$build->whole_rmdup_bam_file};
+
+    my $merged_bam = $build->whole_rmdup_bam_file;
+    my $id = $CGHUB_INFO->{$merged_bam};
     unless (defined $id) {
-        $id = $CGHUB_INFO_BY_TCGA_NAME->{$build->subject->extraction_label};
+        $id = $CGHUB_INFO_BY_TCGA_NAME->{$build->subject->name_in_vcf};
         unless (defined $id) {
-            die("CGHub id could not be resolved for build ".$build->id." with bam file ".$build->whole_rmdup_bam_file);
+            $id = $id_by_bam_base->{basename $merged_bam};
```
apregier: Should this be the first thing tried?  Then we wouldn't have to try $id=$CGHUB_INFO->{$merged_bam} and would only have two 'unlesses' instead of three

apregier:
+1 seems ok to me, as long as we are sure that the bams will be unique by basename.

apregier:
You could add in a check that would die if it saw something that wasn't unique.

dufeiyu:
The merged bam base name should be unique enough because its name format is either <build_id>_merged_rmdup.bam or <merged_result_id>.bam. But just in case I put this way as the last solution in order.

PR #1146, opened 2015-12-04
TCGA Priorities: Put lookup hash by bam base name in cache

dufeiyu:
This is bugfix for https://github.com/genome/genome/pull/1144


dufeiyu:
@apregier  Can you review this ?  Thanks.

apregier:
Is this just for performance?

dufeiyu:
Yes.

PR #1147, opened 2015-12-04
Make name/value accessors consistent for differing metric types.

acoffman:
While working with @susannasiebert today, we noticed that `Genome::Model::Metric`s and `Genome::SoftwareResult::Metric`s have different interfaces. One responds to `metric_{name,value}` and the other `{name,value}` which is frustrating :frowning: Ironically, their column names are consistent with each other, but the properties are named differently in the class definitions.

This PR makes it so both types of metrics respond to both interfaces. There are other small inconsistencies that this does not address (one has read-only properties, the other doesn't for instance). Eventually it seems like we should have one `Metric` class with `SoftwareResult` and `Model` metrics varying only in table name.

Thoughts?

### DIFF
--- a/lib/perl/Genome/Model/Metric.pm
+++ b/lib/perl/Genome/Model/Metric.pm
@@ -26,6 +26,8 @@ class Genome::Model::Metric {
         model_id => { via => 'build' },
         model_name => { via => 'model', to => 'name' },
         data_directory => { via => 'build' },
+        metric_name => { via => '__self__', to => 'name' },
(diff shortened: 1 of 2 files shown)


ebelter:
+1 for @brummett's suggestion.

acoffman:
@ebelter @brummett Agreed, that's much better! I've updated the PR.

tmooney:
+1 :flags:

brummett:
+1

PR #1148, opened 2015-12-08
SomaticValidation SubjectMapping Exporter

tmooney:
The idea of this exporter is to reproduce the file that would have been created for the importer.

* The `SeparatedValueWriter` does not handle this case, since it wouldn't know what to do with the optional tag columns that must only be written if present.
* It might be possible to rewrite this in terms of an existing lister; the use of several hang-off tables complicates that approach.

This approach is very trusting and will still produce a file even if run on the wrong sort of analysis project.  The result in that case won't be very useful (not that it could be).


chrisamiller:
+1

brummett:
If you were interested in using a Lister, you could programatically whip up a class containing the relevant properties that only exists for the life of the program, with properties implemented as via/to props pointing to the real data.  Genome::Config::AnalysisProject::SubjectMapping::Command::Export::SomaticValidation would then be a Lister command with subject_class_name => that temp class.

But unless users are going to want to use the filter/show functionality it's probably overkill, and would certainly be less maintainable.

If this accomplishes what the users need, then +1

PR #1150, opened 2015-12-08
adds config copy/reactivate options

chrisamiller:


chrisamiller:
pushed the wrong thing..

PR #1151, opened 2015-12-08
new config options to copy/reactivate

chrisamiller:


tmooney:
I believe that disabling a config file is intended to be permanent.

To the extent that copying a disabled configuration makes sense, it should probably end up with a different status in the copy straight away, since a new disabled configuration is not useful.

acoffman:
If we do decide to go down the route of re-enabling a config item - the tool should likely schedule existing instrument data for reprocessing, otherwise the enabling won't do much (except for new instrument data coming in from LIMS).

chrisamiller:
The intent was to allow disabled config items to be copied en masse, then
to re-enable them in the new AnP (which may have different instrument
data). I suppose the copy command could set the copied items to active. Is
that definitely what we want here? If we copy config from one AnP to
another, should the state of the config items be retained, or should they
be set to active?

jasonwalker80:
Rather than copy the config from an old, disabled AnP to a new AnP, why not add the config file from the disabled config item on the old AnP?  As you've found copy does not retain tags or subject mappings.  Those all have to be created anew.  

Reactivating a config file should not be allowed.  It's possible that the data for the disabled config item has already been removed.  The policy, albeit not automatically enforced, is to remove software results uniquely associated with disabled config items after 30 days.

PR #1152, opened 2015-12-08
Don't import "compare" since we're defining our own.

tmooney:
We were already using `File::Compare::compare` explicitly in this module.


acoffman:
+1

ebelter:
+1 I prefer the direct 'require' methodology which doesn't have the implications of use, which runs require and import.

tmooney:
I've changed `qw()` to `()` as suggested by [the perldoc](http://perldoc.perl.org/functions/use.html).

@ebelter `use` also has the implication of `BEGIN`, which I think is proper in this circumstance.  This module doesn't work without `File::Compare`.

PR #1153, opened 2015-12-09
update blessed process to account for new N-caller and VAF fixes

ernfrid:
This should fix failing model tests.


tmooney:
+1, that looks like the ID of the run that just completed!

ebelter:
Merging to get tests fixed!

PR #1154, opened 2015-12-10
Remove Genome::Model::Comparison.

tmooney:
This model type was never fully implemented and hasn't been used since the tests in 2013.

This PR removes the model type completely. If accepted, merging it should be accompanied by the deletion of the builds, models, and processing-profiles in the database.  (The main effect of running a build of this type seems to be to launch builds for two underlying models, so those data products would not be affected by this removal.)


susannasiebert:
+1. Before we merge this, should we come up with a script/command to remove the builds, models, processing profiles (and I presume all their meta-data) from the database? This would be needed for other deprecation-related tasks as well.

tmooney:
@susannasiebert I don't believe we need to keep the builds, models, and processing profiles from this type, but it could be a real-world test case for that script/command!

dufeiyu:
+1  begin of the model deprecation

PR #1155, opened 2015-12-10
Remove trash folder support.  Delete purged data immediately.

tmooney:
In practice the trash volumes were a waste of space as data was never reclaimed therefrom.  They're being retired, so now we can remove data directly when purging.


ebelter:
+1 Does the disk group need to be marked for removal? I updated confluence docs.

jasonwalker80:
+1, the trash size has been reduced until this goes live.  Eventually it should be removed entirely.

PR #1157, opened 2015-12-11
TCGA Priorities: make sample get_tcga_names return all possible tcga names

dufeiyu:
For some rare cases, LIMS does not store any TCGA name as external_name in sample attributes, but does store as sample extraction_label instead.


dufeiyu:
CI, test this

apregier:
+1

PR #1158, opened 2015-12-11
Use display_name for printing AnalysisProject when Unarchiving

acoffman:
Currently, the command prints out this:

````
Unarchiving allocation(s) for Analysis Project Genome::Config::AnalysisProject=HASH(0x7c07088)
````

Now it would print out this:

````
Unarchiving allocation(s) for Analysis Project: My super sweet analysis project (id)
````


brummett:
+1

tmooney:
+1, this is clearly better than printing out the memory location!

The colon seems a bit strange to me, though, since the Analysis Project display name does not end the sentence in the `status_message`.

acoffman:
@tmooney I've removed the colon and also removed a now unneeded call to `sprintf`

tmooney:
+1 :incoming_envelope:

PR #1159, opened 2015-12-14
Qc metrics fix

apregier:
I think it is only valid to dump a ref to a file with DumpFile - at least I was having problems loading it in when I had more than one entry in the array.  Or, I might be misunderstanding something about yaml format.  Anyway, this seems to work much better for me.

### DIFF
--- a/lib/perl/Genome/Qc/Command/BuildMetrics.pm
+++ b/lib/perl/Genome/Qc/Command/BuildMetrics.pm
@@ -49,7 +49,7 @@ sub execute {
         die($self->error_message);
     }
 
-    DumpFile($self->output_file,@metrics);
+    DumpFile($self->output_file,\@metrics);
 
     return 1;
 }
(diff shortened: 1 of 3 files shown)


ebelter:
+1

tmooney:
+1.  I tried using `DumpFile` and `LoadFile` with an `@array`, and it seemed to work, but either way is fine as long as we're consistent!

jasonwalker80:
@wschroeder do you know if this would impact the parsing of LIMS QC metrics from APIPE?  Are the metrics dumped to a file at all by LIMS?  Is there an expectation for array or array ref?

tmooney:
I believe `LIMSGenome::Model` calls `metrics_for_build` programmatically rather than using the `YAML` output.

wschroeder:
That is correct.  We do not attempt to parse text or run by script; we leverage the API directly via the lims-genome interface bridge:

    my $build_obj = Genome::Model::Build->get($build_id);
    my $qc_cmd = Genome::Qc::Command::BuildMetrics->create(builds => [$build_obj]);
    my @qc_results = $qc_cmd->metrics_for_build($build_obj) if ($qc_cmd);

We assume the values we retrieve from the keys are numbers or strings, not compound values.  We are only looking at metrics that have been configured/approved, so if you have a metric whose value is an arrayref but is not configured for our analysis protocol, we will ignore it.

jasonwalker80:
+1, Thanks Tom!

apregier:
I'm going to close this - I was mixed up about what LoadFile would return.  When I expect an array, it works fine, as Tom indicates.  Sorry for the confusion!

PR #1161, opened 2015-12-15
Remove LSF queue overrides in Varscan "high confidence" filters.

tmooney:
These don't spawn workflows, so with this change they'll use the default queue for workers specified in [DV2/Filter.pm](https://github.com/genome/genome/blob/2d2f91c10ed2187e5d8346400b61b4edcafb2e69/lib/perl/Genome/Model/Tools/DetectVariants2/Filter.pm#L160).


acoffman:
+1

dufeiyu:
+1

PR #1162, opened 2015-12-16
AnP Reprocess

ebelter:
Allow the reprocessing of select instrument data for an AnP.

### DIFF
--- a/lib/perl/Genome/Config/AnalysisProject/Command/Reprocess.pm
+++ b/lib/perl/Genome/Config/AnalysisProject/Command/Reprocess.pm
@@ -56,8 +56,9 @@ sub _resolve_analysis_project_bridges {
 
     return Genome::Config::AnalysisProject::InstrumentDataBridge->get(
         analysis_project => $self->analysis_project,
-        instrument_data => [ $self->instrument_data ],
+        'instrument_data.id' => [ map { $_->id}  $self->instrument_data ],
     );
 }
 
-1
+1;
+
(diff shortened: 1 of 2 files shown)


ebelter:
The test worked, put in pratcie, the command didn't. Used instdata ids to get bridges.

lib/perl/Genome/Config/AnalysisProject/Command/Reprocess.pm
```diff
@@ -35,12 +42,23 @@ sub valid_statuses {
...
 }
 
+sub _resolve_analysis_project_bridges {
+    my $self = shift;
+
+    return $self->analysis_project->analysis_project_bridges if not $self->instrument_data;
+
+    return Genome::Config::AnalysisProject::InstrumentDataBridge->get(
+        analysis_project => $self->analysis_project,
+        'instrument_data.id' => [ map { $_->id}  $self->instrument_data ],
```
tmooney: `instrument_data_id` is a direct property of the `InstrumentDataBridge`, so we don't need to use the `.` syntax here.

ebelter: @tmooney But I WANT to use it :man_with_gua_pi_mao:

tmooney: :smile:, but it does an extra join!

Another option more parallel to the "all" case might be:
```perl
return $self->analysis_project->analysis_project_bridges(instrument_data_id => [map $_->id, $self->instrument_data]);
```

PR #1163, opened 2015-12-16
Asynchronous Shortcutting when using PTero

davidlmorton:
Instead of sleeping and consuming a shell-command worker while waiting on locks, this PR allows you to configure Genome to exit with a particular exit-code when a lock cannot be obtained.  When submitting the workflow, 'retrySettings' are specified to the shell-command service that specify that when the job exits with that particular exit-code the service should retry the job.

This will allow us to scale up without increasing the number of shell-command workers.

### DIFF
--- a/etc/genome/spec/software_result_async_locking_num_attempts.yaml
+++ b/etc/genome/spec/software_result_async_locking_num_attempts.yaml
@@ -1,0 +1,7 @@
+---
+env: XGENOME_SOFTWARE_RESULT_ASYNC_LOCKING_MAX_ATTEMPTS
+# with software_result_async_locking_max_interval set to 3600 seconds
+# 750 is approximately 1 month worth of attempts.
+default_value: 750
+validators:
+  - Numeric
--- a/etc/genome/spec/software_result_async_locking_num_attempts.yaml
+++ b/etc/genome/spec/software_result_async_locking_num_attempts.yaml
@@ -1,5 +1,5 @@
 ---
-env: XGENOME_SOFTWARE_RESULT_ASYNC_LOCKING_MAX_ATTEMPTS
+env: XGENOME_SOFTWARE_RESULT_ASYNC_LOCKING_NUM_ATTEMPTS
 # with software_result_async_locking_max_interval set to 3600 seconds
 # 750 is approximately 1 month worth of attempts.
 default_value: 750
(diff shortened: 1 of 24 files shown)


etc/genome/spec/sys_lock_exit_code.yaml
```diff
@@ -0,0 +1,3 @@
+---
+env: XGENOME_SYS_LOCK_EXIT_CODE
+default_value: ' '
```
tmooney: It might be nice to add:
```yaml
validator:
  - Numeric
```

It's lame that this has to have a default value to avoid crashing, and if we did validate it we'd be stuck with using something like 0 as a special value.

tmooney:
This seems fine for `shortcut`, but a terrible thing to have happen during `execute`!

etc/genome/spec/sys_lock_exit_code.yaml
brummett: But 0 is a valid exit code value.  You'd have to use -1 or 65536 (something greater than you can store in an int), or something

tmooney:
+1 :sparkles:

etc/genome/spec/software_result_async_locking_num_attempts.yaml
```diff
@@ -0,0 +1,7 @@
+---
+env: XGENOME_SOFTWARE_RESULT_ASYNC_LOCKING_MAX_ATTEMPTS
```
tmooney: This env name doesn't match the YAML file's name.

lib/perl/Genome/SoftwareResult.pm
```diff
@@ -820,16 +820,25 @@ sub _lock {
...
-            scope => 'site',
-        )->lock(
-            wait_announce_interval => 600,
-        );
-        unless ($lock) {
-            $class->error_message("Failed to get existing data lock!");
-            die($class->error_message);
+        if (Genome::Config::get("software_result_async_locking")) {
+            my $exit_code = Genome::Config::get(
+                "software_result_async_locking_exit_code");
```
brummett: This can be simplified to
```
if (my $exit_code = Genome::Config::get('software_result_async_locking')) {
```

The way this is structured, it means if you want it to actually exit early with a 0 exit code, you'd have to specify this value as "0 but true"

tmooney: The latest commit separated these into two variables, a boolean `software_result_async_locking` to control when it is used and `software_result_async_locking_exit_code` to hold the exit code.

mkiwala:
+1 :koala:

PR #1165, opened 2015-12-17
Remove "SimpleAlignment" model type.

tmooney:
This was a not-fully-adopted pipeline from 2010.  It was previously reduced to a stub; this PR removes the remaining stub classes.

(As with other issues in this series (beginning with #1154), merging this should be preceded by removing any existing data from the database for this type.)


susannasiebert:
+1 once the remaining tests are fixed.

dufeiyu:
+1

PR #1166, opened 2015-12-18
allow for multiple reference converters to exist

ernfrid:
The CLE test database priming script was dying because we have multiple converters. This fix simply pushes all converters over. 

A better solution might be to only push those over that don't have test names set?

### DIFF
--- a/lib/perl/Genome/Site/TGI/CleTest/prime-db.pl
+++ b/lib/perl/Genome/Site/TGI/CleTest/prime-db.pl
@@ -111,9 +111,11 @@ sub load_converters {
         for my $to (@reference_ids) {
             if($from != $to) {
                 my @converters = Genome::Model::Build::ReferenceSequence::Converter->get( source_reference_build_id => $from, destination_reference_build_id => $to );
-                die "More than one converter found from: $from to: $to!\n" if (@converters > 1);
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Site/TGI/CleTest/prime-db.pl
```diff
@@ -111,9 +111,11 @@ sub load_converters {
         for my $to (@reference_ids) {
             if($from != $to) {
                 my @converters = Genome::Model::Build::ReferenceSequence::Converter->get( source_reference_build_id => $from, destination_reference_build_id => $to );
-                die "More than one converter found from: $from to: $to!\n" if (@converters > 1);
+                for my $converter (@converters) {
+                #die "More than one converter found from: $from to: $to!\n" if (@converters > 1);
```
tmooney: Do we need to keep this commented-out `die` line?

tmooney:
+1.  I agree it might be better to limit on test name. (Usually `get_with_lock` is used to guarantee a single result, though that requires a sponsor and requestor.)

Is there utility in having the spurious converters present in the test database to make sure nothing in the tests uses them?

ernfrid:
I'm going to open a new pull request with just the query change.

PR #1167, opened 2015-12-21
Only load those converters that don't have test_name set

ernfrid:
We have a few instances of more than one converter for a given pair of reference sequences, but only one shouldn't have test name set. This alters the query to ensure we only pull those converters where test_name is not set and leaves in the check to make sure we only get back one.


tmooney:
+1 :tram:

acoffman:
+1

PR #1168, opened 2015-12-22
Quiet warnings in SyncSysUserWithLdap

tmooney:
Without a return value, the status message is used as the exit code, which triggers a message like:
```
Argument "done- 1 creates, 9000 deletes, 9001 total" isn't numeric in numeric lt (<) at Command/Common.pm line 126.
```
An explicit return was added.

Running the test produced another warning:
```
Scalar value @apipe_db_users[0] better written as $apipe_db_users[0] at Genome/Site/TGI/Command/SyncSysUserWithLdap.t line 36.
```
`perl`'s advice was taken here.


brummett:
+1

tmooney:
#1170 performs a more extensive overhaul of this command; I'll thus close this PR.

PR #1170, opened 2015-12-22
Sync Sys/LDAP Users: Add --max-changes-allowed option to run with more than 10 changes

ebelter:


tmooney:
I have been advocating against naming parameters `--force`.  Perhaps this could be an integer parameter like `--max-allowed-changes` with default of `10`?  Or if we stick with a boolean, maybe `--allow-copious-changes` or something descriptive like that?

ebelter:
@tmooney Sure...done.

acoffman:
+1

tmooney:
+1 :banana:

PR #1171, opened 2015-12-23
Add do_not_create_define_command so they are not made for model class stubs

ebelter:


lib/perl/Genome/Model/Command/Define.pm
```diff
@@ -24,6 +24,7 @@ sub _build_sub_command {
 
     my $model_subclass_name = $class_name;
     $model_subclass_name =~ s/::Command::Define::/::/;
+    return if eval{ $model_subclass_name->do_not_create_define_command; };
```
brummett: This should use Try::Tiny to protect $@.

Alternatively, you could define ```do_not_create_define_command()``` in Genome::Model to return false and do away with the eval/try.

ebelter: I didn't want to pollute the main model namespace with this function, hence the eval. I was going to test if the model class isa 'ModelDeprecated', but that includes a bunch of models that actually aren't deprecated :facepunch: I like the try::tiny addition. I envision this being removed when we are done removing the truly deprecated model types.

lib/perl/Genome/Model/Command/Define.pm
```diff
@@ -24,6 +24,9 @@ sub _build_sub_command {
 
     my $model_subclass_name = $class_name;
     $model_subclass_name =~ s/::Command::Define::/::/;
+    my $do_not_create_define_command;
+    Try::Tiny::try { $do_not_create_define_command = $model_subclass_name->do_not_create_define_command; };
```
tmooney: Should this module `use` or `require` `Try::Tiny` before calling `Try::Tiny::try`?

tmooney:
+1 :grapes:

PR #1174, opened 2015-12-29
New command for exporting the data of a build

susannasiebert:


susannasiebert:
@acoffman I addressed both of your comments. Great idea about the `is_$status` subroutines. I was wondering how to remove that redundancy but wasn't sure how to solve it.

acoffman:
+1

brummett:
+1

PR #1176, opened 2015-12-29
Remove metagenomic composition shotgun model type

ebelter:
Awaiting data removal, or archiving.


lib/perl/Genome/Model/Build/MetagenomicCompositionShotgun.t
```diff
@@ -6,8 +6,5 @@ use warnings;
 use above 'Genome';
 use Test::More tests => 1;
 
-# This test was auto-generated because './Model/Build/MetagenomicCompositionShotgun.pm'
-# had no '.t' file beside it.  Please remove this test if you believe it was
-# created unnecessarily.  This is a bare minimum test that just compiles Perl
-# and the UR class.
+# This model type is deprecated
```
tmooney: Since we're removing the Model class completely, we should also remove the Build class completely!

lib/perl/Genome/ProcessingProfile/MetagenomicCompositionShotgun.t
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env genome-perl
+
+use strict;
+use warnings;
+
+use above 'Genome';
+use Test::More tests => 1;
+
+# This model type is deprecated
```
tmooney: Same story for the ProcessingProfile! :smile:

tmooney:
+1 :fire:

PR #1177, opened 2015-12-29
Now that Genome::Logger uses Genome, only require it to break cycle.

tmooney:
Genome in the TGI environment uses these backends, so if they indirectly use Genome it causes a circular dependency with sometimes indeterminate effects.

This fixes the currently-failing testcase on `master` for me.


brummett:
+1

ebelter:
+1

PR #1179, opened 2015-12-31
Add an LSF Queue specification to GMT: Sam::Readcount

tmooney:
This should take care of another class of jobs that was ending up in the default queue.


ebelter:
+1 :bomb:

acoffman:
+1

PR #1180, opened 2016-01-04
Process Diff Blessed: add diff role and helps

ebelter:


lib/perl/Genome/Role/Comparable/Command/DiffBlessed.pm
```diff
@@ -11,6 +11,14 @@ role Genome::Role::Comparable::Command::DiffBlessed {
     requires => ['new_object'],
 };
 
+sub help_brief : Overrides(Genome::Role::Comparable::Command::Diff) {
```
tmooney: I'm not sure what it means for this to specify an override for a role that it does not itself consume.

tmooney:
I think I'm of the opinion that inheritance was the better relationship for the build and process diff commands.

lib/perl/Genome/Role/Comparable/Command/DiffBlessed.pm
ebelter: This is an artifact of me trying to let this role have the 'diff' role. The methods from the diff role were not conferred to the diff-blessed role, and I abandoned this fix. I will remove it, and have additionally fixed the help to not refer to processes, but objects.

ebelter:
@tmooney I agree. I wanted to see what this looked like mocked up. Close this PR, and revert b251a2a9eed11c399251381776550e2aa6b6b2fa ?

ebelter:
This change has horked some additional tests...

ebelter:
Closing in favor of #1181.

PR #1181, opened 2016-01-04
Revert "Convert Diff/DiffBlessed commands into UR Roles"

ebelter:
This reverts commit b251a2a9eed11c399251381776550e2aa6b6b2fa.

The error of not finding 'has_diffs' on genome process diff blessed causeed investigation into this issue. It was decided that having the diff as roles was not suitable, and they should remain commands that pass on behavior via inheritance.

A fix for this problem was proposed in #1180.


tmooney:
This needs to also undo part (but not all) of 2246a310d98530faec76ede5644c335c54f48e96.

tmooney:
+1 :panda_face:

brummett:
+1

PR #1182, opened 2016-01-05
Calculate the LSF resource for Alignment based on the Instrument Data

tmooney:
This is a feature that RefAlign has but the AlignmentDispatcher did not.  ("AlignAndMerge" is using this same approach for SpeedSeq in SomaticValidation; this does the same thing for regular "Align".)


dufeiyu:
+1

acoffman:
+1

PR #1183, opened 2016-01-05
Remove Genome::Model::Germline

susannasiebert:
This model type is already deprecated and the classes are a stub.

Before merging, all models, builds, processing profiles etc should be deleted from the database.


tmooney:
+1.

The old workflow that this build type ran was in `gmt germline capture-bams` (which itself was deleted in d08d6b84701442346ccfd9076c1be997c608732b in 2013).

jeldred:
+1 :shipit:

susannasiebert:
The data has been deleted. Merging

PR #1184, opened 2016-01-05
Include stalled tickets in failed-model-tickets scripts.

tmooney:
The `failed-model-tickets` script was reporting a few models that were already in a ticket, because that ticket had been stalled.  This updates the search to include stalled tickets so they are not re-reported.  (Additionally, `failed-model-ticket-status` will also include stalled tickets in its default output.)


ebelter:
+2

dufeiyu:
+1

PR #1185, opened 2016-01-05
Don't try to call ptero_workflow_proxy on an undefined value.

tmooney:
Inline builds do not have processes created in [`Genome::Model::Build->_launch`](https://github.com/genome/genome/blob/981439e064d6042a3abb384ccc46bcd52bfa7787/lib/perl/Genome/Model/Build.pm#L1431).

### DIFF
--- a/lib/perl/Genome/Model/Build.pm
+++ b/lib/perl/Genome/Model/Build.pm
@@ -557,6 +557,7 @@ sub workflow_name {
 sub ptero_workflow_proxy {
     my $self = shift;
 
+    return unless $self->process;
     return $self->process->ptero_workflow_proxy;
 }
 


acoffman:
+2

davidlmorton:
+1 Although I don't know who uses inline builds and why.

tmooney:
@davidlmorton It's the [default option](https://github.com/genome/genome/blob/5b9d0b840bc747a29f7b569615196274fc99fdb6/lib/perl/Genome/Model/Command/Define/ImportedReferenceSequence.pm#L86-L93) when importing new reference sequences.

ebelter:
Should the inline option be removed for ref seqs?

acoffman:
It seems like it probably should right? Aren't inline workflows eventually going away once we deprecate legacy workflow?

tmooney:
Probably the `server_dispatch` and `job_dispatch` options need to be taken out of `Build` altogether.  A lot of what's in there can be removed in favour of what `Process` provides.

At the moment there may still exist builds without processes, so I'm going to merge this :smile:

PR #1186, opened 2016-01-06
DV2 Dispatcher uses WorkflowBuilder

tmooney:
What's new:
* Switch to WorkflowBuilder to create the DV2 workflow.

Potential issue:
* This still does `use Workflow;` in order to set its log directory to that of the Workflow it happens to be running inside.

Future development:
* Split this module into one thing that makes the workflow, and another step that does the additional work after the workflow completes.


mkiwala:
+1 :koala:

acoffman:
+1

PR #1187, opened 2016-01-06
Stop using Workflow since this is built with WorkflowBuilder.

tmooney:
This was updated to stop using Workflow in #924, but these straggling references were not changed.


acoffman:
+2

davidlmorton:
+1

PR #1188, opened 2016-01-06
Return a HASH reference from Composite Workflow Generator

tmooney:
In the beginning was `Workflow` and it accepted its inputs as a list, so this produced an `ARRAY` reference.  At some point this was changed to call `Genome::WorkflowBuilder::DAG`, which accepts a hash reference, and a comment was added questioning the reasonableness of the choice of an `ARRAY` reference.  In #924 it was suggested we should change it.

Here are a couple versions (either the first commit or both together) of how it might look to change it.  There's a third possibility where even more things get changed to hashes as well.

### DIFF
--- a/lib/perl/Genome/InstrumentData/Composite/Workflow-unit.t
+++ b/lib/perl/Genome/InstrumentData/Composite/Workflow-unit.t
@@ -119,8 +119,7 @@ sub workflow_and_inputs_for_strategy {
 
     my ($wf, $wf_inputs) = $obj->_generate_workflow($tree);
     isa_ok($wf, "Genome::WorkflowBuilder::DAG", "workflow model");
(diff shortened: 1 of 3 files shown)


davidlmorton:
+1 It makes sense to me that if the data is always made up of key:value pairs that we use a hash or hashref.  Perl's semantics for updating a hash with another hash are a little cumbersome though.

tmooney:
There's another syntax that might be better:
```perl
@inputs{ keys %new_inputs } = values %new_inputs;
```

davidlmorton:
I just meant that my concept of what makes sense is colored by my experiences with python where the syntax is much clearer:

inputs.update(new_inputs)

acoffman:
+1

PR #1189, opened 2016-01-07
Use Genome::Role::CommandWithColor as role

davidlmorton:


tmooney:
+2 :construction_worker:

PR #1190, opened 2016-01-07
`_execute_build`-style builds use WorkflowBuilder

tmooney:
In #1045, `Build` was updated to accept workflows from `WorkflowBuilder`.  This takes advantage of that for our one-step builds (like `ReferenceSequence`).


davidlmorton:
+1

susannasiebert:
+1

PR #1191, opened 2016-01-07
Breakdancer uses WorkflowBuilder for parallel-by-chromosome

tmooney:
A future next step might be to add the call to `MergeFiles` into the workflow itself.

### DIFF
--- a/lib/perl/Genome/Model/Tools/DetectVariants2/Breakdancer.pm
+++ b/lib/perl/Genome/Model/Tools/DetectVariants2/Breakdancer.pm
@@ -272,9 +272,8 @@ sub _parallelize_by_chromosome {
     );
     $dag->add_operation($op);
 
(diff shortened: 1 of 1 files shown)


dufeiyu:
+1  looks good to me. Only concern is Breakdancer.t doesn't have test case for chr eq 'all'. We just need keep a close eye on both jenkins short and long model test. Or more proactive way is to start a de novo build of model 'apipe-test-somatic-variation' (2889153558) from this git branch to see whether build succeeds with no diff or not.

tmooney:
I ran a manual test with chr eq 'all' to make sure it still executes the workflow; the reason it's not included in the unit test is that it takes awhile to get through all the chromosomes.  I did not check to see if there are diffs, though.

susannasiebert:
+1

PR #1195, opened 2016-01-08
RNA Seq Picard Metrics: skip if no annotation build or it does not have requried files

ebelter:


lib/perl/Genome/InstrumentData/AlignmentResult/Command/PicardRnaSeqMetrics.pm
```diff
@@ -78,8 +79,37 @@ sub chart_output_filename {
...
+    my ($self, $annotation_build, $reference_build) = Params::Validate::validate_pos(
+        @_, {isa => __PACKAGE__}, {type => OBJECT}, {type => OBJECT},
+    );
+
+    my @missing_files;
+    for my $file_method ( required_files_from_annotation_build() ) {
+        my $file = $annotation_build->$file_method('gtf', $reference_build->id ,0);
+        if ( not -s $file ) {
+            $file = $annotation_build->$file_method('gtf', undef, 0);
+            if ( not -s $file ) {
```
tmooney: This pattern of trying with the reference build and then without is now repeated thrice in this module.  Would it be worth pulling it out into its own subroutine?

lib/perl/Genome/InstrumentData/AlignmentResult/Command/PicardRnaSeqMetrics.pm
```diff
@@ -78,8 +79,37 @@ sub chart_output_filename {
...
+                push @missing_files, $file_method;
+            }
+        }
+    }
+
+    if ( @missing_files ) {
+        die $self->error_message("Cannot proceed! Missing required files from annotation build: @missing_files",);
```
tmooney: I'd prefer that something named `does...` returns true or false, and that something named `verify...`, `validate...`, `ensure...`, etc. either returns true or `die`s.

Since the names of which files are missing are desired, maybe this subroutine should be modeled on `__errors__` where the return value is anything that went wrong?  Then this command can die with the list, and the `RnaSeq` command can skip calling this command without needing to catch an exception.

lib/perl/Genome/InstrumentData/AlignmentResult/Command/PicardRnaSeqMetrics.t
```diff
@@ -3,11 +3,53 @@
...
+        throws_ok(
+            sub{ $class->does_annotation_build_have_required_files($annotation_build, $reference_build); },
+            qr/Cannot proceed\! Missing required files from annotation build: /,
+            "does_annotation_build_have_required_files fails when annotation build does not have $file_method",
+        );
+        $annotation_build->set_always($file_method, 'foo');
+    }
+
+    map { $annotation_build->set_always($_, $0) } $class->required_files_from_annotation_build;
+    $annotation_build->set_always('annotation_file', $0);
```
tmooney: This line appears to be redundant to the line above it.

This will fail in this pathological case:
`perl -e 'do "Genome/InstrumentData/AlignmentResult/Command/PicardRnaSeqMetrics.t"'` as `$0` will have the value `-e`.  `__FILE__` might be a better choice.  Another option might be `$^X`. (Or write something in `/tmp`!)

lib/perl/Genome/InstrumentData/AlignmentResult/Command/PicardRnaSeqMetrics.t
```diff
@@ -3,11 +3,53 @@
...
-use_ok('Genome::InstrumentData::AlignmentResult::Command::PicardRnaSeqMetrics');
+my $class = 'Genome::InstrumentData::AlignmentResult::Command::PicardRnaSeqMetrics';
+use_ok($class) or die;
+
+subtest 'does_annotation_build_have_required_files' => sub{
+    plan tests => 5;
+
+    my $annotation_build = Test::MockObject->new;
+    $annotation_build->set_always('rRNA_MT_file', 'foo');
+    $annotation_build->set_always('annotation_file', 'foo');
```
tmooney: These may cause failures if the person running tests has created a file named `foo` in their working directory.  (This probably won't catch me out, since it doesn't say `turkey` :smile:)

`/dev/null` could be an option (or naming a non-existent file in a temporary directory!)

lib/perl/Genome/InstrumentData/AlignmentResult/Command/PicardRnaSeqMetrics.pm
ebelter: I originally didn't do it because this module does not have any tests less the ones I just wrote. Went ahead and made it a method.

lib/perl/Genome/InstrumentData/AlignmentResult/Command/PicardRnaSeqMetrics.pm
ebelter: Sure. Would be nice if we all agreed on this sort of behavior. I get the 'does' seem like a check, and would not be so harsh as to die if the check failed.

lib/perl/Genome/InstrumentData/AlignmentResult/Command/PicardRnaSeqMetrics.t
ebelter: Replaced the files with annon subs that perform the expected behavior. Updated to use a real temp file and dev null;

ebelter:
@tmooney Updates made. Take a looky-see when you can!

lib/perl/Genome/InstrumentData/AlignmentResult/Command/PicardRnaSeqMetrics.pm
```diff
@@ -78,21 +79,60 @@ sub chart_output_filename {
...
+    return $file if $file and -s $file;
+
+    return;
+}
+
 sub execute {
     my $self = shift;
-    
+
+    $self->do_input_builds_have_required_files($self->annotation_build, $self->reference_build);
```
tmooney: Do you mean `verify_annotation_build_has_required_files`?

tmooney:
It is important that the logic to skip be run in `execute`.  Having it occur in `shortcut` is just that--a shortcut.

ebelter:
@tmooney Seems like shortcut should always be run before execute.

tmooney:
@ebelter `shortcut` will not be run if a command is used at the command line.  `shortcut` might also fail for some unrelated reason, and then `execute` will be attempted.

As a general rule, I don't think it makes sense for `execute` to fail in a case where `shortcut` would have succeeded.

ebelter:
@tmooney Moved skipping logic into own method called by shortcut and execute. Moved mock annotation build creation into test factory namespace.

PR #1196, opened 2016-01-09
AnP: activate config file cli

ebelter:


lib/perl/Genome/Config/AnalysisProject/Command/ActivateConfigFile.pm
```diff
@@ -0,0 +1,26 @@
...
+use warnings;
+
+use Genome;
+
+class Genome::Config::AnalysisProject::Command::ActivateConfigFile {
+    is => 'Command::V2',
+    has_input => [
+        profile_item => {
+            is => 'Genome::Config::Profile::Item',
+            doc => 'Specified config file will no longer be used',
```
tmooney: This documentation is not accurate.

tmooney:
This brings up the same issues as #1151.

jasonwalker80:
I would reiterate that activating a disabled config should not be allowed.  Is there a use case for activating a config file?  Why was it inactivated or disabled in the first place?

ebelter:
My use case involved reactivating a config that had a processing profile that skipped digital expression. I added a config to reprocess some instrument data that failed with HT Seq. After reprocessing, I inactivated the config file.

I thought about not allowing disabled configs to be activated, and should be restricted.

Additionally, the file could just be added back to the AnP every time, even a disabled one.

ebelter:
Jason and I have talked out a different solution. PR forthcoming.

PR #1198, opened 2016-01-11
Remove CoverageStatsV2.

tmooney:
The non-V2 version is the version used in pipelines; this version was not put into production.  (`AlignmentSummaryV2` was incorporated into the non-V2 version.)


acoffman:
+1

lib/perl/Genome/Model/Tools/BioSamtools/RefCov.pm
```diff
@@ -86,11 +86,8 @@ sub execute {
     my $output_directory = $self->output_directory;
     my $wingspan = $self->wingspan;
     if ($output_directory) {
-        # TODO: Old CoverageStats required running RefCov with multiple Wingspan valude
-        # In reality, this create redundancy across the ROI.
-
-        # Moving forward CoverageStatsV2 will create multiple BED files.
-        # What do we do with this auto-generated subdirectory....
+        # TODO: CoverageStats requires running RefCov with multiple Wingspan valude
```
dufeiyu: A typo here: value.

dufeiyu:
+1  just in case, is @jasonwalker80  ok with this removal ?

lib/perl/Genome/Model/Tools/BioSamtools/RefCov.pm
tmooney: Fixed!

jasonwalker80:
+1

PR #1200, opened 2016-01-11
Specify a default LSF queue for aligners that don't list one explicitly.

tmooney:
In #1182 the resource request generation for alignments was changed.

Because the fallback is now overridden, the queue information can be lost if the aligner-specific implementation doesn't reiterate it.  These parameters establish a default, so the job still goes to the alignment queues if not otherwise specified.


acoffman:
Is the value of `lsf_queue` not inherited by `PerLaneTophat` and `Star`? They are children of `AlignReads` which does have it specified.

tmooney:
It is in fact inherited by the subclasses.  I've updated the PR to only change a single file.

acoffman:
+1

dufeiyu:
+1

PR #1201, opened 2016-01-12
Add support for AnP-based envrionmental configuration.

tmooney:
Have you ever wanted to have one Analysis Project run PTero without making them all do so?  Have you ever wished you could change the LSF resource request for an entire project at once rather than manually starting models?  This is the pull request for you!

Features:
* Adds a new location for `Genome::Config` to search for configuration that takes precedence over the home directory (but still lower precedence than manually set environment variables).

Caveats:
* Won't work unless we update our production users to use a `genome/config.yaml` instead of manually set environment variables.

### DIFF
--- a/lib/perl/Genome/Config.pm
+++ b/lib/perl/Genome/Config.pm
@@ -130,6 +130,12 @@ sub all_keys {
 
 sub config_subpath { Path::Class::File->new('genome', 'config.yaml') }
 
+sub project_dir {
+    my $dir = $ENV{XGENOME_CONFIG_PROJECT_DIR};
+    return $dir if $dir;
+    return;
+}
+
 sub home_dir {
     my $path = ( $ENV{XGENOME_CONFIG_HOME} || File::Spec->join($ENV{HOME}, '.config') );
     return Path::Class::Dir->new($path);
@@ -193,7 +199,7 @@ sub _lookup_value {
     if ($spec->has_env && exists $ENV{$spec->env}) {
         return $ENV{$spec->env};
     }
-    my @files = _lookup_files($config_subpath, home_dir());
+    my @files = _lookup_files($config_subpath, project_dir(), home_dir());
     my $value = _lookup_value_from_files($spec, @files);
     if (defined $value) {
         return $value;
--- a/lib/perl/Genome/Config.pm
+++ b/lib/perl/Genome/Config.pm
@@ -132,7 +132,7 @@ sub config_subpath { Path::Class::File->new('genome', 'config.yaml') }
 
 sub project_dir {
     my $dir = $ENV{XGENOME_CONFIG_PROJECT_DIR};
-    return $dir if $dir;
+    return Path::Class::Dir->new($dir) if $dir;
     return;
 }
 
@@ -150,7 +150,7 @@ sub snapshot_dir {
     my @chop = ('lib', 'perl', split('::', __PACKAGE__));
     my @chopped = splice(@path, -1 * @chop);
     my @base_dir = @path;
-    return File::Spec->join(@base_dir, 'etc');
+    return Path::Class::Dir->new(File::Spec->join(@base_dir, 'etc'));
 }
 
 sub global_dirs {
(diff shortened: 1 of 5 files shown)


lib/perl/Genome/Model/Build/Command/Start.pm
```diff
@@ -143,6 +143,9 @@ sub create_and_start_build {
 
     my $outer_transaction = UR::Context::Transaction->begin();
     $model->build_requested(0);
+    my $anp = $model->analysis_project;
+    my $config_dir = $anp? $anp->environment_config_dir : undef;
```
acoffman: I guess this check is in here to hedge our bets - but it seems like people shouldn't be starting builds that aren't linked to analysis projects anyways. Would it make sense to `die` if an AnP is not present?

lib/perl/Genome/Config.pm
```diff
@@ -130,6 +130,12 @@ sub all_keys {
 
 sub config_subpath { Path::Class::File->new('genome', 'config.yaml') }
 
+sub project_dir {
+    my $dir = $ENV{XGENOME_CONFIG_PROJECT_DIR};
+    return $dir if $dir;
```
acoffman: It may not matter, but it appears that `home_dir` wraps its path in a `Path::Class::Dir` object on return. It seems odd that `project_dir` and `home_dir` would have inconsistent return types.

lib/perl/Genome/Config/AnalysisProject/Command/AddEnvironmentFile.pm
```diff
@@ -0,0 +1,72 @@
...
+sub execute {
+    my $self = shift;
+    my $analysis_project = $self->analysis_project;
+    if($analysis_project->environment_config_dir) {
+        $self->fatal_message('Project %s already has an existing environment file.', $analysis_project->__display_name__);
+    }
+
+    my $allocation = Genome::Disk::Allocation->get(owner => $analysis_project);
+    unless ($allocation) {
+        $allocation = Genome::Disk::Allocation->create(
```
acoffman: Maybe not, but at the point at which we're creating allocations that directly belong to an AnalysisProject, does an AnalysisProject become an `ObjectWithAllocations` and do we need to clean up this allocation when analysis project gets deleted?

lib/perl/Genome/Config.pm
tmooney: Mine matches `snapshot_dir` (which also differs from `global_dirs`).  You're right, though--it's odd that they're inconsistent!  That should be reconciled one way or the other...

lib/perl/Genome/Config/AnalysisProject/Command/AddEnvironmentFile.pm
tmooney: You're right, we should clean this up when it's deleted.  Maybe it does become an `ObjectWithAllocations`.

(Though in practice how often do we delete AnPs?)

lib/perl/Genome/Model/Build/Command/Start.pm
tmooney: I would be happy to make that change if @jasonwalker80 thinks we're at the point where we can enforce the Analysis Project requirement!

jasonwalker80: Without additional PRs or changes to our test suite we couldn't meet this requirement with test builds.  Many of the recent builds without an AnP are created_by apipe-tester:
$ genome model build list --filter 'created_at>=2015-11-01,model.analysis_projects.id=""' --nohead -sh created_by | sort | uniq -c

With some improved documentation describing how to assign a SomVar, ClinSeq or random model to an AnP manually and an appropriate amount of notice before the requirement is released to a live snapshot... yes we should require an AnP for all builds to start.   However, is it worth coming back to this in another PR?

acoffman: Yeah, I don't think we should hold up this PR for what amounts to an unrelated change. Just something I noticed while reading though.

acoffman:
+1

susannasiebert:
+1

PR #1202, opened 2016-01-12
Build volume lister

brummett:
This command lists the disk volumes used by the given builds.  You can filter by build status, or build_id (or both, if you want the intersection).

It looks at the disk allocations used directly by the build, it's software results and instrument data.

### DIFF
--- a/lib/perl/Genome/Model/Build/Command/ListVolumes.pm
+++ b/lib/perl/Genome/Model/Build/Command/ListVolumes.pm
@@ -1,0 +1,92 @@
+package Genome::Model::Build::Command::ListVolumes;
+
+use Genome;
+
+class Genome::Model::Build::Command::ListVolumes {
+    is => 'Command::V2',
+    has_optional => [
+        status => {
+            is                  => 'String',
+            doc => 'Build status to filter on',
+            valid_values => Genome::Model::Build->__meta__->property('status')->valid_values,
+        },
+        build_ids => {
+            is                  => 'String',
+            is_many             => 1,
+            shell_args_position => 1,
+            doc => 'Build ID(s) to use.'
+        },
+    ],
+};
+
+sub help_detail {
+    <<HELP
+Prints the mount paths for volumes used by the given builds directly, and by
+their software results and instrument data.
+
+Builds may be filtered by build_id or by status.
+HELP
+}
+
+sub execute {
+    my $self = shift;
+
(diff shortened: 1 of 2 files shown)


tmooney:
Could this command support using an arbitrary BX for builds at least resembling the default `Command::V2` object resolution?

brummett:
I considered that.  Are people likely to use different kinds of filters to justify the more complicated API?

tmooney:
I'm not really sure who is likely to run this command.

If I'm running it, I probably would prefer it work like most `Command::V2`s, so both `--builds 123456 123457 123458` and `--builds run_by=tmooney,status=Succeeded,'date_completed>2016-01-05T00:00:00Z'` work.

Barring that, I'd probably provide only a list of IDs since I'm going to have to find those IDs another way that probably incorporates any filters already.

brummett:
I'm using this command to know when it's safe to shrink some volumes.  Running it with ```--status=Running``` and seeing if any of the listed volumes is in the list of volumes I want to shrink.

I decided against using Command::V2's built-in object conversion for args (--builds 123456, etc getting changed into build object).  This command is likely to churn through a lot of data, and there's no way to get an iterator from Command::V2 (though that would be a useful feature).

What about passing the arg to this command through ```UR::BoolExpr->resolve_for_string``` and using that as the filter for the iterator?

tmooney:
I think I'd like that--then it behaves pretty much like our listers do.  We could maybe even call it `--filter`!

lib/perl/Genome/Model/Build/Command/ListVolumes.pm
```diff
@@ -0,0 +1,92 @@
...
+    my $print_new_vols = sub {
+        my $path = shift;
+        unless (exists $seen_vols{$path}) {
+            $seen_vols{$path} = undef;
+            print $path,"\n";
+        }
+    };
+
+    BUILD:
+    while(1) {
```
acoffman: Rather than an infinite loop with a label and a `last` would something like `while (my $guard = UR::Context::AutoUnloadPool->create() && $iter->next()) { }` work?

I also wonder if it would make sense to split this out into two different subs: one that calculates the unique list of volumes and returns the set or hash or list (or whatever data structure you want to use) and another that prints it. That leaves the door open for someone to use the former in a script elsewhere if they need the list of volumes for a build.

brummett:
I'm closing this PR in favor of #1283

PR #1207, opened 2016-01-13
BamToCna declares "params" as an Input and Output

tmooney:
WorkflowBuilder is more stringent about validating its components than Workflow.  "params" needs to be declared as an input and output for the workflow that the DetectVariants2 Dispatcher creates to validate.  (The Dispatcher passes "params" to all DV2 detectors--the base Detector class declares "params" appropriately, but this class overrides that definition.)


brummett:
+1

acoffman:
+1

PR #1208, opened 2016-01-13
Update the Varscan SomaticFilterWorkflow to use WorkflowBuilder.

tmooney:
This is run as part of `ClinSeq`'s `IdentifyLoh` step.

### DIFF
--- a/lib/perl/Genome/Model/Tools/Varscan/SomaticFilterWorkflow.pm
+++ b/lib/perl/Genome/Model/Tools/Varscan/SomaticFilterWorkflow.pm
@@ -169,11 +169,13 @@ sub run_format_workflow {
     $w->parallel_by('variants_file');
     $self->set_lsf_queue($w);
 
-    my $result = $self->execute_workflow_for_operation($w, 'format', {
+    my $inputs = {
         'variants_file' => $input->{variants_file},
         'outdir' => $self->outdir,
         'append_line' => 1,
-    });
+    };
+
+    my $result = $self->execute_workflow_for_operation($w, 'format', $inputs);
 
     $self->check_result($result);
 }
@@ -189,9 +191,11 @@ sub run_process_workflow {
     $w->parallel_by('status_file');
     $self->set_lsf_queue($w);
 
-    my $result = $self->execute_workflow_for_operation($w, 'process', {
+    my $inputs = {
         'status_file' => $input->{format_output},
-    });
+    };
+
+    my $result = $self->execute_workflow_for_operation($w, 'process', $inputs);
 
     $self->check_result($result);
 }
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Model/Tools/Varscan/SomaticFilterWorkflow.pm
```diff
@@ -116,52 +115,65 @@ sub check_result {
...
     $w->parallel_by('variants_file');
     $self->set_lsf_queue($w);
-    $w->log_dir($self->outdir);
-    $self->validate_workflow($w);
 
-    # Launch workflow
-    $self->status_message("Launching format workflow now.");
-    my $result = Workflow::Simple::run_workflow_lsf(
-        $w,
+    my $result = $self->execute_workflow_for_operation($w, 'format', {
```
davidlmorton: Perhaps outside the scope of this PR, but I think using named arguments here would be clearer.

lib/perl/Genome/Model/Tools/Varscan/SomaticFilterWorkflow.pm
```diff
@@ -170,22 +182,16 @@ sub run_process_workflow {
...
     $w->parallel_by('status_file');
     $self->set_lsf_queue($w);
-    $w->log_dir($self->outdir);
-    $self->validate_workflow($w);
 
-    # Launch workflow
-    $self->status_message("Launching process workflow now.");
-    my $result = Workflow::Simple::run_workflow_lsf(
-        $w,
+    my $result = $self->execute_workflow_for_operation($w, 'process', {
```
davidlmorton: Here as well, named arguments would be clearer.  And since this method was created in this PR, I don't think its outside the scope. :)

``` perl
use Params::Validate qw(validate_pos :types);

sub execute_workflow_for_operation {
    my $self = shift;
    my %p = validate_pos(@_, {
            operation => {isa => 'Genome::WorkflowBuilder::Detail::Operation'},
            operation_shortname => {type => SCALAR},
            inputs => {type => HASHREF},
        }
    );
    # ...
}
```

tmooney: If I did make `execute_workflow_for_operation` pass parameters by name, I'd want to do it for this whole class.  Having one subroutine that passes by name instead of position would be odd.

I think a lot of the weirdness in my change was that I kept the inline parameter passing from `run_workflow_lsf`.  Now I've moved the input hash construction to before the call.  The variable name thus is a much larger clue what is happening.

If that didn't sufficiently increase the clarity, I'd prefer to make a follow-up PR that cleans up this class more generally.  (I would probably abandon `$w` as a variable name while I'm at it!)

davidlmorton:
+1

mkiwala:
+1 :koala:

PR #1210, opened 2016-01-13
BamQc builds its workflow with WorkflowBuilder.

tmooney:
This was already converting to WorkflowBuilder to run the workflow; now it builds that way, too!


davidlmorton:
+1 You're really cranking these conversions out!

mkiwala:
+1 :koala:

PR #1211, opened 2016-01-14
Remove Msi model commands.

tmooney:
The `Msi` build class was removed in 5f3a96bc2d72199a18f1f8d7ceeb574430c93e03.

These commands were not updated to account for UUID build IDs, so it seems unlikely they have been run recently!  This removes some references to `Workflow::Simple` without updating them to use `Genome::WorkflowBuilder`.


acoffman:
+1

brummett:
+1

PR #1212, opened 2016-01-14
Call recursively_set_log_dir on an object whose class implements it.

tmooney:
#1206 had a bug in it. This corrects the mistake.


acoffman:
+2

PR #1213, opened 2016-01-14
Remove SmrtAnalysis GMT tools.

tmooney:
These tools show no usage since the beginning of recordkeeping in the usage log.  These would need to be updated to `WorkflowBuilder`, but instead this PR deletes them.


acoffman:
+1

`+0 −7,200` :tada:

brummett:
+1

PR #1214, opened 2016-01-14
Remove unused CCSVCF includes.

tmooney:
This doesn't really use `Workflow` since it was converted to a Process as part of #429.


acoffman:
+1

dufeiyu:
+1

PR #1215, opened 2016-01-15
AnP: Replace Model Command

ebelter:
This command takes a model and new config item and copies the model,
replacing the model params as directed be the config.The model is added
to the AnP and queued to build.

### DIFF
--- a/lib/perl/Genome/Config/AnalysisProject/Command/ReplaceModel.pm
+++ b/lib/perl/Genome/Config/AnalysisProject/Command/ReplaceModel.pm
@@ -88,7 +88,7 @@ sub _overrides_for_model {
         next if ref $config_for_model->{$key};
         next if defined $config_for_new_model->{$key} and $config_for_model->{$key} eq $config_for_new_model->{$key};
         my $key_no_id = $key;
-        $key_no_id =~ s/_id//;
+        $key_no_id =~ s/_id$//;
         $overrides{$key_no_id} = $config_for_new_model->{$key};
     }
 
--- a/lib/perl/Genome/Config/AnalysisProject/Command/ReplaceModel.pm
+++ b/lib/perl/Genome/Config/AnalysisProject/Command/ReplaceModel.pm
@@ -64,13 +64,13 @@ sub _verify_new_config {
     my $self = shift;
 
     if ( $self->new_profile_item->status ne 'inactive') {
-        die $self->error_message(
+        $self->fatal_message(
             "Config profile (%s) status (%s) must be 'inactive'.", $self->new_profile_item->id, $self->new_profile_item->status
         );
     }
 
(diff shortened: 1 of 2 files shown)


lib/perl/Genome/Config/AnalysisProject/Command/ReplaceModel.pm
```diff
@@ -0,0 +1,145 @@
...
+   },
+};
+
+sub help_brief {
+    return 'replace a model';
+}
+
+sub help_detail {
+    return <<EOS
+This command takes a model and new config profile item and copies the model, replacing the different items in the new profile config as compared to the model's existing config. The model is queued for building. Additionally, the builds forthe old model shold be abandoned. That way they do not appear in a failed state.
```
tmooney: * `forthe` should be bisected by a ` `.
* `shold` is missing a `u`.

lib/perl/Genome/Config/AnalysisProject/Command/ReplaceModel.pm
```diff
@@ -0,0 +1,145 @@
...
+    my $self = shift;
+
+    my $config_for_model = $self->_load_model_config_from_file( $self->model->config_profile_item->file_path );
+    my $config_for_new_model = $self->_load_model_config_from_file( $self->new_profile_item->file_path );
+    my %overrides;
+    for my $key ( keys %$config_for_model ) {
+        next if ref $config_for_model->{$key};
+        next if defined $config_for_new_model->{$key} and $config_for_model->{$key} eq $config_for_new_model->{$key};
+        my $key_no_id = $key;
+        $key_no_id =~ s/_id//;
```
tmooney: Is this intended to remove trailing `_id`?  If so a `$` would be good.

lib/perl/Genome/Config/AnalysisProject/Command/ReplaceModel.pm
```diff
@@ -0,0 +1,145 @@
...
+        $overrides{$key_no_id} = $config_for_new_model->{$key};
+    }
+
+    if ( not %overrides ) {
+        die $self->error_message("No overrides found for model! %s\nConfig: %s\nNew Config: %s\n",
```
tmooney: We have `$self->fatal_message` now to achieve the effect of `die $self->error_message`.

tmooney:
Should this command check for an existing copy associated with the new config profile item?

ebelter:
@tmooney I did, but it seemed overly complicated to do so. Maybe some of the logic could be pulled out of CQID to help do it.

tmooney:
Yeah, it does start to get very complex and replicate CQID.  This seems fine to help do what we're trying to do, and we can make it better if the need arises.

tmooney:
+1 :sandal:

PR #1216, opened 2016-01-19
It is reference_sequence_build not reference_build

dufeiyu:
This bug of RnaSeq build property name is from #1195 and causing jenkins rnaseq test build failures. It needs a early merge.


susannasiebert:
+2 once tests pass.

PR #1217, opened 2016-01-21
Add lsf_resource_rnaseq_cufflinks to genome config

dufeiyu:
So it is convenient to change lsf limit for cufflink run.


brummett:
+1

tmooney:
+1.  Should this be named `lsf_resource_rnaseq_cufflinks` to more closely match the name of the tool being run?

dufeiyu:
@tmooney it makes sense. I changed name from cufflink to cufflinks.

PR #1218, opened 2016-01-25
Submitting a Process Preserves Environment

tmooney:
In Perl closures close over lexical (`my`) variables but not over `local`ized globals.

This change saves a copy of the environment before creating the submission closure.  This allows the environment to be customized the same way before either `submit`ting or `launch`ing a Process.

This addresses an issue using #1201 with PTero.


davidlmorton:
+1 Nice to remove that big comment.

brummett:
+1

PR #1219, opened 2016-01-26
Don't copy entire allocations in place of symlinks.

tmooney:
The path_for_allocation might be a parent of the file or directory being symlinked.  We were getting infinite recursion when the allocation resolved to a parent of the current allocation's directory.  This
solution ignores that the symlink might be in an allocation and happily copies the contents.


acoffman:
Does this do what we actually want now? In theory a user could export a build, and that build could have symlinks to data still on apipe disk which would then be deleted. I think we do need to follow the symlinks and copy the actual data somehow (if I'm understanding this change correctly).

acoffman:
@tmooney explained what he's doing here to me off ticket. The case that was previously handled by the removed code still gets handled further down in the method.

+1

susannasiebert:
+1

PR #1220, opened 2016-01-26
Pindel VCF Converter uses WorkflowBuilder.

tmooney:
I don't know why this converter contains a workflow at all for its single step, although it has been present since the original version of the converter.


mkiwala:
+1 :koala:

ebelter:
+1 :jeans:

PR #1221, opened 2016-01-27
Sanitize build name for anntoation files

ebelter:


tmooney:
+1.

I prefer to use or require `Genome::Utility::Text`.  The autoloader warns against autoloading for non-methods.

acoffman:
+1

PR #1224, opened 2016-01-28
Remove some redundant methods from Model.pm

tmooney:
This completes a few long-standing TODOs by removing redundant methods.

(It seems like the `${status}_builds` methods in `ModelDeprecated` are also ripe for removal.  Some of them are still used at the moment.)


ebelter:
+1 double checked fo removed methods :statue_of_liberty:

acoffman:
+1

PR #1225, opened 2016-01-29
Lsf resource parsing

mkiwala:
The Genome codebase contained some strings meant to represent lsf resources that were neither valid resource requirement strings or bsub command options.  This PR fixes those problems.


davidlmorton:
+1

acoffman:
+1

PR #1226, opened 2016-01-31
Use pre/post-exec scripts for ptero LSF jobs

davidlmorton:
This should not be merged before:

https://github.com/genome/ptero-perl-sdk/pull/109 is packaged and deployed via genome-snapshot-deps.

and

https://github.com/genome/ptero-lsf/pull/110 is deployed to production.

### DIFF
--- a/lib/perl/Genome/Ptero/ToJSON.t
+++ b/lib/perl/Genome/Ptero/ToJSON.t
@@ -43,6 +43,9 @@ for my $test_directory (glob test_data_directory('*')) {
             qr/^.*"workingDirectory" :.*$/,
             qr/^.*"user" :.*$/,
             qr/^.*"cwd" :.*$/,
+            qr/^.*"postExecCmd" :.*$/,
+            qr/^.*"outFile" :.*$/,
+            qr/^.*"errFile" :.*$/,
         ],
     );
 }
--- a/lib/perl/Genome/Ptero/ToJSONForProcess.t
+++ b/lib/perl/Genome/Ptero/ToJSONForProcess.t
@@ -58,6 +58,9 @@ for my $test_directory (glob test_data_directory('*')) {
             qr/^.*"workingDirectory" :.*$/,
             qr/^.*"user" :.*$/,
             qr/^.*"cwd" :.*$/,
+            qr/^.*"postExecCmd" :.*$/,
+            qr/^.*"outFile" :.*$/,
+            qr/^.*"errFile" :.*$/,
         ],
     );
 }
--- a/lib/perl/Genome/Ptero/workflow_tests/jagged-parallel-models/ptero_workflow.json
+++ b/lib/perl/Genome/Ptero/workflow_tests/jagged-parallel-models/ptero_workflow.json
@@ -75,7 +75,11 @@
                      "XGENOME_PTERO_SHELL_COMMAND_SERVICE_URL" : "http://example.com/v1"
                   },
                   "options" : {
+                     "errFile" : "/tmp/ptero-lsf-logfile-4a3e43b5-7779-4d40-be86-dad0181826f1.err",
                      "numProcessors" : 1,
+                     "outFile" : "/tmp/ptero-lsf-logfile-4a3e43b5-7779-4d40-be86-dad0181826f1.out",
+                     "postExecCmd" : "ptero-lsf-post-exec --stderr /tmp/ptero-lsf-logfile-4a3e43b5-7779-4d40-be86-dad0181826f1.err --stdout /tmp/ptero-lsf-logfile-4a3e43b5-7779-4d40-be86-dad0181826f1.out",
+                     "preExecCmd" : "ptero-lsf-pre-exec",
                      "resReq" : "select[gtmp>1 && mem>3000] span[hosts=1] rusage[gtmp=1 && mem=3000]"
                   },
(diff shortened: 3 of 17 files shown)


tmooney:
Will Genome or another PTero user ever have cause to modify or omit these pre-exec and post-exec scripts?  (I'm wondering why it's exposed as an option that `Genome` sets to a constant every time as opposed to something that happens behind-the-scenes.)

As an aside... in `flow` when pre-exec fails the job will `PEND` in LSF indefinitely.  Will that still happen with this version?

mkiwala:
> As an aside... in flow when pre-exec fails the job will PEND in LSF indefinitely. Will that still happen with this version?

I was also having this conversation with Dave on genome/ptero-perl-sdk#109.  According to the lsf admin guide, "By default, if job pre-execution fails, LSF retries the job automatically. The job remains in the queue and pre-execution is retried 5 times by default, to minimize any impact to performance and throughput."

I checked the puppet-modules repo to see how systems has configured pre-execution parameters:
```
mkiwala@linus82:puppet-modules[master] $ git grep PREEXEC
site/lsf/templates/lsf9/lsb_applications_header.erb:# MAX_PREEXEC_RETRY = integer
site/lsf/templates/lsf9/lsb_queues_header.erb:# MAX_PREEXEC_RETRY specifies the maximum number of times to attempt the
```

It appears that they have the default set, so I would expect the job to fail after the 5th retry.  I tested by running `bsub -E false true` and running `bjobs -l` a few times in quick succession to monitor the job.  The job status appears to flip-flop between the RUN then PEND statuses, as expected, but lsf does not appear to kick the job out of the queue, which is unexpected to me based on my reading of the lsf admin guide and the lsf configuration in the puppet-modules repo.

mkiwala:
> Will Genome or another PTero user ever have cause to modify or omit these pre-exec and post-exec scripts? (I'm wondering why it's exposed as an option that Genome sets to a constant every time as opposed to something that happens behind-the-scenes.)

It seems less presumptuous to me to make the use of the pre and post exec scripts configurable.  Not exposing this an option to the end-user would require that the ptero-perl-sdk has been deployed to the ptero execution hosts and that the end-user does not need the pre- and post-exec scripts for their own custom uses.  By exposing this option to the end-user, we allow for installations that do not have the perl-sdk deployed to the lsf execution hosts and for uses of pre- and post-exec other than for patching ptero-lsf endpoints.

davidlmorton:
Not to beat this issue into the ground but I thought I'd give more rationale for why Genome is/should be responsible for opting in to use pre/post-exec scripts.  It all starts with the restriction that the script designated to lsf be installed globally (to all lsf hosts).  Our lsf service code is not installed globally, the only thing in PTero that is installed everywhere is the PerlSDK.  In terms of abstraction layers, PerlSDK sits atop of the lsf service, meaning that it would be inappropriate for the lsf service to know about the existence of or the API to use the PerlSDK.

One might argue at this point that the responsibility should be in the PerlSDK, where the pre/post scripts are defined.  There are two compelling reasons it should be moved up another level of abstraction to the application layer (Genome in our case).  First, the PerlSDK is job service agnostic when it comes to constructing workflows.  This allows us and other PTero users to define new job services without adding explicit support in the PerlSDK.  Second, the post-exec script has the ability to relay the contents of some specified log files to the lsf service so they are saved in the database.  These file paths are defined in Genome (as outFile and errFile parameters to the lsf service), so it only makes sense to also have Genome specify these paths as command line arguments to the pre/post-exec scripts.

So shifting the responsibility to Genome allows us to keep the lsf service unaware and therefore simpler.  Also, as Michael points out, it allows us or others to utilize the pre/post-exec and errFile/outFile differently if we choose.  It lets us keep the PerlSDK job agnostic (at least for workflow construction).

If another organization comes to mind please don't hesitate to bring it up.

tmooney:
I agree it's nice to preserve the option for users to run whatever pre/post-exec scripts they want.

It seems weird from a black-box perspective that we have specify a command to tell PTero about part of itself to get some behaviour Genome doesn't particularly care about.  The rationale from inside the box why each given part doesn't know about this script makes sense.

From a practical perspective, do we need to worry about cleaning up these log files from `/tmp/`?

mkiwala:
> From a practical perspective, do we need to worry about cleaning up these log files from /tmp/?

Most of the time, the files will get unlinked.   If for some reason the files are not unlinked, we expect the file contents to be small.  And we're using uuid's to name the files, so we expect file names to not clash.

You probably want to see genome/ptero-perl-sdk#109, to see the proposed scripts before we merge them.  In there, you'll find that the ptero-lsf-post-exec script unlinks the log files if it's able to read them.

tmooney:
Ah!  That helped me understand what these stdout and stderr files are for, though I'm even more convinced Genome doesn't care about them.  (Is there a situation where it's useful to have the option to specify a different error file to LSF than to the post-exec script?)

This code looks like it'll work, so +1 for that!

davidlmorton:
> Is there a situation where it's useful to have the option to specify a different error file to LSF than to the post-exec script?

Maybe... You can specify a file for errFile/outFile and use the ptero-lsf-post-exec script without specifying any arguments.  The status updating part of the post-exec script can be useful even without the currying of log data.

mkiwala:
+1 :koala:

PR #1230, opened 2016-02-02
Support Picard v1.130 Changes

jasonwalker80:
The JAVA command line usage changed in Picard versions after 1.123.  No longer are there individual JAR files for each sub-command.  A single JAR is installed and the sub-command name is provided instead of the class name.  These changes are to support the new 1.130 command line structure.


jasonwalker80:
@tmooney and @ebelter please have a look at the latest commit.  I added a test of the new 1.130 version as well as the version_uses_single_jar suggestion.  I also added a die() to run_java_vm(in _java_cmdline) when new versions are used with the old interface.  It seems like all Picard commands should be refactored to use the Picard/Base.pm rather than straight using Picard.pm and the run_java_vm interface.

tmooney:
+1.

I agree that other commands should be updated to Picard/Base.pm.  Eventually Picard.pm would become a `Command::Tree` and not have any of the functionality in it.

ebelter:
+1

PR #1231, opened 2016-02-02
Command to print out all haplotype caller output files

apregier:


tmooney:
+1.

If we find we need a more flexible version, we could make a custom lister based on something like:
`ur list objects --subject-class=Genome::Model::SingleSampleGenotype::Result::HaplotypeCaller -f users.user_id=d9104e55c1cb461f9e405e1d533b7197 -sh id,alignment_result.sample_name,intervals,vcf_file --order-by=intervals`.

susannasiebert:
+1

PR #1232, opened 2016-02-03
Commit after each interval of the HaplotypeCaller runs.

tmooney:
Since each interval is an independent run, we can save the results for those intervals even if a subsequent one does not succeed.

(This would probably be better with genome/UR#62.)

I didn't go with the shell-out approach that some things (like `IntermediateAlignmentResult`) use, but that is a possible alternative if calling `commit` mid-way seems suboptimal.

### DIFF
--- a/lib/perl/Genome/Model/SingleSampleGenotype/Command/HaplotypeCaller/BucketIterator.pm
+++ b/lib/perl/Genome/Model/SingleSampleGenotype/Command/HaplotypeCaller/BucketIterator.pm
@@ -56,6 +56,11 @@ sub _run {
             $self->fatal_message('Could not %s for interval %s.', $mode, $interval);
         }
 
+        unless($ENV{UR_DBI_NO_COMMIT}) {
+            #save our progress after creating a new result
+            UR::Context->commit() or $self->fatal_message('Failed to commit after %s for interval %s.', $mode, $interval);
+        }
+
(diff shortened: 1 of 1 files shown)


brummett:
+1

susannasiebert:
+1

lib/perl/Genome/Model/SingleSampleGenotype/Command/HaplotypeCaller/BucketIterator.pm
```diff
@@ -56,6 +56,11 @@ sub _run {
             $self->fatal_message('Could not %s for interval %s.', $mode, $interval);
         }
 
+        unless($ENV{UR_DBI_NO_COMMIT}) {
+            #save our progress after creating a new result
```
ebelter: Is it necessary to check if no commit is set? Doesn't that happen under the hood in UR?

tmooney: It's not strictly necessary.  I included it to avoid repeatedly syncing databases and then rolling back as it iterated.

brummett: It does, but checking for UR_DBI_NO_COMMIT here allows it to skip preparing/executing all the SQL that's going to get rolled back anyway.

PR #1233, opened 2016-02-03
Update Picard FilterSamReads to use Base class

jasonwalker80:
This ensures this command is compliant with the v1.130 interface. I also added an integration test with a small subset of NA12878 chr22 data.  The previous test only covered the command creation logic which should be covered by the Picard.t.


lib/perl/Genome/Model/Tools/Picard/FilterSamReads.pm
```diff
@@ -53,51 +58,18 @@ sub help_detail {
...
-    my $output_file = $self->output_file;
-
-    my $jar_path = $self->picard_path . '/FilterSamReads.jar';
-    unless ( -e $jar_path ) {
-        die( "Failed to find $jar_path!" .
-          "This command may not be available in version " . $self->use_version );
-    }
+sub minimum_version_required { '1.77'; }
+sub _jar_name { 'FilterSamReads.jar'; }
+sub _java_class { 'net.sf.picard.sam.FilterSamReads'; }
```
tmooney: GMT::Picard::Base takes care of adding the `net.sf` when it is necessary.  The `_java_class` is documented to expect return a list, which the class takes care of joining with `.`s as well.  (The current implementation still works if this is already a joined string, however.)

jasonwalker80: The most recent commit should fix this not only in FilterSamReads but other occurrences in the picard tree.

lib/perl/Genome/Model/Tools/Picard/FilterSamReads.t
```diff
@@ -1,62 +1,49 @@
...
-    ok( $error_str =~ m/may not be available/,
-      "Crashed with wrong Picard version");
-}
+my $test_dir = __FILE__.'.d';
+
+# Params
+my $picard_version = '1.130';
+
+# Inputs
+my $input_bam_file = $test_dir .'/cf651d3d8af5484a80bbe2f1498aab9b_chr22_1k.bam';
```
susannasiebert: I think we generally like to use `File::Spec->join` for file path concatenations.

susannasiebert:
+1

tmooney:
+1 :snowman:

PR #1235, opened 2016-02-03
Modernize input listing for ReferenceAlignment.

tmooney:
This is the first step on the path to modernizing ReferenceAlignment.

This removes support for programmatically defining a build with a name or ID.  (Programmers can easily get the object!)  The command line resolution based on name or ID should still be in place.


susannasiebert:
+1

dufeiyu:
+1

PR #1236, opened 2016-02-03
Remove explicit _id properties that Model.pm will generate for us.

tmooney:
The generated version points at the bridge table and so does not require the manually-generated `add_` methods either.


susannasiebert:
+1

brummett:
+1

PR #1238, opened 2016-02-03
InstData Delete CLI

ebelter:
Remove the 'remove' command and use CRUD built in 'delete'. Additionally, update the instrument data property on the delete command to only retrieve imported data.


tmooney:
+1.

It's interesting that constructing this command programmatically does not enforce the class specification, but it's no different than the old class in that regard!

ebelter:
@tmooney Yeah...I have a test written to make sure that it would enforce the class, but it doesn't. Didn't add it because it didn't prove what I wanted it to.

ebelter:
@tmooney Is this worthy of a changelog?

tmooney:
@ebelter Sounds like a good idea.  I don't think the current remove command is widely used, but perhaps it should be!

ebelter:
@tmooney I got the test to work to ensure only imported insdata is deleted. Review if you'd like.

tmooney:
New test generally looks good.  I think it should enable no-commit to be safe.

PR #1239, opened 2016-02-04
Stop subclassing ReferenceAlignment ProcessingProfiles on sequencing_platform.

tmooney:
We'll need to wait for old snapshots to fall out of use before editing the class of existing profiles and removing the subclasses altogether.


mkiwala:
+1 :koala:

susannasiebert:
+1

PR #1240, opened 2016-02-05
Make test independent of inline dispatch

mkiwala:
This test now depends on the test database instead of the ur object cache to maintain state.

It does create some allocations which `genome-test-env` is not able to clean up: https://jira.gsc.wustl.edu/browse/APIPE-3397

```
=> Removing disk allocations created during genome-env...
Using libraries at /gscuser/mkiwala/Projects/genome/lib/perl
***** GENOME_DEV_MODE (linus82.gsc.wustl.edu) ***** at /gscuser/mkiwala/Projects/genome/lib/perl/Genome.pm line 30.
Issuing rollback() due to DESTROY without explicit disconnect() of DBD::Pg::db handle dbname=genome;host=gms-postgres at /gscuser/mkiwala/Projects/genome/lib/perl/Genome/Model/Command/Admin/RemoveDiskAllocationsFromTestdb.pm line 81.
Removing 0.001 GB in 4 allocations.
```


davidlmorton:
We made a Jira issue for the cleanup bug:  https://jira.gsc.wustl.edu/browse/APIPE-3397

lib/perl/Genome/Utility/Test.pm
```diff
@@ -312,10 +323,27 @@ sub data_dir_ok {
     return $data_dir;
 }
 
+sub get_test_dir {
```
tmooney: It seems like things using this could have used `data_dir_ok` immediately preceding it.

lib/perl/Genome/Model/Command/Define/ImportedReferenceSequence.t
```diff
@@ -17,10 +17,11 @@ if (Genome::Sys->arch_os ne 'x86_64') {
     plan skip_all => 'requires 64-bit machine';
 }
 
+my $version = 1;
 my $cmd_class = 'Genome::Model::Command::Define::ImportedReferenceSequence';
 use_ok($cmd_class);
 
-my $data_dir = File::Temp::tempdir('ImportedAnnotationTest-XXXXX', CLEANUP => 1, TMPDIR => 1);
+my $data_dir = get_test_dir($cmd_class, $version);
```
tmooney: In the past we've tended to move files to the `$GENOME_TEST_INPUTS` when they're too big to include in the repository.  It seems better if the test has fewer external dependencies.  

If the trouble is we need these to be network accessible, maybe a `__FILE__.".d"`-type directory would be appropriate?

lib/perl/Genome/Utility/Test.pm
mkiwala: Good point.  Maybe I should have left `get_test_dir` in the variant reporting framework, and used `data_dir_ok` here.

lib/perl/Genome/Model/Command/Define/ImportedReferenceSequence.t
mkiwala: The trouble is that we need these to be network accessible and I agree that fewer external dependencies is better.  I think that I would change this to a `__FILE__.".d"`-type directory here.

lib/perl/Genome/Utility/Test.pm
davidlmorton: My issue with data_dir_ok is mostly academic, but I don't like the idea of using an 'ok' test for things that aren't actually testing the code we're trying to test.  A 'use_ok' is borderline but I'm alright with it since it actually loads the module we're trying to test.  I prefer to use the `note` function for things like this.

tmooney: I would agree that a lot of our testsuite is a long way from being "unit tests"!

Would it make sense to have an `assert` in `Genome::Utility::Test` that's like `ok`, but is basically:
```perl
sub assert {
  my ($test, $msg) = @_;
  if ($test) { 
    note $msg;
  } else {
    die $msg;
  }
}
```
It could be used for things that should be true, but aren't really tests of the code?  The overall effect will be the same--the test fails--but the specific problem won't get parsed and reported in test reports as a failing testcase.

mkiwala:
Ok -- I believe that I've satisfied all of the remaining requests by creating `assert()` in `Genome::Utility::Test` and undoing the contentious `get_test_dir()` stuff in favor using a `__FILE__` relative test-data directory.

Please let me know what you think now.

tmooney:
+1.

In the past when tests have required something that is not always available (historically this was a 64-bit architecture) they used `skip_all` when the environmental prerequisites were not met.  Would that be appropriate here, or do we think it's better to `die` in these circumstances?  (Should we change other `skip_all`s (not in this PR!)?)

At a more theoretical level, I'm still not sure I like the idea of our `.t` tests spawning LSF jobs and no longer working from the command-line.  (I know I'm a lot less likely to run this test!)

mkiwala:
> In the past when tests have required something that is not always available (historically this was a 64-bit architecture) they used skip_all when the environmental prerequisites were not met. Would that be appropriate here, or do we think it's better to die in these circumstances? (Should we change other skip_alls (not in this PR!)?)

That is good question.  I don't have strong preference right now.

mkiwala:
> At a more theoretical level, I'm still not sure I like the idea of our .t tests spawning LSF jobs and no longer working from the command-line. (I know I'm a lot less likely to run this test!)

I think that, ideally, unit tests should have light-weight dependencies and that LSF is not a light-weight dependency.  So I agree that ideally our unit tests should not depend on LSF.

I think that @davidlmorton had suggested this morning supporting a `inline` setting for `XGENOME_WORKFLOW_BUILDER_BACKEND`.  I'm not quite sure where he was going with that idea, but perhaps it could be used as a way to run limited workflows without depending on PTero or LSF.

mkiwala:
> I think that @davidlmorton had suggested this morning supporting a inline setting for XGENOME_WORKFLOW_BUILDER_BACKEND. I'm not quite sure where he was going with that idea, but perhaps it could be used as a way to run limited workflows without depending on PTero or LSF.

Morton definitely making the inline build thing work by implementing it in WorkflowBuilder.

PR #1241, opened 2016-02-05
Ensure allocations are cleaned up when genome-test-env exits

brummett:
Looks like UR_DBI_NO_COMMIT was still true when the allocation removal ran, so
calling delete() on allocations did not remove the files.

### DIFF
--- a/bin/genome-env
+++ b/bin/genome-env
@@ -494,7 +494,7 @@ function setup_test_db {
 # from the test database, then deletes the test database
 function cleanup_test_db {
     echo "=> Removing disk allocations created during $THIS..."
-    genome model admin remove-disk-allocations-from-testdb
+    UR_DBI_NO_COMMIT=0 genome model admin remove-disk-allocations-from-testdb
     test-db delete database $TESTDBSERVER_DB_NAME > /dev/null
 }
 


tmooney:
+1.

It seems sketchy to selectively turn of no-commit, but I suppose we're only in this situation if someone else did the same thing previously!

mkiwala:
We could avoid selectively turning off no-commit here if we can find another way to approach #1240.  So maybe we should hold-off on merging this until we're sure that we have to selectively turn off UR_DBI_NO_COMMIT in #1240.

davidlmorton:
Is the sketchy-ness reduced somewhat by the fact that this is the last thing we do before deleting the database?

tmooney:
In a world where someone might turn off no commit (at their prompt or otherwise), I think this change may be proper regardless of what comes out of #1240.  Just seems sketchy :smile:

brummett:
It gets in this state when:
a) test environment is started with genome-test-env (or genome-env -t)
b) Something within is run with UR_DBI_NO_COMMIT=0

Normally "a" would turn on no-commit, so no new allocations would get created.  The only way any new allocations would be in the test DB in the first place is if "b" were true.

We need some way to ensure that any allocations created within the genome-env are removed before the test DB is removed.  Otherwise, they become orphaned.

davidlmorton:
+1

PR #1242, opened 2016-02-09
Check for directory existence each time before calling mkdir.

tmooney:
This partially reverts 0a28189020054af452b08c8d496234b1734bf403 from PR #809.  The initial check does not seem to be sufficient to ensure the volume is mounted.


mkiwala:
+1 :koala:

ebelter:
+1

PR #1243, opened 2016-02-09
Disk Allocation Unarchive

ebelter:
Allow unarchive of imported instrument data.
Test updates


lib/perl/Genome/Disk/Command/Allocation/Unarchive.t
```diff
@@ -2,190 +2,213 @@
...
+
+    Sub::Install::reinstall_sub({
+            code => sub { return $volume->mount_path },
+            into => 'Genome::Disk::Volume',
+            as => 'active_volume_prefix',
+        });
+
+};
+
+subtest 'unarchive allocation filas without correct owner' => sub{
```
tmooney: The `a` in `fails` has slipped a few positions.

lib/perl/Genome/Disk/Command/Allocation/Unarchive.t
```diff
@@ -2,190 +2,213 @@
...
+    my $cmd = Genome::Disk::Command::Allocation::Unarchive->create(
+        allocations => [$allocation],
+        analysis_project => $analysis_project,
+    );
+    ok($cmd, 'created unarchive command');
+    throws_ok(sub { $cmd->execute }, qr/currently not handled/, 'command fails with unsupported owner');
+
+};
+
+subtest 'unarchive allocation programatically' => sub{
```
tmooney: `programmatically` has lost an `m`.

ebelter:
CI test this. BATS passed for me locally.

lib/perl/Genome/Disk/Command/Allocation/Unarchive.pm
```diff
@@ -59,6 +59,11 @@ sub _link_allocation_to_analysis_project {
         unless($owner->model->analysis_project) {
             $self->fatal_message('No analysis project set on model for build %s.  Please use `genome analysis-project add-model` to correct this.', $owner->__display_name__);
         }
+    } elsif ( $owner->isa('Genome::InstrumentData::Imported') ) {
+        $self->analysis_project->add_analysis_project_bridge(
```
tmooney: There's a unique constraint on this bridge table, so we'll need to make sure we don't try to add the same instrument data to the same analysis project twice.  (`Genome::SoftwareResult::User` is an `ObjectWithLockedConstruction`, so it will find the existing object when `add_user` is called a second time, but this bridge entity is not.)

tmooney:
+1 :tiger2:

PR #1245, opened 2016-02-10
Complete the removal of RefAlign PP subclasses.

tmooney:
This completes the transition started in #1239.

This should not be merged until all running snapshots are newer than `genome-3681`.  Prior to merging a data migration will be required.


ebelter:
+1

susannasiebert:
+1

PR #1246, opened 2016-02-10
Remove Obsolete RefAlign ProcessingProfile Parameters

tmooney:
They're hold-overs from `maq` or early versions of the pipeline.


mkiwala:
+1 :koala: -- assuming that all of these parameters really are obsolete.

ebelter:
+1 :fire:

PR #1249, opened 2016-02-10
Out of process imported ref seq

mkiwala:
This PR depends on making the Imported Reference Sequence test independent of inline dispatch, so #1240 should be merged before merging this PR.


lib/perl/Genome/Model/Command/Define/ImportedReferenceSequence.pm
```diff
@@ -359,29 +351,17 @@ sub _create_build {
...
-        }
+    if($build->start) {
+        UR::Context->commit or die "Failed to commit.";
+        $self->status_message('Started build.');
     } else {
         $self->error_message("Failed to start build " . $build->build_id . " for model " . $model->genome_model_id . ".");
         return;
     }
 
+    $build->wait;
```
tmooney: Perhaps waiting should be optional if the user doesn't want to wait?  (That was the effect of overriding the server dispatch previously.)

tmooney:
+1 :haircut:

PR #1250, opened 2016-02-10
Use GMT:Joinx::SnvConcordanceByQuality in RefAlign Summary Report.

tmooney:
This quiets this warning:

```
Use of inherited AUTOLOAD for non-method Genome::Model::Tools::Joinx::SnvConcordanceByQuality::parse_results_file() is deprecated at .../Genome/Model/ReferenceAlignment/Report/Summary.pm line 220.
```

### DIFF
--- a/lib/perl/Genome/Model/ReferenceAlignment/Report/Summary.pm
+++ b/lib/perl/Genome/Model/ReferenceAlignment/Report/Summary.pm
@@ -7,6 +7,7 @@ use strict;
 use warnings;
 
 use Genome;
+use Genome::Model::Tools::Joinx::SnvConcordanceByQuality;
 
 use IO::String;
 use Template;


mkiwala:
Given the choice between `use` and `require`, I typically choose `use` because it seems like more idiomatic Perl.  Does `use` not work in this case?

Either way:
+1 :koala:

tmooney:
I think `use` would work splendidly in this case.  I chose `require` since I thought it was more similar to what was previously happening, but I have no objection to changing it.

dufeiyu:
+1  even though I prefer "use" a little bit more.

ebelter:
Aye! Me, too, have switched alliance from require to use.

tmooney:
By popular demand: Changed to `use`!

PR #1253, opened 2016-02-10
SomVar CreateReport uses Build Accessors for BAMs and Reference

tmooney:
The implementation in the build class is designed to handle both `RnaSeq` and `ReferenceAlignment` builds.


ebelter:
+1

chrisamiller:
+1 :hamburger:

PR #1255, opened 2016-02-11
Remove Old Alignment Accessors

tmooney:
An abridged history of `sub libraries` (from 176611fdfd9e3818d5c51e827f17cee6c6a58580):
> The "temp hack" was added in bd502ac71a8a613e95af57f429d98b1d948abc76
on 14 July 2008 to the original `Genome::Model`.  It was then migrated to
`Genome::Model::ShortRead` before moving to its most recent location here
in `Genome::Model::ReferenceAlignment`.

As a bonus, this PR also removes one method from `Genome::Model::Build::RnaSeq`.


mkiwala:
Removed a bunch of functions and tests passed.  Looks good to me!
+1 :koala:

brummett:
+1

PR #1256, opened 2016-02-11
Ensure that qc results are unique

susannasiebert:


tmooney:
+1 :performing_arts:

mkiwala:
Why is it that calling `$build->results` could return the same qc result more than once?

susannasiebert:
Builds results are retrieved via the `Genome::SoftwareResult::User` bridge table. A software result can have multiple associations with a build if the `label` is different.

mkiwala:
Maybe we change how `results()` delegates to `Genome::SoftwareResult::User` to also filter for unique results.

@susannasiebert mentioned offline that it's possible something depends on the current behavior of `results()` and that changing the behavior now could lead to errors in other logic which depends on this behavior.  @tmooney do you think this is likely?

tmooney:
It doesn't seem likely.  In a situation where I can imagine wanting the same result multiple times I'd also want the labels (which requires using `result_users` instead of `results`).

This is currently implemented as a UR `via`/`to` property.  I'm not sure if there's a way to keep it query-able in UR while deduplicating the results.

`Genome::Process` addressed this issue by implementing `unique_results` in 2f044a66b406712efe5f4acc5bea3ec87de6f0e1.

In the case of particular build types, always using the accessors listed under `has_output` will get you unique results, but this tool has to work with multiple types of builds, not all of which are aware they have QC results.

brummett:
Looks like this happens because the 'via' has multiple rows that match each thing the 'to' points to.  There's no current way to ask UR to do the filtering.  It'd be easy to add an 'is_distinct' meta-property that would tell UR's via/to property resolver to apply filtering for you.

Ideally, UR would be able to do a 'select distinct' query, but that would be a _lot_ more work.

susannasiebert:
I like the `is_distinct` solution that @brummett suggested but it seems out of the scope of this PR.

brummett:
You get the same effect by changing Genome::Model::Build's ```results``` property from
```via => 'result_users'```
to
```via => 'result_associations'```

One is an alias of the other.  The difference is that ```result_associations``` has 'reverse_as' metadata, but 'result_users' doesn't, so the bridging code in the ```results``` accessor makes a different choice at [UR::Object::Type::AccessorWriter line 362](https://github.com/genome/UR/blob/master/lib/UR/Object/Type/AccessorWriter.pm#L362)

susannasiebert:
I like that suggestion. Closing this PR.

PR #1257, opened 2016-02-11
AnalysisProject View: Display Environment Config if One Exists.

tmooney:
This provides notice if `genome analysis-project add-environment-file` has been used for a given AnalysisProject.

### DIFF
--- a/lib/perl/Genome/Config/AnalysisProject/Command/View.pm
+++ b/lib/perl/Genome/Config/AnalysisProject/Command/View.pm
@@ -9,6 +9,7 @@ use Genome::Utility::Text qw(justify);
 use List::Util qw(max sum);
 use YAML::Syck qw();
 use List::MoreUtils qw(uniq);
+use Try::Tiny qw(try catch);
 
 class Genome::Config::AnalysisProject::Command::View {
     is => ['Genome::Command::Viewer' ],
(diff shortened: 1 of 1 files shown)


mkiwala:
+1 :koala:

lib/perl/Genome/Config/AnalysisProject/Command/View.pm
```diff
@@ -371,6 +372,35 @@ sub _color_status {
...
+        print $handle "\n";
+    }
+}
+
+sub _write_environment_config_file {
+    my ($self, $handle, $environment_config) = @_;
+
+    my $f = File::Spec->join($environment_config, Genome::Config::config_subpath);
+    unless (-r $f) {
+        $self->warning_message('Could not read environment config <%s>.', $f);
```
ebelter: Since read_file calls validate_file_for reading, could this be written as a try catch?

use Try::Tiny 'try';
my $content = try {
  Genome::Sys->read_file($f)
}
catch {
   $self->warning_message('Could not read environment config <%s>.', $f);
}
...

lib/perl/Genome/Config/AnalysisProject/Command/View.pm
```diff
@@ -371,6 +372,35 @@ sub _color_status {
...
+
+        if ($self->config eq 'verbose' or $self->config eq 'parsed') {
+            $self->_write_environment_config_file($handle, $environment_config);
+        }
+
+        print $handle "\n";
+    }
+}
+
+sub _write_environment_config_file {
```
ebelter: This method name maybe confusing, as it indicates that a file is being written. Maybe adding 'contents' or the like to the end could provide some clarity?

lib/perl/Genome/Config/AnalysisProject/Command/View.pm
tmooney: It could!  And now it is :smile:

lib/perl/Genome/Config/AnalysisProject/Command/View.pm
tmooney: I've swapped `file` for `contents`.

ebelter:
+1 :vhs:

PR #1258, opened 2016-02-15
added mm9 snps to examples

chrisamiller:


tmooney:
+1, new example does point to a feature-list that claims to be one-based!

PR #1259, opened 2016-02-16
SomaticValidation Workflow is Instantiated with WorkflowBuilder.

tmooney:
This change is of little consequence for the moment (since `Build` converts this back into XML regardless), but it is a step on the path to removing `Workflow`!


mkiwala:
+1 :koala:

brummett:
+1

PR #1260, opened 2016-02-16
Remove Old "Default Profile"s

tmooney:
These defaults are not up-to-date, and we now use the `ConfigureQueuedInstrumentData` command to decide what profiles to use in most cases.

### DIFF
--- a/lib/perl/Genome/InstrumentData/Solexa.pm
+++ b/lib/perl/Genome/InstrumentData/Solexa.pm
@@ -1253,10 +1253,9 @@ sub get_default_alignment_results {
 
     unless(@alignment_results) {
         # Get alignment results for this inst data and the default aligner name
-        my $pp = Genome::ProcessingProfile::ReferenceAlignment->default_profile;
         @alignment_results = Genome::InstrumentData::AlignmentResult->get(
             instrument_data_id => $self->id,
-            aligner_name       => $pp->read_aligner_name,
+            aligner_name       => 'bwa',
         );
     }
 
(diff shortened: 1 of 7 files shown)


lib/perl/Genome/InstrumentData/Solexa.pm
```diff
@@ -1253,10 +1253,9 @@ sub get_default_alignment_results {
 
     unless(@alignment_results) {
         # Get alignment results for this inst data and the default aligner name
-        my $pp = Genome::ProcessingProfile::ReferenceAlignment->default_profile;
         @alignment_results = Genome::InstrumentData::AlignmentResult->get(
             instrument_data_id => $self->id,
-            aligner_name       => $pp->read_aligner_name,
+            aligner_name       => 'bwa',
```
dufeiyu: just feel a little odd for this hard coded "bwa" here.  Is the old alignment QC metric still being used after the new QC config is in place ? If so, probably it's better to get aligner_name from a processing_profile that is listed in a "default" analysis menu item like "Human Reference Alignment".

tmooney: I agree it's weird to have the hard-coded "bwa" in here.  It's intended to preserve the value from the old processing-profile `2635769`.

All the code in this method inside this `unless` block is an ugly heuristic for old data not in analysis projects, so I believe it should keep the old value (`bwa`) regardless of what the new default might be.

My preference would be to delete this `unless` block entirely!  LIMS developed this code and is its primary user, so it's really up to them, though.

dufeiyu: yeah, I agree that this "unless" block becomes obsolete these days and should be deleted since all of our instrument data now are associated to AnP. It's ok to leave it for now just in case.

dufeiyu:
+1

ebelter:
+1

PR #1261, opened 2016-02-16
Restore the known_sites Magic to SomaticValidation

tmooney:
`known_sites` is an `is_many` property, and the automatic generation of the id property is not set to happen for is_many inputs.


dufeiyu:
+1  since tests pass now and please merge this asap.

PR #1262, opened 2016-02-17
Change G::M::B results to use result_associations instead of result_users

susannasiebert:
This will ensure that calling results returns a unique list.


brummett:
+1

This change also means calling ```$build->results``` does 2 queries instead of n+1

tmooney:
+1 :arrow_double_up:

PR #1264, opened 2016-02-18
ReferenceAlignment: Use user relationships to find alignments.

tmooney:
This is far more direct than assembling all the aligner params and querying!


ebelter:
+1

acoffman:
+1

PR #1265, opened 2016-02-18
Remove disused event accessors from ReferenceAlignment Model.

tmooney:
The build would've been a more appropriate place for many of these, but we're about to remove the events, anyway!


ebelter:
+1

acoffman:
+1

PR #1266, opened 2016-02-19
Inline workflowbuilder DAG execution

davidlmorton:


mkiwala:
+1 :koala:

susannasiebert:
+1

lib/perl/Genome/Test/Command/RunPteroIntegrationTests.pm
```diff
@@ -2,14 +2,18 @@ package Genome::Test::Command::RunPteroIntegrationTests;
 
 use strict;
 use warnings FATAL => 'all';
-use Genome;
+use above "Genome";
```
tmooney: It's unusual to see `use above` in a `.pm` rather than a Perl script that's using Genome.  Is there something particular about this module that requires it?

davidlmorton: It doesn't strictly require it, but it is designed to run tests (it is just separate from the .t files so it isn't run as a unit-test).  If you want to run these tests against a checkout it is handy to not have to manipulate your PERL5LIB.

tmooney: It seems like the caller should be the thing that does `use above`.  By the time we reach this `use above`, we're already using the version of this module contained in `PERL5LIB` and not the version in the desired checkout.

davidlmorton: That's fair, I'll remove it.

tmooney:
+1; I think it is also safe to remove `etc/genome/spec/ptero_log_directory.yaml` since 31117f4 removes all uses of it.

davidlmorton:
In conjunction with merging this PR, we can (if we choose) set XGENOME_WORKFLOW_BUILDER_BACKEND=inline for both the quick and full test-suite (but not the model-tests).

PR #1267, opened 2016-02-22
don't copy config if to and from projects are the same

chrisamiller:


chrisamiller:
Guess how i figured out you could do this... :rage3:

lib/perl/Genome/Config/AnalysisProject/Command/CopyConfig.pm
```diff
@@ -50,6 +50,11 @@ sub execute {
     my $from_project = $self->from_project;
     my $to_project = $self->to_project;
 
+    if($from_project->id eq $to_project->id){
+        $self->error_message("To and From projects cannot be the same");
+        return 0;
```
tmooney: I think we usually `return;` rather than `return 0;` in this circumstance.

tmooney:
+1, yeah, that's a bummer!

chrisamiller:
Sure thing - updated.

brummett:
+1

PR #1269, opened 2016-02-23
AnP Replace Model Uses Config Profile

ebelter:
Extract the process model get/create from CQID into config profile so that both CQID and the replace model command can use it.

### DIFF
--- a/lib/perl/Genome/Config/AnalysisProject/Command/ReplaceModel.t
+++ b/lib/perl/Genome/Config/AnalysisProject/Command/ReplaceModel.t
@@ -4,18 +4,18 @@ use strict;
 use warnings;
 
 BEGIN {
+    $SIG{__DIE__} = sub{ Carp::confess(@_); };
     $ENV{UR_DBI_NO_COMMIT} = 1;
     $ENV{UR_COMMAND_DUMP_STATUS_MESSAGES} = 1;
     $ENV{UR_COMMAND_DUMP_DEBUG_MESSAGES} = 1;
 };
 
-use Test::Exception;
-use Test::More tests => 4;
-
 use above 'Genome';
-#use Genome::Test::Factory::AnalysisProject;
 use Genome::Test::Factory::DiskAllocation;
 use Genome::Test::Factory::InstrumentData::Solexa;
+
+use Test::Exception;
+use Test::More tests => 4;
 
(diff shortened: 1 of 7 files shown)


ebelter:
As suggested by @tmooney  in https://github.com/genome/genome/pull/1215.

lib/perl/Genome/Config/AnalysisProject/Command/ReplaceModel.t
```diff
@@ -5,25 +5,31 @@ use warnings;
...
+    $ENV{UR_COMMAND_DUMP_STATUS_MESSAGES} = 1;
+    $ENV{UR_COMMAND_DUMP_DEBUG_MESSAGES} = 1;
 };
 
 use Test::Exception;
 use Test::More tests => 4;
 
 use above 'Genome';
-use Genome::Test::Factory::AnalysisProject;
+#use Genome::Test::Factory::AnalysisProject;
```
tmooney: Looks like this line could be removed.

lib/perl/Genome/Config/Profile/Item.pm
```diff
@@ -183,7 +183,7 @@ sub _is_created {
 sub is_current {
     my $self = shift;
 
-    return $self->status ne 'disabled' and $self->analysis_project->is_current;
+    return $self->status ne 'disabled' && $self->analysis_project->is_current;
```
tmooney: Good catch!

tmooney:
+1.

Originally `Genome::Config::Profile` was supposed to be "dumb" rather than knowing what to do with all the configuration, but it seems like a reasonable place to put it since it can't live in CQID anymore.

ebelter:
@tmooney Interesting. I chose Profile because it seemed like a common component. Although it wasn't completely clear where it should go. If you have a suggestion, I would be up for thinking about it.

tmooney:
@ebelter One option might be a new class called `Genome::Config::Profile::Interpreter` or somesuch?  But this seems fine to me as-is, as long as we don't plan on more functionality accreting here in the future.  (Maybe @acoffman has an opinion?)

acoffman:
This looks good to me. As @tmooney mentioned, `Config::Profile` was initially intended to be a dumb class; however, this seems like a legitimate reason for moving this functionality there.

+1

PR #1272, opened 2016-02-24
Remove disused "CompleteGenomics" GMTs.

tmooney:
The underlying binaries required to use these are no longer installed.


acoffman:
+1

susannasiebert:
+1

PR #1276, opened 2016-02-24
Remove unfinished "Sword" GMTs.

tmooney:
This was started in 2013, but it looks like it was never fully implemented.


acoffman:
+2

dufeiyu:
+1

PR #1278, opened 2016-02-25
AlignmentResult uses the AlignmentSummary GMT.

tmooney:
(The alternative here would be to remove the `AlignmentSummary` GMT tree.)


dufeiyu:
+1

susannasiebert:
+1

PR #1279, opened 2016-02-25
Remove Genome::UpdateHistory.

tmooney:
It points to an Oracle table but uses the Postgres datasource, so it seems unlikely anything uses it.


brummett:
+1

susannasiebert:
+1

PR #1280, opened 2016-02-25
Remove Remnants of old Perl-based DruggableGene.

tmooney:
The current version is at [genome/dgi-db](https://github.com/genome/dgi-db).


susannasiebert:
+1

brummett:
+1

PR #1281, opened 2016-02-25
Configure permissions on tempdir when using LSF location.

tmooney:
This is intended to allow for easier inspection and debugging of in-progress workflow steps.


ebelter:
+1

brummett:
+1

PR #1283, opened 2016-02-26
Build volume lister command

brummett:
Another attempt at a lister for allocations/volumes connected to a build

This one inherits from UR::Object::Command::List, but it's complicated by the fact that --filter applies to the builds you're interested in inspecting, it collects Allocation Set objects related to those builds, then --show applies to those Sets.

### DIFF
--- a/lib/perl/Genome/Model/Build.pm
+++ b/lib/perl/Genome/Model/Build.pm
@@ -75,7 +75,8 @@ class Genome::Model::Build {
         subject => { is => 'Genome::Subject', via => 'model' },
         processing_profile => { is => 'Genome::ProcessingProfile', via => 'model' },
         run_by => { is => 'Text', doc => 'the user whose build is run' },
-        status => { is => 'Text', doc => 'the current status of the build' },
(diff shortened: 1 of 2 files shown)


tmooney:
It'd be nice to quiet these warnings:
```
Subroutine Genome::Model::Build::Command::ListVolumes::subject_class_name redefined at .../lib/perl/Genome/Model/Build/Command/ListVolumes.pm line 72.
Subroutine Genome::Model::Build::Command::ListVolumes::subject_class_name redefined at .../lib/perl/Genome/Model/Build/Command/ListVolumes.pm line 63.
Subroutine Genome::Model::Build::Command::ListVolumes::subject_class_name redefined at .../lib/perl/Genome/Model/Build/Command/ListVolumes.pm line 72.
```

+1, otherwise--this was pretty fast and four queries, even when listing volumes for an entire Analysis Project!

brummett:
Oops.  I've been running it with dump-sql turned on and didn't notice the warnings.  Added another commit to fix that

PR #1284, opened 2016-02-26
Epitope length needs to be of Integer data type instead of Text

susannasiebert:
Jasreet tried to run the epitope predicition command with epitope_length set to "8,9,10" expecting this to be an is_many property. Because the data type was set to Text she was able to successfully dispatch the workflow but it failed down the line because the underlying tool expects this to be a single integer. Setting the data type to Integer would throw an error if this parameter was set to something like the above.

Down the line we might want to consider making this an is_many property and running this workflow as a parallel by.


brummett:
+1

acoffman:
+1

PR #1285, opened 2016-02-29
Qc build metric report

jasonwalker80:
Additional option to output metrics as TSV.  Two profiles, one for WGS and one for exome, that determine the output columns for each config type.


lib/perl/Genome/Qc/Command/BuildMetrics.pm
```diff
@@ -118,4 +144,138 @@ sub _calculate_haploid_coverage {
...
+
+    # Initialize as undefined
+    $data->{'microarray_chipmix'} = undef;
+    $data->{'microarray_freemix'} = undef;
+    
+    my $build = Genome::Model::Build->get($qc_metric_result->{build_id});
+
+    my %build_vbid_results;
+    for my $vbid (grep { $_->class eq 'Genome::InstrumentData::VerifyBamIdResult'} $build->results) {
+        # Not sure why but the results are not unique.  Duplicates are returned from build.
```
susannasiebert: There has been a recent change to `$build->results` so that only unique results are returned. See https://github.com/genome/genome/pull/1262

susannasiebert: To go into a bit more detail of why `results` used to be non-unique: There can be multiple `result_users` associations between one software result and one builds using different labels. Changing it to look at  `result_associations` only gives unique results.

jasonwalker80: The comment and dedup via hash was made before the duplicate result issue was known.  A new commit was made to remove the comment and the verifyBamId dedup.

lib/perl/Genome/Qc/Command/BuildMetrics.pm
```diff
@@ -118,4 +144,135 @@ sub _calculate_haploid_coverage {
...
+
+#These results only exist from running SomaticValidation models on the same instrument data
+sub _add_microarray_vbid_result_metrics_to_hash_ref {
+    my $self = shift;
+    my $qc_metric_result = shift;
+    my $data = shift;
+
+    my $build = Genome::Model::Build->get($qc_metric_result->{build_id});
+
+    my $vbid = grep { $_->class eq 'Genome::InstrumentData::VerifyBamIdResult'} $build->results;
```
tmooney: This should be `my ($vbid) = ...` or else it will get the number of matches instead of the result itself.

jasonwalker80: Thanks @tmooney, fixed.

lib/perl/Genome/Qc/Command/BuildMetrics.pm
```diff
@@ -118,4 +144,135 @@ sub _calculate_haploid_coverage {
...
+        for my $qc_metric_result(@$qc_metric_results) {
+            my $data = $self->_base_hash_ref_for_qc_metric_result($qc_metric_result);
+            $data->{'mean_target_coverage'} = $qc_metric_result->{MEAN_TARGET_COVERAGE};
+            $data->{'pct_usable_bases_on_target'} = $qc_metric_result->{PCT_USABLE_BASES_ON_TARGET};
+            $data->{'pct_target_bases_20x'} = $qc_metric_result->{PCT_TARGET_BASES_20X};
+            $data->{'pct_exc_off_target'} = $qc_metric_result->{PCT_EXC_OFF_TARGET};
+            $data->{'pct_exc_dupe'} = $qc_metric_result->{PCT_EXC_DUPE};
+            $data->{'pct_exc_overlap'} = $qc_metric_result->{PCT_EXC_OVERLAP};
+            $data->{'pct_exc_mapq'} = $qc_metric_result->{PCT_EXC_MAPQ};
+            $data->{'pct_exc_baseq'} = $qc_metric_result->{PCT_EXC_BASEQ};
```
tmooney: These could be:
```perl
for my $column (@headers) {
  $data->{$column} = $qc_metric_result->{uc $column};
}
```

tmooney:
The TSV case seems a lot more complicated than the YAML case.

Is the work that it's doing making assumptions about what metrics are included in the QC configuration?  If so that seems like it could cause trouble when we decide to change the default configurations.

lib/perl/Genome/Qc/Command/BuildMetrics.pm
jasonwalker80: It's not quite that simple, but yes all the exome metric hash keys do happen to be the uppercase of the column header.  I made a change to reduce the redundancy.

jasonwalker80:
The tsv IS more complicated.  This commit is intended to replace a perl script that parsed the YAML and did the same thing for the QC group.  I'd prefer not having to maintain that custom perl script in my personal bin.  

It's true that a change in configuration could break this report function.  Any changes in production QC would have to be handled another way or concurrent changes made to this command to support it.  For instance, Aye requested that MEAN_INSERT_SIZE be added as a metric in the TSV output.  It's available as a metric for each library but not reported by default in TSV format.

tmooney:
+1; a QC Config change should just result in `NA`s, so I'm not too worried about it.

jasonwalker80:
@susannasiebert additional comments or suggestions?

susannasiebert:
Looks good to me. +1

PR #1286, opened 2016-02-29
Consolidate GMT TestHelpers into Genome::Utility::Test.

tmooney:
There are a bunch of TestHelpers modules out there.  This removes two of them; more could probably be at least partially consolidated as well.


brummett:
Sounds good.
+1

acoffman:
+1

PR #1287, opened 2016-03-01
Don't die in HeartbeatCheck if no PIDs Found.

tmooney:
This is normal for newly started jobs, so report it and move along.


acoffman:
+1

jeldred:
+1 :shipit:  this seems wholly benign and appropriate!

PR #1288, opened 2016-03-01
WIP - overwrite SSG _compare_output_files

susannasiebert:
This customizes the build diffing during model tests for SSG to compare the merged alignment result bams, the haplotype callers results, and the qc metrics. `_compare_haplotype_caller_results` and `_compare_qc_metrics` are untested.


tmooney:
I'm thinking of writing `genome model build checkout` and implementing it for SSG.  Then this PR mostly becomes a matter of pointing the normal diff logic at checked-out copies of the builds instead of the build directories themselves.  (Plus then users can get a directory of all the results on-demand... but one whose symlinks we never have to worry about updating.)

susannasiebert:
I like that idea but I'm not a huge fan of the `checkout` naming convention. For processes we have a `symlink-results` command. Maybe we should use the same command name here as well?

tmooney:
`symlink-results` sounds good to me--it's nice to be consistent with what's already there!

tmooney:
I've opened #1300 with the `symlink-results`-based approach.

tmooney:
I'm hoping #1300 is the way forward and will close this PR.  If that's not true, we can always reopen this one!

PR #1289, opened 2016-03-02
Migrate ReferenceAlignment Parameter Specification to the Model Class

tmooney:
Now that all the functionality has been stripped out of the RefAlign processing profile, we don't need to maintain a separate class for it.


acoffman:
+1

susannasiebert:
+1

PR #1293, opened 2016-03-04
Common somatic interface to the tumor/normal samples.

jasonwalker80:
This was inspired by resolving a long standing JIRA issue, APIPE-2849.  However, @acoffman  and @susannasiebert  this is a very similar problem to how ClinSeq interfaces with somatic model types.


susannasiebert:
+1 once tests pass.

acoffman:
+1

chrisamiller:
+1 :last_quarter_moon_with_face:

tmooney:
+1.  If these common accessors are going to be used from within a build process, it would be good to also add them to the build classes.  (Since the samples are resolved via inputs, it is possible that the build being used as an input to the given running build has a different value than the current one on the model.)

PR #1294, opened 2016-03-07
Grch38 support for alts (experimental)

ernfrid:
These are small changes in and around aligning to the alts using BWA mem. I added a test speedseq version that utilizes a newer version of Sambamba and BWA so that we can align to both the primary reference and the alts for Build38. This requires a minimum BWA version of 0.7.11 and an additional file (provided for specific references in bwakit). I added this extra file to reference sequence 8741b5363c634ee19fad676157a132de. The test version of speedseq uses BWA mem 0.7.12. See https://github.com/lh3/bwa/blob/master/README-alt.md for more information on BWA behavior and requirements.

This may not be how we want to proceed, but I wanted to put this forwards so that we could discuss the proper way to proceed. We have been testing in CCDG informatics using these changes and the alignments appear to be working as intended.

### DIFF
--- a/lib/perl/Genome/InstrumentData/AlignmentResult/ReliesOnBwa.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult/ReliesOnBwa.pm
@@ -23,7 +23,7 @@ sub prepare_reference_sequence_index {
 
     Genome::Sys->create_symlink($refindex->reference_build->get_sequence_dictionary('sam'), File::Spec->join($staging_dir, 'all_sequences.dict'));
 
-    if (-e File::Spec->join($staging_dir, 'all_sequences.fa.alt')) {
+    if (-e $refindex->reference_build->full_consensus_path('fa.alt')) {
         # An alt file exists so BWA-mem post 0.7.11 can perform alt-aware mapping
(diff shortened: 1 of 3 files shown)


tmooney:
What should we do if a user requests to use one of these newer BWA versions but the alternate file isn't there?  If adding it later could change results, then we'd need to invalidate all results for that reference/aligner version combo at such time as we do add it.

ernfrid:
@tmooney - yeah, that would be a legitimate problem.

ernfrid:
@tmooney - What would our other options be?

Right now, I'm thinking there are only two real options

1. Leave it as is and accept the risk that missing the alts at reference creation will require invalidating results. I think this simpler than the other option, but less clean.
2. Introduce some other mechanism to store the .alt file outside of the reference sequence and then bring it in during BWA index creation. I don't know if such an entity already exists. It seems it would need to know if a file existed for the reference in question. If it does exist then this may be an easy solution.

tmooney:
it occurs to me that adding the alt file later won't regenerate the index result, so it'd be ignored and not used.  Since we have to explicitly generate a new index result, could then either:

1. invalidate the old one (at which point we can find any existing alignments that used it)
2. somehow make it an input to the index result so the results are distinguishable.

It'd be nice to at least have a formal command to add the alternate file, even if all it does is add it to the reference build and reallocate.  That command could also be the thing that does number one above.

PR #1295, opened 2016-03-07
Get E-Mail Address from User Object

tmooney:
Since we now have users spanning multiple domains with possibly inconsistent usernames, we can't assume the `username` plus the configured `email_domain` equals the desired address.

### DIFF
--- a/lib/perl/Genome/Model/Tools/Analysis/LaneQc/CompareSnpsBuildLanes.pm
+++ b/lib/perl/Genome/Model/Tools/Analysis/LaneQc/CompareSnpsBuildLanes.pm
@@ -36,7 +36,7 @@ sub execute {
     my $self=shift;
     my $build_id = "";
     my $dir = $self->analysis_dir;
-    my $user = getlogin || getpwuid($<); #get current user name
+    my $user = Genome::Sys->current_user;
     my $sample_name ="";
     my $genotype_file ="";	 
     my $wgs_model_id = $self->model_id;
@@ -123,7 +123,8 @@ sub execute {
 
 # find reference sequences
     my $reference_file=$model->reference_sequence_build->full_consensus_path('fa') ;
-    print "reference: $reference_file User: $user\n";
+    my $username = $user->username;
+    print "reference: $reference_file User: $username\n";
     for my $data (@instrument_data) {
         my @alignments = $build->alignment_results_for_instrument_data($data);
         for my $alignment (@alignments) {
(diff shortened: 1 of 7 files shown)


lib/perl/Genome/Model/Tools/Analysis/LaneQc/CompareSnpsBuildLanes.pm
```diff
@@ -143,8 +144,8 @@ sub execute {
 samtools pileup -vc -f $reference_file $alignment_file | perl -pe '\@F = split /\\t/; \\\$_=q{} unless(\\\$F[7] > 2);' > $dir/$lane_name.var
 gmt analysis lane-qc compare-snps --genotype-file $genotype_file --variant-file $dir/$lane_name.var > $dir/$lane_name.var.compare_snps
 COMMANDS
-                my $email_domain = Genome::Config::get('email_domain');
-                print `bsub -N -u $user\@$email_domain "$command"`;
+                my $email = $user->email;
+                print `bsub -N -u $email "$command"`;
```
wschroeder: I like the new abstraction.  Question on the bsub usage... Should probably avoid using backticks and leverage param-binding API instead (like IPC::Open3).  I see double quotes around the `$command`, and that might be under sufficient control up to this point that we know no one is -- within the command -- using double quotes, but it is legal for [email addresses](https://en.wikipedia.org/wiki/E-mail_address#RFC_specification) to have double-quoting and spaces, too: `"My Name"@example.com`.  Unlikely here, but hey.  I figure if we can't sanitize the incoming string, we should use Perl's built-in sanitizers where possible.

tmooney: Yeah... this code is old and busted, to be sure!  I'm reluctant to put too much time into fixing it until it survives <https://jira.gsc.wustl.edu/browse/APIPE-3367>.  If we did change it, I might use `Genome::Sys::LSF::bsub` (which itself eventually uses `IPC::System::Simple`).

A lot of the commands under `Genome::Model::Tools` are user-contributed and have been left relatively untouched (save for little fixes like this).  This one is from 2010 and was written by someone who left a few years ago; I'll be curious to check the usage log to see if anyone has still been running it.  (As a side note, the git history is redacted before 2012 on this public repository to remove anything we lack permission to share, but the older history is available from the internal server.)

ebelter:
+1 Nice! I did not expect that he emil was not gotten from the User object. But these are old!

brummett:
+1

PR #1296, opened 2016-03-07
Add a base for the SSG command tree.

tmooney:
This should fix tab-completion and the commands showing up in `genome model --help`.


brummett:
+2

PR #1297, opened 2016-03-07
Add Scheduled as valid build status.

tmooney:
It's very hard to start new builds without being able to Schedule them!


dufeiyu:
+2  tests should pass

tmooney:
Merging early so model tests have a chance :smile:

PR #1298, opened 2016-03-07
Production No Longer Hardcoded to "apipe-builder".

tmooney:
Now there's a `Genome::Sys::User::Role` named `production` when deciding to act.

* The default user for new `production` projects has been changed to `prod-builder`.
* `apipe-builder` remains the user with the power to manipulate others' builds.
* The old heuristic for LIMS to find non-project-based data (in `Genome::InstrumentData::Solexa`) is explicitly not being updated.

### DIFF
--- a/lib/perl/Genome/Config/AnalysisProject/Command/Create.pm
+++ b/lib/perl/Genome/Config/AnalysisProject/Command/Create.pm
@@ -100,7 +100,7 @@ sub _resolve_properties_for_environment {
         }
         when ('production') {
             return (
-                run_as => 'apipe-builder',
+                run_as => 'prod-builder',
                 is_cle => 0,
             );
         }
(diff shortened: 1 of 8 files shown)


wschroeder:
The same algorithm comes up several times:

    my $run_as_user = Genome::Sys::User->get(username => $self->run_as);
    if($run_as_user && $run_as_user->has_role_by_name('production')) {

We should be able to convert any

    my $temp_var = process1();
    # more processing
    if ($condition_based_on_temp_vars($temp_var, ...)) {


to

    if (complex_condition_function_is_true()) {


In this case, to reduce the repetition and extra lines of code, I propose:

    if (Genome::Sys::User->has_role_by_name($self->run_as, 'production')) {

tmooney:
I very nearly did that originally, but by creating`Genome::Sys->user_has_role` since there's an existing `Genome::Sys->current_user_has_role`.  I'll make that change.

(I could also see a good argument that most of the user-related subroutines in `Genome::Sys` should be moved down into `Genome::Sys::User`.  But that'd be a change for a different PR :smile:)

acoffman:
+1

brummett:
+1

PR #1299, opened 2016-03-08
ReferenceAlignment makes one BAM regardless of mixed sample assignment.

tmooney:
This fixes a regression found by the `refalign-germline-short` model test.

ReferenceAlignment should produce a single BAM regardless of how many different samples the instrument data came from.


acoffman:
+1 :goberserk:

dufeiyu:
+1

PR #1300, opened 2016-03-08
SingleSampleGenotype Can Be Checked Out; Uses Said Checkout for Diffs

tmooney:
* New command `genome model build symlink-results` (like `genome process symlink-results`) to create a build-directory like structure of results for a build...currently only implemented for SSG.
* SSG uses one of these build-directory like checkouts to do its diff-ing, since the diff logic is primarily based on directories of results.

(This PR is offered as an alternative to #1288.)

### DIFF
--- a/lib/perl/Genome/Model/Build/Command/SymlinkResults.pm
+++ b/lib/perl/Genome/Model/Build/Command/SymlinkResults.pm
@@ -7,7 +7,7 @@ use Genome;
 
 class Genome::Model::Build::Command::SymlinkResults {
     is => ['Command::V2'],
-    has => [
+    has_input => [
         build => {
             is => 'Genome::Model::Build',
             shell_args_position => 1,
@@ -18,6 +18,12 @@ class Genome::Model::Build::Command::SymlinkResults {
             shell_args_position => 2,
             doc => 'The desired location under which to make a directory of symlink(s)',
         }
+    ],
+    has_transient_output => [
+        output_directory => {
+            is => 'DirectoryPath',
+            doc => 'The directory of symlinks that was created',
+        },
     ],
     doc => 'Creates symlinks to the results of a build.',
 };
@@ -35,14 +41,8 @@ sub execute {
     my $build = $self->build;
     my $destination = $self->destination;
 
-    Genome::Sys->validate_directory_for_read_write_access($destination);
-
-    unless($build->can('symlink_results')) {
-        $self->fatal_message('Symlinking results is unavailable for builds of type "%s"', $build->type_name);
-    }
-
     my $dir = $build->symlink_results($destination);
-    $self->status_message('Results for build %s symlinked to %s', $build->__display_name__, $dir);
+    $self->output_directory($dir);
 
     return 1;
 }
--- a/lib/perl/Genome/Model/Build/Command/SymlinkResults.pm
+++ b/lib/perl/Genome/Model/Build/Command/SymlinkResults.pm
@@ -1,0 +1,50 @@
+package Genome::Model::Build::Command::SymlinkResults;
+
+use strict;
+use warnings;
+
+use Genome;
+
+class Genome::Model::Build::Command::SymlinkResults {
+    is => ['Command::V2'],
+    has => [
+        build => {
+            is => 'Genome::Model::Build',
+            shell_args_position => 1,
+            doc => "The builds that has results you'd like to create symlinks to",
+        },
+        destination => {
+            is => 'DirectoryPath',
+            shell_args_position => 2,
+            doc => 'The desired location under which to make a directory of symlink(s)',
+        }
+    ],
+    doc => 'Creates symlinks to the results of a build.',
+};
+
+sub help_detail {
+    my $self = shift;
+    return <<EOP;
+Creates symlinks to the results of a build.  The <destination> option should
+be an existing directory.
+EOP
+}
+
+sub execute {
+    my $self = shift;
+    my $build = $self->build;
+    my $destination = $self->destination;
+
+    Genome::Sys->validate_directory_for_read_write_access($destination);
+
+    unless($build->can('symlink_results')) {
+        $self->fatal_message('Symlinking results is unavailable for builds of type "%s"', $build->type_name);
+    }
+
+    my $dir = $build->symlink_results($destination);
+    $self->status_message('Results for build %s symlinked to %s', $build->__display_name__, $dir);
+
+    return 1;
+}
+
+1;
--- a/lib/perl/Genome/Model/Build/SingleSampleGenotype.pm
+++ b/lib/perl/Genome/Model/Build/SingleSampleGenotype.pm
@@ -102,4 +102,19 @@ sub symlink_results {
     return $top_directory;
 }
 
+sub _compare_output_files {
+    my ($self, $other_build) = @_;
+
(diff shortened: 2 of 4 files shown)


lib/perl/Genome/Model/Build/Command/SymlinkResults.pm
```diff
@@ -0,0 +1,50 @@
...
+    my $self = shift;
+    my $build = $self->build;
+    my $destination = $self->destination;
+
+    Genome::Sys->validate_directory_for_read_write_access($destination);
+
+    unless($build->can('symlink_results')) {
+        $self->fatal_message('Symlinking results is unavailable for builds of type "%s"', $build->type_name);
+    }
+
```
wschroeder: I am not sure what facilities this object model has for traits/roles/flavors, but it is generally better to validate a class has a particular interface than to trust duck typing (does vs can).  $dog->bark vs $tree->bark and all that.  Would that be possible here?

An alternative to an interface would be to assume that all build types support a symlink_results method, and the default implementation in the base class could be to either serve the fatal message or NOP.

lib/perl/Genome/Model/Build/Command/SymlinkResults.pm
```diff
@@ -0,0 +1,50 @@
...
+    my $dir = $build->symlink_results($destination);
+    $self->status_message('Results for build %s symlinked to %s', $build->__display_name__, $dir);
+
```
wschroeder: Should most of this code be part of symlink_results instead of special knowledge on the command level?  The same comment goes for the validate_directory_for_read_write_access() call; it looks like knowledge about how to do symlinking exists in two different abstraction layers.  Does it make sense to validate the directory before figuring out if this is a build type that even supports symlinking?

Since each build will have to do its own flavor of symlinking, an alternative pattern as per recommendations in POODR would be to have the build classes implement a handler, an interface dictated by the base class that would default to the fatal message I mentioned above:

```perl
sub symlink_results {
    my $self          = shift;
    my $destination   = shift;

    my $dir = $self->perform_symlinking($destination);  # the callback
    $self->status_message('Results for build %s symlinked to %s', $self->__display_name__, $dir);
}
```
Another note... Given the interface here, is it accurate to say that any error that occurs in this process is a thrown exception?  For the interface portion of the command, would it be useful to catch the exception and convert to a fatal_message?

lib/perl/Genome/Model/Build/SingleSampleGenotype.pm
```diff
@@ -71,4 +73,48 @@ sub reference_being_replaced_for_input {
...
+        my $hc_basename = Genome::Utility::Text::sanitize_string_for_filesystem(
+            join('-', $hc_result->intervals, $hc_result->id)
+        );
+        my $hc_dir = File::Spec->join($variant_dir, $hc_basename);
+        Genome::Sys->create_symlink($hc_result->output_dir, $hc_dir);
+    }
+
+    return $top_directory;
+}
+
```
wschroeder: We talked about this yesterday; I am just documenting here what I was saying.  Basically, I saw what felt like a large function and could not understand at a glance the expected result.  Then I realized that the noise here comes from a repeating pattern of File::Spec temp variable stuff, followed by creating symlinks or directories.  Here is the rough (comments afterwards about what to do with the roughness) alternative I suggested:

```perl
sub _create_directory {
    my ($dir, $subdir) = @_;
    Genome::Sys->create_directory( File::Spec->join($dir, $subdir) );
}

sub _create_symlink {
    my ($dir, $subdir, $source) = @_;
    my $new_path = File::Spec->join($dir, $subdir);
    Genome::Sys->create_symlink($new_path, $source);
    $new_path;
}

sub _basename {
    my $hc_result = shift;
    Genome::Utility::Text::sanitize_string_for_filesystem(
        join('-', $hc_result->intervals, $hc_result->id)
    );
}

sub symlink_results {
    my $self          = shift;
    my $destination   = shift;

    my $top_directory = _create_directory($destination, $self->id);
    _create_symlink($top_directory, 'alignments', $self->merged_alignment_result->output_dir);
    _create_symlink($top_directory, 'qc', $self->qc_result->output_dir);

    my $variant_dir   = _create_directory($top_directory, 'variants');
    for my $hc_result ($self->haplotype_caller_result) {
        _create_symlink($variant_dir, _basename($hc_result), $hc_result->output_dir);
    }

    return $top_directory;
}
```

This form of _create_directory() actually arose later, but I asked: Should Genome::Sys->create_directory support concatenation of the multiple parts passed in?  This is a common enough pattern that it may be worth promoting to the Sys function.

We identified that Genome::Sys->create_symlink was inconsistent with Genome::Sys->create_directory in that it returns a status code instead of the created entity.  Also, the status code is not a good choice for this function anyway; any error gets croaked or dies out of the function, so this is actually mixing two approaches towards error handling and not actually providing useful information for the caller.  That's probably why the calls in the original code here don't check the result.

Genome::Sys->create_symlink could probably use the same kind of interface I suggested for Genome::Sys->create_directory, also for consistency and for removing my rough private _create_symlink method.

The other rough helper I wrote was _basename(), and I think that is another candidate for promotion to whatever class is behind $self->haplotype_caller_result.  Our symlink_results() function knows a lot about how to construct specific unique names for this entity.  This code is in G::M::B::SingleSampleGenotype, but it seems like any kind of build that wants to deal with haplotype caller results will either need to copy-paste this code or come up with its own mechanism for distinguishing between the multiple results.

Finally, we talked about the train wreck pattern here of my calls like $self->merged_alignment_result->output_dir.  I don't think this is too bad, since it is basically taking an object (merged_alignment_result) in the current object's context ($self) and making a single accessor call.  If this had any more levels of indirection here, the recommendation I have read from SOLID practitioners (in order to decrease coupling) is to promote knowledge up a class.  For example -- though I don't think this is an improvement HERE --

    $self->merged_alignment_result_output_dir

I'd only recommend that if any client would be accessing this build class to find out the output dir.

Sorry for the essay! :laughing:

lib/perl/Genome/Model/Build/Command/SymlinkResults.pm
tmooney: Yeah, I did it this way because I was looking at `Genome::Model::Command::InstrumentData::AlignmentBams` right before this, and this is what it did.

I'd thought about creating a default `symlink_results` in `Genome::Model::Build`; in the context of your second suggestion about moving the directory validation out of the command, perhaps that makes the most sense!  I'll move some stuff around :smile:

lib/perl/Genome/Model/Build/SingleSampleGenotype.pm
tmooney: I didn't do exactly what was suggested here, but I do think breaking this down a little made the code easier to follow.  Thanks for the discussion!

Some history (some of which is also recording yesterday's discussion): The name `create_symlinks` was taken for parallelism with the already-named feature in `Genome::Process` (which also used the duck-typing approach that this PR has now changed away from).  But it really creates a simulated build directory--the names of the paths and the contents in the directory are arbitrary at the discretion of a particular build type.  Most build types do delegate in order to fill in the contents of particular directories (notably to `DetectVariants2` for the `variants` directory), but SSG is the only thing using `HaplotypeCaller` results for now.

jasonwalker80:
+1

acoffman:
+1

PR #1270, opened 2016-02-23
ReferenceAlignment No Longer Uses a Staged ProcessingProfile.

tmooney:
It's all been leading up to this!

Builds `643aca7d6ab24d9a8e38f1686032ce50` and `c226b740f58d4a989c405a60ae40d18a` demonstrate the function of this new version both with and without a test name set.  (The version with a test name diffs cleanly with the currently blessed build.

The `apipe-test-reference-alignment-lane-qc` test model will need to be removed if this PR is accepted.

PR #1302, opened 2016-03-09
BamQC should only check instrument data for the result it will run on.

tmooney:
This fixes a bug I introduced in #1270 (right [here](https://github.com/tmooney/genome/commit/26017071733a84e4f78094d7762f21c133558940#diff-d38dfe2f448c8bdee9ef7eb70b2eac68R73)).  That germline model test is coming in handy--the regular refalign test didn't catch these!


jasonwalker80:
+1

dufeiyu:
+1

PR #1303, opened 2016-03-09
New chr_length reference converter

jasonwalker80:
Add chr_length converter algorithm to match chromosome names when all chromosome lengths are unique and match between reference builds.

### DIFF
--- a/lib/perl/Genome/Model/Build/ReferenceSequence/Converter.pm
+++ b/lib/perl/Genome/Model/Build/ReferenceSequence/Converter.pm
@@ -31,13 +31,17 @@ class Genome::Model::Build::ReferenceSequence::Converter {
     has_metric => [
         algorithm => { # README - If adding an algorithm, please add to the list of valid algorithms
             is => 'Text',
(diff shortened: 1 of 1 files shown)


jasonwalker80:
I can already think of one issue.  What if there are multiple chromosomes with the same length? Suggestions welcome.

lib/perl/Genome/Model/Build/ReferenceSequence/Converter.pm
```diff
@@ -40,6 +40,10 @@ class Genome::Model::Build::ReferenceSequence::Converter {
             is_optional => 1,
         },
     ],
+    has_transient_optional => [
+        _chr_map_hash_ref => {
+        },
```
tmooney: It'd be nice if this property included:
```perl
is => 'HASH'
```
(The type needn't necessarily be included in the name then, but either way!)

A `doc` line might be good, too, but is less important for internal things.

tmooney:
The `seqdict.sam` does include the MD5s, so we could use those to help properly match.  I don't think there's already a method to pull them out of the file, though.

jasonwalker80:
I'm already on the md5 solution.

PR #1304, opened 2016-03-09
New chr_md5 reference converter

jasonwalker80:
A new reference sequence converter that relies on matching the md5 in each build sequence dictionary to map chromosome names.


jasonwalker80:
This replaces #1303

jasonwalker80:
New commits to address suggestions.

tmooney:
Looks like the bucket-creation tests are failing, because they used mocked-up seqdict files without the additional info we're now expecting.  But +1 once those are fixed!

I like the fact that this will fail unless the two references really are identical.  That does mean this implementation is not compatible with the existing `drop_extra_contigs` implementation.

susannasiebert:
+1

PR #1305, opened 2016-03-10
Zero size variants in CreateReport

jasonwalker80:
Allow for zero size variant files in the Create Report step of Somatic Variation models.


jasonwalker80:
@susannasiebert @acoffman or @tmooney any additional suggestions?

tmooney:
It is usually customary in our pipelines to produce empty files when given an empty input.  (See for example `add_tiers` in this module.)  It might be worthwhile to continue that practice here.

Otherwise, +1.

susannasiebert:
I agree with @tmooney's comment. Otherwise, +1.

jasonwalker80:
Thanks for the suggestions.  A new commit that will touch the expected files rather than skip steps.

PR #1306, opened 2016-03-10
Specify a default job group for PTero-launched jobs.

tmooney:
This will allow the group owner (read: `apipe-builder`) to manipulate LSF jobs for any running `PTero` workflows launched from `Genome`.

### DIFF
--- a/lib/perl/Genome/WorkflowBuilder/Command.pm
+++ b/lib/perl/Genome/WorkflowBuilder/Command.pm
@@ -154,6 +154,12 @@ sub _get_ptero_lsf_parameters {
     $set_lsf_option->('queue', $attributes{lsfQueue});
     $set_lsf_option->('projectName', $attributes{lsfProject});
 
+    my $default_job_group = join('/',
+        Genome::Config::get('lsf_job_group'),
+        Genome::Sys->username,
+    );
+    $set_lsf_option->('jobGroup', $default_job_group);
+
     my ($stderr, $stdout, $postexec) = _get_lsf_log_paths();
     $lsf_params->{options}->{errFile} = $stderr;
     $lsf_params->{options}->{outFile} = $stdout;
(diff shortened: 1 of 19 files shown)


acoffman:
+1

lib/perl/Genome/WorkflowBuilder/Command.pm
```diff
@@ -154,6 +154,12 @@ sub _get_ptero_lsf_parameters {
     $set_lsf_option->('queue', $attributes{lsfQueue});
     $set_lsf_option->('projectName', $attributes{lsfProject});
 
+    my $default_job_group = join('/',
```
davidlmorton: This isn't really the 'default' since it cannot be overwritten.  Perhaps the config could default to the empty string (or some sentinel value) and if it isn't overwritten then this logic could be invoked (using /genome/<username>).

tmooney: Couldn't either the user set it or it could be parsed by `parse_lsf_params` (though that looks like it's currently parsing `-g` into `group` instead of `jobGroup`--maybe that should change!)?

The config value could also be changed, and the job would land in the desired group, and in a subgroup thereof that the user would almost certainly be able to further limit if they desired.

davidlmorton: I didn't read the set_lsf_option closely enough to notice that it would not overwrite previous values.

jeldred:
in 
lib/perl/Genome/Ptero/workflow_tests/n-shaped/ptero_workflow.json
and elsewhere should the value of  
"jobGroup" : "/genome/dmorton",   
be changed to someone else?

jeldred:
assuming a good value for
"jobGroup" : "/genome/dmorton",

+1 :shipit:

tmooney:
@jeldred Not to worry!  That specific value for the jobGroup is ignored by the tests in the addition [here](https://github.com/tmooney/genome/blob/af48387c451e7cc7fdef1b55c55420915945794d/lib/perl/Genome/Ptero/ToJSON.t#L49).

PR #1307, opened 2016-03-11
Make FilterFalsePositives not die on empty input.

gatoravi:
The FilterFalsePositives module is run as a part of the LOH pipeline. Currently
if the variant file that is passed to this module is empty,
the workflow crashes with an error message (bam-readcount is invoked
with an empty file which causes it to die). Instead I think
we'd like this module to spit out a warning and not attempt to
filter if the input is empty.


susannasiebert:
CI, test this.

susannasiebert:
The automated apipe-ci/bats test failed the first time around and there wasn't an obvious failure reason. I restarted the test. +1 once tests pass.

tmooney:
+1 :pear:

PR #1308, opened 2016-03-14
Tiering for GRCh38

jasonwalker80:
Add a UCSC tiering directory for GRCh38 reference sequence build.

### DIFF
--- a/lib/perl/Genome/Model/Build/ReferenceSequence.pm
+++ b/lib/perl/Genome/Model/Build/ReferenceSequence.pm
@@ -191,23 +191,19 @@ sub get{
 #For now, this is hard-coded
 sub get_or_create_ucsc_tiering_directory {
     my $self = shift;
-    if ($self->id eq "10194788") {
-        return "/gscmnt/sata921/info/medseq/make_tier_bed_files/hg18_build36_ucsc_files";
-    }
-    elsif ($self->id eq "106942997") {
-        return "/gscmnt/sata921/info/medseq/make_tier_bed_files/NCBI-human-build37/hg19_files";
-    }
-    elsif ($self->id eq "107494762") {
-        return "/gscmnt/sata921/info/medseq/make_tier_bed_files/NCBI-mouse-build37/mm9";
-    }
-    elsif ($self->id eq "3ca0ea4786dd4ebebaf3935f3d3ccab8") {
-        return "/gscmnt/sata921/info/medseq/make_tier_bed_files/NCBI-mouse-build38/mm10";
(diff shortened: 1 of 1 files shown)


susannasiebert:
Could this be made into a hash instead of an if-else cascade?

jasonwalker80:
new commit with hash suggestion.

susannasiebert:
+1 once tests pass.

lib/perl/Genome/Model/Build/ReferenceSequence.pm
```diff
@@ -191,20 +191,19 @@ sub get{
...
-    }
-    else {
-        $self->status_message("UCSC Tiering Directory not currently available for this species: ".$self->species_name);
+
+    my %tiering_dirs = (
+        '10194788' => '/gscmnt/sata921/info/medseq/make_tier_bed_files/hg18_build36_ucsc_files',
+        '106942997' => '/gscmnt/sata921/info/medseq/make_tier_bed_files/hg18_build36_ucsc_files',
+        '107494762' => '/gscmnt/sata921/info/medseq/make_tier_bed_files/NCBI-mouse-build37/mm9',
+        '3ca0ea4786dd4ebebaf3935f3d3ccab8' => '/gscmnt/sata921/info/medseq/make_tier_bed_files/NCBI-mouse-build38/mm10',
+        '4ec1c5bd1f6941b8a99f2e230217cb91' => '/gscmnt/gc13001/info/model_data/jwalker_scratch/GRCh38',
```
tmooney: Would it make sense to add this to `/gscmnt/sata921/info/medseq/make_tier_bed_files/` like the others?

(It makes me a little sad these are all hard-coded, but that'd be a much bigger change.)

jasonwalker80: I could add them there if I had permissions.  I could easily move them to proposed directory after requesting addition to the medseq group.

jasonwalker80: After requesting the addition to the medseq group, I was made aware that the info_medseq disk groups may be going away.  I pointed out that these data paths would have to be migrated somewhere.  I'm going to leave the files where they are until the location of the existing data is sorted out.

https://jira.gsc.wustl.edu/browse/INFOSYS-18998

jasonwalker80:
@tmooney other suggestions on this PR?

tmooney:
Nope--we'll have to figure out something to do with those files, but we can continue to punt on that. +1.

PR #1309, opened 2016-03-14
added check to see existence of wgs or exome

bainscou:


susannasiebert:
Should the exome model be evaluated first to be closer to the original implementation?

bainscou:
@susannasiebert that sounds reasonable. Just committed that change.

susannasiebert:
+1

lib/perl/Genome/Model/ClinSeq/Command/Converge/Stats.pm
```diff
@@ -128,7 +128,14 @@ sub find_stats_tsv {
     my $m = $mb->{$c}->{model};
     my $model_name = $m->name;
     my $data_directory = $b->data_directory;
-    my $subject_name = $m->exome_model->tumor_model->subject->common_name;
+    my $subject_name;
+    if($m->exome_model){
+        $subject_name = $m->exome_model->tumor_model->subject->common_name;
+    }elsif($m->wgs_model){
+        $subject_name = $m->wgs_model->tumor_model->subject->common_name;
```
tmooney: `SomaticVariation` and `SomaticValidation` models both have `experimental_subject` and `control_subject` accessors directly on them now, so these could be cut back one arrow to something like `$m->wgs_model->experimental_subject->common_name`.

bainscou:
@tmooney how is that?

tmooney:
+1, looks good to me!

PR #1310, opened 2016-03-14
Remove old RefCov result.

tmooney:
This appears to be unused--there are no rows in the DB for this result.  (The pipelines use the CoverageStats result instead.)


wschroeder:
+1 :palm_tree:

brummett:
+1

PR #1311, opened 2016-03-14
Replace NO_LSF with "inline" Backend

tmooney:
Under WorkflowBuilder, the NO_LSF environment variable has no effect if the backend is set to `ptero`.  This explicitly uses the `inline` backend where necessary and removes the variable where it does not appear a workflow is being run.

There is a third class of test that is unchanged by this PR: Tests for tools which are not converted to WorkflowBuilder.  Those must be deleted or upgraded before the `inline` setting has any effect.


mkiwala:
+1 :koala:

PR #1312, opened 2016-03-16
Support Heartbeat Check for PTero-based Builds.

tmooney:
This heartbeat code is pretty ugly. It seems like it should be split out into its own module... but at least with this PR it works on builds run with PTero!

I'm definitely a novice at querying PTero; I'd be quite happy to learn of better ways to get the information.

### DIFF
--- a/lib/perl/Genome/Model/Build.pm
+++ b/lib/perl/Genome/Model/Build.pm
@@ -21,6 +21,9 @@ use Genome::Sys::LSF::bsub qw();
 use Genome::Utility::Email;
 use Genome::Utility::Vcf;
 
+use Ptero::HTTP qw();
+use Ptero::Statuses qw();
+
 require Scope::Guard;
 
 class Genome::Model::Build {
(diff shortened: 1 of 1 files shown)


jeldred:
I'm unclear if the new ptero heartbeat code works ... but it's all new, so passing tests suggest it won't break existing functionality even in test/build czar automation

the changes to exist parts do seem slightly better organized 

+1 :shipit:

tmooney:
Yeah... the lack of a new test for this is a reasonable criticism.  I wasn't sure how to go about making a test for this that didn't either require PTero services or mock so much that it wasn't testing anything useful.  But this tool is only run internally, so it's less of an issue if it breaks.

Empirically it has been working for me when I run from a checkout to the extent that it does not instantly pass all PTero workflows like the old version, and does occasionally point out issues.  I know there's a hole in that it won't detect issues with the shortcut jobs, but the existing Workflow-based heartbeat check has the same shortcoming.

acoffman:
+1

PR #1313, opened 2016-03-16
Update the ClinSeq WGS SomVar processing profile.

jasonwalker80:
It appears that ClinSeq update-analysis was not updated to use the WGS pindel false positive parameter value of 0.2 rather than 0.08 which is appropriate for exome data:
`[-[--variant-freq-cutoff=0.2]-] {+[--variant-freq-cutoff=0.08]+}`

I'm tagging ClinSeq people that should review this.  I'm also not aware of how this would impact ClinSeq update-analysis users as the default value is changed.  Perhaps that's why it has not been updated before...
@malachig @gatoravi and @chrisamiller

### DIFF
--- a/lib/perl/Genome/Model/ClinSeq/Command/UpdateAnalysis.pm
+++ b/lib/perl/Genome/Model/ClinSeq/Command/UpdateAnalysis.pm
@@ -41,7 +41,7 @@ class Genome::Model::ClinSeq::Command::UpdateAnalysis {
         },
         _wgs_somatic_variation_pp_id => {
               is => 'Number',
-              default => '2762562',
+              default => 'b359413673b040d2b29b7810e71c4882',
         },
         wgs_somatic_variation_pp => {
               is => 'Genome::ProcessingProfile',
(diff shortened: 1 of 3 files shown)


lib/perl/Genome/Model/ClinSeq/Command/UpdateAnalysis.pm
```diff
@@ -41,7 +41,7 @@ class Genome::Model::ClinSeq::Command::UpdateAnalysis {
         },
         _wgs_somatic_variation_pp_id => {
               is => 'Number',
```
tmooney: If I may be nitpicky, the new chosen ID is not really a `Number`.

jasonwalker80: You may.  I updated all of the Number attributes to be Text instead.

gatoravi:
Is this in anyway related to https://rt.gsc.wustl.edu/Ticket/Display.html?id=106978, we were(are?) basically seeing zero calls from Pindel in somatic-variation currently I believe. At the time I figured this might have been due to the switch to BWA-mem, but never really got to the bottom of this I don't think. Kai suggested updating the version of Pindel being used in-house.

jasonwalker80:
I doubt this has anything to do with the Pindel results and BWA-MEM BAM files.  The value has been the same since 2012.  I'm not sure why a change from BWA to BWA-MEM would have impacted the VAF of indels which is what this parameter would impact.

Also, pushed new test and new test data to get tests passing.

tmooney:
+1 inasmuch as it's definitely a processing profile!  I make no claims as to the suitability of any parameters specified within it.

jasonwalker80:
I'm going to close this PR in favor of planned updates to use Somatic Validation model types in the next few weeks.  ClinSeq has been using this PP for a while.

PR #1314, opened 2016-03-16
Skip mutation spectrum on empty input

gatoravi:
Currently clinseq attempts to run mutation spectrum sequence
context on an empty input which causes the downstream R module
to die unpredictably. We'd like to skip running this analysis
when the tier1 variant file is empty instead.

### DIFF
--- a/lib/perl/Genome/Model/ClinSeq/Command/CreateMutationSpectrum.pm
+++ b/lib/perl/Genome/Model/ClinSeq/Command/CreateMutationSpectrum.pm
@@ -251,6 +251,13 @@ sub get_mutation_spectrum_sequence_context_result {
   my $final_name = shift;
   my $tier1_snvs = shift;
   my $sub_outdir2 = shift;
+
+  #Check for tier1 variants
+  if(-z $tier1_snvs) {
+      $self->status_message("Tier1 SNVs file empty. Skipping mutation spectrum" .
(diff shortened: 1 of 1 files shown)


tmooney:
+1 :mega:

lib/perl/Genome/Model/ClinSeq/Command/CreateMutationSpectrum.pm
```diff
@@ -251,6 +251,13 @@ sub get_mutation_spectrum_sequence_context_result {
   my $final_name = shift;
   my $tier1_snvs = shift;
   my $sub_outdir2 = shift;
+
+  #Check for tier1 variants
+  if(-z $tier1_snvs) {
+      $self->status_message("Tier1 SNVs file empty. Skipping mutation spectrum" .
+                            "sequence context for tier1.");
+      return;
+  }
```
brummett: Maybe it would be better to switch the test.  That way if more code is added to this method in the future, it'll still get run.
```
if (-s $tier1_snvs) {
    ... # existing code
} else {
    $self->status_message('Skipping mutation spectrum')
}
```

brummett:
Actually, I see that's all that method is used for... This patch is fine as-is.
+1

gatoravi:
Thanks!

PR #1315, opened 2016-03-16
Support Multiple Merge Groups for AlignAndMerge in the AlignmentDispatcher

tmooney:
There were previous assumptions baked into this implementation that meant only one AlignAndMerge could appear within a run of the dispatcher.  The step would also not work properly if meant to only run on a subset of the data.  This PR addresses both of these shortcomings.


susannasiebert:
+1

jasonwalker80:
+1

PR #1316, opened 2016-03-17
Add an ID column to the Ptero Workflow display.

tmooney:
This makes it easier to track down additional information about an execution (e.g. the LSF Job ID), which is otherwise not included in the output.


acoffman:
+1

jeldred:
+1

PR #1317, opened 2016-03-17
A Command to Get Information about a PTero Execution.

tmooney:
This command seems to work given the IDs from #1316, but I wanted to make a PR a bit early to solicit feedback.
* Is there a better name for this than `genome ptero execution-info`?
* How should we _really_ get the execution URL?  (Another config variable? Another method in `Genome::Ptero::Utils`? Both?)
* Is there any other information that ought to be included? (Is the output too noisy to display by default?)
* Between this and #1312, would it be worthwhile to make a more official PerlSDK-ish proxy for Job entities?
* Does something already do most of this, and I'm wasting my time? :smile: 

TODO:
- [x] Write some sort of test!

### DIFF
--- a/lib/perl/Genome/Ptero/ExecutionInfo.pm
+++ b/lib/perl/Genome/Ptero/ExecutionInfo.pm
@@ -1,0 +1,169 @@
+package Genome::Ptero::ExecutionInfo;
+
+use strict;
+use warnings;
+
+use Ptero::HTTP qw();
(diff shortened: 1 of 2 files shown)


tmooney:
I've added a test!  So although I'm still happy for answers to the previously asked questions, I also propose that this could be merged as-is depending on the answers.

jeldred:
:+1 :shipit:

jasonwalker80:
+1

PR #1320, opened 2016-03-21
SV realignment compatibility with SpeedSeq BAM headers

jasonwalker80:
Use whitespace to denote end of string rather than word-boundary.  Non-word characters are excluded from the end of the library string such as double quotation marks.  

The change is triggered by SpeedSeq passing the LIMS BAM header rather than recreating the header from sample meta-data.  I'm not sure why the LIMS BAM header has doube quotes around the library name, but that is the source of this issue.


tmooney:
+1.  I don't look forward to having to re-header all these BAMs later when someone decides the quotes are a problem! :smile:

susannasiebert:
+1

PR #1318, opened 2016-03-17
clinseq fixes for hg38

gatoravi:
Contains fixes for clin-seq to be able to work on hg38. Incorporates suggestions from #1292 which can be closed now.

PR #1321, opened 2016-03-22
Run all ClinSeq related commands through perltidy

susannasiebert:
There will probably be some merge conflicts once #1318 gets merged.


jasonwalker80:
+1, once tests pass.

lib/perl/Genome/Model/ClinSeq/Command/AnnotateGenesByCategory.t
```diff
@@ -77,13 +86,13 @@ ok(-e $log_file2, "Wrote message file from annotate-genes-by-category to a log f
...
-};
+    or do {
+    diag("expected: $expected_output_dir\nactual: $temp_dir\n");
+    diag("differences are:");
+    diag(@diff);
+    my $diff_line_count = scalar(@diff);
+    print "\n\nFound $diff_line_count differing lines\n\n";
+    Genome::Sys->shellcmd(cmd => "rm -fr /tmp/last-annotate-genes-by-category");
+    Genome::Sys->shellcmd(cmd => "mv $temp_dir /tmp/last-annotate-genes-by-category");
+    };
```
tmooney: I wonder why perltidy has the contents of this block and the closing `}` at the same indentation.

tmooney:
+1, even though I think the `do` blocks look funny!

lib/perl/Genome/Model/ClinSeq/Command/AnnotateGenesByCategory.t
susannasiebert: I'm not sure but I bet it has something to do with the `or`. I will take a closer look and make it look nicer.

PR #1322, opened 2016-03-23
True CSV parsing support.

jasonwalker80:
use Text::CSV to parse CSV lines rather than simple split.

### DIFF
--- a/lib/perl/Genome/Utility/IO/SeparatedValueReader.pm
+++ b/lib/perl/Genome/Utility/IO/SeparatedValueReader.pm
@@ -3,6 +3,8 @@ package Genome::Utility::IO::SeparatedValueReader;
 use strict;
 use warnings;
 
+use Text::CSV;
+
 use Genome;
 
 class Genome::Utility::IO::SeparatedValueReader {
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Utility/IO/SeparatedValueReader.pm
```diff
@@ -89,7 +91,21 @@ sub create {
...
+            chomp($line);
+            if ($csv->parse($line)) {
+                return _strip_quotes($csv->fields());
+            } else {
+                $self->fatal_message('Failed to parse CSV line: '. $line);
+            }
+        };
+    } else {
+        $self->{_split} = sub{ return _strip_quotes(split($regexp, $_[0], -1)) };
+    }
```
brummett: Text::CSV can handle all kinds of cases, not just when using ```,``` as the separator.  It can also strip quotes by itself.

I'd imagine this whole module could probably get turned into a simple wrapper around Text::CSV.

Another option might be to turn this into some kind of data source factory or for UR::DataSource::FileSystem data sources or a role using UR::DS::FileSystem.  Then consumers would be able to use the same get() API as everything else, and get back real objects instead of arrays of lines.

jasonwalker80: I agree Text::CSV could be used more in this module.  I was focused on adding support of true CSV parsing that uses text delimiters in addition to the field delimiter.

As for turning each entry into a UR object, I believe this is why some parts of our codebase still use separated value reader instead of filesystem data sources.  It's much faster to read millions of lines when simple hash refs are returned rather than loading real objects for each line of text.  However, I could imagine a single interface that would allow the optional creation of real objects for each line.

brummett:
Then I guess converting this module to use Text::CSV and/or speeding up object loading from files would be good projects for the future.

+1

acoffman:
+1

PR #1323, opened 2016-03-24
Add c_position to variant report

dufeiyu:
c_position column from vep output need to be added to report. This PR also replaces the obsolete CGI with URI::Escape, which fixes a bug that unescapes + into space with CGI.

### DIFF
--- a/lib/perl/Genome/File/Vcf/VepConsequenceParser.pm
+++ b/lib/perl/Genome/File/Vcf/VepConsequenceParser.pm
@@ -5,7 +5,7 @@ package Genome::File::Vcf::VepConsequenceParser;
 
 use Data::Dumper;
 use Carp qw/confess/;
-use URI::Escape;
+use URI::Escape qw/uri_unescape/;
 
 use strict;
 use warnings;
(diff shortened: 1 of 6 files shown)


tmooney:
+1 :fallen_leaf:

lib/perl/Genome/File/Vcf/VepConsequenceParser.pm
```diff
@@ -5,7 +5,7 @@ package Genome::File::Vcf::VepConsequenceParser;
 
 use Data::Dumper;
 use Carp qw/confess/;
-use CGI qw/unescape/;
+use URI::Escape;
```
brummett: Even though it automatically exports ```uri_unescape()```, it's still good practice to put it in the import list
```
use URI::Escape qw(uri_unescape);
```

Did you figure out whether this data is sourced from an actual URL and whether uri_unescape is the right thing to do here?

dufeiyu: ok, I will add to import list.

brummett:
+1

PR #1325, opened 2016-03-24
Use new 'service_data_to_save' feature of PTero

davidlmorton:
This cannot pass tests until genome-snapshot-deps has been updated via this PR: https://github.com/genome/genome-snapshot-deps/pull/67

This will put data from the Job services on the PTero execution's data
field saving us from querying the Job service.


tmooney:
+1 once tests pass.

mkiwala:
+1 :koala:

PR #1326, opened 2016-03-25
Port ClinSeq to use WorkflowBuilder

susannasiebert:
This PR makes ClinSeq use WorkflowBuilder. It also includes some refactoring of the _resolve_workflow_for_build subroutine by split the various parts of the workflow into separate subroutines for clarity (hopefully).

I ran a ClinSeq test build in a custom checkout and the workflows look identical. @jasonwalker80, you mentioned that you were working on a ClinSeq model that would include all possible steps. It might be worthwhile to run a build of that model on this PR to make sure that all steps are still executed.

@acoffman, @gatoravi , @jasonwalker80 Feel free to push additional commits to this PR with any improvements that might be necessary while I'm on vacation.

### DIFF
--- a/lib/perl/Genome/Model/Build/ClinSeq/InputBuilds.pm
+++ b/lib/perl/Genome/Model/Build/ClinSeq/InputBuilds.pm
@@ -30,4 +30,9 @@ sub resolve_rnaseq_builds {
   $rnaseq_builds->{$normal_build->id}{type} = 'rnaseq' if $normal_build;
 }
 
+sub wgs_exome_build {
+    my $clinseq_build = shift;
+    return ($clinseq_build->wgs_build and $clinseq_build->exome_build);
+}
+
 1;
--- a/lib/perl/Genome/Model/ClinSeq.pm
+++ b/lib/perl/Genome/Model/ClinSeq.pm
@@ -690,38 +690,7 @@ sub _resolve_workflow_for_build {
     #GenerateClonalityPlots - Run clonality analysis and produce clonality plots
     my $clonality_op;
     if ($build->wgs_build) {
-        $clonality_op = Genome::WorkflowBuilder::Command->create(
-            name => 'Run clonality analysis and produce clonality plots',
-            command => 'Genome::Model::ClinSeq::Command::GenerateClonalityPlots',
-        );
-        $workflow->add_operation($clonality_op);
-        $workflow->connect_input(
-            input_property       => 'misc_annotation_db',
-            destination          => $clonality_op,
-            destination_property => 'misc_annotation_db',
-        );
-        $workflow->connect_input(
-            input_property       => 'wgs_build',
-            destination          => $clonality_op,
-            destination_property => 'somatic_var_build',
-        );
-        $workflow->connect_input(
-            input_property       => 'clonality_dir',
-            destination          => $clonality_op,
-            destination_property => 'output_dir',
-        );
-        for my $property (qw(common_name bam_readcount_version)) {
-            $workflow->connect_input(
-                input_property       => $property,
-                destination          => $clonality_op,
-                destination_property => $property,
-            );
-        }
-        $workflow->connect_output(
-            output_property => 'clonality_result',
-            source          => $clonality_op,
-            source_property => 'result',
-        );
+        $clonality_op = $self->clonality_op($workflow);
     }
 
     #RunCnView - Produce copy number results with run-cn-view.  Relies on clonality step already having been run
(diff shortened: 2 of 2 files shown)


susannasiebert:
I apologize to the people that I accidentally tagged incorrectly.

lib/perl/Genome/Model/Build/ClinSeq/InputBuilds.pm
```diff
@@ -30,4 +30,9 @@ sub resolve_rnaseq_builds {
   $rnaseq_builds->{$normal_build->id}{type} = 'rnaseq' if $normal_build;
 }
 
+sub wgs_exome_build {
```
tmooney: This name is a little strange.  The other `..._build`s return builds, but this returns a boolean.  I see how it's programmatically convenient in the rest of this commit, though.

lib/perl/Genome/Model/ClinSeq.pm
```diff
@@ -2027,6 +1969,71 @@ sub make_circos_plot_op {
     return $make_circos_plot_op;
 }
 
+sub converge_snv_indel_report_op {
+    my $self = shift;
+    my $workflow = shift;
+    my $iterator = shift;
```
tmooney: This variable seems to hold an index rather than an iterator.

lib/perl/Genome/Model/ClinSeq.pm
```diff
@@ -1062,85 +1062,27 @@ sub _resolve_workflow_for_build {
     #Converge SnvIndelReport
     my @converge_snv_indel_report_ops;
     if ($build->wgs_build || $build->exome_build) {
+        my %variant_sources_ops = (
+            exome => $exome_variant_sources_op,
+            wgs   => $wgs_variant_sources_op,
+        );
         #Create a report for each $bq $mq combo.
         my $iterator = List::MoreUtils::each_arrayref([1 .. @$mqs], $mqs, $bqs);
         while (my ($i, $mq, $bq) = $iterator->()) {
```
tmooney: It looks like `$mq` and `$bq` are unused inside this loop.

lib/perl/Genome/Model/ClinSeq.pm
```diff
@@ -2034,6 +1981,56 @@ sub converge_snv_indel_report_op {
     return $converge_snv_indel_report_op;
 }
 
+sub create_mutation_spectrum_op {
+    my $self = shift;
+    my $workflow = shift;
+    my $sequencing_type = shift;
+    my $iterator = shift;
```
tmooney: Same comment about `$iterator` being an index.

lib/perl/Genome/Model/ClinSeq.pm
```diff
@@ -645,33 +645,18 @@ sub _resolve_workflow_for_build {
...
-    }
-    #TophatJunctionsAbsolute - Run tophat junctions absolute analysis on tumor
-    my $tumor_tophat_junctions_absolute_op;
-    if ($build->tumor_rnaseq_build) {
-        $tumor_tophat_junctions_absolute_op = $self->tophat_junctions_absolute_op($workflow, 'tumor');
-    }
-
     #CufflinksExpressionAbsolute - Run cufflinks expression absolute analysis on normal
     if ($build->normal_rnaseq_build) {
+        my $normal_tophat_junctions_absolute_op = $self->tophat_junctions_absolute_op($workflow, 'normal');
```
tmooney: This variable (and the one right below it) are unused.

jasonwalker80: I presume this is why the scope of this variable is limited to the conditional block.  If the normal is defined then the downstream annotation step is linked to the differential_expression rather than these normal-only steps.  Should these operations be used elsewhere in the workflow?

I'm running a test build now with the normal RNA-seq defined.  The test @susannasiebert was using did not have a normal rnaseq input and may not have covered this case.

lib/perl/Genome/Model/Build/ClinSeq/InputBuilds.pm
jasonwalker80: Are there any suggestions for improving the consistency between these subroutines?  I'm not entirely sure what the rest of ClinSeq expects these to return.  There is a future task to refactor how ClinSeq resolves all of these input properties:
https://jira.gsc.wustl.edu/browse/CI-109

lib/perl/Genome/Model/ClinSeq.pm
jasonwalker80: These cufflinks steps ran successfully in a test ClinSeq model.  Are you suggesting that both variables should be removed since they are not used elsewhere?

>  7118929   execute  succeeded  2016-03-31T11:03:49       00:02:14           . Performing cufflinks expression absolute analysis for normal sample
>   7118933   execute  succeeded  2016-03-31T11:03:51       00:02:23           . Performing cufflinks expression absolute analysis for tumor sample

tmooney: Yeah, I'm only suggesting that, since the return values are not used for anything, assigning them to `my` variables isn't necessary.

lib/perl/Genome/Model/Build/ClinSeq/InputBuilds.pm
susannasiebert: I agree with @jasonwalker80 that this is out of scope for this PR. I do think we should make the subroutine names consistent once we do the bigger ClinSeq refactor.

susannasiebert:
@tmooney I pushed a few commits to address your suggestions.

lib/perl/Genome/Model/ClinSeq.pm
```diff
@@ -944,8 +944,8 @@ sub _resolve_workflow_for_build {
             wgs   => $wgs_variant_sources_op,
         );
         #Create a report for each $bq $mq combo.
-        my $iterator = List::MoreUtils::each_arrayref([1 .. @$mqs], $mqs, $bqs);
-        while (my ($i, $mq, $bq) = $iterator->()) {
+        my $iterator = List::MoreUtils::each_arrayref([1 .. @$mqs]);
+        while (my ($i) = $iterator->()) {
```
tmooney: These two lines could be simplified to: `for (1..@$mqs) {`.

tmooney:
+1 :hamster:

lib/perl/Genome/Model/ClinSeq.pm
susannasiebert: Duh, of course. Fixed!

jasonwalker80:
+1

PR #1327, opened 2016-03-28
DetermineError: Properly handle case where dies on next line of logs.

tmooney:
Otherwise we're really checking the same line a second time, which doesn't get us any additional information.

(This fixes reporting for code that uses the pattern `die $self->error_message`.  The newer `$self->fatal_message` is unaffected.)


acoffman:
+1

brummett:
+1

PR #1328, opened 2016-03-29
Set a default empty value for parent_workflow_log_directory.

tmooney:
This fixes `genome config list`, which chokes on undefined values.


ebelter:
+2

PR #1329, opened 2016-03-29
PteroWorkflowMixin: Pass correct number of parameters in the "else" cases as well.

tmooney:
This corrects an omission for the new column addition in #1316.


acoffman:
+1

jeldred:
+1 :mans_shoe:

PR #1331, opened 2016-03-30
Update blessed process object for Jenkins genome-process-test

dufeiyu:
All diff is caused by the newly added c_posiotion column in report. Now update blessed object. Need merge this asap to make model/process test pass.


tmooney:
+1, the new ID does identify a trio process!

PR #1333, opened 2016-03-30
Remove old ModelGroup commands.

tmooney:
These have no recorded usage in the usage-log (except one attempt to run one with no parameters).  They mostly look to have been written before the great redaction and otherwise mostly neglected.

This PR is an alternative to updating these to WorkflowBuilder.


dufeiyu:
+1   as long as we still keep ModelGroup around

tmooney:
I definitely have no plans to remove `ModelGroup` itself!

chrisamiller:
+1

PR #1334, opened 2016-03-30
Use the new execute_inline for DAG instead of relying on Workflow.

tmooney:
Another step on the road to eliminating `Workflow`!


mkiwala:
+1 🐨

jasonwalker80:
+1

PR #1337, opened 2016-03-30
GenotypeMicroarray uses WorkflowBuilder.

tmooney:
This was already `PTero`-compatible, but now it doesn't depend on `Workflow`.


ebelter:
+1

susannasiebert:
+1

PR #1339, opened 2016-03-30
Use inline workflow backend instead of NO_LSF environment variable.

tmooney:
The `NO_LSF` variable is dependent on Workflow to succeed.  (This was the only place outside of a test where it was still being used.)


jeldred:
+1 :sheep:

sleongmgi:
+1

PR #1340, opened 2016-03-31
Add LSF ID to the Ptero workflow display.

tmooney:
Now that PTero is configured to pass the job ID back to the execution (per PR #1325), we can get this info without an additional API call per job.  Instead it's in the info we're already pulling back!

I promise I updated the `else` cases this time!  (cf. PR #1329)


ebelter:
+1

susannasiebert:
+1

PR #1341, opened 2016-03-31
Remove "ContaminationScreen" GMT tree.

tmooney:
This is another case of deletion as the alternative to updating to `WorkflowBuilder`.

(No recorded uses of this were found in the entirety of the usage log.)


jeldred:
bitmoji

tmooney:
Is that a "+1"?

jeldred:
:+1: :shipit:

ebelter:
+1 :light_rail:

PR #1344, opened 2016-04-06
Model: created_by/run_as/creation_date are required.

tmooney:
These are normally filled automatically in during creation.  Bad things happen if they're absent! (For instance, BQM becomes very confused.)

This change should ideally be followed by a database constraint to enforce these.


brummett:
+1

acoffman:
+1

PR #1345, opened 2016-04-07
change the bsub and wait function

sleongmgi:
change the bsub and wait to use the -K LSF flag instead of doing LSF instead of pre/post callback.

### DIFF
--- a/lib/perl/Genome/Site/TGI/Extension/Sys-lsf_api.t
+++ b/lib/perl/Genome/Site/TGI/Extension/Sys-lsf_api.t
@@ -57,7 +57,7 @@ subtest 'get job statuses' => sub {
 };
 
 subtest 'bsub_and_wait_for_completion' => sub {
-    plan tests => 7;
+    plan tests => 1;
 
     my @cmds = (['ls', '~'],
                 'exit 1',
@@ -67,15 +67,9 @@ subtest 'bsub_and_wait_for_completion' => sub {
     my @statuses = Genome::Sys->bsub_and_wait_for_completion(
                         queue => Genome::Config::get('lsf_queue_short'),
                         cmds => \@cmds,
-                        on_submit => sub { my($idx, $job_id) = @_; $submitted{$idx} = $job_id },
-                        on_complete => sub { my($idx, $job_id) = @_; $completed{$idx} = $job_id },
                     );
     is_deeply(\@statuses,
               [SUCCESSFUL_JOB, FAILED_JOB, SUCCESSFUL_JOB],
               'statuses are correct');
 
-    for (my $i = 0; $i < @cmds; $i++) {
-        ok($submitted{$i}, "on_submit callback fired for command $i");
-        ok($completed{$i}, "on_complete callback fired for command $i");
-    }
 };
--- a/lib/perl/Genome/Site/TGI/Extension/Sys.pm
+++ b/lib/perl/Genome/Site/TGI/Extension/Sys.pm
@@ -78,10 +78,6 @@ sub bsub_and_wait_for_completion {
     ($special_args, %bsub_args) = _bsub_and_wait_for_completion__validate_args(%bsub_args);
     my($commands, $on_submit_cb, $on_complete_cb) = @$special_args{'commands','on_submit','on_complete'};
 
-    my $hostname = hostname;
-    my $waiting_socket = IO::Socket::INET->new(Listen => 5, Proto => 'tcp', LocalHost => $hostname);
-    my $port = $waiting_socket->sockport;
-
     my %seq_to_job_id;
     for(my $seq = 0; $seq < @$commands; $seq++) {
         my $cmd = $commands->[$seq];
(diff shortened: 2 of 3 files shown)


apipe-tester:

apipe-tester:

acoffman:
CI, test this.

tmooney:
Hey CI, test this.  Thanks!

lib/perl/Genome/Site/TGI/Extension/Sys.pm
```diff
@@ -128,6 +123,15 @@ sub _bsub_and_wait_for_completion__validate_args {
...
+sub _bsub_and_wait_for_completion__wait_on_jobs_with_lsf_dependency {
+    my ($class, $job_ids, $bsub_args) = @_;
+    local $ENV{LSF_NIOS_JOBSTATUS_INTERVAL} = 1;
+    $class->bsub(%$bsub_args,
+                 wait_for_completion => 1,
+                 depend_on => (join "&&", map { "ended($_)" } @$job_ids) ,
+                 cmd => "echo DONE");
+}
+
 sub _bsub_and_wait_for_completion__wait_on_jobs {
```
tmooney: If we're removing the part that sets up a socket, we should probably also remove this subroutine that uses the socket.

lib/perl/Genome/Site/TGI/Extension/Sys.pm
```diff
@@ -92,15 +88,14 @@ sub bsub_and_wait_for_completion {
             %this_cmd_bsub_args = ( cmd => $cmd );
         }
 
-        my $job_id = $class->bsub(post_exec_cmd => qq(echo $seq | netcat $hostname $port),
+        my $job_id = $class->bsub(
                                   %bsub_args,
                                   %this_cmd_bsub_args,
                                 );
-        $on_submit_cb->($seq, $job_id) if $on_submit_cb;
```
tmooney: Why are we removing this feature?  It's still used [in Genome::Model::Build::Unarchive](https://github.com/genome/genome/blob/f4f9c91170a5067476829fa65345cb45dd6d26b9/lib/perl/Genome/Model/Build/Command/Unarchive.pm#L195).

lib/perl/Genome/Site/TGI/Extension/Sys.pm
```diff
@@ -78,10 +78,6 @@ sub bsub_and_wait_for_completion {
     ($special_args, %bsub_args) = _bsub_and_wait_for_completion__validate_args(%bsub_args);
     my($commands, $on_submit_cb, $on_complete_cb) = @$special_args{'commands','on_submit','on_complete'};
```
tmooney: If we're removing the `on_complete` callback support, we should clean this and the `__validate_args` subroutines up to no longer mention it as well.

brummett:
Something to watch out for... If the blocking LSF job (the one dependent on the other worker jobs) doesn't get scheduled (maybe from fair queuing restrictions or maybe the queues are just plain full), then work won't proceed.

lib/perl/Genome/Site/TGI/Extension/Sys.pm
```diff
@@ -128,6 +123,15 @@ sub _bsub_and_wait_for_completion__validate_args {
     return(\%special_args, %bsub_args);
 }
 
+sub _bsub_and_wait_for_completion__wait_on_jobs_with_lsf_dependency {
+    my ($class, $job_ids, $bsub_args) = @_;
+    local $ENV{LSF_NIOS_JOBSTATUS_INTERVAL} = 1;
+    $class->bsub(%$bsub_args,
```
tmooney: I don't think it's appropriate to reuse all the same bsub arguments for this job.

lib/perl/Genome/Site/TGI/Extension/Sys.pm
sleongmgi: I agreed.

lib/perl/Genome/Site/TGI/Extension/Sys.pm
sleongmgi: Yes, this should not be removed.

lib/perl/Genome/Site/TGI/Extension/Sys.pm
sleongmgi: What do you mean "work won't proceed"?  Do you mean the worker jobs won't proceed?

sleongmgi: I think I just need the bsub queue argument.

brummett: By "work won't proceed", I mean the original process calling bsub_and_wait_for_completion() will stay blocked there because the blocking process (the new LSF job dependent on all the worker jobs) isn't getting scheduled.  We've had this problem in the past where work gets deadlocked because an LSF job is waiting on another LSF job with equal or lower scheduling priority.

lib/perl/Genome/Site/TGI/Extension/Sys-lsf_api.t
```diff
@@ -67,15 +67,9 @@ subtest 'bsub_and_wait_for_completion' => sub {
     my @statuses = Genome::Sys->bsub_and_wait_for_completion(
                         queue => Genome::Config::get('lsf_queue_short'),
                         cmds => \@cmds,
-                        on_submit => sub { my($idx, $job_id) = @_; $submitted{$idx} = $job_id },
```
tmooney: We could keep the test for `on_submit` as well.

tmooney:
Ci, test this.

tmooney:
This has been replaced with #1347.

PR #1346, opened 2016-04-07
Use Genome::Utility

gatoravi:
This fixes issues in ClinSeq where the path to the junctions file is currently incorrectly resolved.

### DIFF
--- a/lib/perl/Genome/Model/ClinSeq/Command/TophatJunctionsAbsolute.pm
+++ b/lib/perl/Genome/Model/ClinSeq/Command/TophatJunctionsAbsolute.pm
@@ -90,21 +90,14 @@ sub execute {
     my $build_id  = $rnaseq_build->id;
     my $build_dir = $rnaseq_build->data_directory;
     my $model     = $rnaseq_build->model;
-    my $ab        = $model->annotation_build;
-    my $ab_name   = Genome::Utility::Text::sanitize_string_for_filesystem($ab->name);
-    my $test_path   = $working_dir . $ab_name . ".Junction.GeneExpression.top*percent.tsv";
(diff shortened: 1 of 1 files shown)


tmooney:
This should `use Genome::Utility::Text;`  The Autoloader doesn't like to load modules via `::` rather than `->` calls.

Otherwise, +1 (assuming tests pass)!

gatoravi:
Sorry I keep forgetting this, I've added this in now.

acoffman:
+1

gatoravi:
If I'm reading this correctly, there's a mixture of RNAseq builds out there, one set where the junctions filenames look like `NCBI-human.ensembl-67_37l_v2*` and another set that looks like `NCBI-human.ensembl_67_37l_v2.` The Genome::Utility method seems to resolve to the latter, the previous perl sub line in ClinSeq used to refer to the former. Any thoughts on how to make this backwords compatible? Should this be handled in Genome::Utility or in ClinSeq?

jasonwalker80:
Can you provide a list or query that demonstrates the mixed behavior?  Was there a specific commit or genome version where this can be traced back to?

gatoravi:
The test failure in this pull-request is due to one such RNA-seq build I think. Yeah I could add a further check in ClinSeq, I just wanted to make sure this is expected.

gatoravi:
Hi sorry for the trouble, this PR will need another round of review since I just pushed a couple of commits to address the test-failures.

jasonwalker80:
+1 given the mixed bag of naming conventions.

tmooney:
+1.

I wonder if it'd be better to try the `Genome::Utility::Text` version first, but it probably is of little consequence.

PR #1347, opened 2016-04-07
Change bsub wait for completion

sleongmgi:


apipe-tester:

apipe-tester:

tmooney:
Dear CI, test this.  Thanks!

lib/perl/Genome/Site/TGI/Extension/Sys.pm
```diff
@@ -128,21 +126,19 @@ sub _bsub_and_wait_for_completion__validate_args {
...
-        $fh->close;
-        $seq =~ s/\r|\n//g;
-
-        my $job_id = delete $seq_to_job_id{$seq};
-        unless ($job_id) {
-            Carp::croak("Got unexpected response '$seq' while waiting for these jobs:\n",
-                join("\n", map { join(' => ', $_, $seq_to_job_id{$_}) } keys %seq_to_job_id));
+sub _bsub_and_wait_for_completion__wait_on_jobs_with_lsf_dependency {
+    my ( $class, $seq_to_job_id, $bsub_args, $on_complete_cb ) = @_;
+    local $ENV{LSF_NIOS_JOBSTATUS_INTERVAL} = 1;
```
tmooney: What is the significance of overriding a user's config to set this environment variable to `1` inside of here?

lib/perl/Genome/Site/TGI/Extension/Sys.pm
```diff
@@ -92,15 +88,17 @@ sub bsub_and_wait_for_completion {
...
                                   %bsub_args,
                                   %this_cmd_bsub_args,
                                 );
         $on_submit_cb->($seq, $job_id) if $on_submit_cb;
         $seq_to_job_id{$seq} = $job_id;
     }
 
-    $class->_bsub_and_wait_for_completion__wait_on_jobs($waiting_socket, $on_complete_cb, %seq_to_job_id);
+    $class->_bsub_and_wait_for_completion__wait_on_jobs_with_lsf_dependency( \%seq_to_job_id,
+        { queue => $bsub_args{queue} || 'short' },
```
tmooney: This should not hardcode an LSF queue.  We have:
```perl
Genome::Config::get('lsf_queue_short')
```

I would advocate that this always use that queue instead of whatever queue was used for the other jobs.  (For instance, this waiting job needn't go to the `hpss` queue.)

lib/perl/Genome/Site/TGI/Extension/Sys.pm
sleongmgi: This is the flag goes with the LSF -K to do pulling.  The user don't set it usually.  I think this is to prevent the -K to wait forever.  Actually, the docker blade set this automatically now.  This is just to make sure it is set in case docker blades miss it.

lib/perl/Genome/Sys.pm
```diff
@@ -940,7 +940,16 @@ sub set_gid {
 
 sub gidgrnam {
     my $group_name = shift;
-    (getgrnam($group_name))[2];
+
```
tmooney: It seems like this commit belongs in a different pull request.

lib/perl/Genome/Site/TGI/Extension/Sys.pm
tmooney: Setting it in the docker container makes sense, since it's required to make this succeed in the docker container.  I don't think setting it here makes sense, since it's not necessary outside of docker (or possibly even in the future when we're granted better support for opening ports).

sleongmgi: The flag is very important.  It might cause problem if it is not there.  Currently, it is here to be fail safe until we can trust the flag is always set on the blade.

lib/perl/Genome/Sys.pm
sleongmgi: This problem is happening on the bsub job.

tmooney: If it's the new LSF job that you created to wait for all the others, then that makes sense!  If it's something the test was already `bsub`bing, which found this different problem, then I believe it's better as a different PR.  (At this point it's already in here, so I wouldn't worry about changing it!)

lib/perl/Genome/Site/TGI/Extension/Sys.pm
tmooney: If it is important (to us) and we cannot trust our machines, maybe we should set it in `Genome::Site::TGI` instead of here?  That way all Genome code internally gets the benefit, and any theoretical external users aren't affected.

tmooney: (I see this is already in our `Site::TGI` modules, so this is less important than I previously thought.  I won't worry about it.)

tmooney:
+1 :checkered_flag:

brummett:
+1

PR #1348, opened 2016-04-07
DetermineError Handles PTero Builds

tmooney:
This is one of the primary means for users to self-diagnose failures these days, so it'd be nice if it works should they run their builds wtih PTero!


jeldred:
👍 🌐

jasonwalker80:
+1

PR #1349, opened 2016-04-07
View and Diff Commands for QC Config

ebelter:


lib/perl/Genome/Qc/Config.pm
```diff
@@ -36,6 +37,10 @@ class Genome::Qc::Config {
     id_generator => '-uuid',
 };
 
+sub __display_name__ { sprintf('%s for %s (%s)', $_[0]->name, $_[0]->type, $_[0]->id); }
+
+sub config_to_yaml { Dump decode_json $_[0]->config; }
```
tmooney: Should `Genome::Qc::Command::Config::SaveToFile` now call this?

ebelter: I could ask :laughing:  I'll take a look!

ebelter:
@tmooney Save config uses config_to_yaml and I added a view cmd test.

ebelter:
CI, test this. Por favor.

tmooney:
+1 :truck:

PR #1350, opened 2016-04-08
Add build38 options for downloading SV annotation files.

jasonwalker80:
Add build38 options for downloading SV annotation files.

### DIFF
--- a/lib/perl/Genome/Db/Ucsc/Command/Base.pm
+++ b/lib/perl/Genome/Db/Ucsc/Command/Base.pm
@@ -39,6 +39,14 @@ class Genome::Db::Ucsc::Command::Base {
             default_value => 'genome-mysql.cse.ucsc.edu',
         },
     ],
+    has => [
+        sort_position => {
+            is => 'Text',
+            doc => 'The position string used for the linux sort.',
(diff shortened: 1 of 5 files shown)


tmooney:
+1; I didn't check that the chosen sort columns are the right ones, but if they are this looks like it'll use them!

jasonwalker80:
@gatoravi could you take a look at this PR for the SV annotation issue.

gatoravi:
+1
I verified this works by running ` genome db ucsc gap-list  --filename=/tmp/gaplist.csv --reference-name=hg18` this seems to work well. However when I try `genome db ucsc gap-list  --filename=/tmp/gaplist.csv --reference-name=hg38` it seems to result in an empty file and a `ERROR: Empty file passed to sort.`, I hope this is expected.
Also, just as a note to self, the sort columns are different for the `gaplist` module since this module does not request the `bin` column like the other two.

jasonwalker80:
I pushed a commit to add the hg38 table name for the gap-list command.  Nice catch.

gatoravi:
Yeah the `gap-list` command works well now! 👍

PR #1351, opened 2016-04-11
Remove some old Perl 5.8 tests.

tmooney:
Dependencies on the Ptero Perl SDK are now in `Genome::Model::Build`, and the Ptero Perl SDK is only installed for the system Perl.


brummett:
+1
What changed to make these stop working?

tmooney:
In #1312 `Genome::Model::Build` started using some `Ptero::` modules, which aren't installed for 5.8.  These tests didn't show up as affected since they shell out, so only the "full" tests caught them.

acoffman:
+1

PR #1352, opened 2016-04-11
Remove compile-time dependency on Genome::Model::Build.

tmooney:
This lets us keep around the `gmt lims` tree awhile longer.  Alternatively, we could remove that and not make this change.


ebelter:
If you're asking me, I say remove the 'gmt lims' tree. If it is needed, I am creating a new namespace for reference improvement that these can be ported to.

tmooney:
@ebelter Fair enough! I've opened #1353 with that change.  I'm happy to merge whichever one gets to two +1s :smile:

PR #1353, opened 2016-04-11
Remove `gmt lims` tree.

tmooney:
This is an alternative to #1352.  It removes our need to continue running things on Perl 5.8.


ebelter:
+2 💥

PR #1354, opened 2016-04-12
MergeAlignments: Specify an LSF queue configuration.

tmooney:
Production jobs were ending up in the default queue, which is suboptimal.


jasonwalker80:
+1

ebelter:
+1 🍼

PR #1355, opened 2016-04-12
Allow apipe-builder to use AbandonAndQueue.

tmooney:
This used to inspect to see which of its subclasses it was, which is not ideal.  I wanted to add support for "AbandonAndQueue", but it turned out this made the list match all known subclasses, so it's easier to remove the check.


ebelter:
+1

acoffman:
+1

PR #1356, opened 2016-04-12
fix the path.

sleongmgi:
Fix the old path to the new fasta file path.


tmooney:
+1

By the by, for anyone else looking at this:  The test data is being updated because the old version contains references to a volume not available within docker.

acoffman:
+1

PR #1360, opened 2016-04-12
Remove Parallel Annotator.

tmooney:
This was previously used in a refalign event, but that was removed as part of PR #1263.  At any rate, historically using this tool led to more problems than solutions since it just meant we rolled the dice more times on failing per annotation attempt.


jasonwalker80:
+1

dufeiyu:
+1

PR #1364, opened 2016-04-14
WIP: Refactor ClinSeq workflow to use SnvIndelReport instead of ImportSnvsIndels

### LINKED ISSUE #1363: Refactor ClinSeq workflow to use SnvIndelReport instead of ImportSnvsIndels

susannasiebert:
I made this a WIP because this PR results in changes to the ClinSeq build outputs so a new blessed build is needed. The diffs have been documented here: https://jira.gsc.wustl.edu/browse/CI-99. Once I get the ok for the diffs, a new blessed build will become part of this PR.

For the time being I wanted to open up the code review discussion.

This PR closes #1363.

### DIFF
--- a/lib/perl/Genome/Model/ClinSeq.pm
+++ b/lib/perl/Genome/Model/ClinSeq.pm
@@ -781,7 +781,7 @@ sub _resolve_workflow_for_build {
     #Converge SnvIndelReport
     #CreateMutationDiagrams - Create mutation spectrum results for wgs and exome data
     #GenerateSciClonePlots - Run clonality analysis and produce clonality plots
-    my @converge_snv_indel_report_ops;
+    my ($best_converge_snv_indel_report_op, $best_mq_bq_sum) = (undef, -Inf);
     if ($build->wgs_build || $build->exome_build) {
         my %variant_sources_ops = (
             exome => $exome_variant_sources_op,
@@ -810,7 +810,10 @@ sub _resolve_workflow_for_build {
                     destination_property => 'converge_snv_indel_report_result',
                 );
             }
-            push @converge_snv_indel_report_ops, $converge_snv_indel_report_op;
+            if (($mqs->[$i] + $bqs->[$i]) > $best_mq_bq_sum) {
+                $best_mq_bq_sum = $mqs->[$i] + $bqs->[$i];
+                $best_converge_snv_indel_report_op = $converge_snv_indel_report_op;
+            }
             if ($build->wgs_build or $build->should_run_exome_cnv) {
                 my $sciclone_op = $self->sciclone_op($workflow, $i);
                 if ($build->wgs_build) {
@@ -849,14 +852,14 @@ sub _resolve_workflow_for_build {
         #AnnotateGenesByCategory - gene_category_<exome|wgs|wgs_exome>_<snv|indel>_result
         my $annotate_genes_by_category_op = $self->annotate_genes_by_category_op($workflow);
         $workflow->create_link(
-            source               => $converge_snv_indel_report_ops[-1],
+            source               => $best_converge_snv_indel_report_op,
             source_property      => 'final_filtered_coding_clean_tsv',
             destination          => $annotate_genes_by_category_op,
             destination_property => 'infile',
         );
         my $annotate_genes_by_dgidb_op = $self->annotate_genes_by_dgidb_op($workflow, 'final_filtered_coding_clean_tsv', 'dgidb_snv_indel_result');
         $workflow->create_link(
-            source               => $converge_snv_indel_report_ops[-1],
+            source               => $best_converge_snv_indel_report_op,
             source_property      => 'final_filtered_coding_clean_tsv',
             destination          => $annotate_genes_by_dgidb_op,
             destination_property => 'input_file',
@@ -892,7 +895,7 @@ sub _resolve_workflow_for_build {
             );
         }
         $workflow->create_link(
-            source               => $converge_snv_indel_report_ops[-1],
+            source               => $best_converge_snv_indel_report_op,
             source_property      => 'final_filtered_coding_tsv',
             destination          => $mutation_diagram_op,
             destination_property => 'annotated_variants_tsv',
@@ -932,7 +935,7 @@ sub _resolve_workflow_for_build {
             );
         }
         $workflow->create_link(
-            source               => $converge_snv_indel_report_ops[-1],
+            source               => $best_converge_snv_indel_report_op,
             source_property      => 'final_filtered_coding_clean_tsv',
             destination          => $make_circos_plot_op,
             destination_property => 'annotated_variants_tsv',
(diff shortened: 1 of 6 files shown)


vendor/pindel
```diff
@@ -1 +1 @@
-Subproject commit baea5200199ea80c1ae0f7710537508d7dbee267
+Subproject commit 9e46d06c7d5615bcd1daa59f0e6a87238e9316a8
```
tmooney: Is this change meant to be here?

susannasiebert: It is not. I removed it. Good catch.

lib/perl/Genome/Model/ClinSeq/Command/MakeCircosPlot.pm
```diff
@@ -963,43 +972,63 @@ EOS
...
+            $genes_AmpDel{$4}="hs$1\t$2\t$3";
+            print $snv_fh "hs$1 $2 $3 0.5 fill_color=black\n";
+            #        switch($5){
+            #            case "nonsense"            {print $snv_fh "fill_color=goldenrod\n"}
+            #            case "missense"            {print $snv_fh "fill_color=blue\n"}
+            #            case "silent"            {print $snv_fh "fill_color=green\n"}
+            #            case "splice_site"        {print $snv_fh "fill_color=black\n"}
+            #            case "rna"                {print $snv_fh "fill_color=purple\n"}
+            #            case "nonstop"            {print $snv_fh "fill_color=black\n"}
+            #        }
```
tmooney: Do we need to keep this commented-out code around?  (If it's not needed, we could also get rid of `use Switch;` at the top.)

lib/perl/Genome/Model/ClinSeq.pm
```diff
@@ -1000,6 +845,104 @@ sub _resolve_workflow_for_build {
                 );
             }
         }
+        #Add gene category annotations to some output files from steps above. (e.g. determine which SNV affected genes are kinases, ion channels, etc.)
+        #AnnotateGenesByCategory - gene_category_<exome|wgs|wgs_exome>_<snv|indel>_result
+        my $annotate_genes_by_category_op = $self->annotate_genes_by_category_op($workflow);
+        $workflow->create_link(
+            source               => $converge_snv_indel_report_ops[-1],
```
tmooney: What is the special significance of the last converge operation that it's the one used for all subsequent steps?

lib/perl/Genome/Model/ClinSeq/Command/MakeCircosPlot.pm
susannasiebert: That's a very good question. @gatoravi what do you think?

lib/perl/Genome/Model/ClinSeq.pm
susannasiebert: It's our naive attempt at picking the snv indel report that was processed with the highest minimum mapping qualities. sireport_min_mq_bq is supposed to be a ''Comma separated increasing-list of minimum mapping qualities for bam-readcounting." Although I don't believe that that is enforced in any way. I wonder if the parse_qualities should somehow attempt to sort the quality cutoffs. @gatoravi @jasonwalker80 what do you think?

tmooney: If we don't want to sort them, the `for` loop could decide to only keep track of the job for the highest quality threshold, something like:
```perl
my ($best_converge_op, $best_converge_op_quality) = (undef, -Inf);  #or whatever names
for my $i (1..@$mq) {
   ...
   if ($mq[$i] > $best_converge_op_quality) {
     $best_converge_op_quality = $mq[$i];
     $best_converge_op = $converge_snv_indel_report_op;
   }
}
```

susannasiebert: I like this idea. The qualities are actually a set of the mapping quality (mq) and the base quality (bq). So I'm not sure if just looking at the mq is sufficient. @gatoravi what would be the best way to determine the "best" set of mapping and base quality in the sireport_min_mq_bq list? Is looking at just the mqs and picking the highest sufficient?

jasonwalker80: Perhaps the sum of the two scores would work?

lib/perl/Genome/Model/ClinSeq/Command/MakeCircosPlot.pm
gatoravi: I'd say drop this since we can always look at git history later. @bainscou to confirm!

lib/perl/Genome/Model/ClinSeq.pm
gatoravi: I think determining the best set of mapping and base qualities is a project in itself and I don't think we've done any formal evaluations in house based on simulated/validation data.
I think it's safe to pick highest cutoffs for base and mapping qualities and move ahead with those, with the current default processing profiles that would be a base-quality cutoff of 20 and mapping-quality cutoff of 30 which are not outrageous. If the processing profiles had incredibly high cutoffs the user would not see any variants downstream which is perhaps acceptable behavior. 
TL;DR: I like the idea of using the max, I want to make sure @malachig and @obigriffith are ok with this since this would affect some of the results we see.

lib/perl/Genome/Model/ClinSeq/Command/MakeCircosPlot.pm
gatoravi: Ben indicates over personal communication(slack) that it's ok to drop these commented out lines of code.

susannasiebert: Sounds good. I removed these lines.

lib/perl/Genome/Model/ClinSeq.pm
susannasiebert: I went ahead and pushed a commit that implements Tom's suggestion but using the sum as Jason mentioned. We can change it if @malachig or @obigriffith have an alternate approach in mind.

jasonwalker80:
+1 once new blessed build approved

tmooney:
+1 for what's here so far!

susannasiebert:
@jasonwalker80 @tmooney  I just pushed a few commits to fix some bugs as well as update the blessed ClinSeq build. This would be ready for release.

lib/perl/Genome/Model/ClinSeq.pm
```diff
@@ -966,6 +810,10 @@ sub _resolve_workflow_for_build {
                     destination_property => 'converge_snv_indel_report_result',
                 );
             }
+            if (($mqs->[$i-1] + $bqs->[$i-1]) > $best_mq_bq_sum) {
+                $best_mq_bq_sum = $mqs->[$i-1] + $bqs->[$i-1];
```
tmooney: Would it be possible to change the range of `$i` to `(0..$#$mqs)` to avoid these `-1`s?  Or maybe a variable that holds `$i-1` to give meaning to the repeated math?

tmooney:
+1, the new blessed build is in fact a build!

jasonwalker80:
+1

PR #1365, opened 2016-04-14
Annotator updates for Ensembl 84

jasonwalker80:
The annotator is being used less and less, but it's still relied upon by parts of the GMS.  This PR includes additional annotation biotypes and statuses for Ensembl 84 annotation.  Moving the creation of the biotype_class to a class method should help with troubleshooting future annotation builds.


tmooney:
+1, those are quite the lists!

brummett:
+1

PR #1366, opened 2016-04-14
SmallRna uses WorkflowBuilder

tmooney:
This pipeline hasn't been run in awhile and the workflow modules themselves have been mostly neglected since 2012, but I heartell it's not on the list for removal.  So, here's a minimal change to make it use `WorkflowBuilder` instead of `Workflow` (and convert tabs to spaces!  Use `?w=1` for easier diff viewing).


jasonwalker80:
+1

acoffman:
+1

PR #1367, opened 2016-04-18
Add wgs/exome only if input builds exist

gatoravi:
Currently if a clin-seq build only has a exome_model as an input,
the wgs version of create-mutation-spectrum is still added as a
step in the workflow. This pull-request hopes to fix that.

### DIFF
--- a/lib/perl/Genome/Model/ClinSeq.pm
+++ b/lib/perl/Genome/Model/ClinSeq.pm
@@ -943,11 +943,10 @@ sub _resolve_workflow_for_build {
             exome => $exome_variant_sources_op,
             wgs   => $wgs_variant_sources_op,
         );
-        my $sequencing_types = $self->_get_sequencing_types($build);
         #Create a report for each $bq $mq combo.
         for my $i (1..@$mqs) {
             my $converge_snv_indel_report_op = $self->converge_snv_indel_report_op($workflow, $i);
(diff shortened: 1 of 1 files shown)


tmooney:
+1.

This seems fine, but it is a different approach than taken in the other things that iterate over sequencing types (e.g. [here](https://github.com/gatoravi/genome/blob/00711c6188d9707487ed6229908cdbea6d858308/lib/perl/Genome/Model/ClinSeq.pm#L757)).  I wonder if it'd be worthwhile to have a standard way of going about this?

gatoravi:
You're right Tom, I hadn't noticed that and this change seems redundant now,  I could just move the
```
    962                 my $create_mutation_spectrum_op = $self->create_mutation_spectrum_op($workflow, $sequencing_type, $i);
    963                 $workflow->create_link(
    964                     source               => $converge_snv_indel_report_op,
    965                     source_property      => 'result',
    966                     destination          => $create_mutation_spectrum_op,
    967                     destination_property => 'converge_snv_indel_report_result',
    968                 );
``` block up into the `if ($build->$build_accessor) {` block..

gatoravi:
Ok,  perhaps a little more consistent now.

tmooney:
+1, that seems better!

jasonwalker80:
+1

PR #1368, opened 2016-04-18
update the config file that uses sata420.

sleongmgi:
Update the config file path.


tmooney:
+1 assuming tests pass!

Interesting that the `ref_seq.txt` next to the `ref_small.txt` points to a non-existent location.  Does that file even need to be there?

brummett:
+1

PR #1369, opened 2016-04-19
RefAlign Summary Test Doesn't Lock

tmooney:
More accurately, it doesn't use the real lock location so multiple test runs don't collide.

I'd be swell to use `Genome::Config::set_env('site_lock_dir', Genome::Sys->create_temp_directory)` instead, but using `Genome` already takes us past the point where this value gets examined.

### DIFF
--- a/lib/perl/Genome/Model/ReferenceAlignment/Report/Summary.t
+++ b/lib/perl/Genome/Model/ReferenceAlignment/Report/Summary.t
@@ -3,9 +3,14 @@
 use strict;
 use warnings;
 
+use File::Temp;
+
+
 BEGIN {
     #use locks local to this test--must be set before Genome is loaded
-    $ENV{XGENOME_SITE_LOCK_DIR} = '/tmp';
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Model/ReferenceAlignment/Report/Summary.t
```diff
@@ -3,6 +3,11 @@
 use strict;
 use warnings;
 
+BEGIN {
+    #use locks local to this test--must be set before Genome is loaded
+    $ENV{XGENOME_SITE_LOCK_DIR} = '/tmp';
+}
+
```
brummett: To be extra safe, this should probably be something like ```/tmp/$$``` or a real tmpdir (and remove it after the test finishes).  Multiple copies _could_ get scheduled to run on the same machine at the same time.

tmooney: Good point!  I'll make a real tempdir out of this.

brummett:
+1

sleongmgi:
+1

PR #1370, opened 2016-04-19
fix the test that wait forever for input if the files do not exist.

sleongmgi:


lib/perl/Genome/Model/ReferenceAlignment/Report/Summary.pm
```diff
@@ -364,10 +364,13 @@ sub _generate_summary_information
     unless (@lq_files) {
         @lq_files = $build->get_variant_bed_file('snps_all_sequences');
     }
-    my $lq_snp_calls = `wc -l @lq_files | tail -n 1`;
-    $lq_snp_calls =~ s/\s\S+\s*$//i;
-    $lq_snp_calls =~ s/\s//g;
-    my $unfiltered_snp_calls = $filtered_snp_calls + $lq_snp_calls;
+    my ( $lq_snp_calls, $unfiltered_snp_calls );
```
tmooney: `$lq_snp_calls` isn't needed outside of the `if` block.

lib/perl/Genome/Model/ReferenceAlignment/Report/Summary.pm
```diff
@@ -364,10 +364,13 @@ sub _generate_summary_information
...
+    if (@lq_files) {
+        $lq_snp_calls = `wc -l @lq_files | tail -n 1`;
+        $lq_snp_calls =~ s/\s\S+\s*$//i;
+        $lq_snp_calls =~ s/\s//g;
+        $unfiltered_snp_calls = $filtered_snp_calls + $lq_snp_calls;
```
tmooney: `$unfiltered_snp_calls` should probably be initialized to `$filtered_snp_calls` rather than being left undefined when there are no `@lq_files`.  (Then this could be a `+=`.)

lib/perl/Genome/Model/ReferenceAlignment/Report/Summary.pm
```diff
@@ -474,6 +477,7 @@ sub get_contents {
 }
 
 sub commify {
+        return unless(@_ && defined $_[0]);
```
tmooney: This line uses a different spacing than the rest of the subroutine.

tmooney:
+1 🎆

acoffman:
+1

PR #1372, opened 2016-04-19
deal with warning message

sleongmgi:
Deal with the warning message.  Make the debugging message to the standard output to use the diag.

### DIFF
--- a/lib/perl/Genome/Sys/Lock/FileBackend.t
+++ b/lib/perl/Genome/Sys/Lock/FileBackend.t
@@ -24,6 +24,11 @@ use Socket qw(AF_UNIX SOCK_STREAM PF_UNSPEC);
 use Test::More;
 use Time::HiRes qw(gettimeofday);
 
+use Sub::Override;
+
+my $sub_override = Sub::Override->new;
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Sys/Lock/FileBackend.t
```diff
@@ -24,6 +24,11 @@ use Socket qw(AF_UNIX SOCK_STREAM PF_UNSPEC);
 use Test::More;
 use Time::HiRes qw(gettimeofday);
 
+use Sub::Override;
+
+my $sub_override = Sub::Override->new;
+$sub_override->replace('Genome::Logger::warning' => sub { diag($_[1]); }, 'Genome::Logger::info' => sub { diag($_[1]); });
+
```
tmooney: Alternatively, we could turn off printing logs to the screen:
```perl
my $logger = Genome::Logger->get();
$logger->screen(0);
```

sleongmgi: Yes.  I think it might be better to keep the original output.

tmooney: Fair enough!

tmooney:
+1 👌

brummett:
+1

PR #1373, opened 2016-04-20
New GMT : Manta structural variant caller

jasonwalker80:
Initial revision of new tool tree and commands to run the Manta SV caller.

### DIFF
--- a/lib/perl/Genome/Model/Tools/Manta.pm
+++ b/lib/perl/Genome/Model/Tools/Manta.pm
@@ -13,3 +13,6 @@ class Genome::Model::Tools::Manta {
 sub help_detail {
     "These commands are used to execute Manta and related processes";
 }
+
+
+1;
--- a/lib/perl/Genome/Model/Tools/Manta.pm
+++ b/lib/perl/Genome/Model/Tools/Manta.pm
@@ -1,0 +1,15 @@
+package Genome::Model::Tools::Manta;
+
+use strict;
+use warnings;
+
+use Genome;
+
+class Genome::Model::Tools::Manta {
+    is  => 'Command::Tree',
+    doc => 'Toolkit for executing the Manta structural variant call',
+};
+
+sub help_detail {
+    "These commands are used to execute Manta and related processes";
+}
--- a/lib/perl/Genome/Model/Tools/Manta/Base.pm
+++ b/lib/perl/Genome/Model/Tools/Manta/Base.pm
@@ -104,9 +104,12 @@ sub _tool_input_file_metas {
 # formatting (argname=argvalue).
 sub _format_tool_arg {
     my ($type, $name, $value) = @_;
-    return '--'. $name if ($type eq 'Boolean');
-    $value = qq{"$value"} if $type eq 'Text';
-    return sprintf('--%s %s', $name, $value);
+
+    if ($type eq 'Boolean') {
+        return '--'. $name;
+    } else {
+        return '--'.$name, $value;
+    }
 }
 
 # given a property meta object (defining property name, and ostensibly
@@ -201,27 +204,19 @@ sub _shellcmd_extra_params {
     return (
         skip_if_output_is_present => 0,
         keep_dbh_connection_open => 0,
-        set_pipefail => 0,
         input_files => \@input_files,
         output_files => \@output_files,
     );
 }
 
-sub build_cmdline_list {
+sub build_cmdline_array_ref {
     my $self = shift;
 
-    return (
+    return [
         $self->tool_path,
         $self->_tool_param_args,
         $self->_tool_input_args,
-    );
-}
-
-sub build_cmdline_string {
-    my $self = shift;
-
-    my $cmd = join(' ', $self->build_cmdline_list);
-    return $cmd;
+    ];
 }
 
 # override with care.
@@ -231,7 +226,7 @@ sub execute {
     $self->_validate_params;
 
     my %params = (
-        cmd => $self->build_cmdline_string,
+        cmd => $self->build_cmdline_array_ref,
         $self->_shellcmd_extra_params
     );
 
--- a/lib/perl/Genome/Model/Tools/Manta/Base.pm
+++ b/lib/perl/Genome/Model/Tools/Manta/Base.pm
@@ -6,6 +6,10 @@ use warnings;
 use Genome;
 use Carp qw(confess);
 
+my %TOOL_VERSIONS = (
+    '0.29.6'     => '/gscmnt/gc13001/info/model_data/jwalker_scratch/src/manta-0.29.6.centos5_x86_64/bin',
+);
+
 class Genome::Model::Tools::Manta::Base {
     is => 'Command',
     is_abstract => 1,
@@ -23,7 +27,7 @@ class Genome::Model::Tools::Manta::Base {
         version => {
             is => 'Version',
             doc => 'The version of Manta to use.',
-            valid_values => ['0.29.6'],
+            valid_values => [sort keys %TOOL_VERSIONS],
         },
     ],
     has_optional_output => [
@@ -38,15 +42,6 @@ class Genome::Model::Tools::Manta::Base {
 
 sub help_detail {
     "Manta calls structural variants (SVs) and indels from mapped paired-end sequencing reads."
-}
-
-my %TOOL_VERSIONS = (
-    '0.29.6'     => '/gscmnt/gc13001/info/model_data/jwalker_scratch/src/manta-0.29.6.centos5_x86_64/bin',
-);
-
-sub available_versions {
-    my $self = shift;
-    return keys(%TOOL_VERSIONS);
 }
 
 sub path_for_version {
--- a/lib/perl/Genome/Model/Tools/Manta/Base.pm
+++ b/lib/perl/Genome/Model/Tools/Manta/Base.pm
@@ -243,4 +243,5 @@ sub execute {
     return Genome::Sys->shellcmd(%params);
 }
 
+
 1;
--- a/lib/perl/Genome/Model/Tools/Manta/Base.pm
+++ b/lib/perl/Genome/Model/Tools/Manta/Base.pm
@@ -1,0 +1,246 @@
+package Genome::Model::Tools::Manta::Base;
+
+use strict;
+use warnings;
+
+use Genome;
+use Carp qw(confess);
+
+class Genome::Model::Tools::Manta::Base {
+    is => 'Command',
+    is_abstract => 1,
+    attributes_have => [
+        tool_arg_name => {
+            is => 'String',
+            is_optional => 1,
+        },
+        tool_input_file => {
+            is => 'Boolean',
+            is_optional => 1,
+        },
+    ],
+    has_param => [
+        version => {
+            is => 'Version',
+            doc => 'The version of Manta to use.',
+            valid_values => ['0.29.6'],
+        },
+    ],
+    has_optional_output => [
+        output_files => {
+            is_many => 1,
+        },
+        input_files => {
+            is_many => 1,
+        },
+    ],
+};
+
+sub help_detail {
+    "Manta calls structural variants (SVs) and indels from mapped paired-end sequencing reads."
+}
+
+my %TOOL_VERSIONS = (
+    '0.29.6'     => '/gscmnt/gc13001/info/model_data/jwalker_scratch/src/manta-0.29.6.centos5_x86_64/bin',
+);
+
+sub available_versions {
+    my $self = shift;
+    return keys(%TOOL_VERSIONS);
+}
+
+sub path_for_version {
+    my ($class, $version) = @_;
+    unless ($version) {
+        die('No version defined!');
+    }
+    my $path = $TOOL_VERSIONS{$version};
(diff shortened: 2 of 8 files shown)


lib/perl/Genome/Model/Tools/Manta.pm
```diff
@@ -0,0 +1,15 @@
...
+use Genome;
+
+class Genome::Model::Tools::Manta {
+    is  => 'Command::Tree',
+    doc => 'Toolkit for executing the Manta structural variant call',
+};
+
+sub help_detail {
+    "These commands are used to execute Manta and related processes";
+}
```
tmooney: This module is missing a true value (`1;`) on the last line.

(In a more nit-picky comment, some of the other files are missing newlines at the end of them.)

lib/perl/Genome/Model/Tools/Manta/Base.pm
```diff
@@ -0,0 +1,246 @@
...
+        tool_input_file => {
+            is => 'Boolean',
+            is_optional => 1,
+        },
+    ],
+    has_param => [
+        version => {
+            is => 'Version',
+            doc => 'The version of Manta to use.',
+            valid_values => ['0.29.6'],
```
tmooney: It might be nice if this could somehow rely on `&available_versions` (or `sort keys %TOOL_VERSIONS` directly) instead of repeating the values.

lib/perl/Genome/Model/Tools/Manta/Base.pm
```diff
@@ -0,0 +1,246 @@
...
+        },
+    ],
+};
+
+sub help_detail {
+    "Manta calls structural variants (SVs) and indels from mapped paired-end sequencing reads."
+}
+
+my %TOOL_VERSIONS = (
+    '0.29.6'     => '/gscmnt/gc13001/info/model_data/jwalker_scratch/src/manta-0.29.6.centos5_x86_64/bin',
```
tmooney: Are there plans to put this somewhere more standard, whether packaged, under `Genome::Config::get('sw')`, or elsewhere?

lib/perl/Genome/Model/Tools/Manta/Base.pm
```diff
@@ -0,0 +1,246 @@
...
+    return unless defined $self->$ur_name;
+
+    # This works for things that are is_many or not.
+    return map {$_} $self->$ur_name;
+}
+
+# command line args that come after the executable
+#
+# The default implementation just translates all the inputs that have
+# speedseq_param_name defined. Boolean values get translated to true/false,
```
tmooney: This reference to `speedseq` seems like a hold-over from `Genome::Model::Tools::Speedseq::Base`.

lib/perl/Genome/Model/Tools/Manta/Base.pm
```diff
@@ -0,0 +1,246 @@
...
+}
+
+# override with care.
+sub execute {
+    my $self = shift;
+
+    $self->_validate_params;
+
+    my %params = (
+        cmd => $self->build_cmdline_string,
```
tmooney: It'd be better if we use the list form of `shellcmd` for new things.

lib/perl/Genome/Model/Tools/Manta/Config.t
```diff
@@ -0,0 +1,33 @@
...
+my $cmdline_string = $config_manta->build_cmdline_string();
+is($cmdline_string,$expected_cmdline,'expected command-line for version');
+
+$config_manta->_resolve_input_files;
+my @input_files = $config_manta->input_files;                       
+is_deeply(\@input_files,\@expected_input_files,'expected input files');
+
+# TODO: Test the execute subroutine
+
+exit;
```
tmooney: I'm curious if these `exit`s are in here for a particular purpose.

lib/perl/Genome/Model/Tools/Manta/Run.pm
```diff
@@ -0,0 +1,63 @@
...
+use warnings;
+
+use Genome;
+
+class Genome::Model::Tools::Manta::Run {
+    is => 'Genome::Model::Tools::Manta::Base',
+    has_input => [
+        working_directory => {
+            is => 'Text',
+            doc => 'The same working directory used during the Config step.  The directory should contain the  script used to execute Manta',
```
tmooney: There are two spaces between `the` and `script`.

lib/perl/Genome/Model/Tools/Manta/Base.pm
jasonwalker80: I presumed this was the modus operandi based on another request for software install:
https://jira.gsc.wustl.edu/browse/INFOSYS-19053

Packaging seems overly complicated especially given the move toward containers in the future.
I haven't checked in years, but I doubt I have permissions to write to : `Genome::Config::get('sw')`
If there is a better, more centralized location, I'm all for moving this wherever it belongs.

tmooney: Well, it sounds like we're not allowed to do this in a good way anymore... so carry on!

lib/perl/Genome/Model/Tools/Manta/Base.pm
jasonwalker80: Do you have an example of where the list form is successfully used?  With my attempt the command-line is quoted improperly in `Genome::Sys->quote_for_shell`:

ex: 
`/gscmnt/gc13001/info/model_data/jwalker_scratch/src/manta-0.29.6.centos5_x86_64/bin/configManta.py '--normalBam "/gscmnt/gc9022/info/build_merged_alignments/merged-alignment-blade13-4-5.gsc.wustl.edu-jwalker-25191-5e2355d1bb8c4ff78279ddb3947fc0d3/5e2355d1bb8c4ff78279ddb3947fc0d3.bam"' '--referenceFasta "/gscmnt/ams1102/info/model_data/2869585698/build106942997/all_sequences.fa"' '--tumorBam "/gscmnt/gc8002/info/build_merged_alignments/merged-alignment-blade13-2-11.gsc.wustl.edu-jwalker-27691-f1e64d1802b0493095d70b837a467f47/f1e64d1802b0493095d70b837a467f47.bam"' '--runDir "/tmp/"'`

tmooney: Looks like the list implementation in this tool is broken.  Specifically `_format_tool_arg` makes a string instead of building up a list of arguments.

lib/perl/Genome/Model/Tools/Manta/Base.pm
mcallaway: > On Apr 20, 2016, at 2:10 PM, Thomas B. Mooney <notifications@github.com> wrote:
>
> In lib/perl/Genome/Model/Tools/Manta/Base.pm:
>
> > +    has_optional_output => [
> > +        output_files => {
> > +            is_many => 1,
> > +        },
> > +        input_files => {
> > +            is_many => 1,
> > +        },
> > +    ],
> > +};
> > +
> > +sub help_detail {
> > +    "Manta calls structural variants (SVs) and indels from mapped paired-end sequencing reads."
> > +}
> > +
> > +my %TOOL_VERSIONS = (
> > +    '0.29.6'     => '/gscmnt/gc13001/info/model_data/jwalker_scratch/src/manta-0.29.6.centos5_x86_64/bin',
>
> Well, it sounds like we're not allowed to do this in a good way anymore... so carry on!

Can you make paths to executables configurable?

my %TOOL_VERSIONS = (
    '0.29.6'     => ${GENOME_TOOL_EXEC_PATH_MANTA}

/etc/profile.d/genome.sh:
GENOME_TOOL_EXEC_PATH_MANTA=/gscmnt/gc13001/info/model_data/jwalker_scratch/src/manta-0.29.6.centos5_x86_64/bin

OR:

/etc/genome/config.yaml:
GENOME_TOOL_EXEC_PATH_MANTA: /gscmnt/gc13001/info/model_data/jwalker_scratch/src/manta-0.29.6.centos5_x86_64/bin





________________________________
The materials in this message are private and may contain Protected Healthcare Information or other information of a sensitive nature. If you are not the intended recipient, be advised that any unauthorized use, disclosure, copying or the taking of any action in reliance on the contents of this information is strictly prohibited. If you have received this email in error, please immediately notify the sender via telephone or return mail.

tmooney: Sure!  It could be in configuration.  One nice thing about the packaging is we could slightly better guarantee the correct version is being used.  This way we're trusting the user to get it right.

sleongmgi: Since we have the docker queue, should we use the manta docker image instead?

jasonwalker80: @mcallaway the config env variable would have to be specific to each version, ex. `GENOME_TOOL_EXEC_PATH_MANTA_0.29.6` in order to support multiple versions of a tool.  Encoding the software version in the env doesn't feel right.  Using a container rather than packaging seems like a better option, but see comment below.

@sleongmgi please provide an example of how to setup a GMT to run in a docker container inside of DV2 as part of a genome model build.  Using a docker container for the first time seems outside the scope of this PR.  Feel free to submit a pull request to my branch if it's something simple that I'm overlooking.  I'm all for utilizing docker if it's straightforward, but developing the GMS to support it is again outside the scope of this issue, ie. replacing breakdander for SVs.

sleongmgi: Shown below is an example changes I made to demonstrate the use of the manta docker image.

diff --git a/lib/perl/Genome/Model/Tools/Manta/Base.pm b/lib/perl/Genome/Model/Tools/Manta/Base.pm
index 534ace1..cb98198 100644
--- a/lib/perl/Genome/Model/Tools/Manta/Base.pm
+++ b/lib/perl/Genome/Model/Tools/Manta/Base.pm
@@ -7,7 +7,7 @@ use Genome;
 use Carp qw(confess);
 
 my %TOOL_VERSIONS = (
-    '0.29.6'     => '/gscmnt/gc13001/info/model_data/jwalker_scratch/src/manta-0.29.6.centos5_x86_64/bin',
+    '0.29.6'     => 'bsub -Is -q docker -a "docker(registry.gsc.wustl.edu/sleong/manta:0.29.6)" /usr/local/pkg/manta-0.29.6.centos5_x86_64/bin/',
 );
 
 class Genome::Model::Tools::Manta::Base {
@@ -56,7 +56,7 @@ sub path_for_version {
 
 sub tool_path {
     my $self = shift;
-    return File::Spec->join($self->path_for_version($self->version),$self->_tool_subcommand_name);
+    return $self->path_for_version($self->version) . $self->_tool_subcommand_name;
 }
 
 # tool subcommand name as a string (e.g., 'configManta.py')

jasonwalker80: And how would I pass along other LSF resource requirements like RAM, temp space, CPUs, etc. to the docker queue?  Are those not needed?  Is the env of the existing LSF job somehow recreated in the docker queue?

jasonwalker80:
@tmooney  checkout the latest commits to address the list form of shellcmd.

lib/perl/Genome/Model/Tools/Manta/Base.pm
sleongmgi: It is the same as using normal lsf queue.  For example,

`   bsub -R 'select[mem>16000] rusage[mem=16000]' -M 16000000 -Is -q docker -a "docker(registry.gsc.wustl.edu/sleong/manta:0.29.6)" /usr/local/pkg/manta-0.29.6.centos5_x86_64/bin/configManta.py --normalBam "/gscmnt/gc9022/info/build_merged_alignments/merged-alignment-blade13-4-5.gsc.wustl.edu-jwalker-25191-5e2355d1bb8c4ff78279ddb3947fc0d3/5e2355d1bb8c4ff78279ddb3947fc0d3.bam" --referenceFasta "/gscmnt/ams1102/info/model_data/2869585698/build106942997/all_sequences.fa" --tumorBam "/gscmnt/gc8002/info/build_merged_alignments/merged-alignment-blade13-2-11.gsc.wustl.edu-jwalker-27691-f1e64d1802b0493095d70b837a467f47/f1e64d1802b0493095d70b837a467f47.bam" --runDir "/tmp/"`

tmooney: If we're not careful, this introduces another layer of "spawn-and-wait" to everything we do, which seems bad.

acoffman: As @tmooney mentioned, this introduces double queue usage which I don't think we want. Additionally, if we're going to start Dockerizing tools it seems like we need a general strategy for that. Hardcoding `bsub -q docker` everywhere we run something doesn't seem like the correct level of abstraction.

sleongmgi: I agree we need to find a general strategy.

jasonwalker80:
Are there any additional comments or suggestions?  Any +1s out there?

susannasiebert:
+1

acoffman:
+1

PR #1375, opened 2016-04-21
Remove old RnaSeq events.

tmooney:
These haven't been used in a long time and aren't referenced anywhere in the code anymore.  Loading any past events will yield an object of class `Genome::Model::Event::Generic`.


acoffman:
+1

jasonwalker80:
+1

PR #1376, opened 2016-04-22
add the java max memory used flag.

sleongmgi:


acoffman:
+1

lib/perl/Genome/Model/Tools/Varscan.pm
```diff
@@ -66,14 +89,20 @@ my %VARSCAN_VERSIONS = (
...
-    my $path = $self->path_for_version($self->version);
-    my $headers = $self->no_headers ? "--no-headers 1" : "";
-    my $command_line = 'java -jar ' . $path . ' ' . $parameter_string . ' '.$headers;
+    my @java_cmd = ( $self->java_interpreter );
+
+    my $path = $self->path_for_version( $self->version );
+    if ( defined $self->max_memory ) {
+        push @java_cmd, sprintf( "-Xmx%dg", $self->max_memory );
+    }
+    push @java_cmd, "-Djava.io.tmpdir=" . $self->tmp_dir, '-jar' => $path;
```
tmooney: `tmp_dir` is currently not being set anywhere.  Perhaps [this fallback from G/M/T/Gatk/Base.pm](https://github.com/genome/genome/blob/5bf928f396ba5921abc6aaf715ccf43353fdb2eb/lib/perl/Genome/Model/Tools/Gatk/Base.pm#L95-L98) would be useful?

tmooney:
+1.

I'm not keen on `sub max_memory` being copy+pasted wholesale from `Genome::Model::Tools::Gatk::Base`.  If it ends up being needed in a third location, we'll definitely need to come up with a better solution.

sleongmgi:
I agreed.

PR #1378, opened 2016-04-25
Adds command for dumping out VCFs from a build_id into a tar.

dummey:
This PR adds the feature of creating a tar from the VCFs of a build. This is useful when VCFs of a build need to be submitted to repositories like the SRA. 

https://jira.gsc.wustl.edu/browse/LIMS-1508


acoffman:
You can specify multiple `-C` and filenames in one tar command, so you should be able to achieve the flat directory structure you're after without having to shell out separately for each file. Additionally, this will allow you to take advantage of `tar`'s built in gzip support with the `-z` flag.

Something like the following could replace the whole loop:
````perl
my @tar_cmd = ("tar", "-zvcf", $tar_fullpath);
for my $r ( sort { ( $a->intervals )[0] cmp( $b->intervals )[0] } $build->haplotype_caller_result ) {
    push(@tar_cmd, "-C", $dir, $filename);
}

Genome::Sys->shellcmd(cmd => \@tar_cmd);
````

dummey:
The argument that I had for going the append route is to avoid potentially creating a tar command that is over the max terminal length.

wschroeder:
I would have enjoyed a bit more TME in this, but with this small amount of code, meh. :smile:   +1  :fallen_leaf:

acoffman:
@dummey That makes sense to me! +1

PR #1382, opened 2016-04-26
Remove "parallel" blastx tool.

tmooney:
It used old `Workflow` to accomplish its parallelization.  There are no records of it having been run in the usage log.


jasonwalker80:
+1

ebelter:
Feel the #Bern 🔥

PR #1383, opened 2016-04-26
Remove unused include in GenotypeMicroarray.t

tmooney:
This should've been part of #1337, but it was missed then.


jasonwalker80:
+1

ebelter:
+1

PR #1384, opened 2016-04-27
WorkflowBuilder supports references to other XML files

tmooney:
One smaller feature `Workflow` has is that the definition of an operation can be in a separate file.  This brings that feature to `WorkflowBuilder`.


ebelter:
+1 This seems reasonable.

jeldred:
+1

PR #1385, opened 2016-04-27
ViromeScreening uses WorkflowBuilder

tmooney:
This depends on #1384 to pass tests.  This also reverts #1273, which was really used here via files that were previously not in the repo.


jasonwalker80:
Is this tool still in use?

tmooney:
I saw seemingly real uses from this year in the Usage Log, so I took that to mean somebody still cares about it.

ebelter:
Kristine Wylie would be one to contact.

tmooney:
Would the CI test this please?

jasonwalker80:
+1

susannasiebert:
+1

PR #1386, opened 2016-04-27
DV2 Polymutt uses WorkflowBuilder

tmooney:
This replaces the test as well, since the old one depended on deleted production data and was never really run in practice.


jasonwalker80:
+1.  However, the same comment as blat, is polymutt used in the GMS or a tool that anyone plans to use for future common disease genomic studies?  If not, perhaps it should be removed instead.

jeldred:
+1 I agree with Jason's comments

tmooney:
I think I'll merge this one for now, and we can delete Polymutt later if some of the other pipeline removal PRs make it through.

PR #1388, opened 2016-04-27
Remove Blat GMT Tree and Alignment Result

tmooney:
Another use of `Workflow` gone...


jasonwalker80:
+1.  Is blat used anywhere in the GMS?  Is this a candidate for removal rather than updating to WorkflowBuilder?

tmooney:
There's an `AlignmentResult` class for Blat still.  It is quite possibly a candidate for removal otherwise.

jasonwalker80:
My opinion, I would check for any recent (<1 year old) results and remove the AlignmentResult class if it's no longer used.

tmooney:
The AlignmentResult class was used exactly thrice ever.  I've updated this PR to remove the Blat tools completely.

jasonwalker80:
+1

jeldred:
+1

PR #1389, opened 2016-04-28
Manta DV2 Detector

jasonwalker80:
This is considered a WIP because I'm still working on the Manta.t.  I need to create a small slice/subset of HCC1395 tumor/normal BAMs to use as inputs to the test.  Since this PR requires changes to the DV2 dispatcher, I thought I'd get it out early for review.  An internal test model build was run.  ID : 434620ceae4e496e9779188aba24ac61


lib/perl/Genome/Model/Tools/DetectVariants2/Dispatcher.pm
```diff
@@ -1020,38 +1020,60 @@ sub _create_bed_from_vcf {
...
+    my $dir = shift;
+    my $file_path = shift;
+
+    my $file_base;
+    if (-l $file_path) {
+        my $file_link = readlink($file_path);
+        $file_base = basename($file_link);
+    } else {
+        $file_base = basename($file_path);
+    }
```
tmooney: If you want to be sure not to end up with symlinks, this could be
```perl
use Cwd qw();

...

$file_path = Cwd::abs_path($file_path);
my $file_base = basename($file_path);
```

Then it'll still work if the symlink points to another symlink as well.  (This is what `DV2/Detector.pm` does with some of its inputs.)

jasonwalker80: Sadly I'm not entirely sure what the intent is.  I'm not sure what expectations DV2 makes about resolving these paths.  I can make the suggested change and see what happens....

jasonwalker80:
I removed the WIP since the test commit is pushed.  That didn't take nearly as long as I suspected it would.  The test produces an empty file.  A different genomic region with a known SV might be a better test data set, but the exist now exists.

lib/perl/Genome/Model/Tools/DetectVariants2/Dispatcher.pm
tmooney: Hm... looking at the old code, I think it was trying to resolve the symlink that goes to the "versioned" BED file.  Seems overly complicated!

jasonwalker80: I agree this is overly complicated.  For what it's worth, I was planning another PR that would remove the versioned BED files.  No one seems concerned with supporting versions and retaining the many levels of symlinks.  Unless the readlink behavior is required, do you agree we should simply use abs_path instead?

lib/perl/Genome/Model/Tools/DetectVariants2/Manta.pm
```diff
@@ -0,0 +1,120 @@
...
+    # Determine which manta output VCF should be used as the "high-quality" result
+    # If single-sample, assume tumor-only (NOT GERMLINE).  We know somatic if a normal is defined.
+    my $manta_vcf_filename = 'tumorSV.vcf.gz';
+
+    if ( defined($self->control_aligned_reads_input) ) {
+        push @resolved_config_params, ('--normal-bam-file' => $self->control_aligned_reads_input);
+        $manta_vcf_filename = 'somaticSV.vcf.gz';
+    }
+
+    my @dv2_config_params = split(' ',$self->params);
```
brummett: This isn't necessarily safe.  For example, if one of the values is a quoted string with a space in it.

Why not loop over the class's known properties (or ask the class what properties it has), extract their values, and build up a hash of params to pass to ```create()``` on line 65?  Something like:
```
my @dv2_config_params = map { $_ => $self->$_ } @known_params_list;
my $config_cmd = Genome::Model::Tools::Manta::Config->create(@dv2_config_params, @resolved_config_params);
```

lib/perl/Genome/Model/Tools/DetectVariants2/Manta.pm
```diff
@@ -0,0 +1,120 @@
...
+    my ($config_cmd_class, $config_cmd_params) = Genome::Model::Tools::Manta::Config->resolve_class_and_params_for_argv(@dv2_config_params,@resolved_config_params);
+
+    my $config_cmd = Genome::Model::Tools::Manta::Config->create($config_cmd_params);
```
brummett: If you do end up staying with this method, ```$config_cmd_params``` is a hashref here, so shouldn't this be ```...->create(%$config_cmd_params)```

lib/perl/Genome/Model/Tools/DetectVariants2/Manta.pm
jasonwalker80: I understand I could get a list of valid parameters from the tools that I plan to create.  Given that list of parameters I could define a hash of key values as you noted, but I haven't parsed the string `$self->params`.  It does seem dangerous to split on whitespace.  How does the UR command parser get an array of values to pass to resolve_class_and_params_for_argv?

Even with `@known_params_list` I have to parse the value or the existence of the boolean flags in the string `$self->params`

etc/genome/spec/lsf_resource_dv2_manta.yaml
```diff
@@ -1,3 +1,3 @@
 ---
-default_value: "-R 'select[tmp>=10000 && mem>=8000] rusage[tmp=10000, mem=8000] span[hosts=1]' -M 8000000 -n 12"
+default_value: "-R 'select[tmp>=10000 && mem>=24000] rusage[tmp=10000, mem=24000] span[hosts=1]' -M 24000000 -n 12"
```
jasonwalker80: Assuming 2GB per job slot.  See the comment about RAM usage below:
https://github.com/Illumina/manta/issues/38

lib/perl/Genome/Model/Tools/DetectVariants2/Manta.pm
brummett: > How does the UR command parser get an array of values to pass to resolve_class_and_params_for_argv?

UR's "parser" relies on the shell having already split up the command line and populated ```@ARGV``` before the process starts up.  The only parsing it has to do is recognize ```--``` as the preamble to a parameter name, and know that param names can be separated from values with "=".

> Even with @known_params_list I have to parse the value or the existence of the boolean flags in the string $self->params

```@known_params_list``` is supposed to contain only property/param names, not values.  Then the map in my example will build up the right params list suitable for passing to ```create()```, whether the values are boolean or not.  Another option might be to look at UR::Object::copy().  Extract the part of it that gets the list of keys/values into a new method.  Use this new method to get the keys/values out of ```$self``` and pass them into ```Genome::Model::Tools::Manta::Config->create()```.

Maybe I'm missing something... At this point in the code, have the properties of ```$self```  been filled in yet as a result of parsing what's in ```$self->params()```?  If you really must parse this string, Text::Parsewords::shellwords() purports to break it up the same way a shell would.

Comment (jasonwalker80) [inline lib/perl/Genome/Model/Tools/DetectVariants2/Manta.pm]: > Maybe I'm missing something... At this point in the code, have the properties of $self been filled in yet as a result > of parsing what's in $self->params()? If you really must parse this string, Text::Parsewords::shellwords() purports > to break it up the same way a shell would.

The properties never exist on `$self`.  The only property is `$self->params`.  It sounds like there is no way around parsing the 'params' string then.  I'll try using Text::Parsewords::shellwords() for this.

**Comment by jasonwalker80** [comment]
Apologies in advance for so many commits.  I tried to cleanup some of the relative/full path business.  I'm still uncertain why we would need to resolve the relative path simply to append it to the same _temp_staging_directory.  Perhaps that's for another round of cleanup if this passes perl and model tests.

File: lib/perl/Genome/Model/Tools/DetectVariants2/Dispatcher.pm [hunk 5]
```diff
@@ -261,18 +261,18 @@ sub _dump_workflow {
...
     $xml_file->close;
-    #$workflow->as_png($self->output_directory."/workflow.png"); #currently commented out because blades do not all have the "dot" library to use graphviz
+    #$workflow->as_png($self->_temp_staging_directory."/workflow.png"); #currently commented out because blades do not all have the "dot" library to use graphviz
 }
 
 sub _dump_dv_cmd {
     my $self = shift;
     my $cmd = join(" ",@INC)."\n===============================================\n";
-    $cmd .=   "gmt detect-variants2 dispatcher --output-directory ".$self->output_directory;
+    $cmd .=   "gmt detect-variants2 dispatcher --output-directory ".$self->_temp_staging_directory;
```
tmooney: I don't believe this one should have been updated.  This "dump" is meant to represent what the user would type in to run this dispatcher again.

lib/perl/Genome/Model/Tools/DetectVariants2/Dispatcher.pm
```diff
@@ -932,40 +938,39 @@ sub _create_temp_directories {
...
-                if(-l $vcf_link){
+                if (-l $vcf_link) {
                     $source = readlink($vcf_link);
                 } else {
                     $source = $vcf_link;
                 }
-                my $link_target = $output_dir."/$variant_type" . "s.detailed.vcf.gz";
-                my $clipped_vcf = $output_dir."/$variant_type" . "s.vcf.gz";
+                my $link_target = File::Spec->join($self->_temp_staging_directory,$variant_type .'s.detailed.vcf.gz');
+                my $clipped_vcf = File::Spec->join($self->_temp_staging_directory, $variant_type .'s.vcf.gz');
```
tmooney: I don't understand the logic behind making this change.  The final product is a symlink to a location in the `output directory`; it seems merely coincidental if this happens to work when created as a symlink to the `_temp_staging_directory` instead.

lib/perl/Genome/Model/Tools/DetectVariants2/Dispatcher.pm
```diff
@@ -1142,7 +1150,7 @@ sub _generate_standard_files {
             $self->$lq_accessor($lq_result);
 
             my $f = $lq_result->_file_for_type($variant_type);
-            Genome::Sys->create_symlink($lq_result->path($f), $self->output_directory . "/$f");
+            Genome::Sys->create_symlink($lq_result->path($f), $self->_temp_staging_directory . "/$f");
```
tmooney: This one missed the opportunity to be updated to `File::Spec`.

lib/perl/Genome/Model/Tools/DetectVariants2/Dispatcher.pm
jasonwalker80: From what I understand about the Dispatcher class, the _temp_staging_directory and output_directory are always the same, see the _create_directories subroutine.  The intention of this commit was merely to reduce the confusion by using one or the other accessor throughout.  If the output_directory is preferred over the _temp_staging_directory, I'd be happy to update all of the paths to use it instead.  The same goes for the _dump_dv_cmd comment.  I may not understand how the Dispatcher is used and perhaps there are situations when the output_directory and _temp_staging_directory are indeed different.

tmooney: I don't know that there are situations where the two are different.  My comments are only concerned with which one would be correct if they were in fact different.  When reading most of the code we shouldn't need to worry about where each directory happens to point, but instead only which one is correct for what we're trying to accomplish with each use.

In the case of dumped commands or symlinks, we want to use the final user-facing directory.  For creating most files, we can generally stage them and let them be promoted later.

Alternatively we could get rid of the `temp_staging_directory` concept in the dispatcher and only use `output_directory` for everything.

lib/perl/Genome/Model/Tools/DetectVariants2/Dispatcher.pm
jasonwalker80: I have converted this and some additional string concatenation to use `File::Spec`.

lib/perl/Genome/Model/Tools/DetectVariants2/Dispatcher.pm
```diff
@@ -1020,38 +1030,69 @@ sub _create_bed_from_vcf {
...
+                    $self->fatal_message('Unable to locate full path to high-quality '. $variant_type .' variant file!');
+                }
+            } else {
+                my $hq_file_name = $variant_type .'s.hq.bed'; # FIXME this will not be true for polymutt or other detectors that output only vcf
+                my $hq_file_path = File::Spec->join($hq_output_dir,$hq_file_name);
+                $hq_full_path = $self->_resolve_variant_file_full_path($hq_output_dir, $hq_file_path);
+                if ( defined($hq_full_path) ) {
+                    $self->$file_accessor($hq_full_path);
+                } else {
+                    $self->debug_message('Unable to locate full path to high-quality '. $variant_type .' variant file. Continuing with \''. $hq_file_path .'\' since DV2 laters resolves the path using the file basename.');
```
jasonwalker80: Just a quick note on this debug message.  I would have thought this should be a fatal message; however, the test 'CopyCatSomaticWithBamWindow.t' will fail if so.  The test is a bit different than most since it has to run the Dispatcher to get the Samtools detector result that is faked in the test.  The /tmp paths are different between the result and the Dispatcher but somehow the test succeeds and the results are generated if the missing path is returned.

jasonwalker80:
@tmooney  and @brummett new commits to address the concerns about `File::Spec` and the use of `$self->output_directory` instead of `$self->_temp_staging_directory`.

lib/perl/Genome/Model/Tools/DetectVariants2/Dispatcher.pm
tmooney: `laters` should have fewer `s`s.

lib/perl/Genome/Model/Tools/DetectVariants2/Dispatcher.pm
```diff
@@ -300,18 +303,25 @@ sub _dump_dv_cmd {
...
+
+    my $relative_path = $result_path;
+
+    my $dispatcher_path = $self->output_directory;
+    if ($relative_path =~ /$dispatcher_path\/?/ ) {
+        $relative_path =~ s/$dispatcher_path\/?//;
+        return $relative_path;
+    }
+
+    return;
```
brummett: You can just put the ```s///``` inside the if-condition and remove the match.  The substitution will return true if the match portion matched anything.

I know it was there before, but the substitution seems suspect... It's looking for a (partial?) path followed by a slash followed by any character, then cutting it out.  If there was any more of the path after the part that matched, it's now bogus because it's missing that one extra character.

jasonwalker80: The suggestion is implemented.

I'm not surprised that any of this seems suspect.  It's not clear to me when the relative path business is ever needed.  I believe the intent of the regular expression is that the trailing slash is optional.  Is that not the case?  Instead, is the `?` really matching on any single character after the slash?

brummett: I just wasn't thinking right... it's actually looking for the path followed by 1 or 0 slashes.  It's OK after all.  If you want to be extra super duper safe, it probably should be matching a (again, partial?) path in its entirety.  Meaning if $dispatcher path is "foo", it should correctly match and snip out
* /path/to/foo
* /path/to/foo/bar

while leaving
* /path/to/foobar
* /blah/foobar/baz
alone.

In practice, I'm sure it won't come up, but a more complicated regex could handle it.  Maybe something like ```s#(?<=/)$dispatcher_path(?:/|\Z)##;```  That looks for a slash, then $dispatcher_path, then either a slash or the end of the string, but the first slash isn't cut out because it's a look-behind assertion.

tmooney: Maybe this would be a good place to use `Path::Class`.  It has `contains` (or `subsumes`) and then we could use `$file->relative($dir)` to get the relative path.

jasonwalker80:
A new commit using `Path::Class` as suggested.

brummett:
+1

tmooney:
+1 💳

PR #1391, opened 2016-04-28
Slight refactor of qualities iterator to avoid $i-1

susannasiebert:
I missed @tmooney's last comment on #1364 before merging. This PR fixes the iteration over the base/mapping qualities so that we are not repeatedly using `$i-1`.


tmooney:
+1 😎

acoffman:
+1

PR #1392, opened 2016-04-29
Allow ClinSeq to also use Somatic Validation models as inputs

susannasiebert:
This is achieved by implementing a joined `Genome::Model::Build::SomaticInterface` class.

This PR also includes refactors of various ClinSeq commands and utility classes to make the code more readable and to remove redundancies. There are also a few commits that address general changes needed for Somatic Validation builds to work in ClinSeq.


lib/perl/Genome/InstrumentData/AlignmentResult/Merged/CoverageStats.pm
```diff
@@ -196,7 +197,9 @@ sub _dump_bed_file {
     my $merge_status = $self->merge_contiguous_regions;
     my $use_short_names = $self->use_short_roi_names;
 
-    my $bed_file_path = $self->temp_staging_directory .'/'. $roi_set->id .'.bed';
+    unless ($bed_file_path) {
+        $bed_file_path = $self->temp_staging_directory .'/'. $roi_set->id .'.bed';
```
tmooney: This looks like a prime opportunity to switch to `File::Spec->join`!

lib/perl/Genome/Model/Build/ClinSeq.pm
```diff
@@ -29,4 +29,59 @@ sub should_run_exome_cnv {
...
 
+sub input_builds {
+    my $self = shift;
+
+    my @input_builds = ( $self->tumor_rnaseq_build, $self->normal_rnaseq_build );
+    for my $accessor (qw(wgs_build exome_build)) {
+        if ($self->$accessor) {
+            push @input_builds, $self->$accessor;
+            for my $build (qw(tumor_build normal_build)) {
+                if ($self->$accessor->can($build)) {
```
tmooney: This could also be:
```perl
if (my $input_build = $self->$accessor) {
   push @input_builds, $input_build;
   for my $build (qw(tumor_build normal_build)) {
       if ($input_build->can($build)) {
```

lib/perl/Genome/Model/Build/ClinSeq.pm
```diff
@@ -29,4 +29,59 @@ sub should_run_exome_cnv {
...
+            }
+        }
+    }
+
+    return grep {defined($_)} @input_builds;
+}
+
+sub reference_sequence_build {
+    my $self = shift;
+    return $self->model->_resolve_reference;
```
tmooney: I generally don't believe that builds should rely on model inputs to make decisions.  All relevant inputs should be on the build.

lib/perl/Genome/Model/Build/ClinSeq.pm
```diff
@@ -29,4 +29,59 @@ sub should_run_exome_cnv {
...
+}
+
+sub has_microarray_build {
+    my $self = shift;
+
+    my $base_build;
+    if ($self->exome_build) {
+        $base_build = $self->exome_build;
+    }
+    elsif ($self->wgs_build) {
```
tmooney: Just a curiosity... but why doesn't `has_microarray_build` prefer the `best_somatic_build`?

lib/perl/Genome/Model/Build/SomaticInterface.pm
```diff
@@ -0,0 +1,67 @@
...
+}
+
+sub individual {
+    my $self = shift;
+    $self->fatal_message('Abstract: (individual) needs to be defined on class (%s)', $self->class);
+}
+
+sub individual_common_name {
+    my $self = shift;
+    $self->fatal_message('Abstract: (individual_common_name) needs to be defined on class (%s)', $self->class);
```
tmooney: Is there ever a situation where this couldn't be:
```perl
return $self->individual->common_name;
```

lib/perl/Genome/Model/Build/SomaticValidation.pm
```diff
@@ -319,4 +321,48 @@ sub get_feature_list_from_reference {
     return $self->reference_sequence_build->get_feature_list($feature_list_accessor);
 }
 
+sub individual {
+    my $self = shift;
+    return $self->subject;
```
tmooney: Fair warning: this is currently only about 75% reliable.
```
genome=> SELECT subject_class_name,COUNT(subject_class_name) FROM model.model WHERE subclass_name = 'Genome::Model::SomaticValidation' GROUP BY subject_class_name;
 subject_class_name | count 
--------------------+-------
 Genome::Sample     | 13929
 Genome::Individual | 43260
 Genome::Taxon      |  1022
(3 rows)
```

lib/perl/Genome/Model/Build/SomaticValidation.pm
```diff
@@ -319,4 +321,48 @@ sub get_feature_list_from_reference {
...
+            annotation_file => $annotation_vcf,
+            output_file => $annotated_snvs_vcf_file,
+            use_bgzip => 1,
+            info_fields => "",
+            info => "",
+            use_version => Genome::Model::Tools::Joinx->get_default_version,
+        );
+        unless ($vcf_annotator->execute) {
+            $self->fatal_message("Failed to execute Joinx Vcf annotation for vcf file (%s) using db (%s)", $snvs_vcf_file, $annotation_vcf);
+        }
```
tmooney: It's a bit sketchy to be adding things to the build post-hoc like this.  This may need to be careful to:
* cleanup partial files if it fails (assuming we don't get killed)
* reallocate if it succeeds

lib/perl/Genome/Model/ClinSeq.pm
```diff
@@ -249,6 +233,19 @@ sub _resolve_reference {
 sub __errors__ {
     my $self   = shift;
     my @errors = $self->SUPER::__errors__;
+
+    for my $model (qw(wgs_model exome_model)) {
+        if ($self->$model) {
+            unless ($self->$model->isa('Genome::Model::SomaticVariation') || $self->$model->isa('Genome::Model::SomaticValidation')) {
```
tmooney: Seems like there ought to be a way to say "Things that have builds that implement the `SomaticInterface`" rather than calling these two out specifically.

lib/perl/Genome/Model/Build/SomaticValidation.pm
susannasiebert: It looks like there is minimal chance for the same problem in SomVar
```
genome=> SELECT subject_class_name,COUNT(subject_class_name) FROM model.model WHERE subclass_name = 'Genome::Model::SomaticVariation' GROUP BY subject_class_name;
 subject_class_name | count 
--------------------+-------
 Genome::Sample     | 26438
 Genome::Individual |     4
(2 rows)```

File: lib/perl/Genome/Model/ClinSeq.pm [hunk 9]
```diff
@@ -1863,38 +1851,31 @@ sub _infer_annotations_from_input_models {
...
 sub _input_models {
     my $self = shift;
 
-    my @input_models = ( $self->wgs_model, $self->exome_model, $self->tumor_rnaseq_model, $self->normal_rnaseq_model );
-
-    $self->_infer_input_models_for_somatic_model(\@input_models,$self->wgs_model) if $self->wgs_model;
-    $self->_infer_input_models_for_somatic_model(\@input_models,$self->exome_model) if $self->exome_model;
+    my @input_models = ( $self->tumor_rnaseq_model, $self->normal_rnaseq_model );
+    for my $accessor (qw(wgs_model exome_model)) {
+        if ($self->$accessor) {
```
tmooney: (The same comment as for `_input_builds` applies here.)

lib/perl/Genome/Model/ClinSeq.pm
susannasiebert: Would `$self->$model->last_succeeded_build->isa('Genome::Model::Build::SomaticInterface')` do the trick? Maybe with an additional check that `last_succeeded_build` exists? Or does the inheritance hierarchy not support this?

tmooney: That seems unreliable, since it's dependent on a build existing to pass.

A few options:
* Implement `Genome::Model::SomaticInterface` (some of the accessors like `individual` on the build class might end up delegating there)
* Make a `build_class` like `Genome::Model::Build::model_class` and then it's `$self->$model->build_class->isa('Genome::Model::Build::SomaticInterface')`

lib/perl/Genome/Model/ClinSeq/Command/CreateMutationSpectrum.pm
```diff
@@ -7,11 +7,11 @@ use Genome;
...
     is        => ['Command::V2', 'Genome::Model::ClinSeq::Util'],
     has_input => [
-        somvar_build => {
-            is                  => 'Genome::Model::Build::SomaticVariation',
+        somatic_build => {
+            is                  => 'Genome::Model::Build::SomaticInterface',
             shell_args_position => 1,
             require_user_verify => 0,
-            doc                 => 'somatic variation build(s) to ' . 'create mutation spectrum results from',
+            doc                 => 'somatic variation to create mutation spectrum results from',
```
tmooney: I think the wrong word got removed from this doc line.

lib/perl/Genome/Model/ClinSeq/Command/CreateMutationSpectrum.pm
```diff
@@ -301,14 +298,13 @@ sub generate_summarize_mutation_spectrum_result {
...
     my $self                     = shift;
-    my $somvar_build             = shift;
     my $final_name               = shift;
     my $tier1_snvs               = shift;
     my $tier2_snvs               = shift;
     my $tier3_snvs               = shift;
     my $sub_outdir4              = shift;
     my $data_type                = $self->datatype;
-    my $reference_annotation_dir = $somvar_build->model->annotation_build->data_directory;
+    my $reference_annotation_dir = $self->somatic_build->model->annotation_build->data_directory;
```
tmooney: This is another circumstance where I don't believe delegating to the model is appropriate.

lib/perl/Genome/Model/ClinSeq/Command/GetBamReadCounts.pm
```diff
@@ -15,16 +15,16 @@ class Genome::Model::ClinSeq::Command::GetBamReadCounts {
                 . "  (e.g. 5:112176318-112176318	APC	APC	p.R1676T	G	C)"
         },
 
-        wgs_som_var_build => {
+        wgs_somatic_build => {
             is          => 'Genome::Model::Build::SomaticVariation',
```
tmooney: If this accessor's name is changing, should it's `is` also change?

lib/perl/Genome/Model/ClinSeq/Command/GetBamReadCounts.pm
```diff
@@ -15,16 +15,16 @@ class Genome::Model::ClinSeq::Command::GetBamReadCounts {
...
             is_optional => 1,
-            doc         => 'Whole genome sequence (WGS) somatic variation build'
+            doc         => 'Whole genome sequence (WGS) somatic build'
         },
 
-        exome_som_var_build => {
+        exome_somatic_build => {
             is          => 'Genome::Model::Build::SomaticVariation',
```
tmooney: If this accessor's name is changing, should it's `is` also change?

lib/perl/Genome/Model/ClinSeq/Command/GetVariantSources.pm
```diff
@@ -265,25 +300,35 @@ sub createIndelOutfile {
...
         }
-        print INDEL_OUT "$indels->{$indel}{coord_string}\t$indels->{$indel}{line}\t", join(",", @callers),
-            "\t$strelka\t$gatk\t$pindel\t$varscan\t$indels->{$indel}{tier}\n";
+        print INDEL_OUT "$indels{$indel}{coord_string}\t$indels{$indel}{line}\t", join(",", @callers),
+            "\t$strelka\t$gatk\t$pindel\t$varscan\t$indels{$indel}{tier}\n";
     }
     close(INDEL_OUT);
+
+    #Set output files as output to this step
+    die $self->error_message("Trying to set a file as output but the file does not exist: $indel_outfile")
```
tmooney: Obligatory plug for `$self->fatal_message`!

jeldred:
I am not prepared to speak to the correctness of these changes...but this is COOL 😄

lib/perl/Genome/Model/ClinSeq/Command/GetVariantSources.pm
```diff
@@ -297,67 +342,76 @@ sub createSnvOutfile {
...
         }
-        print SNV_OUT "$snvs->{$snv}{coord_string}\t$snvs->{$snv}{line}\t", join(",", @callers),
-            "\t$strelka\t$sniper\t" . "$varscan\t$samtools\t$mutect\t$snvs->{$snv}{tier}\n";
+        print SNV_OUT "$snvs{$snv}{coord_string}\t$snvs{$snv}{line}\t", join(",", @callers),
+            "\t$strelka\t$sniper\t" . "$varscan\t$samtools\t$mutect\t$snvs{$snv}{tier}\n";
     }
     close(SNV_OUT);
+
+    #Set output files as output to this step
+    die $self->error_message("Trying to set a file as output but the file does not exist: $snv_outfile")
```
tmooney: Obligatory plug for `$self->fatal_message`!

lib/perl/Genome/Model/ClinSeq/Command/RunCnView.pm
```diff
@@ -30,7 +30,7 @@ class Genome::Model::ClinSeq::Command::RunCnView {
         cnv_hq_file => {
             is          => 'FilesystemPath',
             is_optional => 1,
-            doc         => 'cnv_hq file from clonality analysis.' . 'if not provided file from wgs_somvar build is used'
+            doc         => 'cnv_hq file from clonality analysis. if not provided file from wgs somaic build is used'
```
tmooney: `somaic` would benefit from an additional `t`.

lib/perl/Genome/Model/ClinSeq.pm
brummett: Another option is that this kind of thing is traditionally what roles are for.  You can check whether a class or instance consumes a role with ```->does('Role::Name')```

acoffman: Our initial thought for the `SomaticInferface` in general was to make it a role, but we ended up going with a class because we wanted to use the built in UR "type checking" on inputs and parameters by updating them to `SomaticInterface`. As far as I know, this would not work if it was a role, but I'm totally ready to be wrong on that! :)

brummett: @acoffman - Type checking doesn't work with roles (yet).  I don't think there's anything stopping it from happening other than someone to do the work.

lib/perl/Genome/Model/Build/ClinSeq.pm
susannasiebert: @gatoravi @jasonwalker80 Do you know why in this subroutine we pick the exome build over the wgs build? Is there any harm in changing it?

This subroutine was originally in G::M::ClinSeq and just got moved as part of this PR. The implementation didn't change.

lib/perl/Genome/Model/Build/SomaticValidation.pm
susannasiebert: @gatoravi @jasonwalker80 Both SomVal and SomVar build can have samples or individuals as their subject so I plan on changing this to a generalized implementation to get to the subject. However, SomVal can also have a taxon as a subject. How should this case be handled?

lib/perl/Genome/Model/Build/ClinSeq.pm
jasonwalker80: I do not know why and agree that it should be consistent with other parts of ClinSeq, like best_somatic_build.

lib/perl/Genome/Model/Build/SomaticValidation.pm
jasonwalker80: Taxon is rarely used and when used is often for strange cases of combining samples from different individuals.  I suggest failing with an appropriate error message rather than supporting SomVal builds with taxon subjects.  ClinSeq is intended for supporting individual-based analysis and taxon is outside the scope.

lib/perl/Genome/Model/Build/ClinSeq.pm
gatoravi: We try to get at the microarray_build here, as $base_model->tumor_model->genotype_microarray. I think we expect the microarray build be set in both exome and wgs ref-aligns, hence the choice of exome first was probably arbitrary in this case. Ideally we'd check both if that makes sense..

susannasiebert: Ok, so we should probably refactor this even further to look at both the wgs and the exome model. What should we do if they both have genotype microarray builds but they are different? Should one be picked over the other or should there be a fatal error?

gatoravi: Yep, that sounds accurate. I think if they are different then a fatal error would be best, I can't think of a situation where we would expect the same sample to have different microarray builds for wgs and exome reference-alignment data.

susannasiebert:
I believe I addressed all comments on this PR aside from the switch to a UR::Role.

lib/perl/Genome/Model/ClinSeq.pm
susannasiebert: Making the `SomaticInterface` into a role sounds like a good idea. I'm willing to tackle the necessary UR changes. @brummett can you provide some pointers of what would need to change  for role type checking to work?

susannasiebert: After meeting with @jasonwalker80 we decided that the switch to UR::Role is something we would tackle outside of this PR.

susannasiebert:
I now also addressed the remaining instances of getting a property via the model instead of the build.

susannasiebert:
CI, test this bitte.

lib/perl/Genome/Model/ClinSeq.pm
jasonwalker80: See new issue #1400

tmooney:
+1 🐐

jasonwalker80:
+1

PR #1394, opened 2016-05-03
Redirect the standard output and error to the /dev/null.

sleongmgi:
Otherwise, the standard output will interfere with the
TAP output.


wschroeder:
+1 :footprints:

tmooney:
+1 🌅

PR #1396, opened 2016-05-04
put an message because it might cause error if it is only one test.

sleongmgi:


wschroeder:
+1 :gem:

brummett:
A better fix might be to check if ```@files_to_compile``` is 0.  If so, then ```plan skip_all => 'no new files to test compiling'```.  Then this test for >= 0 can be removed.

lib/perl/Genome/compile.t
```diff
@@ -15,7 +15,9 @@ my $start_point = resolve_start_point();
 my @files_to_compile = files_to_compile($start_point);;
 plan tests => scalar(@files_to_compile) + 1;
 
-ok(@files_to_compile >= 0); # have to have at least one test for this to pass when there is nothing to compile
+# LSF Have to put a message otherwise, it might
+#     cause an error  if it is only one test.
+ok(@files_to_compile >= 0, 'checked files to compile'); # have to have at least one test for this to pass when there is nothing to compile
```
tmooney: The new comment lines might be better off in the commit message instead of in the code.

Or we could revise the existing comment for the line with to say `have to have at least one test with a message for this to pass when there is nothing to compile`.

sleongmgi:
I tried to keep what the original code logic was.

tmooney:
+1 ⏰

PR #1397, opened 2016-05-05
A new ClinSeq command and software result for annotating snvs vcf file

susannasiebert:
This is needed to fix a problem with #1392. We need to be able to make a software result when annotating the somatic validation snvs vcf.


lib/perl/Genome/Model/ClinSeq/Command/AnnotateSnvsVcf/Result.pm
```diff
@@ -0,0 +1,72 @@
...
+
+sub _run {
+    my $self = shift;
+
+    my $vcf_annotator = Genome::Model::Tools::Joinx::VcfAnnotate->create(
+        input_file => $self->input_file,
+        annotation_file => $self->annotation_file,
+        output_file => $self->_temp_staging_file_path,
+        use_bgzip => $self->use_bgzip,
+        info_fields => $self->info_fields || "",
```
tmooney: Is it okay that this would normalize `0` to the empty string?  (Looks like `info_fields` is a required property; is this trying to catch a particular issue that has happened?)

susannasiebert: When `info_fields` is set to `""` when creating the command it doesn't get passed on to the result at all. However, it looks like `info_fields` can just be made optional all the way to the joinx wrapper. Joinx doesn't require this parameter.

tmooney:
+1 🐫

jasonwalker80:
To summarize a short discussion offline:

1. Rename the class to be SNV agnostic
2. Use the same input file basename as the output file basename

jasonwalker80:
+1

PR #1398, opened 2016-05-05
Split sequences by ns

ebelter:


brummett:
+1

This looks ok as is, but if it'll be dealing with really large files, you'll want to switch to an iterator-based approach.  As long as the writer object is actually streaming the data out, the inner loop splitting on the n's could look something like this:
```
while( $seq->{seq} =~ m/(\w+?)n{$n,}/gi ) {
    $writer->write({ id => join('.', $seq->{id}, ++$cnt),
                             seq => $1});
}
```
And if that regex is too slow, you can get each match position with ```my $pos = $seq->{seq} =~ m/n{$n,}/gi``` and extract substrings from the original string.

lib/perl/Genome/Model/Tools/Sx/Split/ByNs.pm
```diff
@@ -0,0 +1,56 @@
...
+        number_of_ns => {
+            is => 'Integer',
+            doc => 'Number of Ns to split sequence.',
+        },
+    },
+};
+
+sub help_brief { 'Split sequences by Ns' }
+
+sub help_detail { 'Split sequences by a set number of Ns. Does not support splitting qualites.' }
```
tmooney: `qualites` would benefit from an additional `i`.  (Would it make sense to say it only supports splitting on sequence instead?  There are a great many things in the universe that it doesn't support!)

lib/perl/Genome/Model/Tools/Sx/Split/ByNs.t
```diff
@@ -0,0 +1,33 @@
...
+my $pkg = 'Genome::Model::Tools::Sx::Split::ByNs';
+use_ok($pkg);
+
+my $dir = File::Spec->join( Genome::Config::get('test_inputs'), 'Genome-Model-Tools-Sx', 'SplitByNs');
+my $in_fasta = File::Spec->join($dir, 'in.fasta');
+ok(-s $in_fasta, 'in fasta');
+my $expected_fasta = File::Spec->join($dir, 'expected.fasta');
+ok(-s $expected_fasta, 'expected fasta');
+
+my $tmp_dir = File::Temp::tempdir(CLEANUP => 1);
```
tmooney: This could be a `Genome::Sys->create_temp_directory` so LSF cleans it up even if things go south.

ebelter:
@brummett I couldn't implement iterating the way you suggested b/c the sequences may not contain ends. I still used split to remove chunks of valid sequence, keeping track the number of ns. I can then use them to split the qualities.

ebelter:
@tmooney Now it splits qualities 😄

tmooney:
+1 ♠️

PR #1399, opened 2016-05-09
SmallRNA: Remove inputs that are already supplied via other steps.

tmooney:
These extra connections have been in the workflow for a long time, but `WorkflowBuilder` is stricter about validating this.


jasonwalker80:
+1

jeldred:
+1

PR #1401, opened 2016-05-10
ClinSeq's Cufflinks Converge uses Test Data

tmooney:
As part of cleanup of old data we've been removing directories for Abandoned builds.  This is one of a few tests that was using an Abandoned build for its data.  With this change it's not relying on production data at all!

This PR includes a small change to the tool: the hash keys are now sorted so that the output is consistent.

### DIFF
--- a/lib/perl/Genome/Model/ClinSeq/Command/Converge/CufflinksDe.t
+++ b/lib/perl/Genome/Model/ClinSeq/Command/Converge/CufflinksDe.t
@@ -16,13 +16,13 @@ use above "Genome";
 use Test::More tests => 8;  #One per 'ok', 'is', etc. statement below
 use Data::Dumper;
 
+use Genome::Utility::Test;
+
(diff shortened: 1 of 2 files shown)


lib/perl/Genome/Model/ClinSeq/Command/Converge/CufflinksDe.t
```diff
@@ -16,7 +16,8 @@ use above "Genome";
 use Test::More tests => 8;  #One per 'ok', 'is', etc. statement below
 use Data::Dumper;
 
-use_ok('Genome::Model::ClinSeq::Command::Converge::CufflinksDe') or die;
+my $pkg = 'Genome::Model::ClinSeq::Command::Converge::CufflinksDe';
+use_ok($pkg) or die;
```
susannasiebert: Can we remove the `or die` here since it's not necessary?

susannasiebert:
+1

jasonwalker80:
+1

PR #1402, opened 2016-05-10
ClinSeq's SummarizeSnvIndel Converge Test Uses Test Data

tmooney:
A similar story as in #1401... The production data backing this test is gone, so we need to use test data!


susannasiebert:
+1 once tests pass

jasonwalker80:
+1

PR #1404, opened 2016-05-10
Remove SomticVariation Compare Variants.

tmooney:
The test for this was relying on production data.  The only uses of it recorded in the usage log were people running the example in the test.  We have a generic solution for build diffs already, and someone could run joinx themselves if they really want to compare particular files.


wschroeder:
+1 :dancer:

susannasiebert:
+1. Yay code deletion!

PR #1406, opened 2016-05-10
Remove ClinSeq OriginalScripts Directory

tmooney:
These have been hanging out for awhile; a test for one started failing, so I propose removing them rather than fixing the failed test.


gatoravi:
I'm ok with this change, I've put this folder [here](https://github.com/griffithlab/misc-resources/tree/master/OriginalClinSeqScripts) just in case anyone's looking for it down the road.

susannasiebert:
+1. 🍰  for deleting almost 10k loc.

jeldred:
+10k

tmooney:
These scripts will also remain available in the git history here.

malachig:
+1

PR #1407, opened 2016-05-11
Processed BED without chr substitution

jasonwalker80:
No longer remove chr from BED files.  Rely on the refseq converter to translate chromosome names.

This issue has resulted in refseq converters that do not work on the processed_bed_file.  @ebelter @susannasiebert @tmooney @apregier  and @jeldred where all involved in the previous discussion on #970.


tmooney:
I'm generally in favour of this change.

The big question (as seen in the previous PR) is whether or not we need to also make changes in [`Genome::VariantReporting::Suite::Vep::RunResult`](https://github.com/genome/genome/blob/2706cc686b1678c002f04b63191ffc8adbc35193/lib/perl/Genome/VariantReporting/Suite/Vep/RunResult.pm#L212) and in [`Genome::Qc::Tool::Bedtools::Intersect`](https://github.com/genome/genome/blob/2706cc686b1678c002f04b63191ffc8adbc35193/lib/perl/Genome/Qc/Tool/Bedtools/Intersect.pm#L32).  There are a few other places that use this as well.

jasonwalker80:
Please see #1418 as an option that should not change existing behavior for the previously mentioned commands that rely on processed_bed_file.

jasonwalker80:
Closing as #1418 replaces this.

PR #1410, opened 2016-05-11
New HCC1395 SV test data

jasonwalker80:
Add new HCC1395 test data that actually produces a DEL call with Manta.


tmooney:
+1.  (Looks like this is replacing the test data from 2a8ac672b262e92a560254a363b649d4fe2c5a65.)

acoffman:
+1

PR #1413, opened 2016-05-11
Use the subroutines in G::M::B::RunsDV2 instead of implementing our own

susannasiebert:


jasonwalker80:
+1

tmooney:
+1 (assuming tests pass!)

`snvs_annotated_variants_vcf_file` could use `RunsDV2`'s `variants_directory` instead of manually building the path to the variants directory.  It seems like the accessor could also be moved there, but it's not really a part of DV2, is it?

susannasiebert:
Yeah, it's not part of DV2. It's actually a SomVar specific file so the current implementation isn't quite right either. I pushed another commit to clean up that part a bit more.

jasonwalker80:
+1

PR #1414, opened 2016-05-11
Add SpeedseqSv probability curves

jasonwalker80:
Add support for -P param to speedseq sv.  
Cleanup the variable and method names resolving tool parameters.


lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.t
```diff
@@ -49,7 +49,7 @@ my $command2 = $pkg2->create(
 );
 
 ok($command2->execute, 'Executed `gmt detect-variants2 Speedseq` command');
-
+$DB::single=1;
```
tmooney: A breakpoint snuck in here.

jasonwalker80: Thanks.  Removed.

tmooney:
+1 🎣

lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.t
```diff
@@ -36,7 +36,7 @@ symlink($reference_fasta_index,$reference_build->data_directory .'/all_sequences
 
 my $result_users = Genome::Test::Factory::SoftwareResult::User->setup_user_hash(reference_sequence_build_id => $reference_build->id,);
 
-my $params = "-g,-d,-o:Hello";
+my $params = '-P,-g,-d';
```
susannasiebert: What's the reasoning behind removing the `-o:Hello` test case? It would be great to still have a test for that syntax.

jasonwalker80: I agree testing the syntax would be useful, but the param did not work in the first place.  The -o option is to define the output_prefix which is redefined here:
https://github.com/jasonwalker80/genome/blob/28d3c86f0c66f242abf3b343ebacb46f427cc46b/lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.pm#L46

jasonwalker80: The latest commit adds a colon separated (not semi-colon as the commit suggests) param that is parsed and used in the Speedseq SV tool.

susannasiebert:
+1

PR #1416, opened 2016-05-12
Better Intermediate Result Deletion

tmooney:
The addition of the `shortcut` label as this is used by the main AR was preventing its subsequent deletion.  This change broadens the inactivation of users.

(An alternative might be to inactivate all users and forcibly remove the result.  That would eliminate any means of claiming intermediate results, though.)


susannasiebert:
+1 once tests are fixed.

jasonwalker80:
+1

PR #1418, opened 2016-05-12
Processed BED without chr substitution : Option #2

jasonwalker80:
This is a proposed replacement for #1407.

Add option for stripping off the chr in processed_bed_file_content.  By default generate_converted_bed_file will never strip chr from the BED content.

### DIFF
--- a/lib/perl/Genome/FeatureList.pm
+++ b/lib/perl/Genome/FeatureList.pm
@@ -254,10 +254,17 @@ sub processed_bed_file_content {
     unless (defined($track_name)) {
         $track_name = 'target_region';
     }
+
     my $short_name = delete($args{short_name});
     unless (defined($short_name)) {
         $short_name = 1;
     }
+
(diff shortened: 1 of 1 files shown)


susannasiebert:
I like this option but it does change the default behavior. Previously it always stripped `chr` and now it only strips if the option is set. So we would need to pass this option along in the appropirate places (`Genome::VariantReporting::Suite::Vep::RunResult`, `Genome::Qc::Tool::Bedtools::Intersect`, etc.)

jasonwalker80:
The option is optional and if not defined will default to 1, the original behavior.  It might be best to update all of the callers, but not necessary.

susannasiebert:
Ah yes, you're right. I totally missed that. No need to explicitly set it in the callers. +1

tmooney:
As nice as #1407 would be, this seems like a more practical approach given the circumstances. +1 (assuming tests pass 😄)

jasonwalker80:
CI, test this.

PR #1420, opened 2016-05-12
ClinSeq ModelSummary test uses Test Data

tmooney:
Here's another test that has newly fallen victim to the deletion of old production data.


susannasiebert:
+1

jasonwalker80:
+1

PR #1422, opened 2016-05-13
ClinSeq SummarizeSvs uses test data for is test

susannasiebert:
This test is currently failing on master because it relies on deleted production data. This PR fixes that.


acoffman:
+1

jeldred:
+1 

the only thing that caused me any second thoughts was trying to decide if there was a somvar_build2 to go along with 

my $somvar_build_id1 = 119390903;
my $somvar_build1    = Genome::Model::Build->get($somvar_build_id1);

PR #1423, opened 2016-05-13
Clinseq CreateMutationSpectrum uses Test Data

susannasiebert:


lib/perl/Genome/Model/ClinSeq/Command/CreateMutationSpectrum-exome.t
```diff
@@ -14,27 +14,35 @@ BEGIN {
...
+use Sub::Override;
 
-use_ok('Genome::Model::ClinSeq::Command::CreateMutationSpectrum') or die;
+my $pkg = 'Genome::Model::ClinSeq::Command::CreateMutationSpectrum';
+use_ok($pkg);
+
+use Genome::Model::Build::ReferenceSequence;
+my $override = Sub::Override->new(
+    'Genome::Model::Build::ReferenceSequence::full_consensus_path',
+    sub { return '/gscmnt/ams1102/info/model_data/2869585698/build106942997/all_sequences.fa'; }
```
tmooney: I'm curious why these need to hardcode a reference path.

tmooney:
Notwithstanding from my curiosity, this is definitely an improvement, so +1.

lib/perl/Genome/Model/ClinSeq/Command/CreateMutationSpectrum-exome.t
susannasiebert: I was hoping to discuss this with you. The tool needs a reference sequence in order to run. This is the path to the reference sequence that the original test build uses. I don't want to put it in the clinseq test data because it's such a large file so I wasn't sure how to proceed so I hardcoded the path for now.

tmooney: I generally think it is okay to request "well-known" reference sequences in tests by name (e.g. `Genome::Model::Build::ReferenceSequence->get(name => 'GRCh37-lite-build37');`).    They should be included in the test database as well.

Barring that, it might be okay to symlink to the production version in the test data.  If that file moves we've got bigger trouble 😄

wschroeder: I'm not sure the final form of disk virtualization, but I would be more wary of hardcoded paths for now and lean towards lookups as suggested.  It would reduce the amount of possible changes needed in the future.

susannasiebert: Doh, of course. Should I replace the ClinSeq TestData mocked up reference sequence with build37? Or should I use Sub::Override to get the build 37 and return its fasta?

tmooney:
+1, even better now!

jasonwalker80:
+1

PR #1425, opened 2016-05-19
Add lsb sub additional etc spec

sleongmgi:


wschroeder:
+1 if tests pass; looks consistent with surrounding code

tmooney:
+1 📯

PR #1426, opened 2016-05-23
Path for BER has moved.

tmooney:
These modules are slated for eventual removal... but the contents of the allocation on `/gscmnt/gc9002/` are slated for earlier removal than that.


wschroeder:
+1 👘 

My understanding is that they are wrapping up projects using this pipeline in the next three to six months.

jasonwalker80:
+1

PR #1427, opened 2016-05-23
Add speedseq sv verbose argument.

jasonwalker80:


lib/perl/Genome/Model/Tools/Speedseq/Sv.pm
```diff
@@ -110,8 +109,13 @@ class Genome::Model::Tools::Speedseq::Sv {
                     doc => 'output LUMPY probability curves in VCF',
                     is_optional => 1,
                     tool_param_name => 'P',
-                }
-
+                },
+                verbose => {
+                    is => 'Boolean',
+                    doc => 'verbose',
```
tmooney: This `doc` line is rather terse!

jasonwalker80: I agree.  I was merely using the description in the speedseq sv tool.  I updated the `doc` line a bit, but I'm not sure what else to include in the documentation.

tmooney: Ah. Unless we have some special insight into what it does, I guess that works!

tmooney:
+1 🔔

acoffman:
+1

PR #1428, opened 2016-05-23
Stop using "server_dispatch => 'inline'" feature.

tmooney:
The inline workflow_builder_backend takes care of most use cases.  A future PR will remove this feature from Build entirely.

### DIFF
--- a/lib/perl/Genome/Model/Command/Define/ReferenceSequence.t
+++ b/lib/perl/Genome/Model/Command/Define/ReferenceSequence.t
@@ -15,6 +15,8 @@ use File::Spec;
 
 Genome::Report::Email->silent();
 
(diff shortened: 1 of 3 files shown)


wschroeder:
+1 🏦

sleongmgi:
Is the following set the whole workflow backend to inline?  If so, it might override the previous work flow builder backend.

 my $guard = Genome::Config::set_env('workflow_builder_backend', 'inline');

tmooney:
@sleongmgi Yes, it overrides the backend to run the build inline.  The `$guard` restores the original value when it goes out of scope, so this should only affect the particular build we wanted to run inline.

sleongmgi:
+1

PR #1429, opened 2016-05-25
Remove "sequencing-qc" tool and its corresponding result.

tmooney:
This is another tool that uses old `Workflow`.  Removal is offered here as an alternative to updating it to `WorkflowBuilder`.


jasonwalker80:
+1

jeldred:
:shipit:

PR #1430, opened 2016-05-25
Don't set NO_LSF now that this doesn't run a workflow.

tmooney:
This should have been part of #1386.  Polymutt is still on the list for removal... but that'll be a longer process, since it's currently entwined in the FalsePositiveVcf tests.  This at least gets it out of the list of "things still using Workflow-related variables".


wschroeder:
+1 🔕

jasonwalker80:
+1

PR #1431, opened 2016-05-25
CCSVCF tests use "inline" WorkflowBuilder backend instead of NO_LSF variable.

tmooney:
The CCSVCF tools already use WorkflowBuidler and processes.  This updates the tests to the `WorkflowBuilder`-ish way of running inline.


wschroeder:
+1 🍘

jasonwalker80:
+1

PR #1432, opened 2016-05-25
Don't worry about the value of NO_LSF variable in Build Start test.

tmooney:
After #1429, #1430, and #1431 this is the last reference to `NO_LSF` left in Genome.


wschroeder:
+1 👖

jasonwalker80:
+1

PR #1433, opened 2016-05-25
DeNovoAssembly pipeline uses WorkflowBuilder.

tmooney:
I haven't heard it's okay to remove this pipeline, so here's an update to use WorkflowBuilder!


jasonwalker80:
+1, last I understood Newbler was the only assembler needed by the Reference Grant for BAC assemblies.  @ebelter is Newbler still necessary for the reference grant?

ebelter:
I do not know. No one has asked me about it. I am asking if someone is using it.

ebelter:
+1 Additionally, the other assemblers are NOT being used in the pipeline. They could be removed.

PR #1435, opened 2016-05-25
Remove some old capture commands.

tmooney:
Several were flagged as having hardcoded `bsub`s in them and none of them are recently used, so away they go!


jasonwalker80:
+1, @dkoboldt any objection?

wschroeder:
+1 🙈

tmooney:
If there are any objections, we can always restore these from the git history!

PR #1436, opened 2016-05-25
Remove old workflow for Burden Analysis.

tmooney:
This had two versions: one using `Workflow` and one using a bare `bsub`.  We're trying to eliminate both things in the codebase.  Since this hasn't been used recently, this PR proposes removing the tools rather than writing a third, WorkflowBuilder-based version.


wschroeder:
+1 🌁

acoffman:
+1

PR #1438, opened 2016-05-25
Remove BMR Submit commands.

tmooney:
These are wrappers that `bsub` other BMR commands.  The usage-log has no recent mentions, so this PR proposes removal rather than updating.


wschroeder:
+1 🎣

acoffman:
+1

PR #1439, opened 2016-05-26
Add a blessed SSG test build.

tmooney:
This should be the last change to make before enabling the SSG automated model test.


acoffman:
I make no claims about the contents of the build itself, but this does in fact add a blessed one!

+1

jasonwalker80:
+1

PR #1440, opened 2016-05-26
LSB_SUB_ADDITIONAL needs to be set sooner.

tmooney:
This was being set inside the bsubbed `genome ptero wrapper` command.  Now it will instead be set in the environment for when that `genome ptero wrapper` command is bsubbed.

### DIFF
--- a/lib/perl/Genome/WorkflowBuilder/Command.pm
+++ b/lib/perl/Genome/WorkflowBuilder/Command.pm
@@ -114,9 +114,11 @@ sub _get_ptero_shortcut_method {
 sub _get_ptero_execute_method {
     require Ptero::Builder::Job;
 
-
     my $self = shift;
     my $log_dir = shift;
+
+    local $ENV{LSB_SUB_ADDITIONAL} = Genome::Config::get('lsb_sub_additional') // $ENV{LSB_SUB_ADDITIONAL};
+
     my $ptero_lsf_parameters = $self->_get_ptero_lsf_parameters();
     $ptero_lsf_parameters->{command} = sprintf(
         'genome ptero wrapper --command-class %s '
--- a/lib/perl/Genome/WorkflowBuilder/Command.pm
+++ b/lib/perl/Genome/WorkflowBuilder/Command.pm
@@ -117,7 +117,7 @@ sub _get_ptero_execute_method {
     my $self = shift;
     my $log_dir = shift;
 
-    local $ENV{LSB_SUB_ADDITIONAL} = Genome::Config::get('lsb_sub_additional') // $ENV{LSB_SUB_ADDITIONAL};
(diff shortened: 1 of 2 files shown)


sleongmgi:
+1 if all tests pass.

lib/perl/Genome/WorkflowBuilder/Command.pm
```diff
@@ -114,9 +114,16 @@ sub _get_ptero_shortcut_method {
...
     my $log_dir = shift;
+
+    my $guard;
+    if (my $lsb_sub_additional = Genome::Config::get('lsb_sub_additional')) {
+        my $old = $ENV{LSB_SUB_ADDITIONAL};
+        $guard = Scope::Guard->new(sub { $ENV{LSB_SUB_ADDITIONAL} = $old; });
+        $ENV{LSB_SUB_ADDITIONAL} = $lsb_sub_additional;
+    }
+
     my $ptero_lsf_parameters = $self->_get_ptero_lsf_parameters();
```
wschroeder: Is this approach effectively the same?

```perl
local $ENV{LSB_SUB_ADDITIONAL} = Genome::Config::get('lsb_sub_additional') // $ENV{LSB_SUB_ADDITIONAL};
```

tmooney: Indeed!  I'd talked myself out of using `local` for some reason...  I'll change it.

tmooney: I ended up changing `//` to `||` because config has a default value of the empty string.  I'll come to regret this if we ever need to set this to `0`, but that doesn't seem likely to me! 😄

wschroeder: In several other language APIs, I have seen a "default value" concept for retrieving values.  Another option here could be:

```perl
local $ENV{LSB_SUB_ADDITIONAL} = Genome::Config::get('lsb_sub_additional', default => $ENV{LSB_SUB_ADDITIONAL});
```

For extra flexibility for situations where default values have costly derivation logic:

```perl
local $ENV{LSB_SUB_ADDITIONAL} = Genome::Config::get('lsb_sub_additional', default => sub { $ENV{LSB_SUB_ADDITIONAL} });
```

tmooney: `Genome::Config` variables must have values. (My original conditional was pointless, as the tests helpfully pointed out!)

Although the extra parameter might make sense for this particular variable, in general it's not clear to me when we would elect to use the supplied default in alternative to the existing value.

wschroeder: Good learning for me!  So the issue here is that there's an empty string, and since someone intentionally chose that, that should be honored.

wschroeder:
+1 😰

tmooney:
I'll merge this one early since model tests will probably fail without it 😄

PR #1441, opened 2016-05-26
Only pass through boolean tool argument if true

apregier:
Otherwise just pass an empty string

### DIFF
--- a/lib/perl/Genome/Model/Tools/Speedseq/Base.pm
+++ b/lib/perl/Genome/Model/Tools/Speedseq/Base.pm
@@ -128,8 +128,14 @@ sub _tool_input_file_metas {
 # formatting (argname=argvalue).
 sub _format_tool_arg {
     my ($type, $name, $value) = @_;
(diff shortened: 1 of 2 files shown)


lib/perl/Genome/Model/Tools/Speedseq/Base.pm
```diff
@@ -128,7 +128,8 @@ sub _tool_input_file_metas {
 # formatting (argname=argvalue).
 sub _format_tool_arg {
     my ($type, $name, $value) = @_;
-    return '-'. $name if ($type eq 'Boolean');
+    return '-'. $name if ($type eq 'Boolean' and $value);
+    return "" if ($type eq 'Boolean');
```
tmooney: It might be nice to group the Boolean handling in a block or subroutine instead of checking the condition at the end of each line to which it applies.

tmooney:
+1, two possible values for booleans are better than one!

acoffman:
+1

PR #1442, opened 2016-05-27
Use tumor_bam and normal_bam subroutines instead of RefAlign builds

### LINKED ISSUE #1409: Add Somatic Validation tracks to DumpIgvXml ClinSeq command.
The DumpIgvXml command relies on finding the tumor_build and normal_build which only exist as ReferenceAlignment builds on Somatic Variation inputs.  The DumpIgvXml command should resolve the path to the tumor/normal bed/bam file for all somatic model types.

susannasiebert:
Closes #1409

### DIFF
--- a/lib/perl/Genome/Model/ClinSeq/Command/DumpIgvXml.pm
+++ b/lib/perl/Genome/Model/ClinSeq/Command/DumpIgvXml.pm
@@ -194,79 +194,76 @@ sub execute {
             my $wgs_build = $clinseq_build->wgs_build;
 
             #WGS normal refalign build - source of normal WGS BAMs track
-            if ($wgs_build->can('normal_build')) {
-                my $wgs_normal_refalign_build = $wgs_build->normal_build;
-                my $wgs_normal_track = $self->generate_track_xml(
-                    '-build'         => $wgs_normal_refalign_build,
-                    '-resource_type' => 'bam'
-                );
-                my $wgs_normal_xml = $wgs_normal_track->{xml};
-                $wgs_normal_xml = $self->generate_panel_xml(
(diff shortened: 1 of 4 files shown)


jasonwalker80:
+1

lib/perl/Genome/Model/ClinSeq/Command/DumpIgvXml.pm
```diff
@@ -194,79 +194,76 @@ sub execute {
...
-                );
-                my $wgs_normal_xml = $wgs_normal_track->{xml};
-                $wgs_normal_xml = $self->generate_panel_xml(
-                    '-track_xml'   => $wgs_normal_xml,
-                    '-panel_count' => $panel_count++
-                );
-                push(@resources, $wgs_normal_track->{resource});
-                $main_features_track_xml .= $wgs_normal_xml;
-            }
+            my $wgs_normal_build = $wgs_build->can('normal_build') ? $wgs_build->normal_build : $wgs_build;
```
tmooney: Would this be something useful to add to the `Genome::Model::Build::SomaticInterface`? Something like `build_for_normal_bam`?

(I was surprised to see that `tumor_bam` and `normal_bam` aren't really a part of that interface.)

tmooney:
+1 🔥

lib/perl/Genome/Model/ClinSeq/Command/DumpIgvXml.pm
susannasiebert: Great idea. I pushed a commit to implement your suggestion.

jasonwalker80:
+1, I like the suggestion of moving these into the interface.

tmooney:
(+1 again!)

PR #1443, opened 2016-05-27
Speedseq sv config file

jasonwalker80:
Generate a per chromosome FASTA software result and use that directory for cnvnator in a Speedseq config file result that is both speedseq version and reference sequence specific.


susannasiebert:
+1 once tests pass

lib/perl/Genome/Model/Tools/Speedseq/ConfigFile.pm
```diff
@@ -0,0 +1,50 @@
...
+
+sub _run {
+    my $self = shift;
+
+    my $per_chr = $self->reference_sequence_build->per_chromosome_fastas();
+    my $fasta_dir = $per_chr->fasta_directory;
+    my $old_config_path = Genome::Model::Tools::Speedseq::Base->config_path_for_version($self->speedseq_version);
+    my $new_config_path = File::Spec->join($self->temp_staging_directory, 'speedseq.config');
+
+    my $cmd = "sed 's|CNVNATOR_CHROMS_DIR=.*\$|CNVNATOR_CHROMS_DIR=$fasta_dir|' $old_config_path > $new_config_path";
```
tmooney: In the long run, I don't know that it's safe to trust that the allocation for the per-chromosome FASTAs will never move.  It probably won't happen often, though, so maybe we invalidate the result if it does and not worry about it?

tmooney:
+1 📼

lib/perl/Genome/Model/Tools/Speedseq/ConfigFile.pm
jasonwalker80: That is a concern. I think it's acceptable to invalidate the results if the paths indeed change.  With result users hopefully that is not too challenging if in fact we encounter this situation.

PR #1444, opened 2016-05-31
Update annotation builds to some that exist.

tmooney:
I accidentally deleted a few annotation build directories that were in use by these tests.  This updates the test to use others that still exist. 😄 

I'm not keen on the hard-coded path, but this maintains the status quo.


susannasiebert:
How difficult would it be to switch to a version using mock objects?

tmooney:
It turned out to be easy enough to extract the files needed for the Chimerascan index test, which was using a fake annotation build but real data previously.

The other test is using a lot more of a real annotation build.  I'm not sure if it's a good idea to copy so much of it or if it falls under the "well-known data" exception in some way.  We could convert it to ask for the annotation build by name and version instead of by build ID...

susannasiebert:
+1

jeldred:
:+1: ⚽

PR #1446, opened 2016-06-01
Add a config variable for the LSF resource for annotation.

tmooney:
The default 4GB is normally sufficient here, but occasionally we have to increase it.

### DIFF
--- a/etc/genome/spec/lsf_resource_annotate_variants.yaml
+++ b/etc/genome/spec/lsf_resource_annotate_variants.yaml
@@ -1,0 +1,3 @@
+---
+default_value: "-M 4000000 -R 'select[mem>=4000 && tmp>=2000] rusage[mem=4000,tmp=2000]'"
+env: XGENOME_LSF_RESOURCE_ANNOTATE_VARIANTS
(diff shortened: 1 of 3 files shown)


jasonwalker80:
+1

etc/genome/spec/lsf_resource_annotate_variants.yaml
```diff
@@ -0,0 +1,3 @@
+---
+default_value: "-M 4000000 -R 'select[mem>=4000 && tmp>=2000] rusage[mem=4000,tmp=2000]'"
```
ebelter: Are we using tmp or gtmp?

ebelter:
+1

etc/genome/spec/lsf_resource_annotate_variants.yaml
tmooney: I believe our code converts both to whichever one we're using these days.  I've heard discussion on changing which one that is, but I don't think there's been recent movement on that.

FWIW I copied this resource string from another config verbatim.

ebelter: Just for my curiosity! 🐫

PR #1447, opened 2016-06-02
SX Split By Ns UPDATE

ebelter:
use iterator style to get split seqs
change method for finding Ns to accomodate the max allowed characters in a regex
return gap info if requested


tmooney:
+1 🐓

PR #1449, opened 2016-06-06
Keep logs around when Ptero's post-exec fails.

tmooney:
If Ptero's post-exec fails, these logs are never copied out of /tmp/ and into Ptero's DB, so they are lost  exactly when we'd be most interested in them!  This writes them to the Workflow's log directory, so in case of trouble they'll stick around for inspection.  Otherwise our post-exec should clean them up as usual.

### DIFF
--- a/lib/perl/Genome/WorkflowBuilder/Command.pm
+++ b/lib/perl/Genome/WorkflowBuilder/Command.pm
@@ -10,6 +10,7 @@ use Data::UUID;
 use Carp qw();
 use Data::Dump qw(pp);
 use Try::Tiny;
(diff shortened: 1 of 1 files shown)


sleongmgi:
Is it possible to write it to logstash?

tmooney:
Maybe?  I'm not sure what I would be trying to put into logstash from this.

If there's an error condition inside `ptero-lsf-pre-exec` and `ptero-lsf-post-exec`, they might write them to logstash... but that's not part of Genome.

jeldred:
👍  a pure improvement

PR #1451, opened 2016-06-08
Add PindelRegion to DV2

dufeiyu:
There is a cle pipeline need for pindel running on region file.


lib/perl/Genome/Model/Tools/DetectVariants2/PindelRegion.pm
```diff
@@ -0,0 +1,78 @@
...
+    
+    return 1;
+}
+
+
+sub has_version {
+    my ($self, $version) = @_;
+
+    $version = $self->version unless defined $version;
+    my @versions = Genome::Model::Tools::Pindel::RunPindel->available_pindel_versions;
```
tmooney: Should this only check the `pindel_region_compatible_versions` instead?

lib/perl/Genome/Model/Tools/Pindel/RunPindel.pm
```diff
@@ -447,6 +458,25 @@ sub _run_pindel_for_chromosome {
...
 
+sub _run_pindel_for_region {
+    my ($self, $region_file) = @_;
+    
+    my $output_basename = join '/', $self->output_directory, 'all_sequences';
+    my $window_size     = $self->window_size;
+    my $ref_seq         = $self->reference_sequence_input;
+
+    my $cmd = $self->pindel_path . " -f ".$ref_seq. " -i " . $self->_config_file . " -o ". $output_basename . " -j ".$region_file . " -w ".$window_size." -b /dev/null";
+    Genome::Sys->shellcmd(cmd => $cmd);
```
tmooney: I think it's preferable for new `shellcmd` calls to use the array form of `cmd` instead of the string form when possible, i.e.
```perl
my $cmd = [$self->pindel_path, "-f ", $ref_seq, "-i", $self->_config_file, "-o", $output_basename, "-j", $region_file, "-w", $window_size, "-b", "/dev/null"];
```

lib/perl/Genome/Model/Tools/DetectVariants2/PindelRegion.pm
dufeiyu: good point. will make the change.

lib/perl/Genome/Model/Tools/Pindel/RunPindel.pm
dufeiyu: hmm, didn't even know this array ref form. I will make the change.

tmooney:
+1.

Does this require that the region list be a part of the processing profile? That's probably fine if it won't change often.  Otherwise it might be better to get it from a build input somehow.

dufeiyu:
Yes. It needs either region file path or region feature list id inside the parameter (similar to mutect). For cle use, this region file will contain some known mutation "hot spots" and won't be changed often.

PR #1452, opened 2016-06-09
Fix ptero-lsf-{pre,post}-exec within docker

tmooney:
LSF doesn't automatically launch the pre- and post-exec commands within the chosen docker container, so we've got to call the docker_run.py script ourselves.

In the future I'd like to make some ptero-lsf docker images to accomplish the pre- and post-exec tasks rather than specifying these commands manually within the main container, but this gets these back to their main task of maintaining workflow status.


jeldred:
:shipit: bitmoji

acoffman:
+1

Out of curiosity, why is one of the commands invoked directly as an executable, while the other is called via bash -C ?

tmooney:
That is a very good question! I wish I knew the answer... I just tried to keep it how it was.

acoffman:
Hah, fair enough!

PR #1454, opened 2016-06-09
Remove old "PolymuttPipeline" workflow tool.

tmooney:
There are still some other Polymutt tools left to be removed, but this one uses old `Workflow` so it gets removed first!


acoffman:
+1 ☀️

PR #1455, opened 2016-06-09
`gmt annotate reference-genome` uses WorkflowBuilder.

tmooney:
This is somehow a part of ImportedAnnotation builds.  I'm not sure if it's really still needed, but converting it to `WorkflowBuilder` was faster than figuring that out 😄


wschroeder:
+1  Not familiar with some of the packages and methods in use here, but I especially like the reduction in the function's cyclic complexity.

acoffman:
+1

PR #1456, opened 2016-06-09
`gmt hgmi` uses WorkflowBuilder

tmooney:
It's my understanding that these tools are still in active use for the next few months, so here's a PR to update them to `WorkflowBuilder`!


wschroeder:
+1 Consistent with the other workflow conversion change.

acoffman:
+1

PR #1458, opened 2016-06-13
Make trio wrapper work on somatic cases

dufeiyu:
Currently trio wrapper works on three samples: normal, tumor and followup. But there are cases in cle that only have normal and tumor available. This PR is to tweak the trio command and its supporting modules to make trio work on cases without followup.

### DIFF
--- a/lib/perl/Genome/VariantReporting/Command/Wrappers/ModelPairFactory.pm
+++ b/lib/perl/Genome/VariantReporting/Command/Wrappers/ModelPairFactory.pm
@@ -79,13 +79,12 @@ sub get_model_pairs {
                 label => 'discovery',
             );
         }
+        elsif (@models != 2) { 
+            $self->warning_message("Skipping models for ROI %s because there are not either one or two models: %s",
+                $roi, join(", ", map {$_->__display_name__} @models));
+            next;
+        }
         else {
-            unless (@models == 2) {
-                $self->warning_message("Skipping models for ROI %s because there are not exactly two models: %s",
-                    $roi, join(", ", map {$_->__display_name__} @models));
-                next;
-            }
-
             my @discovery_models = grep { $self->is_model_discovery($_) } @models;
             my @validation_models = grep { $self->is_model_followup($_) } @models;
 
--- a/lib/perl/Genome/VariantReporting/Command/Wrappers/ModelPairFactory.pm
+++ b/lib/perl/Genome/VariantReporting/Command/Wrappers/ModelPairFactory.pm
@@ -13,7 +13,7 @@ class Genome::VariantReporting::Command::Wrappers::ModelPairFactory {
             is_many => 1,
         },
         discovery_sample => { is => 'Genome::Sample', },
-        followup_sample => { is => 'Genome::Sample', },
+        followup_sample => { is => 'Genome::Sample', is_optional => 1},
         normal_sample => { is => 'Genome::Sample',},
         other_input_vcf_pairs => { is => 'Hashref', default_value => {}},
     },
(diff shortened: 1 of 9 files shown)


lib/perl/Genome/VariantReporting/Command/Wrappers/ModelPairFactory.pm
```diff
@@ -68,58 +68,76 @@ sub get_model_pairs {
...
+            $discovery_build = $models[0]->last_succeeded_build;
+            push @model_pairs, Genome::VariantReporting::Command::Wrappers::ModelPair->create(
+                common_translations => $self->get_common_translations(),
+                plan_file_basename => "cle_somatic_TYPE_report.yaml",
+                discovery => $discovery_build,
+                label => 'discovery',
+            );
         }
+        else {
+            unless (@models == 2) {
```
tmooney: It's a little confusing that the first thing in this `else` is another check on the same variable as in the `if`.  It might be clearer to make `@models != 2` its own `elsif` case and then leave the other logic in the `else`.

dufeiyu: It makes sense. I will make the change.

lib/perl/Genome/VariantReporting/Command/Wrappers/ModelPairFactory.t
```diff
@@ -51,8 +38,8 @@ subtest "Three models for an roi" => sub {
     );
 
     my @pairs = @{$factory->get_model_pairs};
-    ok(@pairs == 0, "Factory with only one model returned no pairs");
-    ok($factory->warning_message =~ /Skipping models for ROI $roi_name because there are not exactly two models/,
+    ok(@pairs == 0, "Factory with three model returned no pairs");
```
tmooney: This should probably say `three models`.

lib/perl/Genome/VariantReporting/Command/Wrappers/ModelPairFactory.pm
```diff
@@ -68,58 +68,75 @@ sub get_model_pairs {
...
                 common_translations => $self->get_common_translations(),
                 discovery => $discovery_build,
-                followup => $validation_build,
                 plan_file_basename => "cle_docm_report_TYPE.yaml",
                 label => $other_input_vcf_pair,
                 other_snvs_vcf_input => $self->other_input_vcf_pairs->{$other_input_vcf_pair}->[0],
                 other_indels_vcf_input => $self->other_input_vcf_pairs->{$other_input_vcf_pair}->[1],
             );
+            if ($validation_build) {
+                $params{followup} = $validation_build
```
tmooney: I wonder if it would be better to move:
```perl
plan_file_basename => "cle_docm_report_TYPE.yaml",
```
inside of here as well.

tmooney:
+1 notwithstanding my nitpicking 😄

lib/perl/Genome/VariantReporting/Command/Wrappers/ModelPairFactory.t
dufeiyu: good catch. I will make the change.

lib/perl/Genome/VariantReporting/Command/Wrappers/ModelPairFactory.pm
dufeiyu: I agree. It would be clearer to move "cle_docm_report_TYPE.yaml" here to match "cle_somatic_docm_report_TYPE.yaml" in downstairs else block. I will make the change.

dufeiyu:
@apregier  Alison, can you take a look at this ? thanks.

dufeiyu:
CI, test this

apregier:
The changes look good to me.  What it doesn't have is an integration-type test.  The only real test for the Trio wrapper was moved to the 'process' tests in jenkins, and without adding another one with only two models, there may be additional problems you aren't catching

dufeiyu:
I agree that an integration test is probably needed for duo/somatic cases of trio wrapper. Ideally this "duo" process can be added to jenkins in the future. There is a subtest "one discovery model for an roi" in ModelPairFactory.t related to this change. On the other hand, most tweak of this PR is to adding conditions to command to make followup optional, which shouldn't have much impact on individual process and output/results. The real process/command/result tests of this duo should already be included in trio jenkins test since duo is part of trio. I did run test on a somatic case whose "normal + discovery" are from a real trio. The directory diff is fine.

PR #1459, opened 2016-06-13
BER/Prediction Workflow uses WorkflowBuilder

tmooney:
I believe these are the last places in `Genome` using `Workflow::Simple` aside from our build start logic (and other open PRs).


wschroeder:
+1 📟

jeldred:
:+1: why are these still in the repo?

tmooney:
We have been asked to keep the GenePrediction pipeline around for now, so here we are patching it up!

PR #1460, opened 2016-06-15
Remove uses of Workflow::Model in GenePrediction.

tmooney:
These didn't `use Workflow` or its derivatives but still referred to `Workflow::Model`.


wschroeder:
+1 🌖

PR #1461, opened 2016-06-15
SX: Trim By Position

ebelter:


lib/perl/Genome/Model/Tools/Sx/Trim/ByPosition.pm
```diff
@@ -0,0 +1,140 @@
...
+use Params::Validate (qw/ :types validate_pos /);
+use Regexp::Common;
+use Try::Tiny;
+
+class Genome::Model::Tools::Sx::Trim::ByPosition {
+    is => 'Genome::Model::Tools::Sx::Base',
+    has => {
+        positions_path => {
+            is => 'File',
+            doc => 'Path to he file of positions to trim.',
```
tmooney: `the` has lost a `t`.

lib/perl/Genome/Model/Tools/Sx/Trim/ByPosition.pm
```diff
@@ -0,0 +1,140 @@
...
+        },    
+    },
+    has_transient_optional => {
+        trim_positions => { is => 'HASH', },
+    },
+};
+
+sub help_brief { "Trim sequences by positions" }
+sub help_detail { help_brief() }
```
tmooney: Is the format for the positions file documented somewhere or well-known?

lib/perl/Genome/Model/Tools/Sx/Trim/ByPosition.pm
```diff
@@ -0,0 +1,140 @@
...
+        if ( not $trim_positions ) {
+            # pcap naming
+            my $pcap_seq_id = $seq->{id};
+            $pcap_seq_id =~ s/scaffold/Contig/g;
+            $trim_positions = $self->trim_positions->{$pcap_seq_id};
+            return 'ALL' if not $trim_positions; # keep it!
+        }
+    }
+
+    return if not ref $trim_positions; # trim all
```
tmooney: Just for my own clarification: Does this `return` happen when `load_positions` sets the value to `ALL`? And that's the opposite of what happens if `keep_positions_for_sequence` returns `ALL`?  Are there other cases when `$trim_positions` might not be a `ref`?

lib/perl/Genome/Model/Tools/Sx/Trim/ByPosition.pm
```diff
@@ -0,0 +1,140 @@
...
+
+    my @keep_positions;
+    my $current_pos = 0;
+    for my $set ( @$trim_positions ) {
+        push @keep_positions, [ $current_pos, ( $set->[0] - $current_pos - 1 ) ];
+        $current_pos = $set->[1];
+    }
+
+    # keep the last part
+    push @keep_positions, [ $current_pos, ( length($seq->{seq}) - $current_pos ) ];
```
tmooney: What does this do when `$stop` is `end`?

lib/perl/Genome/Model/Tools/Sx/Trim/ByPosition.pm
ebelter: Ah - no. This is a lift over a Kyung script. I deplore the format, but it's consistency was insisted upon. I will add docs.

lib/perl/Genome/Model/Tools/Sx/Trim/ByPosition.pm
ebelter: I believe the ref check is a hold over from when I was developing this,. I will remove it.

lib/perl/Genome/Model/Tools/Sx/Trim/ByPosition.pm
ebelter: I'll add a condition to only add this if the length is > 0.

lib/perl/Genome/Model/Tools/Sx/Trim/ByPosition.pm
ebelter: I will clarify this in the code.

lib/perl/Genome/Model/Tools/Sx/Trim/ByPosition.pm
```diff
@@ -0,0 +1,152 @@
...
+        }
+    }
+
+    return if not @$trim_positions; # sequence indicated, but no trim positions given means trim all/keep none
+
+    my @keep_positions;
+    my $current_pos = 0;
+    for my $set ( @$trim_positions ) {
+        push @keep_positions, [ $current_pos, ( $set->[0] - $current_pos - 1 ) ];
+        $current_pos = $set->[1];
```
tmooney: Maybe this would be a good place to convert [the stop value `'end'`](https://github.com/ebelter/genome/blob/9b94526cc4f7aa12acc6ebf33950146a8423c7ab/lib/perl/Genome/Model/Tools/Sx/Trim/ByPosition.pm#L86) to an actual number? Like
```perl
$current_pos = length($seq->{seq}) if $current_pos eq 'end';
```

I guess the below condition would work anyway, but with warnings about `'end'` not being numeric.

ebelter: Ah! I decided not to support the 'end' condition. I have removed it, and all positions must be numeric.

lib/perl/Genome/Model/Tools/Sx/Trim/ByPosition.pm
```diff
@@ -0,0 +1,152 @@
...
+        if ( not $positions ) {
+            $trim_positions{$seq_id} = [];
+            next;
+        }
+
+        for my $set ( split(/,/, $positions) ) {
+            my ($start, $stop) = split(/\-/, $set);
+            if ( $RE{num}{int}->matches($stop) ) {
+                $class->fatal_message('Invalid positions for %s! %s', $seq_id, $set) if $start == $stop or $start > $stop;
+            } elsif ( $stop ne 'end' ) {
```
tmooney: I'm putting a comment here just so I can see it vanish when the `end` support is dropped :smile:

ebelter: :P Commit THEN push.

tmooney:
+1 💴

ebelter:
CI, test this!

PR #1462, opened 2016-06-20
Specify queues in Mutect and SomVal's VerifyBam.

tmooney:
The default queue is no longer guaranteed to work, which makes it easier to spot cases where the queue specification has been omitted.


wschroeder:
+1 if tests pass.

jeldred:
+1 🌴

PR #1463, opened 2016-06-20
Increase annotation memory usage from 4g to 8g

dufeiyu:
Recently there are quite a few of cle somatic validation builds failed at annotation step due to out of memory. This PR doubles the default memory usage for variant annotation.


tmooney:
+1.

This seems like it's been a frequent problem specifically for CLE.  I wonder if it might be better in the CLE user's `.config/genome/config.yaml`.  (It's not a huge memory request, so I'm not too worried about it either way.)

PR #1464, opened 2016-06-20
Remove References to Volumes Being Retired

tmooney:
Having a bunch of hard-coded paths around in the code is sub-optimal!  Most of these were in comments or examples, but a few remove defaults.

Subsequent PRs will address other affected modules for which I think more expansive changes are required.

### DIFF
--- a/lib/perl/Genome/Model/Tools/Somatic/FilterMouseBases.pm
+++ b/lib/perl/Genome/Model/Tools/Somatic/FilterMouseBases.pm
@@ -23,7 +23,6 @@ class Genome::Model::Tools::Somatic::FilterMouseBases {
            is_input => 1,
            is_output => 1,
            doc => 'UCSC liftover chain file for human to mouse',
-           example_values => ['/gscmnt/sata112/info/medseq/reference_sequences/Homo_sapiens/liftOver_files/hg19_GRCh37_Build-37/hg19ToMm9.over.chain.gz'],
        },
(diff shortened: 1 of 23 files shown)


lib/perl/Genome/Model/Tools/Somatic/FilterMouseBases.pm
```diff
@@ -23,7 +23,6 @@ class Genome::Model::Tools::Somatic::FilterMouseBases {
            is_input => 1,
            is_output => 1,
            doc => 'UCSC liftover chain file for human to mouse',
-           example_values => ['/gscmnt/sata112/info/medseq/reference_sequences/Homo_sapiens/liftOver_files/hg19_GRCh37_Build-37/hg19ToMm9.over.chain.gz'],
        },
```
wschroeder: Should this use the `/path/to/` approach instead of completely vanishing?

wschroeder:
+1 after the quick-tests issue is fixed.  I like the removal of all these hardcoded paths and defaults.

lib/perl/Genome/Model/Tools/Somatic/FilterMouseBases.pm
tmooney: The `/path/to` approach is useful in the synopsis, but I prefer that `example_values` be things that one can really use as the value.

jeldred:
💯

PR #1465, opened 2016-06-20
Remove old CopyNumber tools that depend on volumes being retired.

tmooney:
These have several references to `sata921` and occasionally user home directories. (@chrisamiller may have an opinion about these?)


wschroeder:
+1

chrisamiller:
Yeah, these are outdated and can be nuked. Thanks, Tom. +1

PR #1467, opened 2016-06-20
Remove old plotSV.pl script.

tmooney:
It has references to both homedirs and a volume being retired.


wschroeder:
+1 🍂

jeldred:
:shipit: : 🚢 🐑

PR #1468, opened 2016-06-20
UCSC annotation file changes for ReferenceSequence

tmooney:
This changes two things:
* The UCSC conservation directory must be manually linked to each annotation build that wants it.
* The UCSC tiering files for a reference should be linked into the reference build instead of using a hard-coded path.


tmooney:
(This depends on #1470 to pass tests.)

jasonwalker80:
I presume the existing tiering files have been moved to the appropriate refseq locations?

I don't believe conservation scores are used anymore.  So manual linking seems fine if it's really desired in the future.

tmooney:
Yep! I put the files in the new expected locations before opening this PR.

jasonwalker80:
+1

susannasiebert:
+1

PR #1469, opened 2016-06-21
Speedseq SV remove sort and symlink VCF.

jasonwalker80:
1. Skip the sort since the output is VCF.
2. Generate a symlink for the VCF (and index) so the DV2 Dispatcher will find the output file.
3. Increase RAM to 16GB

### DIFF
--- a/lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.pm
+++ b/lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.pm
@@ -7,7 +7,6 @@ use Genome;
 
 use File::Basename qw(fileparse);
 use File::Spec;
-use Cwd;
 
 class Genome::Model::Tools::DetectVariants2::SpeedseqSv {
     is => 'Genome::Model::Tools::DetectVariants2::Detector',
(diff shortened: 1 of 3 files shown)


lib/perl/Genome/Model/Tools/DetectVariants2/SpeedseqSv.pm
```diff
@@ -67,7 +67,9 @@ sub _detect_variants {
     # Make a symlink to the actual VCF as the DV2 Dispatcher expects an output of svs.hq rather than BED or VCF output
     my ($vcf_file) = glob($self->_sv_staging_output .'*.vcf.gz');
     if ($vcf_file) {
-        symlink($vcf_file,$self->_sv_staging_output);
+        my $rel_path = File::Spec->abs2rel( $vcf_file, $self->_temp_staging_directory ) ;
+        Genome::Sys->create_symlink($rel_path,$self->_sv_staging_output .'.vcf.gz');
+        Genome::Sys->create_symlink($rel_path,$self->_sv_staging_output .'.vcf.gz.tbi');
```
tmooney: Should this be adding a `.tbi` to the `$rel_path` as well?

jasonwalker80: good catch,  thanks tom!

tmooney:
+1 🔋

It might be nice to add a test to make sure the standard files are around after it's done running.  Something like:
```perl
ok(-e $command->sv_output, 'output file created as expected');
```
(I think the test I wrote there may fail.  But that's the idea.)

jasonwalker80:
Thanks for the suggestion @tmooney.  Tests added.

susannasiebert:
+1

PR #1470, opened 2016-06-21
Use GDC Portal instead of TCGA Portal

tmooney:
The [TCGA Data Portal](https://tcga-data.nci.nih.gov) has been down several days and contains a note directing users to the GDC Portal instead.  This updates our TCGA barcode query to retrieve the UUID from the GDC API.

### DIFF
--- a/lib/perl/Genome/Model/Tools/Vcf/Convert/Base.pm
+++ b/lib/perl/Genome/Model/Tools/Vcf/Convert/Base.pm
@@ -11,6 +11,7 @@ use Params::Validate qw(validate CODEREF SCALAR);
 use POSIX 'strftime';
 use LWP::UserAgent;
 use Carp qw(croak);
+use URI::Escape qw(uri_escape);
 
 class Genome::Model::Tools::Vcf::Convert::Base {
     is => 'Command',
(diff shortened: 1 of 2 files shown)


lib/perl/Genome/Model/Tools/Vcf/Convert/Base.pm
```diff
@@ -389,20 +389,21 @@ sub query_tcga_barcode_error_template {
...
     my $agent = LWP::UserAgent->new();
 
     # based on instrumentation the upper bound on response times is 700 ms
     $agent->timeout(5);
 
-    my $request = HTTP::Request->new(POST => $url);
-    $request->content_type('text/plain');
-    $request->content($barcode_str);
+    my $query = join('%2C', map { '%22' . $_ . '%22' } @$barcodes);
+    my $filters = '%7B%22op%22%3A%22in%22%2C%22content%22%3A%7B%22field%22%3A%22samples.portions.analytes.aliquots.submitter_id%22%2C%22value%22%3A%5B' . $query. '%5D%7D%7D';
```
acoffman: I might URL encode this on the fly so that the string is a little more readable in code. Likewise with the filters, is it possible to store them in some sort of data structure rather than a string, at least until they're needed?

Basically, is there an OO-ier way of building this query that isn't just a bunch of string concatenation?

tmooney: Fair point!  I had been lazy and copy+pasted this escaped version from the docs and edited only the query.  I've now changed it convert to JSON on the fly.

acoffman:
+1

ebelter:
+1

PR #1471, opened 2016-06-22
FeatureList Import: Skip blank lines instead of crashing.

tmooney:
I'm not sure why there'd be blank lines here to begin with, but at least there's a straightforward thing to do with them. 😄


wschroeder:
+1 🌼

ebelter:
+1 👒

PR #1472, opened 2016-06-22
Only check for the annotated vcf if the build exists

acoffman:
@kkrysiak ran into the following error in production:

```
Could not start build 34774bf01c0d472480b8cd6ca3a86c7c, reason: Can't call method "has_snvs_annotated_variants_vcf_file" on an undefined value at /gsc/scripts/opt/genome/snapshots/genome-3714/lib/perl/Genome/Model/ClinSeq.pm line 784.
```

This should hopefully remedy that!

### DIFF
--- a/lib/perl/Genome/Model/ClinSeq.pm
+++ b/lib/perl/Genome/Model/ClinSeq.pm
@@ -781,10 +781,10 @@ sub _resolve_workflow_for_build {
     my ($best_converge_snv_indel_report_op, $best_mq_bq_sum) = (undef, -Inf);
     if ($build->wgs_build || $build->exome_build) {
         my ($wgs_annotate_snvs_vcf_op, $exome_annotate_snvs_vcf_op);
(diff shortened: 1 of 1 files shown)


tmooney:
I think at this point it might be clearer to move the construction of this step inside the check that happens on [line 804](https://github.com/genome/genome/blob/226f545b79ef90919e038c29bc0f660ea3c6231d/lib/perl/Genome/Model/ClinSeq.pm#L804).

Since [Line 815](https://github.com/genome/genome/blob/226f545b79ef90919e038c29bc0f660ea3c6231d/lib/perl/Genome/Model/ClinSeq.pm#L815) looks like the only place using these newly constructed steps, that ought to simplify some of the logic in here and we could eliminate the `%annotate_snvs_vcf_ops` hash altogether.

acoffman:
Good suggestion @tmooney, I believe I've addressed it.

tmooney:
+1 with passing tests!

jasonwalker80:
+1

PR #1473, opened 2016-06-23
Add Genome::Sys::LSF::bsub Docker Support

tmooney:
We switched our testing user to default to the docker queues.  That pointed out this bsub wasn't setting the requisite environment variable for it to work.


tmooney:
This depends on #1474 and #1475 to pass tests.

wschroeder:
+1 after dependencies are working

PR #1474, opened 2016-06-23
Better Model/RnaSeq.t Diffs

tmooney:
The existing diff logic was dependent on configuration having certain values, which defeats the purpose of configuration!  (We recently changed away from the assumed values, exposing this flaw.)


ebelter:
+1

PR #1475, opened 2016-06-23
Don't worry about matching queues and resource requests.

tmooney:
This is the same concept as #1474 but for a different test.


ebelter:
+1

wschroeder:
+1 ⏩

PR #1478, opened 2016-06-28
Replace hard-coded NHLBI and 1000 Genomes files with parameters.

tmooney:
The hard-coded versions are on volumes slated for removal, and hardcoded paths are bad, anyway!

This tool appears to be in active use, so a CHANGELOG entry is included with this commit.


wschroeder:
+1 📛

acoffman:
+1

PR #1479, opened 2016-06-28
Remove GMTs that relied on Workflow integration.

tmooney:
These used a `Workflow`-specific way of parallel_by that effectively made them workflows unto themselves.  The underlying commands are still around for use if needed for now.


wschroeder:
I have been advocating for us leveraging Ptero over the Workflow stuff, but I am concerned about removing tests for existing packages that may warrant future maintenance.  Should we delete the tools as well?  Is someone going to replace the tests with more modern tests?

tmooney:
@wschroeder I'd be happy to delete more tools!  (or write new tests as needed :smile:) I'm not sure which tests you're concerned about, though.

wschroeder:
That's because I didn't check all the package names there.  How silly!  Ignore my comment.  Instead:

+1 🎃

acoffman:
+1

PR #1480, opened 2016-06-28
SomVar uses WorkflowBuilder

tmooney:
The base class has long since accepted either a `Genome::WorkflowBuilder::DAG` or a `Workflow::Model`. This PR switches SomaticVariation from the latter to the former.


wschroeder:
+1 💹

acoffman:
+1

PR #1482, opened 2016-06-29
Convert RnaSeq Workflow Generation to WorkflowBuilder.

tmooney:
`Genome::Model::Build->start` was already converting the resulting workflow to `WorkflowBuilder` when starting the build, but now it uses it the whole time.  Another step on the road to removing `Workflow`!


jasonwalker80:
+1

susannasiebert:
+1

PR #1484, opened 2016-06-29
Remove `genome individual cpu-disk-summary`.

tmooney:
This command is coupled to Workflow builds, and so it does not work in our new more Ptero-iffic world.  The usage log reports no uses, so remove it rather than fixing it up.


wschroeder:
+1 🎀

acoffman:
+1

PR #1485, opened 2016-06-29
ClinSeq: Remove false comment.

tmooney:
This was converted to WorkflowBuilder in #1326, but this comment was missed.  (This now returns a `Genome::WorkflowBuilder::DAG`, but I don't think a comment to that effect is necessary.)


wschroeder:
+2

PR #1486, opened 2016-06-29
Remove `gmt bwa align-reads` and unused "use Workflow;" includes.

tmooney:
This removes Genome/Model/Tools/Bwa/AlignReads.pm (which is not part of `Genome::InstrumentData::AlignmentResult::Bwa`) as well as a bunch of spurious `use Workflow;`s in modules that don't appear to really use `Workflow`.

There may be some other `use Workflow;`s to remove later.  Most of the rest are in modules already slated for removal in other open PRs, so I'm holding off on including them here to avoid conflicts.


wschroeder:
+1 🐺

acoffman:
+1

PR #1487, opened 2016-06-29
Specify a default queue in several SomVal steps.

tmooney:
I am surprised so many SomaticValidation build steps didn't already have a queue set; I'd thought this was required awhile ago.  At any rate, now `WorkflowBuilder` will specify a queue for any steps that don't have one set.

### DIFF
--- a/lib/perl/Genome/Model/SomaticValidation/Command/IdentifyDnp.pm
+++ b/lib/perl/Genome/Model/SomaticValidation/Command/IdentifyDnp.pm
@@ -34,6 +34,11 @@ class Genome::Model::SomaticValidation::Command::IdentifyDnp {
             doc => "Somatic Validation build that the DNP result will belong to and which procides the reads and proportion parameters.",
         },
     ],
+    has_param => [
+        lsf_queue => {
+            value => Genome::Config::get('lsf_queue_build_worker_alt'),
+        },
+    ],
     doc => 'identify DNPs',
 };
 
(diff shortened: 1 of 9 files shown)


lib/perl/Genome/WorkflowBuilder/Command.pm
```diff
@@ -247,6 +247,11 @@ sub operation_type_attributes {
             $attributes{$expected_attributes{$name}} = $value;
         }
     }
+
+    unless (defined $attributes{'lsfQueue'}) {
+        $attributes{'lsfQueue'} = Genome::Config::get('lsf_queue_build_worker');
+    }
```
wschroeder: Should this be "lsf_queue"?

tmooney: I believe this key needs to match the value on [line 215](https://github.com/genome/genome/pull/1487/files/155a71a73af43a4d3124e87e54d0a84249609fad#diff-bdb6fd10fa2c2b74b604776c9d33f79dR215).

tmooney: At any rate, it looks like changing this default will force me to update the expected output of almost twenty tests (which highlights plenty of cases not setting defaults!), so I think I'll pull this change into its own PR.

sleongmgi:
+1

wschroeder:
+1 🐺

PR #1488, opened 2016-06-30
AlignmentResult: Check the reference and annotation builds when seeking our Merged result(s)

tmooney:
This reduces the number of false-positives when the same data has been aligned to many references, saving us from collecting all their individual alignments when we know they won't match.

(This is a bit of a selfish improvement, since the most drastic impact will be on test data that gets aligned to many, many references 😄 )


acoffman:
+1

jasonwalker80:
+1

PR #1489, opened 2016-06-30
SomaticFilterWorkflow test: Quote parameters to `diff`.

tmooney:
If files in the current directory matched the patterns, the shell would expand them.  If not, then the shell would pass them along as-is to `diff`. This test would thus pass or fail depending on the place from which it was launched.


dufeiyu:
+1

jasonwalker80:
+1

PR #1490, opened 2016-06-30
If we don't get a username back from `getpwuid`, try again!

tmooney:
It seems our systems are sometimes unreliable at returning a value here, causing some build failures.  In my testing, it will fail to return a value on the first attempt, but the second attempt will succeed.  Trying this five times is thus probably overkill(!), but hopefully is sufficient to make `Genome::Sys->username` reliable.

~~If things really go poorly, the user_id is often "good enough" as a unique per-user value... just one less friendly to the viewer.  However, it's insufficient to locate a `Genome::Sys::User` object, so anything depending on real user objects would still fail.~~

### DIFF
--- a/lib/perl/Genome/Sys.pm
+++ b/lib/perl/Genome/Sys.pm
@@ -1379,6 +1379,9 @@ sub username {
         my $username = getpwuid($user_id);
         return $username if $username;
 
+        #This is a hack!
+        #Sometimes our machines fail to get a result from getpwuid() on the first try.
+        #If we wait a second, it tends to work on subsequent attempts!
         $class->warning_message('Failed attempt %s to resolve username for user ID %s', $try, $user_id);
         sleep 1;
     }
--- a/lib/perl/Genome/Sys.pm
+++ b/lib/perl/Genome/Sys.pm
@@ -1372,9 +1372,11 @@ sub user_id {
 sub username {
     my $class = shift;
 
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Sys.pm
```diff
@@ -1371,8 +1371,18 @@ sub user_id {
...
+    for my $try (1..5) {
+        my $username = $ENV{'REMOTE_USER'} || getpwuid($user_id);
+        return $username if $username;
+
+        $class->warning_message('Failed attempt %s to resolve username for user ID %s', $try, $user_id);
+        sleep 1;
+    }
+
+    $class->warning_message('Giving up and returning user_id instead of name for %s' , $user_id);
+    return $user_id;
```
sleongmgi: Should this be "die/terminated" instead of return $user_id?

tmooney: That's a good idea!  I had leaned towards preserving the existing behavior of always returning something, but crashing is probably safer.

sleongmgi:
+1

brummett:
If retrieving it is unreliable, how about caching the value once it's been successfully resolved?

wschroeder:
+1 🌁

tmooney:
@brummett That's not a bad thought!  Whatever's going wrong here, it seems to work once it starts working, but since I don't really understand the underlying problem that sounds like a good way to mitigate potential issues.  I already merged this so we can get a passing build, but I'm still looking into this issue so I can add that in a follow up.

PR #1491, opened 2016-06-30
Cache the username in Genome::Sys->username.

tmooney:
Per @brummett's suggestion, save the username if we succeed to save us from further travails if `getpwuid` is acting up.


brummett:
+1

acoffman:
+1

PR #1493, opened 2016-07-05
Use Config to Select a Default LSF Queue if None Specified

tmooney:
This was split off from #1487 after it led to a large number of test failures.  This version includes updates to applicable tests to ignore whatever queue might be specified.  (Since it's decided by config, we shouldn't rely on it having any particular value!)

### DIFF
--- a/lib/perl/Genome/Ptero/ToJSON.t
+++ b/lib/perl/Genome/Ptero/ToJSON.t
@@ -15,6 +15,11 @@ use File::Basename qw(dirname basename);
 use File::Spec qw();
 use File::Copy qw(copy);
 
+use Sub::Override;
+use Genome::ConfigValidator::LSFQueue;
+
+#since we kill the environment in this test, the LSFQueue check cannot pass
+my $override = Sub::Override->new('Genome::ConfigValidator::LSFQueue::check', sub { 1 });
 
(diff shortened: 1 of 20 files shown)


lib/perl/Genome/Ptero/ToJSON.t
```diff
@@ -15,12 +15,18 @@ use File::Basename qw(dirname basename);
...
+#since we kill the environment in this test, the LSFQueue check cannot pass
+my $override = Sub::Override->new('Genome::ConfigValidator::LSFQueue::check', sub { 1 });
 
 for my $test_directory (glob test_data_directory('*')) {
     my $test_name = basename($test_directory);
     my $workflow = Genome::WorkflowBuilder::DAG->from_xml_filename(workflow_xml_file($test_name));
 
     my $json_filename = do {
+        my $path = $ENV{PATH};
         local %ENV; # otherwise the JSON contains the current user's environment.
```
wschroeder: Unused variable?

tmooney: Quite right!  That was leftover from my attempt to get the LSF queue validation to work before deciding to skip it.  I'll remove that.

sleongmgi:
+1

PR #1494, opened 2016-07-05
Use lower queue for BamQc "server" process.

tmooney:
This shouldn't use the same queue as its workers, lest they potentially deadlock.


dufeiyu:
+1

jasonwalker80:
+1

PR #1495, opened 2016-07-05
Reuse "rederivable" Reference Sequences

tmooney:
This will save us from creating extra copies of reference sequences and spending time and disk space indexing them for alignment when they match a reference we've already created.  (This means that `SomaticValidation` builds could possibly shortcut some of the validation realignment steps.)

This change would result in builds possibly being returned with model names and/or versions that do not match the supplied "create" parameters.  This is generally fine for the existing `rederivable` uses we have for reference sequences, but it's conceivable some future use might violate this assumption.


lib/perl/Genome/Model/Command/Define/ImportedReferenceSequence.pm
```diff
@@ -197,12 +197,26 @@ sub execute {
         $taxon = $taxons[0];
     }
 
+    my $fasta_md5;
+    if ($self->is_rederivable) {
+        $DB::single = 1;
```
jasonwalker80: Is this intended to be here?

tmooney: Nope!  I'll take that out :)

jasonwalker80:
+1, other than the break point

susannasiebert:
+1

PR #1497, opened 2016-07-06
Use config to decide the LSF queue (Ptero NullCommand)

tmooney:
Somehow this hardcoded queue in the Ptero tests snuck by us!


tmooney:
Looks like this depends on #1493 to pass tests (since that PR updates the tests to ignore the specific queue that has been specified).

wschroeder:
+1 after tests pass

acoffman:
+1 🐻

PR #1499, opened 2016-07-07
Add error to instantiation failure message.

tmooney:
This error message came up today, and it would've been helpful to see why without having to reconstruct the situation!

While we're here, update the `die $self->error_message()` to use `$self->fatal_message()` instead.


acoffman:
+1

dufeiyu:
+1

PR #1501, opened 2016-07-12
Only create one annotate_snvs_vcf_op per type.

tmooney:
In #1472 I made a suggestion that made this not work quite right, because it might make duplicate operations.  This should fix that issue!

(This is another PR that benefits from the `?w=1` option.)


acoffman:
+1

susannasiebert:
+1

PR #1502, opened 2016-07-13
ClinSeq MicroarrayCnv takes a Build

tmooney:
Now that this command is part of the ClinSeq build process (#1392), we can't assume that the model will have a last_succeeded_build to work with.  Instead, let's use our current build to do the work.

### DIFF
--- a/lib/perl/Genome/Model/ClinSeq.pm
+++ b/lib/perl/Genome/Model/ClinSeq.pm
@@ -1375,9 +1375,9 @@ sub microarray_cnv_op {
         destination_property => 'outdir',
     );
     $workflow->connect_input(
-        input_property       => 'model',
+        input_property       => 'build',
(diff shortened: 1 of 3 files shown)


jasonwalker80:
+1

susannasiebert:
When I updated the other ClinSeq tests I made them get the data from the test db. Is that something we an do here as well?

tmooney:
@susannasiebert Yep!  The "hack" commit is to get around the fact that the test db snapshot's build is not one that has the required input set.  (That is to say, the test worked fine on the production database.)

It'd be much better to use the test factory and not have to worry about a primed database at all!  I'll see if I can change it...

tmooney:
Updated to use the `Genome::Test::Factory`!

susannasiebert:
+1

PR #1504, opened 2016-07-14
Make trio wrapper merge work if one of input reports is empty

dufeiyu:
Currently trio process will crash at merge report step if one of input bed file is empty (refer to https://jira.gsc.wustl.edu/browse/CI-167). This PR fixes this.


tmooney:
+1.  Is there a test case that covers this situation?

dufeiyu:
Good point. I just find out that the actual unit test for Genome/VariantReporting/Framework/Component/Report/MergedReport.pm is in Genome/VariantReporting/Framework/MergeReports.t.  I will add test case there.

tmooney:
+1, again! 😄

PR #1506, opened 2016-07-18
Update CLE plan files with new filters

dufeiyu:
Add n-callers and vaf-cutoff filters to cle snvs plan files and annotation-category filter to cle indels plan file.


tmooney:
Will this require updating the expected case in the regression test?

dufeiyu:
Yeah, both jenkins cle regression test and trio_main process test are expected to have diff after this merges. Two new PR will be created for updating both.

tmooney:
Sounds like a plan! +1.

PR #1507, opened 2016-07-18
RefSeq Converter: Use names of references instead of object hashes in error message.

tmooney:
I recently encountered a build with this error message in it, but 
```
Failed to convert from bed (...) with reference (Genome::Model::Build::ImportedReferenceSequence=HASH(0x70802e0)) to bed (...) with reference (Genome::Model::Build::ImportedReferenceSequence=HASH(0x70260c0)
```
is not the most informative!


jasonwalker80:
+1

susannasiebert:
+1

PR #1508, opened 2016-07-20
Stop warning about undefined environment variables.

tmooney:
This doesn't affect the operation of the workflow, and users don't really care about the limitations of the environment compared to Perl hashes.


jasonwalker80:
+1

susannasiebert:
+1

PR #1509, opened 2016-07-20
Defer to new Ptero feature to decide the log paths.

tmooney:
This uses the new features in genome/ptero-lsf#124 and genome/ptero-perl-sdk#117 to ensure a unique filename is chosen for each job, even during a parallel-by.  As a bonus that filename's UUID will now be the Job ID making it easier to match to the job to which it belongs.

(This should not be merged until the aformentioned PRs are deployed.)


wschroeder:
+1

jasonwalker80:
+1

PR #1510, opened 2016-07-21
Don't die just because we can't revert the logging.

tmooney:
Letting a few messages leak into the task's log files is far better than throwing out all of our work when there's an issue reopening the original log file!

(#1509 is a long-term solution to the present problem causing the errors reopening the log file, but we probably shouldn't fail in any event.)


jeldred:
+1 bitmoji

jasonwalker80:
+1

PR #1511, opened 2016-07-25
Require a newer SSL implementation to perform the web queries.

tmooney:
The GDC portal now requires newer SSL to perform the handshake, so disable the TCGA support for "older" systems.  (Running with an outdated OpenSSL is scary, and one probably shouldn't do it, anyway!)


ebelter:
+1 👘

jasonwalker80:
+1

PR #1512, opened 2016-07-26
bedtools merge requires the file to be sorted.

tmooney:
Although imported feature-lists often are sorted, we don't guarantee that, so better sort here.


jasonwalker80:
+1

dufeiyu:
+1  just curious, is this change tested in FeatureList.t ?

tmooney:
The code is exercised in a couple places including `Genome/FeatureList/Command/DumpMergedList.t` and `Genome/InstrumentData/AlignmentResult/Merged/CoverageStats.t`.  In both of those cases, the input list is already sorted (as we were expecting), so it doesn't test that we get a different result.

PR #1514, opened 2016-07-27
CQID: Give all the "new" things a try first.

tmooney:
If someone manually resets the status to `new` (for example, via `genome analysis-project reprocess`) they probably expect us to try their instrument data at least once in short order!  This is also a good clue that they may have corrected whatever issues were causing the previous failures.

If it still fails, its `fail_count` will be incremented as usual and it'll fall back to its previous spot in line.


ebelter:
+2

PR #1516, opened 2016-07-29
Update blessed process for Jenkins genome-process-test

dufeiyu:
All diff is caused by the newly added n-callers and vaf-cutoff filters in snv plan files and annotation-category filter in indel plan file. Need merge this asap to make model/process test pass.


susannasiebert:
+1

PR #1518, opened 2016-08-04
Convert chrM to MT.

tmooney:
The sequences in our references don't seem to really match, but I'm told that these are interchangeable.


ebelter:
+1

jasonwalker80:
+1.  This worried me a bit at first.  However, convert_chrXX_to_GL is pretty specific to human and I doubt many new converters use this algorithm anyway.  It should be safe for existing reference converters that use this algorithm.

PR #1519, opened 2016-08-04
Update blessed process for CLE regression test

dufeiyu:
This is an update to the CLE regression test. Medical Director of the CLE, Dave Spencer, approved this.


tmooney:
+1. ID does appear to be of a process of class `Genome::VariantReporting::Process::Trio`!

PR #1522, opened 2016-08-08
Remove old scripts in "CaptureList" directory.

tmooney:
These all look pretty old and have references to users and paths that no longer exist.


ebelter:
+1

jasonwalker80:
+1

PR #1523, opened 2016-08-08
Allow Running Builds with Missing Instrument Data BAMs

### LINKED ISSUE #1521: Allow builds to start with missing instrument data files.
Instead of failing immediately with an Unstartable status, allow builds to start with missing instrument data files.  Often the data is not realigned, but reprocessed through a new pipeline type or with parameters that do not change the alignment strategy.  If the data IS realigned or the raw instrument data IS required by a downstream step, fail at that point of the pipeline instead.

tmooney:
Since they're being very aggressively archived, we're frequently running into the situation where we want to run builds for instrument data that has already vanished.  This allows us to potentially re-use the existing alignments (if they're available) to avoid referring back to the original instrument data.

It may lead to some more inscrutable errors from tools if in fact we do need the instrument data BAMs, but hopefully a real tiered disk architecture will soon end the need for our concern about explicit archiving!

Closes #1521.

### DIFF
--- a/lib/perl/Genome/InstrumentData/AlignmentResult/Bwa.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult/Bwa.pm
@@ -92,7 +92,8 @@ sub tmp_megabytes_estimated {
 
         my $bam_bytes = -s $bam_path;
         unless ($bam_bytes) {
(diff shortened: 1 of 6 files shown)


ebelter:
+1

jasonwalker80:
+1

PR #1524, opened 2016-08-10
WIP : Predict Subject Mappings 

jasonwalker80:
Initial version of a new command to predict subject mappings using the common names listed in ClinSeq updated analysis.

### DIFF
--- a/lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Predict/SomaticValidation.pm
+++ b/lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Predict/SomaticValidation.pm
@@ -102,12 +102,20 @@ sub execute {
             push(@tumor_samples,  $individual_dna_sample) if ($dna_sample_common_name =~ /$tumor_sample_common_names/i);
         }
 
-        if (scalar(@normal_samples) > 1) {
+        my $normal_sample_count = scalar(@normal_samples);
+        my $tumor_sample_count = scalar(@tumor_samples);
+        my $labeled_sample_count = $normal_sample_count + $tumor_sample_count;
+
+        if ($normal_sample_count > 1) {
             $self->fatal_message('More than one normal DNA sample was specified for this individual: '. $individual->__display_name__);
         }
-        if (scalar(@tumor_samples) > 1) {
+        if ($tumor_sample_count > 1) {
             $self->fatal_message('More than one tumor DNA sample was specified for this individual: '. $individual->__display_name__);
         }
+        if ($labeled_sample_count ne $individual_dna_sample_count) {
+            $self->fatal_message('Failed to find a tumor/normal label for all DNA samples of individual: '. $individual->__display_name__);
+        }
+
         my %data;
         if (@normal_samples && @tumor_samples) {
             %data = (
--- a/lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Predict/SomaticValidation.pm
+++ b/lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Predict/SomaticValidation.pm
@@ -1,0 +1,163 @@
+package Genome::Config::AnalysisProject::SubjectMapping::Command::Predict::SomaticValidation;
+
+use strict;
+use warnings;
+
+use Genome;
+
+class Genome::Config::AnalysisProject::SubjectMapping::Command::Predict::SomaticValidation {
+    is => 'Genome::Config::AnalysisProject::Command::Base',
+    has_input => [
+        file_path => {
+            is => 'Text',
+            shell_args_position => 2,
+            doc => 'output path for a newline-delimited, tab-separated list of samples and required whitespace'
+        }
+    ],
+    has_optional => [
+        normal_sample_common_names => {
+            is      => 'Text',
+            default => Genome::Model::ClinSeq::Command::UpdateAnalysis->__meta__->property('normal_sample_common_names')->default_value,
+            doc     => Genome::Model::ClinSeq::Command::UpdateAnalysis->__meta__->property('normal_sample_common_names')->doc,
+        },
+        tumor_sample_common_names => {
+            is      => 'Text',
+            default => Genome::Model::ClinSeq::Command::UpdateAnalysis->__meta__->property('tumor_sample_common_names')->default_value,
+            doc => Genome::Model::ClinSeq::Command::UpdateAnalysis->__meta__->property('tumor_sample_common_names')->doc,
+        },
+        sample_identifier => {
+            is => 'Text',
+            doc => 'The method used to resolve the identifier of the sample.',
+            default => 'name',
+            valid_values => ['name','id'],
+        },
+    ],
+};
+
+sub help_brief {
+    return 'Predict sample pairings in bulk for an AnalysisProject';
+}
+
+sub help_synopsis {
+    return "genome config analysis-project subject-mapping predict somatic-validation <analysis_project_id> <TSV file path>";
+}
+
+sub help_detail {
+    return <<"EOS"
+This command allows you to predict subject mapping information for an AnalysisProject.
+
+The output file is a 6 column, tab separated format with the following columns:
+
+tumor_subject normal_subject snv_variant_list_id indel_variant_list_id sv_variant_list_id tag
+
+EOS
+}
+
+sub valid_statuses {
+    return ("Pending", "Hold", "In Progress");
+}
+
+my @headers = ('tumor_sample', 'normal_sample','snv_variant_list_id', 'indel_variant_list_id', 'sv_variant_list_id','tag');
+
+sub execute {
+    my $self = shift;
+
+    my $analysis_project = $self->analysis_project;
+
+    my $normal_sample_common_names = $self->normal_sample_common_names;
+    my $tumor_sample_common_names  = $self->tumor_sample_common_names;
+
+    my $id_method = $self->sample_identifier;
+
+    my @all_dna_samples = grep {$_->sample_type =~ /dna/i} $analysis_project->samples;
+    my %dna_samples = map { $_->id => $_ } @all_dna_samples;
+    my @dna_sample_ids = sort keys %dna_samples;
+    $self->status_message('Found '. scalar(@dna_sample_ids) .' DNA samples');
+
+    my %dna_samples_by_individual_id;
+    for my $dna_sample_id (@dna_sample_ids) {
+        my $dna_sample = $dna_samples{$dna_sample_id};
+        push @{$dna_samples_by_individual_id{$dna_sample->individual->id}}, $dna_sample;
+    }
+    my @individual_ids = sort keys %dna_samples_by_individual_id;
+    $self->status_message('Found '. scalar(@individual_ids) .' indivudals');
+
+    my @subject_mappings;
+    for my $individual_id (@individual_ids) {
+        my $individual = Genome::Individual->get($individual_id);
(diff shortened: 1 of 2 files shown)


lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Predict/SomaticValidation.pm
```diff
@@ -0,0 +1,163 @@
...
+            is => 'Text',
+            doc => 'The method used to resolve the identifier of the sample.',
+            default => 'name',
+            valid_values => ['name','id'],
+        },
+    ],
+};
+
+sub help_brief {
+    return 'Predict sample pairings in bulk for an AnalysisProject';
```
tmooney: It might be nice if this help contains a mention of `SomaticValidation` (to distinguish it from the `doc` for the command one level up the tree).

lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Predict/SomaticValidation.pm
```diff
@@ -0,0 +1,163 @@
...
+    my $id_method = $self->sample_identifier;
+
+    my @all_dna_samples = grep {$_->sample_type =~ /dna/i} $analysis_project->samples;
+    my %dna_samples = map { $_->id => $_ } @all_dna_samples;
+    my @dna_sample_ids = sort keys %dna_samples;
+    $self->status_message('Found '. scalar(@dna_sample_ids) .' DNA samples');
+
+    my %dna_samples_by_individual_id;
+    for my $dna_sample_id (@dna_sample_ids) {
+        my $dna_sample = $dna_samples{$dna_sample_id};
```
tmooney: It seems like `%dna_samples` and `@dna_sample_ids` aren't necessary.  This loop could be:
````perl
for my $dna_sample (@all_dna_samples) {
````

File: lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Predict/SomaticValidation.pm [hunk 3]
```diff
@@ -0,0 +1,163 @@
...
+            $self->fatal_message('This command currently does not support more than 2 DNA samples per individual.');
+        }
+
+        my @normal_samples;
+        my @tumor_samples;
+
+        for my $individual_dna_sample (@individual_dna_samples) {
+            my $dna_sample_common_name = $individual_dna_sample->common_name || "NULL";
+            push(@normal_samples, $individual_dna_sample) if ($dna_sample_common_name =~ /$normal_sample_common_names/i);
+            push(@tumor_samples,  $individual_dna_sample) if ($dna_sample_common_name =~ /$tumor_sample_common_names/i);
```
tmooney: These regexes are a little sketchy (or at least the default values are).  They probably work in practice...but hopefully we never have a `methodically collected normal` :smile:.

lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Predict/SomaticValidation.pm
jasonwalker80: The issue is that calling `$analysis_project->samples` returns one sample object per instrument data bridge resulting in many duplicate objects.  A better method of deduplication might be possible, suggestions welcome.

tmooney: One option might be:
```perl
use List::MoreUtils qw(uniq);
for my $dna_sample (uniq @all_dna_samples) {
```

lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Predict/SomaticValidation.pm
jasonwalker80: I agree there are better solutions, but we don't necessarily have control over the vocabulary or nomenclature provided to us.  This is the number one reason for only writing a command that predicts subject mappings.  A user should still confirm that the subject mappings are indeed correct.

lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Predict/SomaticValidation.pm
jasonwalker80: I tried this originally on $analysis_project->samples.  For whatever reason it didn't work or I may have had the syntax wrong.  It works correctly now and the suggested changes have been made.

lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Predict/SomaticValidation.pm
```diff
@@ -0,0 +1,162 @@
...
+            %data = (
+                tumor_sample => $tumor_samples[0]->$id_method,
+                normal_sample => '',
+                snv_variant_list_id => '',
+                indel_variant_list_id => '',
+                sv_variant_list_id => '',
+                tag => 'tumor-only',
+            );
+        }
+        # TODO: See if subject mapping already exists before adding to the list
```
jasonwalker80: Is this the right place to worry about creating duplicate subject mappings?  Are there database constraints that prevent us from making duplicate subject mappings?  Should there be or should the import check to see if the subject mappings exist before creating them?  @acoffman I know we've had discussions about this in the past.  If I recall duplicates do not necessarily hurt anything or cause errors, but they do exist.

lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Predict/SomaticValidation.pm
```diff
@@ -0,0 +1,162 @@
...
+    my @subject_mappings;
+    for my $individual_id (@individual_ids) {
+        my $individual = Genome::Individual->get($individual_id);
+        my @individual_dna_samples = @{$dna_samples_by_individual_id{$individual_id}};
+        my $individual_dna_sample_count = scalar(@individual_dna_samples);
+        $self->status_message('Found '. $individual_dna_sample_count .' samples for individual '. $individual->__display_name__);
+
+        if ( $individual_dna_sample_count > 2) {
+            $self->fatal_message('This command currently does not support more than 2 DNA samples per individual.');
+        }
```
tmooney: Should we warn about this individual but continue to create the rest of the predictions?

lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Predict/SomaticValidation.pm
```diff
@@ -0,0 +1,162 @@
...
+    my @all_dna_samples = grep {$_->sample_type =~ /dna/i} $analysis_project->samples;
+    my @dna_samples = List::MoreUtils::uniq(@all_dna_samples);
+    $self->status_message('Found '. scalar(@dna_samples) .' DNA samples');
+
+    my %dna_samples_by_individual_id;
+    for my $dna_sample (@dna_samples) {
+        push @{$dna_samples_by_individual_id{$dna_sample->individual->id}}, $dna_sample;
+    }
+    my @individual_ids = sort keys %dna_samples_by_individual_id;
+    $self->status_message('Found '. scalar(@individual_ids) .' indivudals');
```
tmooney: `indivudals` should be `individuals`.

jasonwalker80: fixed

lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Predict/SomaticValidation.pm
jasonwalker80: I agree, the warning message includes information about the individual as well.

lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Predict/SomaticValidation.pm
```diff
@@ -0,0 +1,163 @@
...
+        }
+        my %data;
+        if (@normal_samples && @tumor_samples) {
+            %data = (
+                tumor_sample => $tumor_samples[0]->$id_method,
+                normal_sample => $normal_samples[0]->$id_method,
+                snv_variant_list_id => '',
+                indel_variant_list_id => '',
+                sv_variant_list_id => '',
+                tag => 'somatic',
```
tmooney: Just a note: of the three tags used in this code, only `germline` already exists in our database.

lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Predict/SomaticValidation.pm
```diff
@@ -0,0 +1,163 @@
...
+        } elsif (@tumor_samples) {
+            %data = (
+                tumor_sample => $tumor_samples[0]->$id_method,
+                normal_sample => '',
+                snv_variant_list_id => '',
+                indel_variant_list_id => '',
+                sv_variant_list_id => '',
+                tag => 'tumor-only',
+            );
+        }
```
tmooney: Should we warn if we don't get either `@tumor_samples` or `@normal_samples`?

lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Predict/SomaticValidation.pm
tmooney: Should we check to see that every sample in the individual lands in one of the two buckets?  It seems likely to miss if it falls to the `NULL`, which might lead us to produce a prediction with only one of the samples.

lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Predict/SomaticValidation.pm
tmooney: Subject Mappings store their Subjects and their Tags in hang-off tables, so there's no constraint on duplicates.  I believe the downside of duplicates is that it makes CQID take longer as it tries to get or create the same model once per mapping.

lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Predict/SomaticValidation.pm
jasonwalker80: Correct, I need to both add the tags and setup configuration that uses them.  This does raise a need for a previously discussed feature.  When a configuration template is defined or the `copy` command is used, can we retain tags?

lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Predict/SomaticValidation.pm
jasonwalker80: I believe the latest commit will throw an error when the count of samples that match either tumor/normal does not match the total count of samples for an individual.

lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Predict/SomaticValidation.pm
jasonwalker80: The latest commit `d880b06` should address this.

tmooney: If I have two samples, one with common name `methodically collected normal` and one without a common name (thus `NULL`), this could lead to a subject mapping where we have one sample in both the tumor and normal position and the other sample silently omitted.

PR #1527, opened 2016-08-10
Make AlignmentSummaryV2 the default in Coverage Reports

tmooney:
The individual result will now only display the V2 metrics, and the set page will now have the toggle default to the `new` setting.


jasonwalker80:
+1

dufeiyu:
+1

PR #1528, opened 2016-08-11
Institute a delete command and lister for subject mappings.

tmooney:
The lister is less friendly but more powerful than the view command.


acoffman:
+1

jasonwalker80:
+1

Issue #1525, opened 2016-08-10
Remove references to Workflow


jasonwalker80:
finish taking out references to `Workflow` in `Genome` (and thus end support for the `genome config set-env workflow_builder_backend workflow` case)

tmooney:
After #1529, #1530, and #1531 are all resolved I have a branch ready to remove the remaining references to `Workflow` from the tree.  After that, we can move on to #1526.

PR #1529, opened 2016-08-11
Deprecate ProteinAnnotation pipeline.

tmooney:
In order to facilitate #1525, this pipeline either needs to be upgraded or removed.  Until removal is approved, this at least stops it from being run.

Alternatives include: continuing to support it and removing it completely!


jasonwalker80:
+2

PR #1533, opened 2016-08-16
Predict subject mappings

jasonwalker80:
This replaces the WIP PR #1524 

The latest version of this command uses Set::Scalar to compare both the tumor/normal sets as well as comparing against existing subject mappings to avoid duplicates.  Also, the command now allows for multiple tumor samples.

This doesn't address the pattern matching issue, but users should be responsible for validating the predicted subject mappings before importing them.


lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Predict/SomaticValidation.pm
```diff
@@ -0,0 +1,247 @@
...
+        if ($normal_sample_set && $tumor_sample_set) {
+            my ($normal_sample) = $normal_sample_set->members();
+            for my $tumor_sample ($tumor_sample_set->members) {
+                my %data = (
+                    tumor_sample => $tumor_sample->$id_method,
+                    normal_sample => $normal_sample->$id_method,
+                    snv_variant_list_id => '',
+                    indel_variant_list_id => '',
+                    sv_variant_list_id => '',
+                    tag => 'somatic',
```
tmooney: These hashes could perhaps be constructed by some sort of helper, like:
```perl
sub _data_hash {
  my ($tumor, $normal, $tag) = @_;
  my $id_method = $self->sample_identifier;

  my %data = (
    tumor_sample  => $tumor->$id_method;
    normal_sample => $normal? $normal->$id_method : '',
    snv_variant_list_id => '',
    indel_variant_list_id => '',
    sv_variant_list_id => '',
    tag => $tag,
  );

  return \%data;
}
```

Then we have less duplication in all these cases and there's a single spot to add support for variant lists later if we so choose.

lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Predict/SomaticValidation.pm
```diff
@@ -0,0 +1,247 @@
...
+
+    return $self->_existing_subject_mapping_set;
+}
+
+sub new_subject_mapping_set {
+    my $self = shift;
+    if ($self->_new_subject_mapping_set) { return $self->_new_subject_mapping_set; }
+    my $new_subject_mapping_set = Set::Scalar->new();
+    $self->_new_subject_mapping_set($new_subject_mapping_set);
+    return $self->_new_subject_mapping_set;
```
tmooney: This could use a property also named `new_subject_mapping_set` and then call `__new_subject_mapping_set` inside of here to get at the built-in accessor.

lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Predict/SomaticValidation.pm
```diff
@@ -0,0 +1,247 @@
...
+        return;
+    }
+    $self->new_subject_mapping_set->insert($data);
+    return 1;
+}
+
+sub resolve_samples_by_individual_id {
+    my $self = shift;
+
+    my @all_dna_samples = grep {$_->sample_type =~ /dna/i} $self->analysis_project->samples;
```
tmooney: This matches `cdna`, which is included in the results for `$sample->is_rna`.  Is that intended?

lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Predict/SomaticValidation.pm
jasonwalker80: I went a little further and added this to add_subject_mapping.  It will create the data hash and add it to the new set.

lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Predict/SomaticValidation.pm
jasonwalker80: I'm not sure I follow this suggestion.  I think this is the same pattern used for `existing_subject_mapping_set` unless I'm missing something.

lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Predict/SomaticValidation.pm
jasonwalker80: That is not intended behavior and I've added a check for not RNA.

lib/perl/Genome/Config/AnalysisProject/SubjectMapping/Command/Predict/SomaticValidation.pm
tmooney: UR has a built-in way for there to be a `sub` that overrides the default behavior of a property accessor.

One example of it being used is [here](https://github.com/genome/genome/blob/118f82cc36adca2f31caef605591c609a4bd1b0a/lib/perl/Genome/VariantReporting/Command/CreateReport.pm#L90).  In that example `plan` is both the name of the property and the `sub`, and within the subroutine `$self->__plan` is used to access the property value directly.

My suggestion is that such an approach could also work here.

jasonwalker80: Thanks!  The example clarifies things for me.  I'll take a shot at the suggestion for both 'new' and 'existing'.

jasonwalker80: Suggestion implemented.

tmooney:
It'd be nice to have a test for this new command, but otherwise +1.

jasonwalker80:
@tmooney a test has been added.

jasonwalker80:
CI, test this

jasonwalker80:
CI, test this

jasonwalker80:
Finally, @tmooney now the new test is passing after successfully updating the test db.

tmooney:
+1 :shipit:

acoffman:
+1 👍

PR #1534, opened 2016-08-17
Unarchive command for AnPs using a Site::TGI Extension for LIMS data.

jasonwalker80:


lib/perl/Genome/Site/TGI/Extension/UnarchiveInstrumentData.pm
```diff
@@ -0,0 +1,196 @@
...
+sub unarchive_lims_data {
+    my $self = shift;
+    my $instrument_data = shift;
+
+    my $lims_unarchive_script_path = $self->resolve_lims_unarchive_script_path();
+
+    my $dv = $self->mount_path;
+    if ($instrument_data->class ne 'Genome::InstrumentData::Solexa') { return 1; }
+
+    my $bam_path = $instrument_data->bam_path;
```
jasonwalker80: In what situation would solexa instrument_data have an undefined `bam_path`.  I have already encountered one with ID 2819728424.

lib/perl/Genome/Site/TGI/Extension/UnarchiveInstrumentData.pm
```diff
@@ -0,0 +1,196 @@
...
+    if (-e $bam_path) { return 1; }
+
+    $self->status_message('Working on missing BAM %s for instrument data %s', $bam_path, $instrument_data->id);
+
+    my ($bam_filename, $bam_dirname) = File::Basename::fileparse($bam_path);
+    $bam_dirname =~ s/\/$//;
```
jasonwalker80: The $bam_dirname is from the GMS instrument data.  If the GMS bam_path was ever updated manually to something other than `csf_#####` then the later glob is likely not going to work.

lib/perl/Genome/Site/TGI/Extension/UnarchiveInstrumentData.pm
```diff
@@ -0,0 +1,196 @@
...
+    my @bam_dirnames = File::Spec->splitdir($bam_dirname);
+
+    my ($unarchived_bam_path) = glob(File::Spec->join($dv->mount_path,'*','csf_*',$bam_filename));
```
jasonwalker80: See above comment on line36.  Perhaps the `csf_*` part of the glob is not necessary?

lib/perl/Genome/Site/TGI/Extension/UnarchiveInstrumentData.pm
tmooney: When it pre-dates the switch from FASTQs to BAMs.  Try `archive_path` instead.

lib/perl/Genome/Site/TGI/Extension/UnarchiveInstrumentData.pm
tmooney: The `csf_*` here is part of the LIMS archive path, so it should always be there... unless it gets changed in the LIMS system. 😄

I'm not super-keen on possibly searching every directory on the volume looking for our file, though, so something to improve this glob would be nice!

lib/perl/Genome/Site/TGI/Extension/UnarchiveInstrumentData.pm
tmooney: Should we avoid using `$bam_dirname` for anything? We could replace the `$bam_dirnames[-1]` down below with an `$unarchived_bam_dirnames[-1]`, which should be the `csf_*` value we match on from the archive.

jasonwalker80: Yes, that should work.

lib/perl/Genome/Site/TGI/Extension/UnarchiveInstrumentData.pm
jasonwalker80: This is no longer an issue if we use the LIMS `$unarchive_bam_dirname` to resolve the directory.

lib/perl/Genome/Config/AnalysisProject/Command/UnarchiveInstrumentData.pm
```diff
@@ -0,0 +1,51 @@
...
+    my $anp = $self->analysis_project;
+
+
+    for my $instrument_data ($anp->instrument_data) {
+        my @allocations = Genome::Disk::Allocation->get(owner_id => $instrument_data->id, owner_class_name => $instrument_data->class);
+
+        if (@allocations) {
+            for my $allocation (@allocations) {
+                if ($allocation->is_archived) {
+                    say 'genome disk allocation unarchive --analysis-project ' . $anp->id . ' ' . $allocation->id;
```
jasonwalker80: Do we need to unarchive the GMS allocations?  I'm finding a bunch of MAQ alignments to build36 for this data.  I do NOT want those.

tmooney: This should only unarchive allocations whose owner is the instrument data itself.  Are the MAQ alignments in that category? If so, we'll definitely have to figure out something different for this.  That's a little scary!

jasonwalker80: Yes, the owner for these allocations are the instrument data themselves.  Indeed scary.

tmooney: I guess on the plus side, nobody needs to run that unarchive command just because we print it out!  But I wonder if there's a good way to distinguish between ones we care about and ones we don't.

lib/perl/Genome/Disk/Volume.pm
```diff
@@ -374,4 +374,18 @@ sub is_near_soft_limit {
...
 }
 
+sub _resolve_param_value_from_text_by_name_or_id {
+    my $class = shift;
+    my $param_arg = shift;
+
+    #First try default behaviour of looking up by name or id
+    my @results = Command::V2->_resolve_param_value_from_text_by_name_or_id($class, $param_arg);
+
+    #If that didn't work, and the argument is a filename, see if we have a feature list matching the provided file.
```
jasonwalker80: The comment seems to be left over from a copy/paste out of `Genome::FeatureList`.

tmooney: Indeed; I've revised it.

jasonwalker80:
I made a change to account for one more level of sub-directory.  This worked, but now I have an issue with permissions:
https://jira.gsc.wustl.edu/browse/LS-7109

lib/perl/Genome/Site/TGI/Extension/UnarchiveInstrumentData.pm
```diff
@@ -0,0 +1,196 @@
...
+    my $self = shift;
+    my $instrument_data = shift;
+
+    my $lims_unarchive_script_path = $self->resolve_lims_unarchive_script_path();
+
+    my $dv = $self->volume;
+    if ($instrument_data->class ne 'Genome::InstrumentData::Solexa') { return 1; }
+
+    my $bam_path = $instrument_data->bam_path;
+    if (-e $bam_path) { return 1; }
```
jasonwalker80: @tmooney I thought I made a commit to address data that has an archive_path instead of a bam_path?  Are we missing a commit, or did this never get pushed to your branch before the PR?

tmooney: I've grabbed the changes from your checkout and committed them to my branch.

lib/perl/Genome/Site/TGI/Extension/UnarchiveInstrumentData.pm
```diff
@@ -0,0 +1,209 @@
...
+                target_directory => $allocation_bam_dirname,
+            )
+          ) {
+            $self->fatal_message('Failed to rsync from %s to %s', $unarchived_bam_dirname, $allocation_bam_dirname);
+        }
+        unless ($ENV{UR_DBI_NO_COMMIT}) {
+            Genome::Sys::CommitAction->create(
+                on_commit => sub {
+                    unless(Genome::Sys->remove_directory_tree($unarchived_bam_dirname)) {
+                        $self->fatal_message('Failed to remove directory: %s', $unarchived_bam_dirname);
```
jasonwalker80: Should this be a warning_message() instead of fatal_message().  You don't want the directory to be removed until commit, but I don't think we want the commit to fail because the directory isn't removed.  The rsync worked, so the copy in the allocation should be fine.  One reason this could happen is due to permissions on the unarchived files/directories.  Currently this requires root access to modify.  If the file was removed, but not the directory... and the commit does not save, then the unarchive has to be rerun and permissions modified PROactively rather than retroactively.  The downside is that directories and/or files could be left around unknowingly.

jasonwalker80: ...or is there an after_commit commit action rather than on_commit?

tmooney: An observer like [this one](https://github.com/genome/genome/blob/d3a38ac13cb372bf19b00784a897624aff8dd26d/lib/perl/Genome/Role/ObjectWithAllocations.pm#L45) should trigger after commit.  (There's also a "precommit" option.)  The `Genome::Sys::CommitAction` object uses a different data-source based approach, which happens during commit.

tmooney: Pushed a change to adopt your suggestion to make this a `warning_message`.  Seems better than complicated observers that still end up failing! 😄

jasonwalker80:
+1

lib/perl/Genome/Config/AnalysisProject/Command/UnarchiveInstrumentData.pm
```diff
@@ -0,0 +1,51 @@
...
+    doc => 'a helper that produces commands that can be run to unarchive data',
+};
+
+sub execute {
+    my $self = shift;
+    my $anp = $self->analysis_project;
+
+
+    for my $instrument_data ($anp->instrument_data) {
+        my @allocations = Genome::Disk::Allocation->get(owner_id => $instrument_data->id, owner_class_name => $instrument_data->class);
```
ebelter: Would owner id be safer alone without including the class?

lib/perl/Genome/Config/AnalysisProject/Command/UnarchiveInstrumentData.pm
```diff
@@ -0,0 +1,51 @@
...
+
```
ebelter: Couldn't we use  Genome::Role::ObjectWithAllocations property?
```
$intrument_data->disk_allocations
```

lib/perl/Genome/Site/TGI/Extension/UnarchiveInstrumentData.pm
```diff
@@ -0,0 +1,196 @@
...
+
+__PACKAGE__->__meta__->property('mount_path')->is_optional(0);
+
+sub unarchive_lims_data {
+    my $self = shift;
+    my $instrument_data = shift;
+
+    my $lims_unarchive_script_path = $self->resolve_lims_unarchive_script_path();
+
+    my $dv = $self->mount_path;
```
ebelter: This could be declared where used below.

lib/perl/Genome/Config/AnalysisProject/Command/UnarchiveInstrumentData.pm
tmooney: Yes we could--updated!

lib/perl/Genome/Site/TGI/Extension/UnarchiveInstrumentData.pm
tmooney: Indeed--moved!

lib/perl/Genome/Site/TGI/Extension/UnarchiveInstrumentData.pm
tmooney: I've made the change to use `@unarchived_bam_dirnames` instead.

lib/perl/Genome/Site/TGI/Extension/generate_lims_disk_unarchive.pl
```diff
@@ -0,0 +1,39 @@
...
+
+use strict;
+use warnings;
+
+use Data::Dumper;
+use Carp qw/confess/;
+
+use GSCApp;
+
+my $ii = GSC::IndexIllumina->get(analysis_id => $ARGV[0]);
```
ebelter: Should check if $ARGV[0..1] are defined. Since you've gone into GSCApp land, you could use App::GetOpt 😄

lib/perl/Genome/Site/TGI/Extension/generate_lims_disk_unarchive.pl
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env lims-perl
+
+use strict;
+use warnings;
+
+use Data::Dumper;
+use Carp qw/confess/;
+
+use GSCApp;
+
```
ebelter: Need App->init ?

lib/perl/Genome/Site/TGI/Extension/UnarchiveInstrumentData.pm
```diff
@@ -0,0 +1,196 @@
...
+        }
+        ($allocation) = Genome::Disk::Allocation->get(allocation_path => $params{allocation_path});
+        unless ($allocation) {
+            $self->fatal_message('Could not find allocation for instrument data: %s', $instrument_data->id);
+        }
+    }
+    return $allocation;
+}
+
+sub validate_bam_is_complete {
```
ebelter: This method is getting a little long and hard to follow with the nested if/else's.

lib/perl/Genome/Site/TGI/Extension/UnarchiveInstrumentData.pm
```diff
@@ -0,0 +1,196 @@
...
+            my $actual_flagstat = Genome::Model::Tools::Sam::Flagstat->parse_file_into_hashref($bam_path .'.flagstat');
+            unless ($actual_flagstat) {
+                $self->fatal_message('Fail to parse flagstat data file: %s.flagstat', $bam_path);
+            }
+            if ($calculated_flagstat->{total_reads} ne $actual_flagstat->{total_reads}) {
+                $self->fatal_message('BAM file total_reads does not match expected total reads: %s', $bam_path);
+            }
+        } else {
+            if ($calculated_flagstat->{total_reads} ne $instrument_data->read_count) {
+                $self->fatal_message('BAM file total_reads doe snot match expected instrument data read count: %s', $bam_path);
```
ebelter: 🐌 doe snot? 😄

lib/perl/Genome/Site/TGI/Extension/UnarchiveInstrumentData.pm
```diff
@@ -0,0 +1,196 @@
...
+    return $allocation;
+}
+
+sub validate_bam_is_complete {
+    my $self = shift;
+    my $bam_path = shift;
+    my $instrument_data = shift;
+
+    if (-e $bam_path .'.md5') {
+        print STDERR 'Calculating md5 for BAM: '. $bam_path ."\n";
```
ebelter: Is there a reason this is a print vs status_message?

lib/perl/Genome/Site/TGI/Extension/generate_lims_disk_unarchive.pl
tmooney: Empirically no, since we've been using this successfully so far. :smile:  It's a good idea, though, so I've added it!

lib/perl/Genome/Site/TGI/Extension/generate_lims_disk_unarchive.pl
tmooney: Added a check for `$ARGV[0]` and `[1]` to be defined.  Normally the user won't call this script directly, so I'm not fussed about making it too fancy!

lib/perl/Genome/Site/TGI/Extension/UnarchiveInstrumentData.pm
tmooney: Fixed.  (Hm, no deer emoji!)

lib/perl/Genome/Site/TGI/Extension/UnarchiveInstrumentData.pm
tmooney: Only that I missed this one one when converting the script to a real module... Changed!

lib/perl/Genome/Site/TGI/Extension/UnarchiveInstrumentData.pm
tmooney: I agree.  `sub unarchive_lims_data` is also pretty long, though it has less branching.

A mitigating factor might be that a new disk system is in the works which eliminates the need for this command entirely, so oughtn't to need to maintain it for even an entire year.  (Famous last words... 😄)

ebelter:
+1 This seems reasonable. Hopefully it will work! 🎆

PR #1535, opened 2016-08-17
InstDAta Import: Fix RevComp Bug

ebelter:
The importer was attempting to revcomp sequences, but only after the bit flag had been altered. This led to sequences not being correctly written.

In response to https://rt.gsc.wustl.edu/Ticket/Display.html?id=111023


tmooney:
+1 🐛

PR #1536, opened 2016-08-22
Delete the tag bridges in advance of deleting the subject mapping.

tmooney:
Otherwise the delete will fail due to the FK constraint.


acoffman:
+1

brummett:
The whole eval might be better written as
```
foreach my $hangoff_type ( qw( subject_bridges inputs tag_bridges) ) {
    $_->delete foreach $self->$hangoff_type;
}
```

Any idea why the eval is even there?  Right after, it's re-throwing the exception.

tmooney:
@brummett Updated!

I don't know what the `eval` was for. It doesn't seem like it does anything, so I took it out.

brummett:
+1

PR #1538, opened 2016-08-24
CCSVCF Result: "command" property should be optional.

tmooney:
We won't have the Command around if we load this from the database, which prevents us from `commit()`ing in contexts where one of these results has been loaded.

(I ran across this while trying to archive some of these results.  They might similarly be resistant to purging presently.)


jasonwalker80:
+1, even though I think CreateCrossSampleVcf could be removed at this point.

jasonwalker80:
+2

PR #1539, opened 2016-08-25
Require an Analysis Project to run Builds.

tmooney:
The long-awaited requirement!  A Model must belong to an Analysis Project (and one that's not Disabled at that!) in order to be built or queued for building.


jasonwalker80:
+2

PR #1541, opened 2016-08-25
Use a template to draw the alignment summary table.

tmooney:
By having a separate table for each version, each datatable independently paginates and the export SWF buttons grab the proper set of data.


jasonwalker80:
+1, @jmcmichael any thoughts on this PR?

jmcmichael:
+1 looks good to me, elegant use of XSL templates to repeat the same table!

PR #1542, opened 2016-08-29
Remove "dev_mode" configuration.

tmooney:
This was previously used to switch between hard-coded server values in the code, but we are (or should be!) using config to set those server values directly instead.


jasonwalker80:
+1

ebelter:
+1 Yay! Config!

PR #1543, opened 2016-08-29
Create a system Analysis Project

tmooney:
This is intended to be the AnP home for GMS-wide data like references and annotation data.

If this PR is accepted we'll need to actually make the AnP to match our chosen System AnP name!

(This or something like it is a pre-requisite for #1539 passing tests.)


lib/perl/Genome/Model/Build/ImportedAnnotation.t
```diff
@@ -20,6 +21,9 @@ my @versions = ('12_34', '56_78');
 my $ann_pp = Genome::ProcessingProfile::ImportedAnnotation->create(name => 'test_ann_pp', annotation_source => 'test_source');
 my $data_dir = File::Temp::tempdir('ImportedAnnotationTest-XXXXX', CLEANUP => 1, TMPDIR => 1);
 
+Genome::Test::Factory::AnalysisProject->setup_system_analysis_project;
```
jasonwalker80: If this returns the `$anp`, why get a new one on the next line?

lib/perl/Genome/Model/SomaticValidation/Command/ValidateSvs/CreateAssembledContigReference.t
```diff
@@ -21,7 +22,9 @@ use_ok($pkg) or die();
 
 my $test_dir = Genome::Utility::Test->data_dir_ok($pkg);
 
+my $anp = Genome::Test::Factory::AnalysisProject->setup_object();
```
jasonwalker80: Would you need to call `setup_system_analysis_project` first?

lib/perl/Genome/Model/Build/ImportedAnnotation.t
tmooney: The system AnP is only automatically applied when going through the `define` command interface, so it has to manually attached here.

I wanted to test that it works with any arbitrary AnP.  It should be completely equivalent, though, so perhaps there's no particularly good reason to use a different one.

lib/perl/Genome/Model/SomaticValidation/Command/ValidateSvs/CreateAssembledContigReference.t
tmooney: Nope; this test doesn't depend on the system AnP.

The `SomaticValidation` build passes along its model's AnP when creating the reference for realignment, so it will belong to the specific project for which it was created instead of the generic system.

jasonwalker80:
+2

PR #1544, opened 2016-08-30
Refseq Converter: Sort after lifting over.

tmooney:
LiftOver might scramble up the regions as they've moved between references, but we expect a sorted file to come out of the converter.


jasonwalker80:
+1

jasonwalker80:
@susannasiebert could you take a look at this PR?

susannasiebert:
+1

PR #1545, opened 2016-08-31
Add a blacklist for locations from which to not report errors.

tmooney:
"Genome/Ptero/Wrapper.pm" catches exceptions to clean up and then re-`die()`s, so we should keep looking for the original error location instead.


jasonwalker80:
+1

jasonwalker80:
+2

Issue #1546, opened 2016-08-31
Remove old mysql code


jasonwalker80:
There are a few remaining commands that either do no use the db_ensembl_* config or make direct connections to mysql servers.  For the former, they should be updated to use the config variables.  For the latter, we should consider removing the code or disabling the connection to mysql servers for old data sets like build36.

Consider updating:
/Model/Tools/Ensembl/Base.pm
./Model/Tools/ImportAnnotation/UpdateAnnotation.pm

Consider removing or disabling mysql connections from:
./Model/Tools/Pcap/Ace.t
./Model/Tools/Pcap/Config.pm
./Model/Tools/Sv/BreakAnnot.pl
./Model/Tools/Sv/SvAnnot.pm

ebelter:
Removal of the Pcap tree should be OK.

jasonwalker80:
I didn't look carefully before, it appears the `Model/Tools/ImportAnnotation` tree could be removed entirely as well.

tmooney:
SvAnnot.pm is part of the DV2 TigraValidation filter.

jasonwalker80:
Instead of removing the files, I was thinking we'd remove the code that connects to the DB.  For example, in SvAnnot perhaps we remove the `repeat_mask` option.  It only works for build36 anyway which is no longer supported.

jasonwalker80:
For BreakAnnot.pl that uses a database named `ucsc` which doesn't appear to be versions by reference build.  Is this still run during DV2 TigraValidation.  If so, I can't imagine how it is still working.

jasonwalker80:
For BreakAnnot.pl it appears it's used to get repeat masker annotation.  This will be WRONG for newer reference sequence builds.  I would suggest similar to SvAnnot, we remove the annotation of repeat regions from this tool.  Something like removing lines 240, 244 and the subroutine(s) that are called from 240.

jasonwalker80:
I agree BreakAnnot.pl should go.  I confirmed the ucsc database in mysql is from build36 as well.  The knownGene entry for MUC6 matched the UCSC hg18 build.

tmooney:
All associated have been merged.

PR #1547, opened 2016-08-31
Use config to get DB connection details.

tmooney:
This is part of #1546.

### DIFF
--- a/lib/perl/Genome/Model/Tools/Ensembl/Base.pm
+++ b/lib/perl/Genome/Model/Tools/Ensembl/Base.pm
@@ -6,9 +6,6 @@ use warnings;
 use Genome;
 
 my $DEFAULT_VERSION = '58_37c';
-my $DEFAULT_HOSTNAME = 'mysql1.gsc.wustl.edu';
-my $DEFAULT_USERNAME = 'mse';
-my $DEFAULT_PASSWORD = '';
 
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Model/Tools/ImportAnnotation/UpdateAnnotation.pm
```diff
@@ -64,8 +64,8 @@ sub execute
     my $flatfile = $self->derive_flatfile_location();
     my $gbff = "/gscmnt/sata835/info/medseq/annotation_data/human.rna.gbff"; # hard code this for now
 
-    my $ensembl_host = "mysql1"; # or mysql2?
-    my $ensembl_user = "mse";
+    my $ensembl_host = Genome::Config::get('db_ensembl_host');
+    my $ensembl_user = Genome::Config::get('db_ensembl_user');
```
jasonwalker80: Do we need to keep `Genome/Model/Tools/ImportAnnotation` around?  it looks really old.

tmooney: It does look old! This file's last update was in August 2009.  I only changed it because it was listed in #1546.

I agree that nobody should ever run this.  I'll take it out of this PR and remove it instead.

jasonwalker80:
+1

ebelter:
+1

PR #1548, opened 2016-08-31
Remove old `gmt import-annotation` tree.

tmooney:
This code has mostly been ignored for years. I'm impressed it survived so long.  (This is also part of #1546.)


jasonwalker80:
+1

acoffman:
+1

PR #1549, opened 2016-08-31
TigraValidation no longer uses MySQL

tmooney:
* This means build36 `TigraValidation` results can no longer be generated with RepeatMasker information included.
* It doesn't seem like `BreakAnnot.pl` is really used for anything, so it's removed completely.

This is part of #1546.


jasonwalker80:
+1

acoffman:
+1

PR #1556, opened 2016-09-06
Update blessed somatic-variation-short build to account for #1549.

tmooney:
RepeatMasker support was removed in #1549, which altered the headers of the SV output.  No data except for the headers `diff`ed between the current and proposed blessed builds.


jasonwalker80:
+1

PR #1558, opened 2016-09-07
Quote extraction types when determining is_rna

ebelter:
This came up when configuring an AnP. Seems qw() was used when quotes were needed.


tmooney:
+1 ⛲

lib/perl/Genome/Library.t
```diff
@@ -2,37 +2,62 @@
...
+};
+
+subtest 'is_rna' => sub{
+    plan tests => 4;
+
+    $library->sample->extraction_type('dna');
+    is($library->is_rna, 0, 'is NOT rna when extraction_type is dna and transcript_strand is UNDEF');
+
+    $library->transcript_strand('unstranded');
+    is($library->is_rna, 1, 'is rna when extraction_type is DNA and transcript_strand is unstranded');
```
jasonwalker80: Do we expect DNA extraction type AND a transcript_strand that is defined?  I guess any time transcript_strand is set is_rna should return true, but not sure how practical this is.

jasonwalker80:
+1, the DNA/transcript_strand test is peculiar, but I don't think the behavior is incorrect.

lib/perl/Genome/Library.t
ebelter: Probably not, I just added a test that preserved the logic as is in the method. The best place to validate this would be with an '__error__' to have transcript strand set and be 'DNA' on DB save.

PR #1559, opened 2016-09-08
Remove support for "Workflow".

tmooney:
This removes all references to `Workflow` and support for `workflow_builder_backend: workflow`.  If this is merged, a subsequent commit will be required to remove the workflow submodule from the root of the repository (after various automated processes are updated to no longer expect it).

Support for `view`ing builds run with older workflows is also removed.  (If this is merged, all data associated with `Workflow`s could therefore be deleted from the database.)


jasonwalker80:
I have started to review this commit-by-commit to make it a bit more digestible (at the suggestion of @tmooney ).  Most of it is deletions.  I have some background with old workflow, but little with Flow and PTero.  @acoffman and @ebelter would you have time to help review this?  If you all prefer, we could organize a group code-review with a projector in a conference room.  Possibly on Thursday?  Otherwise if you have time to review on your own before then that would be great.

PR #1561, opened 2016-09-13
Make normal/tumor sample field real

dufeiyu:
GatkSomaticIndel sometimes swaps the normal and tumor columns in its own vcf output that this vcf converter is based on. Aka, column 9 is not always for tumor. This fix replaces the hardcoded normal/tumor column index with a better way to secure the normal/tumor columns.


dufeiyu:
CI, test this

brummett:
Are you sure the column you're looking for will always be column 9 or 10?

If you run this command on multiple files within the same process that have different column orderings, then it'll do the wrong thing on the first read of the first file after the first read of the second file; it'll have stored the indexes of the second file when reading the first.  You'll have to cache these values on the object instance, not globally.

dufeiyu:
There is always only one input_file for G::M::T::Vcf::Convert::Snv/Indel under DV2 and there are always 11 columns in the original GatkSomaticIndel output vcf file.

PR #1562, opened 2016-09-15
Don't bother intersecting this file with itself. Return the one copy.

tmooney:
`bedtools intersect` seems to not be stable with regard to self-intersection.  It's adding duplicate rows to the resulting file, which causes trouble for our R script.


jasonwalker80:
+2

PR #1564, opened 2016-09-15
Remove out-dated bash completion helper.

tmooney:
It contains a hard-coded `/gsc/` path, and it doesn't even point to the correct file path within the repository assuming someone did set it up that way!


jasonwalker80:
+2

PR #1566, opened 2016-09-15
Remove old Varscan Copy-Number tools.

tmooney:
The usage log showed no use in the past 15 months for these commands.

The plain `CopyNumber` tool includes a reference to `/gsc/bin/sh` (and proceeds to ignore it by calling `bash` explicitly).  This removes the command rather than bothering to fix it.


jasonwalker80:
+2

PR #1568, opened 2016-09-15
Remove old CopyNumber ReadDepth tools.

tmooney:
The `bam-window` parts of these is not currently in a functional state, so I propose this removal rather than fixing it up.


jasonwalker80:
+2

PR #1572, opened 2016-09-16
Add FT format field to pindelRegion vcf

dufeiyu:
Adding PASS to FT format field for somatic variants can preserve this pass status through vcf filter and joinx vcf-merge later.

### DIFF
--- a/lib/perl/Genome/Model/Tools/DetectVariants2/PindelRegion.t
+++ b/lib/perl/Genome/Model/Tools/DetectVariants2/PindelRegion.t
@@ -25,11 +25,12 @@ my $refbuild_id = 101947881;
 my $ref_seq_build = Genome::Model::Build::ImportedReferenceSequence->get($refbuild_id);
 ok($ref_seq_build, 'human36 reference sequence build') or die;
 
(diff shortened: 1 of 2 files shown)


lib/perl/Genome/Model/Tools/DetectVariants2/PindelRegion.t
```diff
@@ -25,11 +25,12 @@ my $refbuild_id = 101947881;
 my $ref_seq_build = Genome::Model::Build::ImportedReferenceSequence->get($refbuild_id);
 ok($ref_seq_build, 'human36 reference sequence build') or die;
 
-my $test_base_dir = Genome::Utility::Test->data_dir($pkg);
-my $region_file   = File::Spec->join($test_base_dir, 'region_file.bed');
+my $test_base_dir  = Genome::Utility::Test->data_dir($pkg);
+my ($test_input_dir, $test_out_dir) = map{$test_base_dir.'/'.$_}qw(input expected_2);
```
tmooney: This could also use `File::Spec->join` like the lines below it.

dufeiyu: fixed

PR #1574, opened 2016-09-20
Update blessed builds

dufeiyu:
All diff is caused by newly fixed GatkSomaticIndel vcf format (https://github.com/genome/genome/pull/1561)and the diff is fine. This PR is to update the blessed builds so Jenkins Genome-Model-Tests can pass.


dufeiyu:
Early merge to make Jenkins model test pass.

PR #1576, opened 2016-09-21
Update homo_polymer list with min_length 7

dufeiyu:
The current homo polymer list has homo polymers of min_length 6, which filters out some real indel variants. The new list has homo polymers of min_length 7 with about 4 million length_6 polymers removed. The indel counts of trio report using the new list look fine. Refer to https://jira.gsc.wustl.edu/browse/CI-191  Diff is expected for regression test.


dufeiyu:
Diff is expected for regression test.

PR #1582, opened 2016-09-24
add_format_field first messes up vcf entry format_field_index

dufeiyu:
This causes entry sample_field failing to find value for given field name.


tmooney:
Does this point to a bug in `Genome::File::Vcf::Entry`?  I'm not understanding why the ordering of these should make a difference.

dufeiyu:
Actually it is a bug in Genome::File::vcf::Entry. I will close this PR and make a new one to fix that bug.

PR #1584, opened 2016-09-27
Reset _format_key_to_idx hash properly and avoid check with autovivified hash

dufeiyu:
This PR is to fix the bug. This is causing $entry->sample_field not returned correctly if add_format_field first.

### DIFF
--- a/lib/perl/Genome/File/Vcf/Entry.pm
+++ b/lib/perl/Genome/File/Vcf/Entry.pm
@@ -592,7 +592,8 @@ sub add_format_field {
     else {
         my $idx = scalar @{$self->{_format}};
         push @{$self->{_format}}, $field;
-        $self->{_format_key_to_idx}{$field} = $idx;
+        my $format_idx = $self->format_field_index;
+        $self->{_format_key_to_idx} = {%$format_idx, $field => $idx};
         return $idx;
     }
 }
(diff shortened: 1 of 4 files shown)


lib/perl/Genome/File/Vcf/Entry.pm
```diff
@@ -592,7 +592,8 @@ sub add_format_field {
     else {
         my $idx = scalar @{$self->{_format}};
         push @{$self->{_format}}, $field;
-        $self->{_format_key_to_idx}{$field} = $idx;
+        my $format_idx = $self->format_field_index;
+        $self->{_format_key_to_idx} = {%$format_idx, $field => $idx};
```
ebelter: Could this be more understandable written as:
```
$format_idx->{ $field } = $idx;
 $self->{_format_key_to_idx} = $format_idx;
```
It also avoids copying the hash into another when using %$format_idx.

dufeiyu: makes sense. I will make the change.

lib/perl/Genome/File/Vcf/Entry.pm
```diff
@@ -622,7 +622,9 @@ an example.
 
 sub format_field_index {
     my ($self, $key) = @_;
-    if (!exists $self->{_format_key_to_idx}) {
+    my $format_idx = $self->{_format_key_to_idx};
+
+    unless (scalar keys %$format_idx) {
```
dufeiyu: Makes sense. I will make the change.

dufeiyu:
@tmooney  Can you review this too ?  This is the one to fix the bug in Genome::File::Vcf::Entry that we discussed in https://github.com/genome/genome/pull/1582    thanks

PR #1593, opened 2016-10-10
Coverage Report: Use a wrapping <div> for the hiding. Apply dataTables to the <table>.

tmooney:
We need the datatables buttons to be toggled when performing the algorithm swap, so that only the appropriate set is visible.


jasonwalker80:
+1, @jmcmichael care to take a quick look at this?

PR #1594, opened 2016-10-12
Add detailed.pass.vcf to DV2 output

dufeiyu:
Currently cle trio report uses detailed.vcf as input, which takes long time to run through Vep and bamreadcount. A detailed.pass.vcf that only keeps final pass variants is needed as DV2 output and should be used for trio reporting.


lib/perl/Genome/Model/Tools/Vcf/FinalPassOnly.pm
```diff
@@ -0,0 +1,72 @@
+package Genome::Model::Tools::Vcf::FinalPassOnly;
+
+use Genome;
+use strict;
+use warnings;
+use Genome::File::Vcf::Reader;
+use Genome::File::Vcf::Writer;
+
+class Genome::Model::Tools::Vcf::FinalPassOnly {
+    is => 'Command',
```
tmooney: New commands should probably use `Command::V2`.  There's less of an effect here since none of the parameters are objects, though.

lib/perl/Genome/Model/Tools/Vcf/FinalPassOnly.pm
```diff
@@ -0,0 +1,72 @@
...
+        next if $entry->is_filtered;
+        my $filters = $entry->filters;
+        if ($entry->sample_field($index, 'FT') eq 'PASS' or $filters ~~ ['PASS']) {
+            $out_vcf->write($entry);
+        }
+    }
+    $out_vcf->close;
+
+    if ($self->tabix_index) {
+        my $cmd = Genome::Model::Tools::Tabix::Index->create(
```
tmooney: The docs for output file say it can be either `.vcf` or `.vcf.gz`, but indexing requires a compressed file.  Would it be possible to either skip this (even if `$self->tabix_index` is true) when `not Genome::Sys->file_is_gzipped($out_file)` or provide an error message about the issue when the command starts?

lib/perl/Genome/Model/Tools/Vcf/FinalPassOnly.pm
dufeiyu: oops, copy/paste wrong thing, will make the change.

lib/perl/Genome/Model/Tools/Vcf/FinalPassOnly.pm
dufeiyu: Good point. I will make the change.

PR #1597, opened 2016-10-18
Remove old build workflow configuration.

tmooney:
Support for `Workflow` is gone as of #1559, so this variable is no longer used by anything.


acoffman:
+1

jasonwalker80:
+1

PR #1599, opened 2016-10-20
Update blessed builds

dufeiyu:
5 blessed builds are updated. The diffs are caused by recent DV2 output addition of .detailed.pass.vcf. https://github.com/genome/genome/pull/1594  This PR needs early merge to make test pass.


jasonwalker80:
+1

PR #1600, opened 2016-10-20
Export DeNovoAssembly Builds

ebelter:
This was a home directory comand. I have converted and updated it to live in Genome. The command takes a work order (Genome::Proejct) and export finishing style projects.

### DIFF
--- a/lib/perl/Genome/Model/DeNovoAssembly/Command/ExportAsGscProject.pm
+++ b/lib/perl/Genome/Model/DeNovoAssembly/Command/ExportAsGscProject.pm
@@ -66,51 +66,8 @@ sub execute {
     my $ans = <STDIN>;
     exit unless $ans =~ /yes/i;
 
-# copy succeeded builds;
     for my $build( @succeeded_builds ) {
-        my $build_dir = $build->data_directory;
-        die "No data directory found, $build_dir\n" unless -d $build_dir;
-
-        # project dir
-        my $output_dir = $copy_dir.'/'.$build->model->subject->name;
-        $output_dir =~ s/\s+/_/g;
-        Genome::Sys->create_directory( $output_dir ) unless -d $output_dir;
-
-        # create build id empty file
-        my $build_id_file = $output_dir."/build_id=".$build->id;
-        Genome::Sys->touch( $build_id_file ) if not -e $build_id_file;
-
-        my $assembler = $build->model->processing_profile->assembler_name;
-        # create subdirs
-        my @sub_dirs = subdirs_for_assembler( $assembler );
-        die "Can't determine which subdirs to create for assembler, $assembler\n" unless @sub_dirs;
-        for my $dir_name( @sub_dirs ) {
-            my $sub_dir = $output_dir."/$dir_name";
-            Genome::Sys->create_directory( $sub_dir ) unless -d $sub_dir;
-        }
-
-        # edit_dir files to copy
-        my $builds_edit_dir = assemblers_edit_dir( $assembler );
-        die "Can't determine locations of builds edit_dir for assembler, $assembler\n" unless $builds_edit_dir;
-        for my $file( glob("$build_dir/$builds_edit_dir/*") ) {
-            my $file_name = basename( $file );
-            my $to = $output_dir.'/edit_dir/'.$file_name;
-            print "skipping, $file already exists\n" and next if -s $to;
-            print "Copying $file to $to\n";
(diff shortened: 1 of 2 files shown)


lib/perl/Genome/Model/DeNovoAssembly/Command/ExportAsGscProject.pm
```diff
@@ -0,0 +1,176 @@
...
+        subdirs => [qw/ edit_dir chromat_dir phd_dir phdball_dir /],
+        edit_dir => File::Spec->join('consed', 'edit_dir'),
+    },
+);
+sub supported_assemblers { keys %supported_assemblers }
+sub is_model_supported {
+    my ($self, $model) = validate_pos(@_, {isa => __PACKAGE__}, {type => OBJECT});
+
+    my $assembler = $model->processing_profile->assembler_name;
+    return if ! List::MoreUtils::any { $assembler eq $_ } supported_assemblers();
```
tmooney: This is rather inconsequential for `N = 2`, but it would be "faster" to check for the key's existence in the hash rather than to create and then iterate over a list of all keys.

lib/perl/Genome/Model/DeNovoAssembly/Command/ExportAsGscProject.pm
```diff
@@ -0,0 +1,176 @@
...
+
+    my @inst_data = $model->instrument_data;
+    return if $assembler =~ /newbler/i && @inst_data > 1;
+    return 1;
+    #return if ! exists $subjects_to_copy{ $model->subject->name };
```
tmooney: Can this commented-out line be removed?

lib/perl/Genome/Model/DeNovoAssembly/Command/ExportAsGscProject.pm
```diff
@@ -0,0 +1,176 @@
...
+    $self->status_message('Existing Models: %s ', scalar(@models));
+
+    my @succeeded_builds;
+    for my $model ( @models ) {
+        next if ! $self->is_model_supported($model);
+        push @succeeded_builds, $model->last_succeeded_build;
+    }
+
+    $self->status_message('Succeeded builds: %s', scalar(@succeeded_builds));
+    $self->fatal_message('No succeeded builds found!') if not @succeeded_builds;
```
tmooney: These two lines could probably be swapped to avoid redundant messaging.

lib/perl/Genome/Model/DeNovoAssembly/Command/ExportAsGscProject.pm
```diff
@@ -0,0 +1,176 @@
...
+
+    return \@succeeded_builds;
+}
+
+sub _export_build {
+    my ($self, $build) = @_;
+
+    my $build_dir = $build->data_directory;
+    $self->fatal_message("Build data directory does not exist! $build_dir") unless -d $build_dir;
```
tmooney: Would it be worthwhile to check the archive here?

lib/perl/Genome/Model/DeNovoAssembly/Command/ExportAsGscProject.pm
```diff
@@ -0,0 +1,176 @@
...
+
+    # project dir
+    my $copy_dir = $self->directory;
+    my $output_dir = File::Spec->join($copy_dir, $build->model->subject->name);
+    $output_dir =~ s/\s+/_/g;
```
tmooney: This seems like a job for `Genome::Utility::Text`'s `sanitize_string_for_filesystem`!  (I'd be tempted to include the ID in the path to avoid collisions, though.)

ebelter: Sounds good. Including the id in the path is a good idea, except this is how they expected the directory to be named.

lib/perl/Genome/Model/DeNovoAssembly/Command/ExportAsGscProject.pm
ebelter: These builds are pulled over quickly after sequencing, so it's OK for now.

lib/perl/Genome/Model/DeNovoAssembly/Command/ExportAsGscProject.pm
ebelter: Yeah, I was encapsulating it in the above method. Changed.

PR #1601, opened 2016-10-25
fixing issues with target-region filtering, header, and indels

chrisamiller:


lib/perl/Genome/Model/Tools/Validation/ProcessSomaticValidation.pm
```diff
@@ -751,30 +705,86 @@ sub execute {
...
+  #-------------------------------------------------
+  #filter out the off-target regions, if target regions are available
+  if($self->restrict_to_target_regions){
+      $self->status_message("Filtering out off-target regions...");
+      
+      my $featurelist;
+      if($self->target_regions) {
+          $featurelist = $self->target_regions;
+      } else { 
+          $featurelist = $model->target_region_set->file_path;
```
tmooney: It's a corner case, but it's possible for `$model` and `$build` to have a different `target_region_set` set.  Would it be better to prefer the `$build`'s set?

lib/perl/Genome/Model/Tools/Validation/ProcessSomaticValidation.pm
```diff
@@ -751,30 +705,86 @@ sub execute {
...
+              $chr =~ s/^chr//g;
+              print FEATFILE join("\t",( $chr, $start, $stop, @rest)) . "\n";
+          }
+          close($inFh);
+          close(FEATFILE);
+          my $cmd;
+          $cmd = "joinx sort '" . $output_dir . "/" . $sample_name . "/featurelist.tmp'" . " >'" . $output_dir . "/" . $sample_name . "/featurelist'";
+          `$cmd`;
+          $cmd = "rm -f '" . $output_dir . "/" . $sample_name . "/featurelist.tmp'";
+          `$cmd`;
```
tmooney: If `Genome::Sys->shellcmd` is overkill, I'd still recommend using `system` with the params in an `@array`).  Maybe even better would be to `Genome::Utility::Text::sanitize_name_for_filesystem` the `$sample_name`!

PR #1610, opened 2016-11-08
Make it possible to disable local filesystem caching.

tmooney:
Set config to empty string or non-existent location to disable cache.  (This will fall back to original location if cache is temporarily unavailable as well.)


bderickson:
+1 :)

PR #1617, opened 2016-11-16
BWA-MEM: Only pass parameter for paired FASTQs if "-p" flag used.

tmooney:
Otherwise we get no output on fragment data!


dufeiyu:
+1

PR #1619, opened 2016-11-16
Avoid using `each()` in loop where we insert into the hash.

tmooney:
Newer Perls provide this warning:
> Use of each() on hash after insertion without resetting hash iterator results in undefined behavior


brummett:
+1, though the sort is unnecessary

PR #1620, opened 2016-11-16
Test::Builder: Don't return before evaluating the entire expression.

tmooney:
This module really shouldn't be in the repository to begin with... but as long as we're supporting old systems it shall remain.


brummett:
+1

PR #1621, opened 2016-11-21
Remove imported restriction to assiging instrument data.

jasonwalker80:
Allow the assignment of all instrument data to analysis projects regardless of class.


tmooney:
+1.  We still won't synch new production data without an AnP being associated with a Work Order, so this only helps add existing data to projects.

PR #1622, opened 2016-11-21
Update IGV ref name example values.

jasonwalker80:
Add hg38 and mm10 as example values for IGV references.


jasonwalker80:
Updated sort order based on relevance.

PR #1626, opened 2016-11-21
Compare Analysis Project Instrument Data 

jasonwalker80:
This command was tucked away in one of my git repos.  Any suggestions or comments on whether it would be valuable.

initial revision of a command to compare AnP instrument data.

### DIFF
--- a/lib/perl/Genome/Config/AnalysisProject/Command/CompareInstrumentData.pm
+++ b/lib/perl/Genome/Config/AnalysisProject/Command/CompareInstrumentData.pm
@@ -6,23 +6,24 @@ use warnings;
 use Genome;
 
 use Set::Scalar;
+use Data::Dumper;
 
 class Genome::Config::AnalysisProject::Command::CompareInstrumentData {
     is => ['Command::V2'],
     has => [
-        analysis_project_a => {
+        first_analysis_project => {
             is => 'Genome::Config::AnalysisProject',
             shell_args_position => 1,
         },
-        print_a_diff => {
+        print_first_diff => {
             is => 'Boolean',
             default_value => 0,
         },
-        analysis_project_b => {
+        second_analysis_project => {
             is => 'Genome::Config::AnalysisProject',
             shell_args_position => 2,
         },
-        print_b_diff => {
+        print_second_diff => {
             is => 'Boolean',
             default_value => 0,
         },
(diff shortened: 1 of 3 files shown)


lib/perl/Genome/Config/AnalysisProject/Command/CompareInstrumentData.pm
```diff
@@ -0,0 +1,70 @@
...
+use warnings;
+
+use Genome;
+
+use Set::Scalar;
+
+class Genome::Config::AnalysisProject::Command::CompareInstrumentData {
+    is => ['Command::V2'],
+    has => [
+        analysis_project_a => {
```
tmooney: It might be nice if the two parameter names differed at the beginning.  For instance, `--first-analysis-project` and `--second-analysis-project` can be shortened to `--first` and `--second`.  The current names require the entire parameter to be spelled out each time.

lib/perl/Genome/Config/AnalysisProject/Command/CompareInstrumentData.pm
```diff
@@ -0,0 +1,70 @@
...
+    return 1;
+}
+
+sub print_diff {
+    my $diff_set = shift;
+
+    for my $diff_member ( $diff_set->members) {
+        my @parts = Genome::ProjectPart->get(entity_id => $diff_member->id);
+        my @projects = Genome::Project->get([map $_->project_id, @parts]);
+        print $diff_member->sequencing_platform ."\t". $diff_member->sample->name ."\t". $diff_member->id ."\t". $diff_member->ignored ."\t". $diff_member->flow_cell_id ."\t". join(',',sort(map {$_->id} @projects)) ."\n";
```
tmooney: This could leverage the existing lister (with a default `--show` of `sequencing_platform,sample.name,id,ignored,flow_cell_id,projects.id` or similar) so as to allow customizing the output columns and formatting easily.

lib/perl/Genome/Config/AnalysisProject/Command/CompareInstrumentData.pm
jasonwalker80: 14ff0a4 should address this.

lib/perl/Genome/Config/AnalysisProject/Command/CompareInstrumentData.pm
jasonwalker80: b886faf should address the intersection feature request.

jasonwalker80:
b886faf should address the intersection feature request.

lib/perl/Genome/Config/AnalysisProject/Command/CompareInstrumentData.pm
jasonwalker80: @tmooney is there an example you could point me to?  I'm not entirely sure how to leverage the lister class here.

jasonwalker80: Something like `Genome::InstrumentData::Command::List->execute()`? What would I use as the `filter`?

tmooney: Here are a few commands that leverage listers in different ways:
* [`genome analysis-project show-config`](https://github.com/genome/genome/blob/76dde10a4d90259b8c3f35d2ae008d69e4521d4b/lib/perl/Genome/Config/AnalysisProject/Command/ShowConfig.pm)
* [`genome model instrument-data list`](https://github.com/genome/genome/blob/76dde10a4d90259b8c3f35d2ae008d69e4521d4b/lib/perl/Genome/Model/Command/InstrumentData/List.pm)
* [`genome config grep`](https://github.com/genome/genome/blob/93bcf79e30e3960d1d528a875ff65652c82c9d7f/lib/perl/Genome/Config/Command/Grep.pm)

The first two are implemented as listers with different forms of custom filtering.  The last one calls the lister as part of running the command--this might be the easiest approach to take here?

tmooney: If you call the lister's execute, the filter would end up being the list of IDs, like [here](https://github.com/genome/genome/blob/93bcf79e30e3960d1d528a875ff65652c82c9d7f/lib/perl/Genome/Config/Command/Grep.pm#L67).

jasonwalker80: thanks, I started down the road of calling execute on the lister.  Mostly there. thanks again.

jasonwalker80: d9036df now uses the object lister.

jasonwalker80:
CI, test this.

lib/perl/Genome/Config/AnalysisProject/Command/CompareInstrumentData.pm
```diff
@@ -0,0 +1,119 @@
...
+            shell_args_position => 1,
+        },
+        second_analysis_project => {
+            is => 'Genome::Config::AnalysisProject',
+            doc => 'The second Analysis Project to compare Instrument Data.',
+            shell_args_position => 2,
+        },
+    ],
+    has_optional => [
+        list_intersection => {
```
susannasiebert: Do we expect the user to only ever pick one of these options or do we think they might pick a mix or all of them. If the first is the expected use case we might consider changing this to be a `mode` that can take one of `intersection`, `first_only`, or `second_only`.

susannasiebert: Alternatively, this could be replaced with a generic `list_details` option that then determines whether or not the `list_set` calls are executed.

```
if ($self->list_details && $set_x->size > 0) {
    $self->list_set($set_x);
}
```
etc.

PR #1631, opened 2016-11-29
WIP: Use samtools sort instead of Picard.

jasonwalker80:
@ebelter please provide feedback on how to move forward.

It appears that samtools sort doesn't play nice with Picard Downsample:
commit a93daccec5838d39259f1092c18ee495c9d0ee40
Author: Eddie Belter <ebelter@genome.wustl.edu>
Date:   Tue Mar 29 13:52:02 2016 -0500

   Switch to picard sort so picard downsample does not fail


ebelter:
The downsampling tests will continue to fail. Picard and samtools sort slightly differently by name, which is what causes this issue when downsampling. You could have 2 sorters, and use Picard when downsampling.

jasonwalker80:
#1633 takes this in a different direction.  However, neither samtools or sambamba correctly order the reads by queryname AND read pair( read1 then read2).

PR #1632, opened 2016-11-30
Debug message for the resource when locking.

tmooney:
This should make it easier to find out what we're waiting on when something else has the lock.  This replaces the earlier pre-backend message that would output the entire contents of the file lock.  (Unlike the previous message, this message gets included even if we do get the lock immediately.)


tmooney:
(This is failing tests due to #1635.)

PR #1636, opened 2016-11-30
Trust that symlinks are mounted.

tmooney:
The `is_mounted` method is premised that each mount path is automounted.  If a mount_path is a symlink, instead assume that it is part of a larger mount that already is mounted--someone should check that larger mount's path directly if they care.

(I'm not sure this is the best solution to this problem, but it seems to solve the problem of making allocations under the new scheme.)

### DIFF
--- a/lib/perl/Genome/Disk/Volume.pm
+++ b/lib/perl/Genome/Disk/Volume.pm
@@ -273,6 +273,10 @@ sub is_mounted {
         return 1;
     }
 
+    if (-l $self->mount_path) {
+        return 1; #trust that symlinks are mounted
+    }
+
     # We can't use Filesys::Df::df because it doesn't report the mount path only the stats.
     my $mount_path = $self->mount_path;
     my @df_output = qx(df -P $mount_path 2> /dev/null);


brummett:
Maybe this whole method could change...

What if it did something like this instead:
* Get the device number of the root filesystem (0th item returned by ```stat('/')```)
* Follow symlinks of the mount path to the final non-symlink thing
* Get the device number of that final thing
The volume is then mounted if the second succeeds _and_ the device number is different than /'s device number

tmooney:
Here's what I get in a test:
```
30 Nov 17:17 [11] $ (stat('/'))[0], (stat('/gscmnt'))[0], (stat('/gscmnt/gc7007'))[0], (stat('/vol/aggr14/gc7007'))[0]
$VAR1 = 2051;
$VAR2 = 30;                                                                                                                                                                                                                                 
$VAR3 = 29;                                                                                                                                                                                                                                 
$VAR4 = 29;
```
If it weren't mounted, I'd expect the third one to also be 30, but the last one wouldn't be found.

I think [the volume lister](https://github.com/genome/genome/blob/c3c9b745f8f62ec9ebc2ba3f63c763a4cd084039/lib/perl/Genome/Disk/Command/Volume/List.pm#L76) will also need updates.  It has a slightly different concept of `is_volume_mounted` since it wants to avoid actions that will cause mounting to occur.

brummett:
Here's what I get on blade17-1-1, which seems to be using the new symlink setup:
```
/: 2049
/gscmnt: 2049
/gscmnt/gc7007: 44
/vol/aggr14/gc7007: 44
```
And on a blade with the old setup:
```
/: 2067
/gscmnt: 30
/gscmnt/gc7007: 41
Use of uninitialized value $dev in concatenation (.) or string at try.pl line 6.
/vol/aggr14/gc7007:
```

```stat()``` follows symlinks automatically.  If you want to stat the link, then you'd use ```lstat()```.  So that would simplify the new code - you wouldn't have to follow the symlink chain by hand.

Here's the problem.  Statting an unmounted dir triggers it to be automounted, which defeats the purpose...

PR #1639, opened 2016-12-01
Use config to resolve disk group for unarchives.

jasonwalker80:


jasonwalker80:
See the latest commit for changes to kb requested.

PR #1644, opened 2016-12-12
Treat "errored" PTero status as a failure condition.

tmooney:
We will now display the "execute" attempt if the "shortcut" attempt errors out.  (Previously, the errored shortcut would be displayed even if the execute is running smoothly.)


jasonwalker80:
+2

PR #1646, opened 2016-12-20
Make column order in snv-indel-report more predictable

malachig:
Previously the ordering of the snv-indel-report columns was sometimes based on a different value than what was actually used for the column header.  This lead to some confusion, in particular in cases where sample name sorting resulted in [tumor then normal] in one case and [normal then tumor] in the next case.  Sorting and column display are now based on the same values.


malachig:
Hey @tmooney any thoughts on why the the `apipe-ci/quick-tests` is failing?  I don't quite follow the error logs. Errors related to cloning 'origin'?

tmooney:
@malachig I believe this is a side-effect of the slow disk issues.  We started seeing these `ERROR: Timeout after 10 minutes` messages after the maintenance window.  Unsatisfying answer: rebuild the tests enough times and it should go through!

malachig:
Hi @tmooney, thanks. Naive question, how do I trigger a new test attempt again?

tmooney:
If you are logged into the Jenkins instance, there should be a "Rebuild" option in the lefthand column of the Jenkins page for the test.  I started one this morning when commenting, though it's not looking too good so far.

malachig:
Hey @tmooney.  I logged in and poked around Jenkins a bit.  I don't see the rebuild option though...  Maybe still not looking in the right place? 

Screenshot:
screen shot 2017-01-09 at 1 49 17 pm

tmooney:
CI, test this.

jasonwalker80:
CI, test this

apipe-tester:

apipe-tester:

PR #1650, opened 2017-01-05
Remove code for discontinued "GenePrediction" model types.

tmooney:
This removes altogether what was retired in #1605.


jasonwalker80:
+1

PR #1652, opened 2017-01-12
More careful splitting of lock information.

tmooney:
* PIDs and LSF Job IDs will be numeric, so we can reliably pull them off as the last two fields.
* Underscores are not a valid hostname character, so that will be the first field.
* Usernames might have underscores in them, so take whatever's left.


brummett:
It would be nice to have a test that includes strings that previously failed.

I suspect a regex could extract the fields more efficiently, though maybe at the cost of readability.

PR #1653, opened 2017-01-18
Specify four CPUs in Strelka request.

tmooney:
This should match the number of jobs specified for `make` [in the detector](https://github.com/genome/genome/blob/3dec8bf28dbfe9f74e6006c2604b01d1237f0dc4/lib/perl/Genome/Model/Tools/DetectVariants2/Strelka.pm#L107).


susannasiebert:
Should we update the make file to get the number of jobs from the env, if possible?

PR #1654, opened 2017-01-19
Analysis Projects Require Environment Specifications

tmooney:
* Rename the "production" option in create to "prod-builder".
* Only allow an Analysis Project to be released if it has an environment configuration specified.

This PR doesn't enforce the contents of the environment configuration in any way.

### DIFF
--- a/lib/perl/Genome/Config/AnalysisProject/Command/Release.pm
+++ b/lib/perl/Genome/Config/AnalysisProject/Command/Release.pm
@@ -56,7 +56,17 @@ sub __errors__ {
                 desc => "Can't release analysis project " . $anp->__display_name__ . " with status: $status"
             );
         }
+
+        my $config = $anp->environment_config_dir;
+        unless ($config) {
+            push @errors, UR::Object::Tag->create(
+                type => 'error',
+                properties => ['analysis_projects'],
+                desc => "Can't release analysis project " . $anp->__display_name__ . " until an environment configuration has been set.",
+            );
+        }
     }
+
     return @errors;
 }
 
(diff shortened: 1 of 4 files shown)


lib/perl/Genome/Config/AnalysisProject/Command/Release.pm
```diff
@@ -46,15 +46,27 @@ sub execute {
...
             push @errors, UR::Object::Tag->create(
                 type => 'error',
                 properties => ['analysis_projects'],
-                desc => "Can't release analysis project with status: $status"
+                desc => "Can't release analysis project " . $anp->__display_name__ . " with status: $status"
+            );
+        }
+
+        my $config = $anp->environment_config_dir;
+        unless ($config) {
```
acoffman: Should it maybe check that there's a non-empty file in there?

tmooney: The [`environment_config_dir`](https://github.com/tmooney/genome/blob/63fb747959461b19b699507aefc568bfdf8dcea3/lib/perl/Genome/Config/AnalysisProject.pm#L151) method does validate the path exists before returning it, but it doesn't check that there's really anything specified.

Would it be beneficial to add some extra checking there?  I considered parsing the environment in advance to look for particular keys, but it's not guaranteed that we'll need to use them for a particular build type.  We could at least look for some valid configuration file at all, though?

acoffman: Yeah, validating the contents of the config seems like a losing battle at this point. I was just debating seeing if there was a non-empty file of some type there, but if that's already being done in the accessor, I think its fine "as-is."

acoffman:
+1

PR #1655, opened 2017-01-19
Escape Sample Name Used as Filename in bam-readcounts

tmooney:
A `/` came up in a sample name, which was causing trouble!


ebelter:
+1 Although I dread samples named like this.

PR #1656, opened 2017-01-19
Remove all trailing slashes from host-groups... not just one!

tmooney:
Tests started failing here due to a change in the host string being returned in production.  This is likely to cause real alignment failures as well...


dufeiyu:
+100   This is causing CLE_regression_test som-val build failing.

PR #1657, opened 2017-01-24
Fix/Disable ClinSeq tests using purged data

tmooney:
A bunch of tests were using production data, which has been removed.  This updates one of them to use test data and disables the rest.

### DIFF
--- a/lib/perl/Genome/Model/Build/ClinSeq/FileAccessors.t
+++ b/lib/perl/Genome/Model/Build/ClinSeq/FileAccessors.t
@@ -10,117 +10,180 @@ BEGIN {
 use above 'Genome';
 use Test::More;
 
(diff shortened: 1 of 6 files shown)


tmooney:
Switching out the ClinSeq model from that AnP yields an error about missing files.  I started down the path of converting this to use the test data, but I'm not sure where to get some of the files it seems to require...

The command is still tested by the `apipe-test-clinseq-wer` model test.  The individual components that make up the SnvIndelReport (like `AnnotateSnvsVcf`, `AddRsid`, and `SummarizeSnvINdelReport`) do have tests that still run.

It's definitely not ideal to disable any of these tests!  I would be reluctant to replace any of them with new live data, but if we find suitable live data perhaps it could be copied to a test data directory.

PR #1659, opened 2017-02-06
fixing problem where new build causes multiple matches for the mm9 build query

chrisamiller:


tmooney:
I don't think it's a good idea to have two reference builds with the same name...  Could we distinguish them somehow?

chrisamiller:
I agree - no idea where they came from:
```
$ use Genome;
$ my @builds = Genome::Model::Build::ReferenceSequence->get(name => "UCSC-mouse-buildmm9")
"my" variable @builds masks earlier declaration in same scope at (eval 1887) line 6.
$Genome_Model_Build_ImportedReferenceSequence1 = Genome::Model::Build::ImportedReferenceSequence=HASH(0x8a92580);
$Genome_Model_Build_ImportedReferenceSequence2 = Genome::Model::Build::ImportedReferenceSequence=HASH(0x8a94b90);

$ map {$_->id} @builds
$VAR1 = 107494762;
$VAR2 = 'f39a0212bf5c406cbafda080644b70de';
```

Looks like the second build was created by @jasonwalker80. Jason, any thoughts on why this second build exists and has the same name?  
```
$ gmb list id=f39a0212bf5c406cbafda080644b70de -sh id,model_name,run_by,date_scheduledID                                 MODEL_NAME   RUN_BY    DATE_SCHEDULED
--                                 ----------   ------    --------------
f39a0212bf5c406cbafda080644b70de   UCSC-mouse   jwalker   2016-08-04 18:56:59
```

jasonwalker80:
I can't recall specifically.  My guess is that the UCSC build we had didn't contain all of the chromosomes.  Around that time (August 2016) I was making a lot of reference sequence converters to handle the transition from one source to another (ex. UCSC-> Ensembl) and/or to a new build (mm9 -> mm10).

jasonwalker80:
I believe the reason for the new build is that the chromosome name for mitochondria was incorrectly 'MT' in the original 2011 version. In 2016, when making additional automated GMS build converters, I imported the correct FASTA files with 'M' as the chromosome name.

jasonwalker80:
Also, I should have mentioned that the md5 for all chromosomes is identical.  The only difference is the mitochondrial chromosome name.

chrisamiller:
Can we rename one of the two, or is that going to cause a huge headache?

jasonwalker80:
Build `f39a0212bf5c406cbafda080644b70de` has it's build_name renamed to `UCSC-mouse-buildmm9-2016`

jasonwalker80:
Can this issue be closed?

chrisamiller:
Yep - let's close it!  Original code should work fine now!

PR #1660, opened 2017-02-08
Remove remaining references to `/gsc/var/`.

tmooney:
This removes lingering references to `/gsc/var/` from the codebase.
 * Stops saving logs for `genome model build unarchive`.
 * Uses the `site_lock_dir` for testing network disk.
 * Removes `gmt sam r1` (from 2010).


susannasiebert:
+1

PR #1661, opened 2017-02-09
Fixing problems with rsID annotation code

kkrysiak:
Allows for the use of both older-style dbSNP VCFs that use the GMAF convention and newer ones with CAF convention. 

In both cases, we are now also ignoring indels. This code was never designed to annotate indels. 

The older versions also often provide the incorrect allele frequency because the minor allele is actually the reference allele and only 1 AF is provided in GMAF-style VCFs. It is definitely advisable to use the CAF versions. Just a warning in case anyone else ever sees this.

### DIFF
--- a/lib/perl/Genome/Model/Tools/Annotate/AddRsid.pm
+++ b/lib/perl/Genome/Model/Tools/Annotate/AddRsid.pm
@@ -79,37 +79,56 @@ sub execute {
             my @var_alleles = @tmp1;
             @var_alleles = @tmp2 if (scalar(@tmp2) > scalar(@tmp1));
 
+            my @dbSNPids = split_dbSNPBuildID($INFO);
+            my $RSid_var_allele;
+
             # Gets the list of allele frequencies if CAF nomenclature is used
             if($INFO =~ /(CAF=[0-9.,]+)/) {
                 $caf_string = ($INFO =~ /(CAF=[0-9.,]+)/)[0];
                 @af = split(/,/, $caf_string);
                 # The first allele frequency provided is the reference allele which we don't want so drop it
                 shift @af;
-            }
-
-            my @dbSNPids = split_dbSNPBuildID($INFO);
-
-            for (my $i = 0; $i < @var_alleles; $i++) {
-                # Reassign GMAF value to the CAF value, if one was found. Pull the appropriate var allele
-                if(scalar(@af) > 0){
-                    $GMAF = $af[$i];
-                    unless($GMAF){
-                        $self->warning_message("CAF undefined for allele, i: $i, CAF values undefined: $caf_string, non-reference AFs: @af");
+            
+                for (my $i = 0; $i < @var_alleles; $i++) {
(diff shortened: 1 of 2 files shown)


lib/perl/Genome/Model/Tools/Annotate/AddRsid.pm
```diff
@@ -79,39 +92,70 @@ sub execute {
...
-                    unless($GMAF){
-                        $self->warning_message("CAF undefined for allele, i: $i, CAF values undefined: $caf_string, non-reference AFs: @af");
-                    }
-                    if($GMAF eq '.'){
-                        $GMAF = "-";
-                    }   
+            if($file_format eq "CAF") {
+                # Gets the list of allele frequencies if CAF nomenclature is used
+                if($INFO =~ /(CAF=[0-9.,]+)/) {
+                    $caf_string = ($INFO =~ /(CAF=[0-9.,]+)/)[0];
```
tmooney: If you like, the capture and `if` could be combined like:
```perl
if (($caf_string) = $INFO =~ /(CAF=[0-9.,]+)/) {
```

lib/perl/Genome/Model/Tools/Annotate/AddRsid.pm
```diff
@@ -79,39 +92,70 @@ sub execute {
...
-                $caf_string = ($INFO =~ /(CAF=[0-9.,]+)/)[0];
-                @af = split(/,/, $caf_string);
-                # The first allele frequency provided is the reference allele which we don't want so drop it
-                shift @af;
+            # Script not designed to handle indels, ignores them if present
+            my $is_indel = 0;
+            for (@var_alleles) {
+                $is_indel = 1 if length($_) > 1;
+            }
+            if (length($ref) > 1 || $is_indel == 1) {
```
tmooney: The `== 1` here isn't strictly necessary.

One possible alternative is to eliminate `$is_indel` by labelling the while loop and changing the for loop like:
```perl
LINE: while (my $line = $vcf_fh->get_line) {
...
  for (@var_alleles) {
    next LINE if length($_) > 1;
  }
```

lib/perl/Genome/Model/Tools/Annotate/AddRsid.pm
kkrysiak: Thanks for the tips. Definitely appreciated.

kkrysiak:
Added a bug fix that @acoffman helped me find. Resulted in headers accidentally being printed twice on tool output.

jasonwalker80:
@kkrysiak are you waiting for more tests or code review before merging this?

kkrysiak:
Wasn't sure who should do the merge. Everything appears to be working properly and tests have passed.

lib/perl/Genome/Model/Tools/Annotate/AddRsid.pm
kkrysiak: Crap, except I forgot to implement these changes before merging the pull request.

PR #1668, opened 2017-02-20
AnP Disk Usage

jasonwalker80:
Initial revision of command to report disk usage. WIP.

### DIFF
--- a/lib/perl/Genome/Config/AnalysisProject/Command/DiskUsage.pm
+++ b/lib/perl/Genome/Config/AnalysisProject/Command/DiskUsage.pm
@@ -69,7 +69,7 @@ sub execute {
                 if ($total_bp) {
                     $bytes_per_base = ($total_kb * 1024) / $total_bp;
                 }
-                my %data = [
+                my %data = {
                     'analysis_project' => $anp->id,
                     'config_item' => $config_item->id,
                     'subclass_name' => $model_type,
@@ -78,7 +78,7 @@ sub execute {
                     'total_kb' => $total_kb,
                     'total_bp' => $total_bp,
                     'bytes_per_bp' => $bytes_per_base, 
-                ];
+                };
                 push @data, \%data; 
             }
         }
--- a/lib/perl/Genome/Config/AnalysisProject/Command/DiskUsage.pm
+++ b/lib/perl/Genome/Config/AnalysisProject/Command/DiskUsage.pm
@@ -22,6 +22,8 @@ sub help_detail {
     return <<EOHELP
 This command will take several minutes to run.
 
+Running builds will not be accounted for properly as the allocations are not complete.
+
 The output is a tab-delimited disk usage summary for each model type and configuration item per analysis project.
 
 The sum of the model type disk usage is not necessarily the total disk usage for an Analysis Project.  Results can be shared between multiple model types.  
--- a/lib/perl/Genome/Config/AnalysisProject/Command/DiskUsage.pm
+++ b/lib/perl/Genome/Config/AnalysisProject/Command/DiskUsage.pm
@@ -34,10 +34,9 @@ EOHELP
 sub execute {
     my $self = shift;
 
-    # Remove normal BAM from disk usgage for somatic
+    my @data;
     for my $anp ($self->analysis_projects) {
         for my $config_item ($anp->config_items) {
-            # TODO: Add model_type
             my %model_type_disk_usage;
             my $model_count = 0;
             my $build_count = 0;
@@ -45,7 +44,6 @@ sub execute {
                 my $subclass_name = $model->subclass_name;
                 $model_type_disk_usage{$subclass_name}{'model_count'}++;
                 for my $build ($model->builds) {
-                    #unless ($build->status eq 'Succeeded') { next; }
                     $model_type_disk_usage{$subclass_name}{'build_count'}++;
                     my @allocations = $build->disk_usage_allocations;
                     for my $allocation (@allocations) {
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Config/AnalysisProject/Command/DiskUsage.pm
```diff
@@ -0,0 +1,72 @@
...
+use feature qw(say);
+
+use Genome;
+
+class Genome::Config::AnalysisProject::Command::DiskUsage {
+    is => 'Command::V2',
+    has => [
+        analysis_projects => {
+            is                  => 'Genome::Config::AnalysisProject',
+            doc                 => 'the analysis projects to release for processing',
```
susannasiebert: This description seems incorrect for this tool.

lib/perl/Genome/Config/AnalysisProject/Command/DiskUsage.pm
```diff
@@ -0,0 +1,72 @@
...
+            for my $model_type (sort keys %model_type_disk_usage) {
+                my $total_kb = 0;
+                map { $total_kb += $_ } values(%{$model_type_disk_usage{$model_type}{'allocations'}});
+
+                my $total_bp = 0;
+                map { $total_bp += $_ } values (%{$model_type_disk_usage{$model_type}{'instrument_data'}});
+
+                my $bytes_per_base = ($total_kb * 1024) / $total_bp;
+
+                print $anp->id ."\t". $config_item->id ."\t". $model_type_disk_usage{$model_type}{'model_count'} ."\t". $model_type_disk_usage{$model_type}{'build_count'} ."\t". $total_kb ."\t". $total_bp ."\t". $bytes_per_base ."\n";
```
susannasiebert: A `join` might be nicer here than concatenating with `\t`.

jasonwalker80:
thanks @susannasiebert. I pushed some untested commits to address your suggestions.  I haven't tested these yet because a `git checkout` is unresponsive on the MGI filesystem.

lib/perl/Genome/Config/AnalysisProject/Command/DiskUsage.pm
```diff
@@ -0,0 +1,101 @@
...
+            my $model_count = 0;
+            my $build_count = 0;
+            for my $model ($anp->models) {
+                my $subclass_name = $model->subclass_name;
+                $model_type_disk_usage{$subclass_name}{'model_count'}++;
+                for my $build ($model->builds) {
+                    $model_type_disk_usage{$subclass_name}{'build_count'}++;
+                    my @allocations = $build->disk_usage_allocations;
+                    for my $allocation (@allocations) {
+                        $model_type_disk_usage{$subclass_name}{'allocations'}{$allocation->id} = $allocation->kilobytes_used;
```
tmooney: One corner case to be aware of: `kilobytes_used` will be `0` for any allocations that have never reallocated.  (For example, new allocations from running builds.)

susannasiebert:
Looks good. Great use of the SeparatedValueWriter. +1 again.

lib/perl/Genome/Config/AnalysisProject/Command/DiskUsage.pm
```diff
@@ -0,0 +1,107 @@
...
+                map { $total_kb += $_ } values(%{$model_type_disk_usage{$model_type}{'allocations'}});
+
+                my $total_bp = 0;
+                map { $total_bp += $_ } values (%{$model_type_disk_usage{$model_type}{'instrument_data'}});
+
+                my $bytes_per_base = 'na';
+                if ($total_bp) {
+                    $bytes_per_base = ($total_kb * 1024) / $total_bp;
+                }
+                my %data = {
```
tmooney: This still isn't quite right.

`%data` expects `( ... )`.

This could also be `my $data = { ... } ; push @data, $data;`

jasonwalker80: yeah, I've been working on this and when with your latter example a while ago.  Still trying to test and move my code changes around on the MGI filesystem....

lib/perl/Genome/Config/AnalysisProject/Command/DiskUsage.pm
```diff
@@ -0,0 +1,102 @@
...
+    my @lines;
+    for my $anp ($self->analysis_projects) {
+        for my $config_item ($anp->config_items) {
+            my %model_type_disk_usage;
+            for my $model ($config_item->models) {
+                my $subclass_name = $model->subclass_name;
+                $model_type_disk_usage{$subclass_name}{'model_count'}++;
+                for my $build ($model->builds) {
+                #my $build = $model->last_succeeded_build;
+                #if ($build) {
```
tmooney: This commented-out code should probably be removed (unless we want to make this an option for the command?)

jasonwalker80: Lines removed.  I think all builds is a more accurate approach that just the last_succeeded_build.

lib/perl/Genome/Config/AnalysisProject/Command/DiskUsage.pm
jasonwalker80: I added a note about this in the docs. I don't think we'd encounter Running builds often, but that depends on how this is used. The current use-case is to asses approximate usage from complete AnPs.

jasonwalker80:
Improved docs to highlight potential for error if the size of all models is summed. A few examples of how files can be shared across model types (even config for that matter).

lib/perl/Genome/Config/AnalysisProject/Command/DiskUsage.pm
```diff
@@ -0,0 +1,107 @@
...
+sub help_detail {
+    return <<EOHELP
+This command will take several minutes to run.
+
+Running builds will not be accounted for properly as the allocations are not complete.
+
+The output is a tab-delimited disk usage summary for each model type and configuration item per analysis project.
+
+The sum of the model type disk usage is not necessarily the total disk usage for an Analysis Project.  Results can be shared between multiple model types.  
+For instance, alignment files, BAMs, can be shared between SingleSampleGenotype, SomaticValidation and/or ReferenceAlignment depending on the processign profiles used.
```
tmooney: `processign` needs its final two letters switched.

lib/perl/Genome/Config/AnalysisProject/Command/DiskUsage.pm
```diff
@@ -0,0 +1,107 @@
...
+The same goes for SomaticVariation and SomaticValidation results. VCF and variant files can be shared between the two model types.
+
+The colums of the output file are:
```
tmooney: `colums` is missing its silent `n`.

lib/perl/Genome/Config/AnalysisProject/Command/DiskUsage.pm
```diff
@@ -0,0 +1,107 @@
...
+analysis project ID, configuration item ID, subclass name, model count, build count, total KB, total base pair, bytes per base
```
tmooney: How about `total base pairs` instead of `total base pair`?

jasonwalker80:
Spelling/grammar fixed.

PR #1680, opened 2017-03-10
Check for build and QC result instrument data does not apply to somatic build

dufeiyu:
For somatic builds, the instrument data count of one QC result is not equal to that of build.

### DIFF
--- a/lib/perl/Genome/Qc/Command/BuildMetrics.pm
+++ b/lib/perl/Genome/Qc/Command/BuildMetrics.pm
@@ -24,11 +24,6 @@ class Genome::Qc::Command::BuildMetrics {
             doc => 'The output file format.',
             valid_values => ['yaml','tsv'],
         },
-        somatic => {
-            is => 'Boolean',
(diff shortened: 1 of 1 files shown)


susannasiebert:
No, we just check with `unless ($self->$model->isa('Genome::Model::SomaticVariation') || $self->$model->isa('Genome::Model::SomaticValidation'))` when we assign the builds in ClinSeq. There is a `Genome::Model::Build::SomaticInterface` that is a parent class for SomaticVariation and SomaticValidation which might be useful in this specific context.

jasonwalker80:
@dufeiyu I'd suggest looking at the model type to infer the "somatic" status rather than relying on the user input.

dufeiyu:
@jasonwalker80  It makes sense. I will make the change.

dufeiyu:
CI, test this

dufeiyu:
CI, test this

dufeiyu:
CI, test this

PR #1687, opened 2017-03-24
Set SSG Test Names

jasonwalker80:
- a command to set test names on all SSG results (excluding aligner index, etc.)
- wait to abandon builds until we have a complete list and test names are set

### DIFF
--- a/lib/perl/Genome/Model/SingleSampleGenotype/Command/SetTestName.pm
+++ b/lib/perl/Genome/Model/SingleSampleGenotype/Command/SetTestName.pm
@@ -21,7 +21,7 @@ class Genome::Model::SingleSampleGenotype::Command::SetTestName {
         },
         abandon_builds => {
             is => 'Boolean',
-            doc => 'attempt to abando all of the builds used by these results',
+            doc => 'attempt to abandon all of the builds used by these results',
             default_value => 0,
         },
     },
--- a/lib/perl/Genome/Model/SingleSampleGenotype/Command/SetTestName.pm
+++ b/lib/perl/Genome/Model/SingleSampleGenotype/Command/SetTestName.pm
@@ -33,8 +33,8 @@ sub execute {
     my $result_classes_to_set_test_name = $self->_result_classes_to_set_test_name;
     my @results_to_set_test_name;
     for my $build ( $self->builds ) {
-        my @results = $build->results;
-        push @results_to_set_test_name, grep { $result_classes_to_set_test_name->{$_->class} } @results;
+        my @disk_usage_results = $build->disk_usage_results;
+        push @results_to_set_test_name, @disk_usage_results;
     }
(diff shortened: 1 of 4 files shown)


lib/perl/Genome/Model/SingleSampleGenotype/Command/SetTestName.pm
```diff
@@ -0,0 +1,67 @@
...
+            is      => "Genome::Model::Build::SingleSampleGenotype",
+            is_many => 1,
+        },
+        new_test_name => {
+            is  => 'String',
+            doc => 'The new test name string to set on all results'
+        },
+        abandon_builds => {
+            is => 'Boolean',
+            doc => 'attempt to abando all of the builds used by these results',
```
tmooney: `abandon` is short its final `n`.

lib/perl/Genome/Model/SingleSampleGenotype/Command/SetTestName.pm
```diff
@@ -0,0 +1,67 @@
...
+};
+
+sub execute {
+    my $self = shift;
+
+    my $result_classes_to_set_test_name = $self->_result_classes_to_set_test_name;
+    my @results_to_set_test_name;
+    for my $build ( $self->builds ) {
+        my @results = $build->results;
+        push @results_to_set_test_name, grep { $result_classes_to_set_test_name->{$_->class} } @results;
```
tmooney: The database/UR can do this restriction for us, e.g.:
```perl
$build->results(subclass_name => [
  'Genome::Qc::Result',
  'Genome::InstrumentData::AlignmentResult::Speedseq',
  'Genome::InstrumentData::AlignmentResult::Merged::Speedseq',
  'Genome::Model::SingleSampleGenotype::Result::HaplotypeCaller'
]);
```

jasonwalker80:
@tmooney will the latest commit work for "whitelisting" result classes?

lib/perl/Genome/SoftwareResult/Command/SetTestName.pm
```diff
@@ -46,21 +49,29 @@ sub execute {
...
         }
     }
+    if ($self->abandon_builds) {
+        my @ids = keys %builds;
+        my @builds = values %builds;
+        $self->status_message('Abandoning builds: '. join(',',@ids));
+        my $abandon_cmd = Genome::Model::Build::Command::Abandon->create(
+            builds => \@builds,
+            header_text => 'Build Abandoned - SR Test Name',
+            body_text => 'Software result '. $software_result->id .' has new test name '. $self->new_test_name,
```
tmooney: With this pulled out of the loop `$software_result` is no longer available.

jasonwalker80:
Yes, the intent is to eventually define `_disk_usage_result_subclass_names` for all model types and `disk_usage_allocations` could then rely on that instead. I have a feeling there will be issues with backward compatibility for old Workflow/Flow builds and such, but at some point we can ignore that.

PR #1697, opened 2017-04-06
Stop swallowing errors in Model create.

tmooney:
The `Scope::Guard` was preventing the message from `Carp::confess` from making it to the user.

### DIFF
--- a/lib/perl/Genome/Model.pm
+++ b/lib/perl/Genome/Model.pm
@@ -477,7 +477,7 @@ sub create {
     }
 
     my $tx = UR::Context::Transaction->begin(commit_validator => sub { 1 });
-    my $guard = Scope::Guard->new(sub { $tx->rollback });
+    my $guard = Scope::Guard->new(sub { local $@; $tx->rollback });
 
     my $self = $class->SUPER::create($bx);
     unless ($self) {


chrisamiller:
+1

brummett:
Which confess() was getting swallowed?

If it's something inside the guard/rollback, then yeah, this is probably the right change.  The guard is implemented with a destructor, and exceptions in destructors aren't fatal.

This one https://github.com/tmooney/genome/blob/04334e95de4e2d75b5884d91ae636a1662e07292/lib/perl/Genome/Model.pm#L519
could get fixed by better manipulation of `$!` while keeping the guaranteed rollback the Guard gets you.

tmooney:
@brummett, It was exactly the `confess` you mentioned that's at issue. (In the motivating case, `$reason` is defined, so we weren't ever looking at `$@`.)
```perl
Carp::confess $reason || $@
```
doesn't seem right, anyway, so I should probably change that, too.

tmooney:
I pushed an update that keeps the Guard, but localizes `$@` inside of it.  The "real issue" was that after the code `die`s inside confess, the handler was overwriting `$@` before `Command::Common` caught it in its `eval`.

(This will almost certainly fail tests without #1706.)

PR #1700, opened 2017-04-24
Site: Avoid using `/vol/` paths as canonical.

tmooney:
This is an attempt to keep `/vol/` paths out of our local database by un-resolving the `/gscmnt/` symlinks.


tmooney:
This will require #1701 to pass tests.

PR #1698, opened 2017-04-19
WIP: Move AnP allocations for a build

jasonwalker80:
This is an initial attempt at developing a command for #1691 

The current version does nothing more than compare the list of class names between getting ALL associated allocations and querying for specific result classes directly.

The output of a test run shows several result classes are missed by the query:
```
genome model build move --analysis-project=081827f844f84b10992b316c878965ed aee9c92c0d2f4d02a2c4a29037eff25b
Using libraries at /gscmnt/gc2764/cad/jwalker/git/genome/lib/perl
'analysis_project' may require verification...
Resolving parameter 'builds' from command argument 'aee9c92c0d2f4d02a2c4a29037eff25b'... found 1
Resolving parameter 'analysis_project' from command argument '081827f844f84b10992b316c878965ed'... found 1
move allocation(s) for Analysis Project Aurora US - Miller Mardis (081827f844f84b10992b316c878965ed) at the request of Jason Walker.
Moving allocations for 1 builds

Working on build aee9c92c0d2f4d02a2c4a29037eff25b (1 of 1)
Skipping the following classes not defined as disk_usage_results:
Genome::SoftwareResult::ImportedFile
Genome::Model::Build::SomaticVariation
Genome::Model::Build::ImportedVariationList
Genome::Model::Build::ReferenceSequence::ConvertedBedResult
Genome::InstrumentData::AlignmentResult::Merged::BamQc
Genome::FeatureList
Genome::Model::Tools::DetectVariants2::Result::Combine::UnionSv
Genome::InstrumentData::AlignmentResult::BwamemStream
Genome::Model::Build::ReferenceSequence::AlignerIndex
Genome::Model::Build::ReferenceSequence::PerChromosomeFastas
Genome::Db::Ensembl::AnnotationStructures
Genome::Model::Tools::DetectVariants2::Result::Manual
Genome::Model::Build::ReferenceAlignment::Solexa
Genome::Model::Build::ImportedAnnotation
Genome::InstrumentData::AlignmentResult::Merged
Genome::Model::Build::ImportedReferenceSequence
```

PR #1702, opened 2017-04-25
Command to move "all" allocations for a build.

tmooney:
It might work for SSG builds... other types probably require more tweaks to their classes first.

This is an adaptation of #1698.

### DIFF
--- a/lib/perl/Genome/Model/Build/Command/MoveAllocations.pm
+++ b/lib/perl/Genome/Model/Build/Command/MoveAllocations.pm
@@ -63,6 +63,7 @@ sub execute {
         my $class_set_by_results = Set::Scalar->new(sort keys %result_classes);
 
         my $diff_associated_allocations = $class_set_by_associated_allocations->difference($class_set_by_results);
+        $diff_associated_allocations->delete($build->_disk_usage_result_subclass_names); #associated_disk_allocations can pull in allocations for classes we handle, but that are not results for this build. Don't report these classes as skipped.
         if (my @m = $diff_associated_allocations->members) {
             $self->status_message('Skipping the following classes not defined as disk_usage_results: '. "\n". join("\n", grep { $_ ne $build->class } @m));
         }
--- a/lib/perl/Genome/Model/Build/Command/MoveAllocations.pm
+++ b/lib/perl/Genome/Model/Build/Command/MoveAllocations.pm
@@ -77,6 +77,9 @@ sub execute {
 
         my $disk_group = $self->disk_group // $self->_resolve_disk_group_for_build($build);
 
+        #Don't try to move things that are already in the desired location!
+        @allocations_to_move = grep { $_->disk_group_name ne $disk_group->disk_group_name } @allocations_to_move;
+
         local $ENV{UR_NO_REQUIRE_USER_VERIFY} = ($ENV{UR_NO_REQUIRE_USER_VERIFY} // 1);
         my $move_cmd = Genome::Disk::Command::Allocation::Move->create(
             allocations => \@allocations_to_move,
--- a/lib/perl/Genome/Model/Build/Command/MoveAllocations.pm
+++ b/lib/perl/Genome/Model/Build/Command/MoveAllocations.pm
@@ -87,6 +87,8 @@ sub execute {
         }
         $move_cmd->execute or
             $self->fatal_message('Failed to move allocations for build %s.', $build->__display_name__);
+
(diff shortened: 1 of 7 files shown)


lib/perl/Genome/Model/Build/Command/MoveAllocations.pm
```diff
@@ -77,6 +77,9 @@ sub execute {
 
         my $disk_group = $self->disk_group // $self->_resolve_disk_group_for_build($build);
 
+        #Don't try to move things that are already in the desired location!
+        @allocations_to_move = grep { $_->disk_group_name eq $disk_group->name } @allocations_to_move;
```
jasonwalker80: $disk_group->disk_group_name

lib/perl/Genome/Model/Build/Command/MoveAllocations.pm
```diff
@@ -0,0 +1,117 @@
...
+            $self->fatal_message('Found additional classes: '. "\n". join("\n", $diff_results->members));
+        }
+
+        my @allocations_to_move = grep { $result_classes{$_->owner_class_name} } @associated_allocations;
+        push @allocations_to_move, $build->disk_allocation;
+
+        my $disk_group = $self->disk_group // $self->_resolve_disk_group_for_build($build);
+
+        #Don't try to move things that are already in the desired location!
+        @allocations_to_move = grep { $_->disk_group_name eq $disk_group->disk_group_name } @allocations_to_move;
```
jasonwalker80: should this be `ne` instead of `eq`?

tmooney: That would make a lot more sense, woudln't it!

lib/perl/Genome/Model/Build/SomaticVariation.pm
```diff
@@ -484,4 +484,29 @@ sub build_for_tumor_bam {
...
+        'Genome::InstrumentData::AlignmentResult::BwamemStream',
+        'Genome::InstrumentData::AlignmentResult::Merged',
+        'Genome::InstrumentData::AlignmentResult::Merged::BamQc',
+        'Genome::Model::Tools::DetectVariants2::Classify::Loh',
+        'Genome::Model::Tools::DetectVariants2::Classify::PreviouslyDiscovered',
+        'Genome::Model::Tools::DetectVariants2::Classify::Tier',
+        'Genome::Model::Tools::DetectVariants2::Result',
+        'Genome::Model::Tools::DetectVariants2::Result::Combine::IntersectSnv',
+        'Genome::Model::Tools::DetectVariants2::Result::Combine::LqUnion',
+        'Genome::Model::Tools::DetectVariants2::Result::Combine::UnionSv',
```
jasonwalker80: I see this in the list of classes skipped:
`Genome::Model::Tools::DetectVariants2::Result::Combine::UnionSv`

lib/perl/Genome/Model/Build/SomaticVariation.pm
```diff
@@ -484,4 +484,29 @@ sub build_for_tumor_bam {
     return $self->tumor_build;
 }
 
+sub _disk_usage_result_subclass_names {
+    my $self = shift;
+
+    my @disk_usage_result_classes = (
+        'Genome::InstrumentData::AlignmentResult::BwamemStream',
```
jasonwalker80: This also shows up in the list of skipped class names:
`Genome::InstrumentData::AlignmentResult::BwamemStream`

jasonwalker80: Ignore this. alignments should be in RefAlign. We should however remove it from the list then.

lib/perl/Genome/Model/Build/SomaticVariation.pm
tmooney: Looks like `associated_disk_allocations` is overly eager.  It finds results that don't really belong to the build in question, but are in a chain with it.

This may lead to false positive listings when they are properly excluded from the `disk_usage_results`.

chrisamiller:
I think the Result::Manual class is for variant lists imported with `genome model somatic-validation manual-result`.  Those are not really used any more, but there may be some old-ass models hanging around that do have it.

lib/perl/Genome/Model/Build/SomaticVariation.pm
tmooney: I've added a change to not display class names that are in the list of handled classes.

PR #1703, opened 2017-04-25
Don't use docker for archive processes!

tmooney:
Now that there's a default docker specification, we have to explicitly exclude it here.


tmooney:
(This will require #1706 to pass tests.)

PR #1706, opened 2017-05-04
Fixes for more tests due to external changes.

tmooney:
* Control the choice of `lsb_sub_additional` in Ptero tests so that result is not dependent on configuration environment.
* Remove a Polymutt filter test. (The tool itself will be removed eventually...)


jasonwalker80:
+2

PR #1710, opened 2017-05-17
Draft CWL Pipeline type.

tmooney:
Here's what one version of a CWL model type could look like.  It needs a few more commands (convenient ways to define PPs and models) to be useful, and the YAML generation needs some fleshing out.  But this combined with relevant environment variables could theoretically be used to run a CWL pipeline tracked as a build within the GMS.

### DIFF
--- a/lib/perl/Genome/Model/CwlPipeline.pm
+++ b/lib/perl/Genome/Model/CwlPipeline.pm
@@ -9,6 +9,10 @@ class Genome::Model::CwlPipeline {
         main_workflow_file => {
             is => 'Text',
         },
+        primary_docker_image => {
+            is => 'Text',
+            doc => 'docker image for the main toil worker jobs',
+        },
     },
 };
 
--- a/lib/perl/Genome/Model/CwlPipeline.pm
+++ b/lib/perl/Genome/Model/CwlPipeline.pm
@@ -3,6 +3,9 @@ package Genome::Model::CwlPipeline;
 use strict;
 use warnings;
 
+use feature qw(switch);
+use Genome;
+
 class Genome::Model::CwlPipeline {
     is => 'Genome::Model',
     has_param => {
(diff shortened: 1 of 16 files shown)


susannasiebert:
Nice work! With this implementation what were your thought on how extend it in order to have different types of pipelines? Would those be child classes that inherit from this?

tmooney:
In this draft version, each different pipeline would be a different ProcessingProfile (with a different `main_workflow_file` param).  I'd prefer to avoid making different subclasses to handle each different pipeline if we can get away with it.

It's then an open question how to best handle inputs.  This draft version will accept whatever inputs are on the model and generate the YAML (assuming everything is a simple value!).
* It could possibly parse the CWL to find out what inputs to expect?
* It could be made to handle different input classes in standard ways (for example, `UR::Value::FilePath` then becomes a file-type input and `Genome::InstrumentData::Solexa` becomes a file-type input to the BAM path)?
* It could possibly be added to the ProcessingProfile params what value needs to be taken from each input to generate the YAML as well as the filetypes?

jasonwalker80:
This is great! A few *general* comments before I actually dive in for a full review:

1) Support for WDL workflows as well as CWL. I'm not sure if one or both will persist, but I could see the advantage of supporting both as long as the backend platform supports either or both. In my head, I'm thinking we'd pull either workflow description language from DockStore and execute one or the other on a supported backend platform.

2) Support for multiple backends, ex. toil, cromwell, arvados, etc. For example could `Genome::Model::CwlPipeline::Command::Run;` have some common interface and a specific toil implementation? 

3) If we use Genome::Workflow are we essentially still relying on PTero to run a one step workflow? I'm not suggesting we shouldn't in an initial implementation, but we should consider removing that dependency long term.

None of these comments are really specific to this initial PR, but general thoughts on the topic. Hopefully I'll have some additional time to review this soon with constructive feedback on the actual code.

tmooney:
Re 3.: we're not specifically dependent on PTero, as the `inline` WorkflowBuilder backend is also available.  Practically speaking, though, unless we change BQM queued builds would end up using Ptero to bsub the main toil process and take care of updating the build status as necessary on success/failure.

Re 2.: If in the future we do want to support multiple workflow technologies, it might be nice for there to be something WorkflowBuilder-like with backends corresponding to `toil` and others!  That's hopefully all something that could change behind the scenes here.

Re 1.: It might make sense for WDL pipelines to be a separate subclass since they have different setup and requirements from CWL?  I'm not sure how much of what's here would be shared by both.

tmooney:
To test this out with the existing model:
* Check out a copy of the PR to a network location.
* Set the `lsb_sub_additional` config to `docker(registry.gsc.wustl.edu/tmooney/genome_perl_environment-test)`
* Run `genome model build start e125c9b8dd114938bdd8d6be0a4156a8`.  This can be done from the normal docker image, but the jobs it submits must use the test image, which includes `cwltool` and `toil`.

To define a new model of this PP to perform different RnaSeq analysis:
* `genome model define cwl-pipeline --processing-profile 9e62edbe0b564dcab132f3a2692ddaca ...`
* Manually assign inputs.  (For this workflow, inputs named `instrument_data` should point to `Genome::InstrumentData` objects and one named `reference_index` should be the path for a suitable reference index (e.g. `/gscmnt/gc2764/cad/HCC1395/rna_seq/GRCh37/refseq/GRCh37_tran`)  This currently may involve a `re.pl` to create the `Genome::Model::Input` objects.

To create a new PP to run a different workflow:
* `genome processing-profile create cwl-pipeline --cwl-directory [path] --main-workflow-file [filename.cwl] ...`
* The directory should have all the CWL files as well as a `process_inputs.pl` that takes a build ID and produces a YAML file suitable for running the pipeline.  (e.g. `/gscmnt/gc2560/core/processing-profile/cwl-pipeline/9e62edbe0b564dcab132f3a2692ddaca/process_inputs.pl`)  Please don't make this script depend on anything other than the build or its inputs to determine the YAML :smile:

Some of these lists definitely suggest future development, but as a Proof of Concept it seems to work!

tmooney:
Recent commits add:
* A CLI for adding/listing/removing inputs from CwlPipeline models.
* A means of specifying arbitrary inputs for CwlPipeline models in an AnP configuration file (via the `input_data:` key).  An example AnP configuration on which this was tested is in project `392239153b234cc3a2c457917efe411b`.
* Some tests :smile:

tmooney:
CI, test this

jasonwalker80:
+2

PR #1719, opened 2017-05-25
First version of a merged AlignmentResult importer.

tmooney:
Further development will be required for this to be usable in existing pipelines.  (At the least, we'll need a way to short-circuit alignment to use these instead of creating the dispatcher workflow.)

### DIFF
--- a/lib/perl/Genome/InstrumentData/AlignmentResult/Merged/External.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult/Merged/External.pm
@@ -1,0 +1,127 @@
+package Genome::InstrumentData::AlignmentResult::Merged::External;
+
+use strict;
+use warnings;
+
+use Genome;
+
+use File::Basename qw();
+use File::Spec;
+
+
+class Genome::InstrumentData::AlignmentResult::Merged::External {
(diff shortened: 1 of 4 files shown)


lib/perl/Genome/InstrumentData/AlignmentResult/Merged/External.pm
```diff
@@ -0,0 +1,133 @@
...
+            is => 'Text',
+            doc => 'brief information about the alignment data',
+        },
+        filetype => {
+            is => 'Text',
+            doc => 'the type of alignment (expressed as file extension)',
+        },
+    ],
+    has_constant => [
+        bam_path => {
```
jasonwalker80: For anyone else reviewing, there was already discussion to keep using `bam_path` and later refactor to something like `alignment_path` and support both BAM and CRAM filetypes.

tmooney:
Currently, our AnP config won't do anything with these, since they're not themselves instrument data.  There is an optional parameter for linking to the applicable instrument data used to generate the alignments, but there's not another way to specify attributes in this version.

If we want automatic processing of alignments (as opposed to of instrument data), that's a new class of AnP assignment, so we should think about what that means and what we would want to have happen.

PR #1720, opened 2017-05-25
Support for Shortcutting the Alignment Process in SomaticValidation

tmooney:
(Combined with #1719 this should allow SomaticValidation to run on externally-produced alignments.)

Issues:
* Should this really just ignore the `alignment_strategy` and assigned `instrument_data` when pre-aligned data are available?  Currently that's a "yes".
* Some of the validation realignment steps operate on instrument data, so they'd crash or use any assigned instrument data while ignoring the pre-aligned data.
* This hasn't been tested by running a build yet.

### DIFF
--- a/lib/perl/Genome/Model/SomaticValidation/Command/AlignReads.pm
+++ b/lib/perl/Genome/Model/SomaticValidation/Command/AlignReads.pm
@@ -51,8 +51,18 @@ class Genome::Model::SomaticValidation::Command::AlignReads {
 
 sub sub_command_category { 'pipeline steps' }
 
+sub shortcut {
+    my $self = shift;
+
+    return $self->_find_existing_alignments;
+}
+
+
 sub execute {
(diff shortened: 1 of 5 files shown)


lib/perl/Genome/Model/SomaticValidation/Command/AlignReads.pm
```diff
@@ -105,5 +109,66 @@ sub execute {
     return 1;
 }
 
+sub _find_existing_alignments {
```
brummett: I'd suggest naming this something like "_found_existing_alignments" or "_has_existing_alignments", since it's returning true/false, not the actual alignments it found.

tmooney: I changed this to return the alignments instead.  I'd like to avoid an innocuous boolean-sounding name since this has a lot of side-effects!

PR #1723, opened 2017-06-01
Specify a job group to limit concurrent MethRatio processes.

tmooney:
Running too many of these is implicated in performance problems in the cluster.


susannasiebert:
+1 once tests pass.

PR #1724, opened 2017-06-14
Support the notion of alternative filetypes for alignment files.

tmooney:
Nothing uses this...yet.  It creates feature-parity with the proposed filetype field for external merged results in #1719.


chrisamiller:
+1

PR #1727, opened 2017-06-22
Only attempt to move allocations that are still active.

jasonwalker80:
@chrisamiller this should resolve: https://jira.gsc.wustl.edu/browse/CIS-54


jasonwalker80:
A live test worked:
`genome model build move-allocations 119857056`

Merging early.

PR #1729, opened 2017-06-22
clarify docs for disk group when moving allocations.

jasonwalker80:


chrisamiller:
+1 - appreciated!

lib/perl/Genome/Model/Build/Command/MoveAllocations.pm
```diff
@@ -19,7 +19,7 @@ class Genome::Model::Build::Command::MoveAllocations {
         disk_group => {
             is                  => 'Genome::Disk::Group',
             is_optional         => 1,
-            doc => 'A specific disk group to move all data to (defaults to AnP configured location)',
+            doc => 'A specific disk group to move all data to. Defaults to AnP configured location unless disk_group_models is already defined in the users environment',
```
tmooney: It might be nice if this were rendered as "user's".

jasonwalker80: will it work to escape it? `user\'s`?

jasonwalker80: yep, that works

tmooney: It seems to work in a quick test!  Could also use `q{}` or the like instead.

PR #1735, opened 2017-07-19
First version of BAM <=> CRAM conversion for Merged AlignmentResults

tmooney:
Here are a couple commands for converting merged alignment results between BAM and CRAM format.  This still needs to be tested more before being widely used!

Potential Shortcomings:
* There's no automatic CRAM => BAM reconversion for converted things if the result is called in a situation that requires BAM.
* It might be nice to be more thorough with verifying the transformation.  This could incorporate a flagstat comparison at the cost of spinning up a second/third LSF job.

### DIFF
--- a/lib/perl/Genome/InstrumentData/AlignmentResult/Command/BamToCram.pm
+++ b/lib/perl/Genome/InstrumentData/AlignmentResult/Command/BamToCram.pm
@@ -1,0 +1,45 @@
+package Genome::InstrumentData::AlignmentResult::Command::BamToCram;
+
+use strict;
+use warnings;
+
+use Genome;
+
+class Genome::InstrumentData::AlignmentResult::Command::BamToCram {
+    is => 'Genome::InstrumentData::AlignmentResult::Command::ConvertBase',
+};
+
+sub shortcut {
+    my $self = shift;
+
+    return $self->_is_currently_cram;
+}
+
+sub execute {
+    my $self = shift;
+
+    return 1 if $self->_is_currently_cram;
+
+    unless($self->_is_currently_bam) {
+        $self->fatal_message("This result cannot be converted to CRAM.");
+    }
+
+    my $result = $self->alignment_result;
+    my $bam_file = $result->bam_file;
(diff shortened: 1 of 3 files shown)


lib/perl/Genome/InstrumentData/AlignmentResult/Command/CramToBam.pm
```diff
@@ -0,0 +1,47 @@
...
+
+sub execute {
+    my $self = shift;
+
+    my $guard = $self->_lock()->unlock_guard;
+
+    return 1 if $self->_is_currently_bam;
+
+    unless($self->_is_currently_cram) {
+        $self->fatal_message("This result cannot be converted to BAM.");
```
chrisamiller: I'd suggest: "This result is already in BAM format!", which seems a little more transparent to the end user.

lib/perl/Genome/InstrumentData/AlignmentResult/Command/BamToCram.pm
```diff
@@ -0,0 +1,47 @@
...
+
+sub execute {
+    my $self = shift;
+
+    my $guard = $self->_lock()->unlock_guard;
+
+    return 1 if $self->_is_currently_cram;
+
+    unless($self->_is_currently_bam) {
+        $self->fatal_message("This result cannot be converted to CRAM.");
```
chrisamiller: I'd suggest: "This result is already in CRAM format!", which seems a little more transparent to the end user.

lib/perl/Genome/InstrumentData/AlignmentResult/Command/CramToBam.pm
tmooney: If it's already BAM we would've hit the "return 1" up above.  This error case protects against a mysterious other format if one happens to exist in the future.

chrisamiller: ah, I see now - comments retracted!

PR #1741, opened 2017-08-14
Move new BAM <=> CRAM conversion commands for command-line use.

tmooney:
There are two similar command directories, but only one of them is included when running `genome`.


chrisamiller:
+1

PR #1744, opened 2017-08-16
Manual Snapshot Promotion

tmooney:
There are a few failing model tests, which prevent the automated snapshot approval from occurring. 
These failures are understood and believed to not be related to the changes contained in this code, so this PR is for a manual approval.


jasonwalker80:
💯

PR #1756, opened 2017-08-24
Import Genotype Data, AnP links, and Sample links for Work Order.

tmooney:
The first version of this command worked to import instrument data assuming the synchronization cron was otherwise running to bring across other associations.  This adds features to make the data import more complete independent of synching.


tmooney:
A few edits have been made after testing this on a work order.

PR #1761, opened 2017-08-31
Final Pass VCF: Ask the sample for the correct sample name for the VCF file.

tmooney:
This tool was failing to find data for the sample if the sample had a different `name_in_vcf` from the sample name itself (e.g., a TCGA name was used instead).


tmooney:
It might be preferable to use #1762 instead?

chrisamiller:
From my shallow understanding, I think #1762 makes more sense.

tmooney:
Yeah, I'm inclined to agree. I'll close this one.

PR #1762, opened 2017-08-31
Pass the corrected sample name to the FinalPassOnly tool.

tmooney:
This is an alternate to #1761.  It has more parallels to the `_create_bed_from_vcf` in the dispatcher. It also allows the final pass tool to remain unaware of the complexities of samples, since it is only passed a string.


chrisamiller:
+1

PR #1766, opened 2017-09-07
More Efficient Synch; Temporarily Stop LimsProjectInstrumentData Synch

tmooney:
This PR includes a few changes to stop loading as much unnecessary data during synch.  Additionally, it stops synching `LimsProjectInstrumentData` for now, as this table cannot be examined in a reasonable timeframe for the operation.


tmooney:
This requires #1756 to succeed at synch.

PR #1771, opened 2017-09-20
Use new canonical DGIdb.org domain.

tmooney:
The old URL is returning 308 redirects, which this tool reports as errors.


tmooney:
This requires griffithlab/dgi-db#293 in order to pass tests.

jasonwalker80:
it looks like #293 was merged? does this need to run tests again?

tmooney:
griffithlab/dgi-db#297 was merged to `staging`, but griffithlab/dgi-db#293 remains open.  The production version hasn't been fixed yet.

PR #1774, opened 2017-10-04
(Re)index the converted alignment file.

### LINKED ISSUE #1773: Handle index during BAM/CRAM conversion
The index file is not useful after going round trip BAM->CRAM->BAM. The BAM file must be reindexed. Appropriately handle or fix the index creation to make this easier on the end user. For now, indexing can be handled by running `samtools index`.

In fact, there is a `crai` index we should make that allows for viewing of the CRAM directly in IGV 3 or greater.

tmooney:
The conversion process may not produce an identical file, so we can't trust the existing index to still be valid.

Closes #1773.

### DIFF
--- a/lib/perl/Genome/InstrumentData/Command/AlignmentResult/Convert/Base.pm
+++ b/lib/perl/Genome/InstrumentData/Command/AlignmentResult/Convert/Base.pm
@@ -92,6 +92,11 @@ sub _run_conversion {
     );
 
     $self->_reindex($destination_file);
+
+    my @alloc = $self->alignment_result->disk_allocations;
+    for (@alloc) {
+        $_->reallocate;
+    }
 }
 
 sub _reindex {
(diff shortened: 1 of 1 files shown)


tmooney:
It seemed to work when I tried it :smile:

The docs don't mention any special options for CRAM <http://www.htslib.org/doc/samtools.html>:
```
index
samtools index [-bc] [-m INT] aln.bam|aln.cram [out.index]

Index a coordinate-sorted BAM or CRAM file for fast random access. (Note that this does not work with SAM files even if they are bgzip compressed — to index such files, use tabix(1) instead.)

This index is needed when region arguments are used to limit samtools view and similar commands to particular regions of interest.

If an output filename is given, the index file will be written to out.index. Otherwise, for a CRAM file aln.cram, index file aln.cram.crai will be created; for a BAM file aln.bam, either aln.bam.bai or aln.bam.csi will be created, depending on the index format selected.

Options:

-b
Create a BAI index. This is currently the default when no format options are used.

-c
Create a CSI index. By default, the minimum interval size for the index is 2^14, which is the same as the fixed value used by the BAI format.

-m INT
Create a CSI index, with a minimum interval size of 2^INT.
```

PR #1777, opened 2017-10-09
First version of a gnomad file importer.

tmooney:
This version doesn't include a command interface or accessors for the file, but it does successfully download the files and store them in an allocation.  (Current example is `/gscmnt/gc2560/core/model_data/genome-db-ensembl-gnomad/2dd4b53431674786b760adad60a29273`.)


kkrysiak:
+1

PR #1781, opened 2017-10-26
adding some feedback to gmb queue

chrisamiller:


lib/perl/Genome/Model/Build/Command/Queue.pm
```diff
@@ -50,7 +50,7 @@ sub execute {
 
     my @models = $self->models;
     map($_->build_requested(1, $self->reason), @models);
-
+    $self->status_message("Successfully queued " . scalar(@models) . " models.");
```
tmooney: It'd be nice to have `model` vs `models` in the status message as appropriate.  (`Lingua::EN::Inflect` could be used for this purpose.)

PR #1783, opened 2017-10-31
Fill in `result_users` hash before getting/creating alignments

tmooney:
This codepath is only used by processing profiles that perform Tophat 1 alignments.


chrisamiller:
+2 :)

PR #1784, opened 2017-11-01
Add a mechanism for specifying arbitrary object inputs in CwlPipeline.

tmooney:
The motivation here is to be able to use database objects as inputs instead of hardcoding file paths.


tmooney:
I should add a testcase for this before merging it.

apipe-tester:

apipe-tester:

PR #1787, opened 2017-11-16
Point to test inputs for test FASTA file.

tmooney:
We don't really use it in this test, anyway, but at least this location exists!


apipe-tester:

apipe-tester:

PR #1786, opened 2017-11-16
Remove GMT::Predictor::Ber tree.

tmooney:
This depends on a now-removed directory.

PR #1788, opened 2017-11-27
Fix setting of purge reason for `Genome::Disk::Allocation->purge`

tmooney:
The object expects to receive a string, not a hash.

(Good thing we didn't purge a half million results before noticing this issue or anything like that 😉)


tmooney:
There's a good chance this won't pass tests without #1786 and #1787.

apipe-tester:

apipe-tester:

PR #1789, opened 2017-11-28
Hackily replace deleted test data with some that's still around.

tmooney:
This would still be better without relying on real data!  This should hopefully be sufficient to get tests to pass again, though!


apipe-tester:

apipe-tester:

PR #1793, opened 2017-12-04
FASTQ Importer with less RAM

jasonwalker80:
Remove unarchive of FASTQ to avoid storing entire contents of the file in RMA.

Use the latest version of Picard for FastqToSam to ensure reading gzipped files works correctly. If I recall correctly older versions of Picard had issues with gzipped inputs. I'm sure this could be standardized for all Picard commands, but for now this seems to work.

### DIFF
--- a/lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/FastqsToBam.pm
+++ b/lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/FastqsToBam.pm
@@ -6,8 +6,6 @@ use warnings;
 use Genome;
 
 use Data::Dumper 'Dumper';
-use Archive::Extract
-$Archive::Extract::PREFER_BIN = 1;
 require File::Basename;
 require File::Spec;
 require Genome::Utility::Text;
@@ -48,9 +46,6 @@ sub execute {
     my $self = shift;
     $self->debug_message('Fastqs to bam...');
 
-    my $unarchive_if_necessary = $self->_unarchive_fastqs_if_necessary;
-    return if not $unarchive_if_necessary;
-
     my $get_fastq_read_counts = $self->_get_fastq_read_counts;
     return if not $get_fastq_read_counts;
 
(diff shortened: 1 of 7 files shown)


apipe-tester:

apipe-tester:

apipe-tester:

lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/FastqsToBam.pm
```diff
@@ -127,29 +86,25 @@ sub _fastqs_to_bam {
...
-        fastq => $fastqs[0],
-        output => $output_bam_path,
-        quality_format => 'Standard',
-        sample_name => $self->library->sample->name,
-        library_name => $self->library->name,
-        read_group_name => UR::Object::Type->autogenerate_new_object_id_uuid,
-        use_version => '1.113',
-    );
+
+    my $cmd = '/gapp/x64linux/opt/java/jre/jre1.8.0_31/bin/java -Xmx16g -jar /gscmnt/gc2560/core/software/picard/2.15.0/picard.jar FastqToSam O='. $output_bam_path .' SM='. $self->library->sample->name .' LB='. $self->library->name .' RG='. UR::Object::Type->autogenerate_new_object_id_uuid .' F1='. $fastqs[0];
```
tmooney: Do we care about continuing to support `gmt picard` in general?  `GMT::Picard::Base` supports specifying a Java interpreter and, if told about this new Picard version, would probably work to continue being a viable wrapper for `FastqToSam`.  We'd have to update this particular GMT to the `Base` version, though.

apipe-tester:

lib/perl/Genome/Sys.pm
```diff
@@ -871,7 +871,12 @@ sub rsync_directory {
 
 sub line_count {
     my ($self, $path) = @_;
-    my ($line_count) = qx(wc -l $path) =~ /^(\d+)/;
+    my $line_count;
+    if ( $path !~ /\.gz$/ ) {
```
tmooney: How about using `file_is_gzipped` here in `Genome::Sys`?

lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/FastqsToBam.pm
jasonwalker80: I envision the tool tree as-is going away and being replaced by a dockerized version of tools/images. The transition of these changes can take considerable time so maintaining the picard tool tree is still required.

However, I think the other issue was with the JAR files. After 1.123 I think Picard switched from tool specific JARs to a single `picard.jar` where you define the sub-command passed as an arg to the main JAR.  It was always my understanding this was the limitation to updating Picard in the GMT tree. We'd have to rewrite how the JAR and subcommand names are defined.

tmooney: We do at least try to handle the single-JAR case (as seen [here](https://github.com/genome/genome/blob/68b99044894435f7d08d9621e4d8d606f23201d1/lib/perl/Genome/Model/Tools/Picard.pm#L246)).  Version 1.130 is installed on our systems via a package as well.  The current code expects any newer versions to also be installed via Debian package, so we'd need to do something special if we're going to put them in an allocation instead.

apipe-tester:

apipe-tester:

apipe-tester:

apipe-tester:

apipe-tester:

apipe-tester:

lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/FastqsToBam.pm
jasonwalker80: I don't view debian packaging as a reasonable solution. All effort for future development should be directed toward supporting Docker images instead.

apipe-tester:

jasonwalker80:
The two failed tests both pass when I run them locally. Suggestions?

tmooney:
The two tests that are failing are comparing the binary versions of the BAMs with `File::Compare`.  There could be something going differently in the compression between your local run and the server-launched one?  The test passed for me locally and run as {{apipe-tester}} inside a {{genome-test-env}} on the Jenkins checkout as well...

I'd be inclined to replace these two tests with something that examines the contents of the BAM files instead.

ebelter:
Tom is right. I have seen issues with comparing compressed bams. Maybe dump to sam/fastq, then compare? Rely on flagstat?

apipe-tester:

apipe-tester:

lib/perl/Genome/InstrumentData/Command/Import/WorkFlow/FastqsToBam.pm
tmooney: In this case "something special" might mean adding to [this array](https://github.com/genome/genome/blob/68b99044894435f7d08d9621e4d8d606f23201d1/lib/perl/Genome/Model/Tools/Picard.pm#L147) or having a standard way to look up allocations for tool versions.  (The hacky version of the latter might be to have the tool be the `owner_class_name` and the version be the `owner_id`, but we could do better :smile:)

If we had `docker run` I could imagine all sorts of ways to update the GMS!

PR #1794, opened 2017-12-06
Update RefAlign test builds for run without genotype data.

tmooney:
The old genotype data these tests used has been purged, and the corresponding model was old enough that we don't run them that way anymore.


apipe-tester:

apipe-tester:

PR #1795, opened 2017-12-06
Better CwlPipeline model naming.

tmooney:
Since CwlPipeline encompasses many different workflows, it is more common to have multiple models per subject.  Having them named "cwl_pipeline-1" through "cwl_pipeline-6" is less than ideal, so this provides a way to distinguish them based on the workflows they run.


apipe-tester:

apipe-tester:

apipe-tester:

tmooney:
Yep! I'd definitely want to fill them in for the existing pipelines we're using.

PR #1799, opened 2017-12-07
PrepForTransfer Test

jasonwalker80:
test for PrepForTransfer. Use absolute path before symlink.


apipe-tester:

apipe-tester:

lib/perl/Genome/Model/CwlPipeline/Command/PrepForTransfer.pm
```diff
@@ -45,9 +45,10 @@ sub execute {
     if ($self->md5sum) {
         push @headers, 'md5sum';
     }
-    if (-d $self->directory) {
-        $self->fatal_message('Unable to use existing directory: '. $self->directory);
-    } else {
+    if (-e $self->directory && !(Genome::Sys->directory_is_empty($self->directory)) ) {
+        $self->fatal_message('Unable to use existing non-empty directory: '. $self->directory);
+    }
+    unless (-d $self->directory) {
```
tmooney: Do we want to worry about the case where `$self->directory` points to an existing non-directory?  (Amusingly `Genome::Sys->directory_is_empty` returns true for files.)

lib/perl/Genome/Model/CwlPipeline/Command/PrepForTransfer.t
```diff
@@ -0,0 +1,59 @@
...
+   builds => \@builds,
+   directory => $output_dir,
+   md5sum => 1,
+);
+
+ok($cmd, "created command");
+ok($cmd->execute, "executed command");
+
+compare_ok($expected_file, $output_file);
+
```
tmooney: Another thing we could check here is if the command created a symlink per original file in the output directory.

lib/perl/Genome/Model/CwlPipeline/Command/PrepForTransfer.pm
jasonwalker80: I assumed that Genome::Sys->create_directory would fail if the existing non-directory where in the way. I guess that was a bad assumption?

lib/perl/Genome/Model/CwlPipeline/Command/PrepForTransfer.t
jasonwalker80: I thought about checking the symlinks. I already had enough trouble making them. suggestions for checking the existence of the symlink *AND* that it's pointing at the right target?

lib/perl/Genome/Model/CwlPipeline/Command/PrepForTransfer.pm
tmooney: It will.  You'll get a weird error like:
```Runtime error: directory does not exist: myfile.txt at Genome/Sys.pm line 935.```

jasonwalker80: that is a strange/ambiguous error. Do you think I should handle an existing non-directory here?

tmooney: To be user-friendly I'd want to handle these two cases:
* `-e $dir && !-d $dir`
* `-d $dir && !Genome::Sys->directory_is_empty($dir)`

`Genome::Sys->create_directory($dir)` is fine with the directory already existing.

jasonwalker80:
One more review of the new test and improved directory handling?

apipe-tester:

apipe-tester:

PR #1800, opened 2017-12-11
Canonicalize the data path.

tmooney:
If the test is run as, e.g., "./ConcatenateMetricFiles.t", the data directory would have the "./" but the files written out would not, thwarting the replace in the comparison.


apipe-tester:

apipe-tester:

PR #1804, opened 2017-12-14
AnP Disk Group Usage

jasonwalker80:
initial revision. Performance and user interface suggestions are welcome.

Main use case, is CIS-241.

### DIFF
--- a/lib/perl/Genome/Config/AnalysisProject/Command/DiskGroupUsage.pm
+++ b/lib/perl/Genome/Config/AnalysisProject/Command/DiskGroupUsage.pm
@@ -31,7 +31,9 @@ Running builds will not be accounted for properly as the allocations are not com
 The output is a tab-delimited disk usage summary for each disk group used per analysis project.
 
 The columns of the output file are:
-analysis project ID, configuration item ID, subclass name, model count, build count, total KB, total base pairs, bytes per base
+analysis project ID, disk group name, total KB
+
+OPTIONAL: Each allocation ID is printed out, tab indented after the disk group summary line
 
 EOHELP
 }
--- a/lib/perl/Genome/Config/AnalysisProject/Command/DiskGroupUsage.pm
+++ b/lib/perl/Genome/Config/AnalysisProject/Command/DiskGroupUsage.pm
@@ -14,8 +14,8 @@ class Genome::Config::AnalysisProject::Command::DiskGroupUsage {
             is_many             => 1,
             shell_args_position => 1,
         },
-        print_allocation_ids_for_disk_group_name => {
-            is => 'Text',
+        print_allocation_ids_for_disk_group => {
+            is => 'Genome::Disk::Group',
             is_optional => 1,
         },
     ],
@@ -69,7 +69,7 @@ sub execute {
                 'total_kb' => $total_kb,
             };
             $writer->write_one($data);
-            if ($self->print_allocation_ids_for_disk_group_name && ($self->print_allocation_ids_for_disk_group_name eq $disk_group_name) ) {
+            if ($self->print_allocation_ids_for_disk_group && ($self->print_allocation_ids_for_disk_group->name eq $disk_group_name) ) {
                 for my $allocation_id (sort keys %{$disk_group_usage{$disk_group_name}} ) {
                     print "\t". $allocation_id ."\n";
                 }
--- a/lib/perl/Genome/Config/AnalysisProject/Command/DiskGroupUsage.pm
+++ b/lib/perl/Genome/Config/AnalysisProject/Command/DiskGroupUsage.pm
@@ -14,8 +14,10 @@ class Genome::Config::AnalysisProject::Command::DiskGroupUsage {
             is_many             => 1,
             shell_args_position => 1,
         },
-        print_allocation_ids_for_disk_group => {
-            is => 'Genome::Disk::Group',
+        print_allocation_ids => {
+            is => 'Boolean',
+            doc => 'This option will turn on printing ALL allocation IDs for each disk group as a comma delimited list. BEWARE, this could be a very long list of IDs.',
+            default_value => 0,
             is_optional => 1,
         },
     ],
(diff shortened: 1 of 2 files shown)


apipe-tester:

apipe-tester:

lib/perl/Genome/Config/AnalysisProject/Command/DiskGroupUsage.pm
```diff
@@ -0,0 +1,85 @@
...
+    is => 'Command::V2',
+    has => [
+        analysis_projects => {
+            is                  => 'Genome::Config::AnalysisProject',
+            doc                 => 'the analysis projects to report disk usage',
+            is_many             => 1,
+            shell_args_position => 1,
+        },
+        print_allocation_ids_for_disk_group_name => {
+            is => 'Text',
```
tmooney: This could be `print_allocation_ids_for_disk_group` (`is => 'Genome::Disk::Group'`) so users can identify disk group objects by any means they like.  (We would want to make sure that supplying a bare name would still find the object like it does for models.  I think we might need to give it a little help since the property is not currently called `name`, but we could add an alias.)

lib/perl/Genome/Config/AnalysisProject/Command/DiskGroupUsage.pm
```diff
@@ -0,0 +1,85 @@
...
+sub help_detail {
+    return <<EOHELP
+This command will take several minutes to run.
+
+Running builds will not be accounted for properly as the allocations are not complete.
+
+The output is a tab-delimited disk usage summary for each disk group used per analysis project.
+
+The columns of the output file are:
+analysis project ID, configuration item ID, subclass name, model count, build count, total KB, total base pairs, bytes per base
```
tmooney: This section appears to be a holdover from another command.

lib/perl/Genome/Config/AnalysisProject/Command/DiskGroupUsage.pm
```diff
@@ -0,0 +1,85 @@
...
+        separator => "\t",
+        headers => \@headers,
+    );
+    for my $anp ($self->analysis_projects) {
+        my %disk_group_usage;
+        for my $config_item ($anp->config_items) {
+            for my $model ($config_item->models) {
+                for my $build ($model->builds) {
+                    my @allocations = $build->disk_usage_allocations;
+                    for my $allocation (@allocations) {
```
tmooney: This is missing any allocations due to instrument data assigned to the AnP.

lib/perl/Genome/Config/AnalysisProject/Command/DiskGroupUsage.pm
```diff
@@ -0,0 +1,85 @@
...
+        total_kb
+    /;
+    my $writer = Genome::Utility::IO::SeparatedValueWriter->create(
+        separator => "\t",
+        headers => \@headers,
+    );
+    for my $anp ($self->analysis_projects) {
+        my %disk_group_usage;
+        for my $config_item ($anp->config_items) {
+            for my $model ($config_item->models) {
```
tmooney: If we're not breaking the data down by config item, we could loop over `$anp->models` instead.

lib/perl/Genome/Config/AnalysisProject/Command/DiskGroupUsage.pm
```diff
@@ -0,0 +1,85 @@
...
+            map { $total_kb += $_ } values %{$disk_group_usage{$disk_group_name}};
+            my $data = {
+                'analysis_project' => $anp->id,
+                'disk_group_name' => $disk_group_name,
+                'total_kb' => $total_kb,
+            };
+            $writer->write_one($data);
+            if ($self->print_allocation_ids_for_disk_group_name && ($self->print_allocation_ids_for_disk_group_name eq $disk_group_name) ) {
+                for my $allocation_id (sort keys %{$disk_group_usage{$disk_group_name}} ) {
+                    print "\t". $allocation_id ."\n";
```
tmooney: Would it make sense to add the allocations as a column in the TSV?

apipe-tester:

apipe-tester:

apipe-tester:

lib/perl/Genome/Config/AnalysisProject/Command/DiskGroupUsage.pm
jasonwalker80: I could use some help resolving the object by name:
`Resolving parameter 'print_allocation_ids_for_disk_group' from command argument 'prod-builder'... none found.`

apipe-tester:

lib/perl/Genome/Config/AnalysisProject/Command/DiskGroupUsage.pm
tmooney: jasonwalker80/genome/pull/3 is one possible solution.

apipe-tester:

lib/perl/Genome/Config/AnalysisProject/Command/DiskGroupUsage.pm
jasonwalker80: I'm mixed about this. I like having things in a true tab-delimited format. I'm a little worried about:
1.  The length of the string (comma-delimited IDs?)
2. What value to print for the other disk groups? Or should I print the allocation IDs for each disk group in the full TSV format?

tmooney: The main thing that seemed odd to me is that these allocation IDs could end up in the middle of the TSV report and they don't really follow the format.  A few possible options for reporting these:
* In their own file
* comma-delimited for all disk groups as a "verbose" option
* tacked on as extra columns in the row to which they apply
* always printed at the end of the report

apipe-tester:

jasonwalker80:
@tmooney I'm starting to believe this should be a build command instead of an AnP command. Then the input to the command could be `model.analysis_project.id=$ANP_ID` to list the allocations and disk usage by group for each build rather than by the entire AnP. Thoughts?

jasonwalker80:
I guess it depends on whether the user needs accurate accounting at the disk group level. Perhaps both commands will have utility?

tmooney:
At the build level we could possibly make something more detailed along the lines of the existing `genome model build list-volumes` (which probably doesn't capture the same set of information as these loops do).  At any rate, I agree that this command is still possibly useful even if someone might also like build-level details.

apipe-tester:

apipe-tester:

PR #1809, opened 2017-12-18
Avoid calling delete() on a Set::Scalar::Null.

tmooney:
Although one can create a null `Set::Scalar` which can have `delete()` called on it, the `difference()` method of `Set::Scalar` returns a special subclass, `Set::Scalar::Null`, that doesn't have `delete()` defined.  This avoids this issue by skipping the deletion if the set is already null.  (In that case, there's nothing to delete, anyway!)


apipe-tester:

apipe-tester:

PR #1815, opened 2017-12-28
Remove intermediate output directories after CwlPipeline run.

tmooney:
This would reduce disk usage until we can get to a better solution for storing results of individual steps.


apipe-tester:

apipe-tester:

chrisamiller:
+1  - appears to do what it says on the tin

PR #1816, opened 2018-01-02
Remove "ignored" from solexa update tree.

tmooney:
This always has an initial value so cannot be updated due to the `only_if_null` restriction on solexa updates.  The generic instrument-data updater can be used to modify this value.


apipe-tester:

apipe-tester:

PR #1818, opened 2018-01-03
CwlPipeline Cleanup

jasonwalker80:
A quick tool to help cleanup older CwlPipeline builds.

I'm getting errors like:
```
2018/01/03 15:39:50 Genome::Sys: Problems encountered removing /gscmnt/gc2568/techd/model_data/4e699a3f32a243c48cda2ecaa374da56/build65eb643b64ac4ed38475751b8285bb75/results/tmp4CoNwC
File /gscmnt/gc2568/techd/model_data/4e699a3f32a243c48cda2ecaa374da56/build65eb643b64ac4ed38475751b8285bb75/results/tmp4CoNwC, message: cannot remove directory: Permission denied
```

NOTE: This was run as the prod-builder user

When I cleanup the tmp* directories as prod-builder user I usually have to:
chmod u+w $data_directory/results
chmod -R u+w $data_directory/results/tmp*
rm -fR $data_directory/results/tmp*
chmod u-w $data_directory/results

I'm not entirely sure how `Genome::Sys->remove_directory_tree()` works on restricted permission files and directories. After looking at an example directory that failed to be removed the files are gone.
  
See #1815 for real solution.

### DIFF
--- a/lib/perl/Genome/Model/CwlPipeline/Command/Cleanup.pm
+++ b/lib/perl/Genome/Model/CwlPipeline/Command/Cleanup.pm
@@ -40,12 +40,14 @@ sub execute {
         my $results_dir = File::Spec->join($data_directory, 'results');
 
         my $tmp_dir = Genome::Sys->create_temp_directory($build->id);
+
         my $results_dir_mode = mode($results_dir);
         $results_dir_mode->add_user_writable();
+        my $guard = Scope::Guard->new(sub { $results_dir_mode->rm_user_writable(); });
+
         unless (Genome::Model::CwlPipeline::Command::Run->cleanup($tmp_dir, $results_dir)) {
             $self->fatal_message("Failed to cleanup build tmp dir '%s' and results dir '%s'", $tmp_dir, $results_dir);
         }
-        $results_dir_mode->rm_user_writable();
 
         my $allocation_path = File::Spec->join('model_data',$build->model->id,'build'. $build->id);
         my $allocation = Genome::Disk::Allocation->get(allocation_path => $allocation_path);
--- a/lib/perl/Genome/Model/CwlPipeline/Command/Cleanup.pm
+++ b/lib/perl/Genome/Model/CwlPipeline/Command/Cleanup.pm
@@ -29,7 +29,10 @@ sub execute {
     my $self = shift;
 
     for my $build ($self->builds) {
-
+        unless ($build->status eq 'Succeeded') {
+            $self->fatal_message("Unable to run cleanup on '%s' build with id '%s'. For Failed builds, please abandon instead once troubleshooting is complete.",$build->status,$build->id);
+        }
+        
         my $data_directory = $build->data_directory;
 
         my $results_dir = File::Spec->join($data_directory, 'results');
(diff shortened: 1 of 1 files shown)


jasonwalker80:
After running a live test, it seems like we need the permissions of the results directory to be updated before removing tmp* and possibly the permissions of tmp* as well... I tried the latter first and then the former before it worked.

apipe-tester:

apipe-tester:

jasonwalker80:
@tmooney I'm not sure if we want to merge this or leave as an open PR until after we've cleaned up all existing CwlPipeline Succeeded builds (and the snapshot with build cleanup is live).

Also, do we need to worry about the permissions with real running CwlPipeline builds? Or do the permission changes happen later?

lib/perl/Genome/Model/CwlPipeline/Command/Cleanup.pm
```diff
@@ -0,0 +1,62 @@
...
+
+        my $data_directory = $build->data_directory;
+
+        my $results_dir = File::Spec->join($data_directory, 'results');
+
+        my $tmp_dir = Genome::Sys->create_temp_directory($build->id);
+        my $results_dir_mode = mode($results_dir);
+        $results_dir_mode->add_user_writable();
+        unless (Genome::Model::CwlPipeline::Command::Run->cleanup($tmp_dir, $results_dir)) {
+            $self->fatal_message("Failed to cleanup build tmp dir '%s' and results dir '%s'", $tmp_dir, $results_dir);
```
tmooney: Failure here will leave the permissions loosened on the directory!

apipe-tester:

apipe-tester:

lib/perl/Genome/Model/CwlPipeline/Command/Cleanup.pm
```diff
@@ -0,0 +1,62 @@
...
+        }
+        $results_dir_mode->rm_user_writable();
+
+        my $allocation_path = File::Spec->join('model_data',$build->model->id,'build'. $build->id);
+        my $allocation = Genome::Disk::Allocation->get(allocation_path => $allocation_path);
+        unless ($allocation) {
+            $self->fatal_message("Failed to find allocation by allocation path '%s'.", $allocation_path);
+        }
+        $allocation->reallocate();
```
tmooney: Since the goal is to reallocate, we could skip all of this and use `$build->reallocate_disk_allocations` from `Genome::Role::ObjectWithAllocations`.

lib/perl/Genome/Model/CwlPipeline/Command/Cleanup.pm
jasonwalker80: I assume I want to have some kind of scope guard or use an eval to cleanup a failed call. Suggestions?
```
eval { cleanup() }
if ($@) { $results_dir_mode->rm_user_writable(); $self->fatal_message();}
```

tmooney: A `Scope::Guard` is probably good enough.  Otherwise I'd suggest `Try::Tiny` for the `finally`.

apipe-tester:

apipe-tester:

jasonwalker80:
Closing PR, if this tool is needed see tmp_cwl_cleanup branch on jasonwalker80/genome.

PR #1819, opened 2018-01-04
Update blessed clinseq-wer build.

tmooney:
Files are being moved, so the paths in the IGV XML are changing.  (This seems likely to be a recurring issue for awhile as data is migrated.)


apipe-tester:

apipe-tester:

PR #1820, opened 2018-01-08
Only check for directory existence if we got a path back.

tmooney:
Some builds never had data directories to begin with.


apipe-tester:

apipe-tester:

PR #1823, opened 2018-01-16
Add UnionCnv to list of DV2 results that should get moved.

tmooney:
These are currently getting left behind by `genome model build move-allocations`.


apipe-tester:

apipe-tester:

PR #1826, opened 2018-02-02
WIP: Set the AnP environment before restarting a build.

jasonwalker80:
Tests fail since the AnP used doesn't have config. Is there an example of configuring a test AnP to have the environment config?

### DIFF
--- a/lib/perl/Genome/Model/CwlPipeline/Command/Restart.pm
+++ b/lib/perl/Genome/Model/CwlPipeline/Command/Restart.pm
@@ -32,6 +32,16 @@ sub execute {
     my $build_count = scalar(@builds);
     my @errors;
     for my $build (@builds) {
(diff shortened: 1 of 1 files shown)


apipe-tester:

apipe-tester:

lib/perl/Genome/Model/CwlPipeline/Command/Restart.pm
```diff
@@ -32,6 +32,16 @@ sub execute {
     my $build_count = scalar(@builds);
     my @errors;
     for my $build (@builds) {
+        my $anp = $build->model->analysis_project;
```
tmooney: Should we put this inside `_restart_build` so that each build can be independently tried?

jasonwalker80:
Closing in favor of #1828

PR #1830, opened 2018-02-14
Another place where we need "defined".

tmooney:
Hot on the heals of #1829 we have another location that needed fixing.


apipe-tester:

apipe-tester:

PR #1831, opened 2018-02-15
Fix for assigning new instrument data to specified AnP.

tmooney:
The `@data` passed in here is not the Genome object but the synch object, so we need to look up the Genome entitiy for the add command.


apipe-tester:

apipe-tester:

PR #1837, opened 2018-02-22
`genome model cwl-pipeline` usability improvements

tmooney:
* Make `builds` a positional parameter when useful.
* Add the base of the command tree so tab-completion can find these commands and they appear in help.


apipe-tester:

apipe-tester:

PR #1839, opened 2018-02-26
Look for the latest workflow or LSF job.

tmooney:
A build could have multiple processes (and workflows) since it can be restarted.  For most purposes (like `genome model build stop`) the most recent process is the correct one to consider.


chrisamiller:
+3(?)

PR #1845, opened 2018-03-27
Only allow restart by the user who originally started a build.

tmooney:
Restricted permissions in the Toil directories will typically mean a different user would be unable to properly continue the attempt.


chrisamiller:
+1

PR #1846, opened 2018-03-28
Remove the source file's index when we remove the source file.

tmooney:
This is a rather brute-force implementation :smile:  Alternatively, we could examine the filename or call `_is_currently_bam` and `_is_currently_cram` to decide which index to remove.


chrisamiller:
+1

PR #1849, opened 2018-03-29
Command for importing LIMS BAMs from the new archive system.

tmooney:
This command has only been tested under a `xenial`-based environment.  It almost certainly doesn't work with older systems, thanks to the fancy ACLs in place in the new archive system.


tmooney:
I've added a commit to remove the allocation if something goes wrong (and also continue on to the next instrument data in the list).

What is the error message you're seeing about missing fields?  Maybe we could still quiet it if it's not truly problematic.

jasonwalker80:
I don't have the error message anymore. I can't find the logs for those. It only happened with one data set. I haven't seen it since. I'd move on with this PR.

PR #1857, opened 2018-05-25
WIP: Short Name option for interval list creation

jasonwalker80:
add short_name option for interal list generation.

I labeled this WIP since I can't confirm this worked. when I went to set a test name I couldn't determine the existing result ID.


tmooney:
We'll need to backfill and rehash all the existing results to account for the new input.

tmooney:
I believe that #1911 implements the intent of this PR, so I'll close this one.

PR #1862, opened 2018-08-21
Make sure $self gets passed to _resolve_bam_path.

tmooney:
This fixes the errors recently seen in the synch cron.


jasonwalker80:
+2

PR #1863, opened 2018-08-22
Add new option whether or not to remove "chr"s during conversion.

tmooney:
The original behavior was a holdover from when we assumed human build36 or 37, but these days we need to also support references with `chr`s in them intentionally! :smile:


tmooney:
Hello CI, test this please!

PR #1869, opened 2018-10-10
Support Cromwell as a CWL runner.

tmooney:
This PR introduces new configuration variables that should be set before using cromwell:
* `cromwell_auth`: The database password for the Cromwell database.
* `cromwell_server`: The `jdbc` connection string for the Cromwell database.
* `cromwell_truststore_auth`: The password for the truststore necessary to connect over SSL to the Cromwell database. (defaults to the default Java password of `changeit`).
* `cromwell_truststore_file`: The truststore necessary to connect over SSL to the Cromwell database.  (Must be explicitly set even if using the system default truststore.)
* `cromwell_user`: The database user for the Cromwell database.  (Probably `cromwell`, but maybe somebody needs it to be different.)
* `cwl_runner`: Set this to `cromwell` to try out this PR.  Defaults to `toil` (the existing CWL runner) for backwards-compatibility.

It'd probably be a good idea to try this out before merging the PR.  Just wanted to get the code out here early for a look-see!


lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
```diff
@@ -25,8 +29,16 @@ sub execute {
 
     my $yaml = $self->prepare_yaml;
     my ($tmp_dir, $results_dir) = $self->prepare_directories;
-    $self->run_toil($yaml, $tmp_dir, $results_dir);
-    $self->cleanup($tmp_dir, $results_dir);
+    my $cwl_runner = Genome::Config::get('cwl_runner');
+    if ($cwl_runner eq 'cromwell') {
+        $self->run_cromwell($yaml, $tmp_dir, $results_dir);
+        #TODO oull results out of cromwell's hierarchy and cleanup
```
jasonwalker80: typo in comment `oull` -> `pull`

tmooney:
I added some new draft changes to stage the outputs to the `results/` dir.  Unlike the previous changes, they're completely untested as of yet except that they survive `perl -c`!

jasonwalker80:
The latest commits look good as well. Although, maybe we wait to see if a CWL implementation of the output gathering, flattening works out:
https://github.com/genome/analysis-workflows/pull/488

PR #1878, opened 2018-10-31
Look up the chain files instead of using hardcoded paths.

tmooney:
I could imagine expanding this system to support many other chain files we have, but for now this at least works right.  (I could also see removing the shortcuts and always requiring a chain file, too... how
often do we need to convert to hg18 these days?)

### DIFF
--- a/lib/perl/Genome/Model/Tools/LiftOver.pm
+++ b/lib/perl/Genome/Model/Tools/LiftOver.pm
@@ -5,6 +5,8 @@ use strict;
 use IO::File;
 use Genome;
 
+use File::Spec;
+
 class Genome::Model::Tools::LiftOver {
     is => 'Command::V2',
     has_input => [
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Model/Tools/LiftOver.pm
```diff
@@ -100,10 +102,16 @@ sub execute {
...
         if( $lift_direction eq 'hg19ToHg18' ) {
-            $chain_file = "/gscmnt/gc12001/info/model_data/chain_files/hg19Tohg18/hg19ToHg18.over.chain";
+            my $alloc = Genome::Disk::Allocation->get(allocation_path => 'model_data/chain_files/hg19Tohg18');
+            $self->fatal_message("Couldn't find chain file for hg19Tohg18") unless $alloc;
+
+            $chain_file = File::Spec->join($alloc->absolute_path, "hg19ToHg18.over.chain");
         }
         elsif( $lift_direction eq 'hg18ToHg19' ) {
-            $chain_file = "/gscmnt/gc12001/info/model_data/chain_files/hg18Tohg19/hg18ToHg19.over.chain";
+            my $alloc = Genome::Disk::Allocation->get(allocation_path => 'model_data/chain_files/hg18Tohg19');
```
susannasiebert: I think we could avoid some code duplication here by just setting a `$allocation_path` variable and doing the the rest after the else block

tmooney: The allocation path and the filename are both different in each case.  However, with a little lowercasing hack I've combined the two.

PR #1896, opened 2018-12-04
Use Local `disk.` Tables

tmooney:
In preparation for the LIMS changing how they manage disk space, use our own copies of the tables so we can manage this independently. The DB already had tables created long ago, but we had never populated them until now.


tmooney:
This currently fails tests because the test DB has empty disk group and volume tables.  I'll have to backfill those, too.

tmooney:
This finally passes tests!

[I'm not a fan of the diff in the new `force-pushed` notice.  It includes irrelevant changes just because it's been rebased on a newer master!]

PR #1900, opened 2018-12-20
Support importing LIMS data that may or may not be BAMs.

tmooney:
Now we can import FASTQ directories too!  Nothing in the system knows how to use them yet, but they're there.  For now, can reach the FASTQ directory using something like:

```perl
my $dir = $instrument_data->disk_allocation->absolute_path;
```

It's then up to the script to decide how to find/use the files it wants.  The existing `Genome::InstrumentData::Solexa` fastq methods don't know what to do with the newer-style directories yet.

### DIFF
--- a/lib/perl/Genome/Site/TGI/Command/ImportDataFromLims.pm
+++ b/lib/perl/Genome/Site/TGI/Command/ImportDataFromLims.pm
@@ -1,4 +1,4 @@
-package Genome::Site::TGI::Command::ImportBamFromLims;
+package Genome::Site::TGI::Command::ImportDataFromLims;
 
 use strict;
 use warnings;
@@ -9,7 +9,7 @@ use Try::Tiny qw(try catch);
 
 use Genome;
 
-class Genome::Site::TGI::Command::ImportBamFromLims {
+class Genome::Site::TGI::Command::ImportDataFromLims {
     is => 'Command::V2',
     has_input => {
         instrument_data => {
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Site/TGI/Command/ImportDataFromLims.pm
```diff
@@ -73,20 +73,31 @@ sub _process_instrument_data {
 
 
     try {
-        my ($bam_file, $lims_source_dir) = File::Basename::fileparse($lims_path);
+        my ($bam_file, $lims_source_dir);
+        if ($lims_path =~ /\.bam$/) {
+            ($bam_file, $lims_source_dir) = File::Basename::fileparse($lims_path);
+        } elsif (-d $lims_path) {
```
acoffman: Should we check not only that this is a directory, but that the files in there are fastq and in the correct format?

lib/perl/Genome/Site/TGI/Command/ImportDataFromLims.pm
```diff
@@ -131,7 +142,7 @@ sub _resolve_lims_bam_path {
     my $path;
     while (!$path and @data) {
         my $next = shift @data;
-        $path = $next if $next =~ m!^/gscmnt/!;
+        $path = $next if ($next =~ m!^/gscmnt/! and $next !~ m!^/storage./!);
```
acoffman: I see that this is accounting for the new storage path structure, but it seems odd in general to have to loop over the lines in the file. (Not that that's related to this PR in particular...)

tmooney: It sure does!  It's because the first lines in the file are messages from `docker_run.py` instead of the output of our command.  We could probably work around this by `bsub`bing a short shell script that writes to the file instead of using the output redirection built into LSF.  Then we ought to be able to accept the first line of the file and run with it.

lib/perl/Genome/Site/TGI/Command/ImportDataFromLims.pm
tmooney: Maybe!  I'd say the current version of our goal with this is to "accept whatever they've generated".  If it's not a BAM (or a .tar.gz of FASTQs) then none of our other code knows what to do with it anyway.  I've thought about adding better support for the current FASTQ format that we're likely to get, but it might not be worth plugging into all our legacy methods in `Genome::InstrumentData::Solexa`.

PR #1907, opened 2019-01-10
More Tolerant Staging and Handling of Strange Outputs

tmooney:
* I don't know why someone would want to make dangling symlinks as the output of their workflow, but we encountered this in the wild... so handle it!
* Sometimes there are arrays of things that aren't JSON objects.  For simple values (i.e. numbers and strings), go ahead and skip them.
* Allow for duplicate filenames as long as the contents of the files are the same.

### DIFF
--- a/lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
+++ b/lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
@@ -333,6 +333,7 @@ sub _stage_cromwell_output {
         return 1;
     }
 
+    return unless ref $item eq 'HASH';
     my $location = $item->{location};
     return unless $location;
 
--- a/lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
+++ b/lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
@@ -8,6 +8,7 @@ use Genome;
 use Genome::Utility::File::Mode qw();
 use YAML;
 use JSON qw(to_json);
+use File::Compare qw();
 
 class Genome::Model::CwlPipeline::Command::Run {
     is => 'Command::V2',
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
```diff
@@ -342,7 +344,12 @@ sub _stage_cromwell_output {
 
         my $destination = File::Spec->join($results_dir, $file);
         if (-e $destination) {
-            $self->fatal_message('Cannot stage results. Multiple outputs with identical names: %s', $file);
+            if( File::Compare::compare($source, $destination) == 0 ) {
```
acoffman: It looks like this does a line-by-line comparison - how big are the files in question and do we already have a hash or something we could use instead? This seems like it could be quite slow depending on the file size.

lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
```diff
@@ -332,6 +333,7 @@ sub _stage_cromwell_output {
         return 1;
     }
 
+    return unless ref $item eq 'HASH';
```
acoffman: Is there ever a time when we'd want to persist these "simple" outputs in a file or somesuch? I guess without context like a label or something, a random number isn't all that useful..

tmooney: Interesting question!  They're all in the Cromwell database (where we got this data we're sifting through), so if we decide we do want to do that it's not too late to go back and backfill later 😄

lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
tmooney: Only `File::Compare::compare_text` goes line-by-line.  Plain old `compare` uses lower-level I/O to go through chunks of the file at a time.  It is still potentially slow (far slower than comparing existing hashes would be!), but I don't think we're currently hashing every output file.  That itself would be slower in practice, since we don't expect name collisions.  (And, if we do have name collisions, we normally expect the files to differ.)

In the one case we've encountered matching files like this, they were less than 1 KB.  (Only headers with no content.)

PR #1909, opened 2019-01-15
Pass color to _format_is_many instead of as an extra sprintf argument.

tmooney:
Newer Perl version helpfully highlighted this bug with a warning!


chrisamiller:
nice  +1

PR #1911, opened 2019-01-16
Add option to preserve region names to the IntervalList dumper.

tmooney:
This will come in handy for the per-target metric reporting!

### DIFF
--- a/lib/perl/Genome/FeatureList/Command/DumpIntervalList.pm
+++ b/lib/perl/Genome/FeatureList/Command/DumpIntervalList.pm
@@ -26,6 +26,11 @@ class Genome::FeatureList::Command::DumpIntervalList {
             doc => 'whether to merge adjacent regions of the BED file',
             default => 1,
         },
+        preserve_region_names => {
+            is => 'Boolean',
+            doc => 'whether to retain the region names from the feature-list (or replace them with "short" names)',
+            default => 0,
+        },
     },
     has_optional_output => {
         output_path => {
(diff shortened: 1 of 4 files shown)


lib/perl/Genome/FeatureList/Command/DumpIntervalList.pm
```diff
@@ -38,6 +43,10 @@ class Genome::FeatureList::Command::DumpIntervalList {
 sub execute {
     my $self = shift;
 
+    if ($self->preserve_region_names and $self->merge) {
+        $self->fatal_message('The --preserve-region-names option has no effect unless --nomerge is also specified.  Running the region merger reassigns the region names.');
```
jasonwalker80: Is this true? When I have run the BED dump-interval-list command I often used `--merge` and `--noshort-name`.

tmooney: I was surprised by this, too, but the testcase failed when I flipped `merge => 0` to `merge => 1`.  The long names were not preserved.

tmooney: Looks like the trouble I was seeing is due to a long-standing issues with FeatureLists.  Asking to dump the merged list with different options does not actually do the work a second time--it returns the already-merged list from the first call.

At any rate, I've removed this error message from this command!

tmooney:
CI, test this.

lib/perl/Genome/FeatureList/Command/DumpIntervalList.pm
```diff
@@ -38,6 +43,9 @@ class Genome::FeatureList::Command::DumpIntervalList {
 sub execute {
     my $self = shift;
 
+#    if ($self->preserve_region_names and $self->merge) {
```
acoffman: If this is going to be commented out, it should probably just be deleted?

tmooney: Oops. It totally should!

PR #1913, opened 2019-01-24
Allow for an optional prefix when staging cromwell outputs.

tmooney:
This is one way to address genome/analysis-workflows#546.

To satisfy that issue, an AnP config would have something like this included:

```yaml
instrument_data_properties:
  input_data:
    output_prefix: sample_name
```

### DIFF
--- a/lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
+++ b/lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
@@ -6,6 +6,7 @@ use warnings;
 use File::Spec;
 use Genome;
 use Genome::Utility::File::Mode qw();
+use Genome::Utility::Text qw();
 use YAML;
 use JSON qw(to_json);
 use File::Compare qw();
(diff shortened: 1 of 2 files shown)


lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
```diff
@@ -314,22 +315,39 @@ sub _stage_cromwell_outputs {
...
     return 1;
 }
 
+sub _determine_output_prefix {
+    my $self = shift;
+
+    my $build = $self->build;
+
+    my $prefix;
+    my (@prefix_input) = grep { $_->name eq 'output_prefix' } $build->inputs;
```
acoffman: Can there be multiple prefixes? Or is this just defensive coding?

tmooney: Yes and yes.

Build input need not be unique, so there's nothing in the system that precludes multiple `output_prefix` inputs.  I don't expect there to be more than one in practice, but I wanted to handle the possibility anyway.

PR #1921, opened 2019-02-26
Set default memory request of 4096MB.

tmooney:
This is the companion to genome/analysis-workflows#589.  It sets a default for any CWL steps that don't have a defined memory request.


tmooney:
@apaul7 Cromwell's config has a separate launch section for jobs with no docker image specified.  In this case we have filled in the `primary_docker_image` from the processing profile.  This should be enough to support the toil-style workflows where everything is in the one image.

https://github.com/genome/genome/blob/7a4e2f3a6db7fb8b17eaa6e1474d4c4286c9c7d3/lib/perl/Genome/Model/CwlPipeline/Command/Run.pm#L223

PR #1923, opened 2019-02-28
Display Cromwell workflows in `genome model build view`.

tmooney:
For CWL, this will be a lot more informative than the Ptero workflow (if one even exists)!

### DIFF
--- a/lib/perl/Genome/Command/CromwellWorkflowMixin.pm
+++ b/lib/perl/Genome/Command/CromwellWorkflowMixin.pm
@@ -1,0 +1,174 @@
+package Genome::Command::CromwellWorkflowMixin;
+
+use strict;
+use warnings;
+
+use Genome;
+use Genome::Utility::Text qw(justify strip_color);
+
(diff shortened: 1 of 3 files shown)


lib/perl/Genome/Command/CromwellWorkflowMixin.pm
```diff
@@ -0,0 +1,184 @@
...
+my $INDENTATION_STR = '. ';
+class Genome::Command::CromwellWorkflowMixin {
+    is => ['Command::V2'],
+    roles => ['Genome::Role::CommandWithColor'],
+    has => [
+        summary_threshold => {
+            is => 'Int',
+            is_optional => 1,
+            default_value => '5',
+            doc => "Scatter stages with thsi many steps will be summarized. Use zero to summarize every step. Use a negative value to never summarize.",
```
apaul7: I think I did the suggestion thing right: ```suggestion
            doc => "Scatter stages with this many steps will be summarized. Use zero to summarize every step. Use a negative value to never summarize.",
```

PR #1924, opened 2019-03-05
Site Feature: Import instrument data from LIMS in CQID.

tmooney:
This will cause CQID failures when the data could not be imported.  But that's probably fine, since the builds would themselves fail later!

There is an additional complication in that this would need to be run as a user with access to import the data.

### DIFF
--- a/lib/perl/Genome/Site/TGI/Extension/ConfigureQueuedInstrumentData.pm
+++ b/lib/perl/Genome/Site/TGI/Extension/ConfigureQueuedInstrumentData.pm
@@ -1,0 +1,29 @@
+package Genome::Config::Command::ConfigureQueuedInstrumentData;
+
+use strict;
+use warnings;
+
+use Genome;
+use Genome::Config::Command::ConfigureQueuedInstrumentData; #load real module first
+
+use Sub::Install;
+
+Sub::Install::reinstall_sub ({
+    as => '_import_instrument_data',
+    code => \&_import_lims_data,
+
+});
+
+sub _import_lims_data {
+    my $self = shift;
+    my $current_pair = shift;
+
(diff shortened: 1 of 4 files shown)


lib/perl/Genome/Site/TGI/Extension/ConfigureQueuedInstrumentData.pm
```diff
@@ -0,0 +1,29 @@
...
+
+use Sub::Install;
+
+Sub::Install::reinstall_sub ({
+    as => '_import_instrument_data',
+    code => \&_import_lims_data,
+
+});
+
+sub _import_lims_data {
```
acoffman: Is there a particular reason to use `Sub::Install` here rather than normal inheritance with an overridden method (or similar?)

tmooney: `Sub::Install` is a nice way to avoid the method-redefinition warning.  The site extensions are kind of class monkey-patching for things peculiar to a particular location.  In this case, nobody else in the world has this LIMS system!  There's no subclassing happening, so I'm not sure what it would look like to try to use inheritance here.

(It's an open question to me whether it's worthwhile to maintain the notion that there is both a core system-independent GMS and also site-specific extensions (especially since it's partially broken already with some of the `Genome::Sys` extensions in use).  As long as it's not hard to keep up, I'm not motivated to intentionally break it.)

acoffman: Cool, makes sense to me - I was imagining `Site::TGI::Command::CQID` as possibly just being a subclass of main `CQID` that implements this method, but this makes sense given the context. Looks good to me!

PR #1926, opened 2019-03-11
Ensure only the LSF Job ID gets into the process note.

tmooney:
There's logic to prepend any notes with a sudo message when running as another user, but that's not helpful here. (Really, storing this as a note is not great--it should just be a column on the table.)

### DIFF
--- a/lib/perl/Genome/Process.pm
+++ b/lib/perl/Genome/Process.pm
@@ -355,7 +355,8 @@ sub lsf_job_id {
         if ($existing) {
             $existing->delete;
         }
-        $self->add_note(header_text => $self->lsf_job_id_header, body_text => $new_value);
+        my $n = $self->add_note(header_text => $self->lsf_job_id_header, body_text => $new_value);
+        $n->body_text($new_value); #ignore any system-generated sudo message for this note.
     } else {
         my $id_note = $self->notes(header_text => $self->lsf_job_id_header);
         if ($id_note) {


lib/perl/Genome/Process.pm
```diff
@@ -355,7 +355,8 @@ sub lsf_job_id {
         if ($existing) {
             $existing->delete;
         }
-        $self->add_note(header_text => $self->lsf_job_id_header, body_text => $new_value);
+        my $n = $self->add_note(header_text => $self->lsf_job_id_header, body_text => $new_value);
+        $n->body_text($new_value); #ignore any system-generated sudo message for this note.
```
acoffman: Am I reading this incorrectly, or are we just assigning the same value twice in a row (once in `add` and once manually). What is causing the value to be different?

tmooney: This line in `create` is causing the value to be different:
https://github.com/genome/genome/blob/385fe08f29f6d8a93d57143b9f483b3345f6aa25/lib/perl/Genome/MiscNote.pm#L59
It's not correct IMO to be using a note to store this value... but here we are! So at least for now it'd be better if it gets stored correctly without the extra sudo message.

acoffman: Cool, makes sense! Would it be clearer/less confusing for the reader if we didn't assign the value initially and then just assigned it manually on the next line so there's only one "assignment?"

tmooney: I'm not convinced it would be.  The question then becomes "Why wasn't this assigned above?" instead of "Why is this assigned twice?" :smile:

acoffman: Hah, fair! Yeah, the real problem seems to be the constructor mutating a value you gave it without you asking it to do so - but I imagine changing that would have larger consequences.

PR #1928, opened 2019-03-25
Only consider active allocations when deciding whether to import.

tmooney:
A previous import might have been purged, in which case new analysis would require reimporting the data. (Users can then also use `genome analysis-project reprocess` if they need to pull back the data to build again later.)


tmooney:
There are some other places we'd need to make changes to make this effective.  (Notably the importer itself performs a similar check, and I believe there are things that expect instrument data to only have one allocation.)

PR #1929, opened 2019-03-26
Symlink Prefix for PrepForTransfer

jasonwalker80:
Add an option to use the subject name as the prefix for the symlinks default to build ID.


jasonwalker80:
Closing for now.

PR #1931, opened 2019-04-16
New command to import a directory of instrument data.

tmooney:
No validation is performed whatsoever... it trusts you and the data!

If it happens to conform to an expected standard, it'll work with existing tools.  If it doesn't, it won't!  But in the age of CWL pipelines, someone could make something that handles it.

(This is kind of the general-purpose version of the site-specific `import-data-from-lims` command.)


chrisamiller:
+1 - looks great

lib/perl/Genome/InstrumentData/Command/Import/TrustedData.t
```diff
@@ -0,0 +1,69 @@
...
+    import_source_name => 'TrustedData.t',
+    library => $library,
+    source_directory => $test_dir,
+    import_format => 'numbers',
+    description => 'junk data for testing the importer',
+    instrument_data_properties => [map { join('=', $_, $extra_properties{$_}) } sort keys %extra_properties],
+);
+isa_ok($cmd, $class, 'created command object');
+
+ok($cmd->execute, 'executed command');
```
apaul7: Not 100% sure but is this is the part the CI is failing on?

tmooney:
The CI is failing on the {{rsync}} because this command depends on the "new" image but the tests run on the "legacy" image.  The older {{rsync}} doesn't have one of the options.

chrisamiller:
We're ready to use this on a bunch of single cell data if it's ready to be merged!

apaul7:
I've ran into a few issues when running this from a checked out version. 
Originally ran the import without providing a description as it was an optional input but the import fails because of an `undef` error. It does create the allocation and rsyncs the data, which gets removed after the failure.
```
ERROR: Invalid data for save!
ERROR: Genome::InstrumentDataAttribute identified by 552193bcd4ed4d789505140aa6cdf7f3 description  WUGC has problems on
INVALID: property 'attribute_value': No value specified for required property
    Current state:
$VAR1 = bless( {
                 '__get_serial' => 7330,
                 'instrument_data_id' => '552193bcd4ed4d789505140aa6cdf7f3',
                 'attribute_label' => 'description',
                 '_change_count' => 1,
                 'attribute_value' => undef,
                 'nomenclature' => 'WUGC',
                 'id' => '552193bcd4ed4d789505140aa6cdf7f3      description             WUGC'
               }, 'Genome::InstrumentDataAttribute' );
Use of uninitialized value in concatenation (.) or string at /gsc/scripts/opt/genome/snapshots/genome-3782/lib/perl/Command/Dispatch/Shell.pm line 46.
ERROR: Failed to commit!:  at /gsc/scripts/opt/genome/snapshots/genome-3782/lib/perl/Command/Dispatch/Shell.pm line 46.
```
When you provide a description it successfully creates the instrument data and adds it to an AnP but the eventual builds are stuck at `unstartable` 

```
Error Message: Could not start build, reason: Build f373dffb838145d9ab9af7cc4361abcd of TWDS-80075999-999PB040716.prod-cwl could not be validated for start!
ERROR: property 'instrument_data': read count for instrument data (888095452aaf456c9d08968fae25a91f) has not been calculated, use `genome instrument-data calculate-read-count` to correct this. at /gsc/scripts/opt/genome/snapshots/genome-3782/lib/perl/Genome/Model/Build/Command/Start.pm line 155.
```
The `calculate-read-count` command fails because it can't locate the bam. Still looking for solutions. Maybe manually set the read count?

tmooney:
Added a commit to only set a description if there is one (and it's a true value 😄).

Addressing the validation of build start seems like a separate issue, so that can happen in a followup PR if we decide to remove that check for CWL builds.  (Alternatively, we could make `read_count` mandatory here since all builds currently require it?  Not sure how useful that is.)

apaul7:
all other import methods either pull it from another source(LIMS) or the importer calculates it as the data is converted to a bam? if so I would say make it a required input for the trusted data import

PR #1932, opened 2019-04-23
Make sure we're in the AnP environment when looking up the disk group.

tmooney:
Otherwise we're in the default environment, which never has a configured disk group.


tmooney:
This fixes an issue introduced in #1930.

PR #1933, opened 2019-05-13
Tool to abandon prior builds for a model.

tmooney:
If a model has many builds, sometimes we only want to keep the recent ones.  This facilitates abandoning the others for a large project.

(This tool's creation was motivated by BGAS-657, wherein a project of 900 models had run out of disk space but had many prior successes and failures.)


tmooney:
Yeah, they'll be sorted that way because of the way the relationship is defined.  I was slightly worried that a change to the definition might cause this to break... but lots of other things would also break.  So let's just never change that :smile:

PR #1937, opened 2019-06-19
Create a separate allocation for scratch space for CWL builds.

tmooney:
This way we can have a pool of scratch space and final project disk groups can be sized to only hold the final data products.

If this is deployed, a value for `disk_group_scratch` will need to be defined in configuration for builds to not fail!


tmooney:
Dearest CI, test this please.

PR #1940, opened 2019-07-22
Better symlink handling in CWL output staging.

tmooney:
Perl's `File::Copy::move` doesn't work to move symlinks between mounts.  `File::Copy::Recursive` handles this, but makes a copy even when moving on the same mount, so isn't ideal for regular files.  Thus we'll use one for symlinks and one for everything else.

Additionally, fix symlink support in our prefix adding.  (Absolute symlinks will get broken, but those don't make much sense in the context of a Cromwell workflow... so hopefully none exist!)  This shortcoming was uncovered when adding a test for symlinks :smile:

### DIFF
--- a/lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
+++ b/lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
@@ -10,6 +10,7 @@ use Genome::Utility::Text qw();
 use YAML;
 use JSON qw(to_json);
 use File::Compare qw();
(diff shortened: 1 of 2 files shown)


tmooney:
@acoffman, yeah I'm wondering if it wasn't a bad choice to use `File::Copy::move` for non-files to begin with, and we were just getting lucky before.  It kind of stinks that the version implemented here performs a copy even on the same volume when a `rename` would do, so I think we'd want a better solution if we implement it in general.

PR #1941, opened 2019-08-02
Add is_10x_ support

apaul7:
This PR adds `is_10x_atac`, `is_10x_gex`, and `is_10x_vdj` support for libraries and tests to confirm they work. Having these will help simplify the configs necessary to run the 10x cellranger pipelines.

One thing I wasn't sure was okay was switching to double quotes `"` because the library protocol names we have contain a single quote `'`.

### DIFF
--- a/lib/perl/Genome/Library.pm
+++ b/lib/perl/Genome/Library.pm
@@ -154,6 +154,45 @@ sub is_rna {
     return 0;
 }
 
+sub is_10x_atac {
+    my $self = shift;
+
+    return 1 if Genome::Utility::List::in(
+        $self->protocol, (
+            "10x_SC_ATAC_SEQ"
+        )
+    );
+
+    return 0;
+}
+
+sub is_10x_gex {
+    my $self = shift;
+
+    return 1 if Genome::Utility::List::in(
+        $self->protocol, (
+            "10x_SC-5'GEX",
+            "10x_SC-3'GEX V3",
+            "10x_SC-5'GEX"
+        )
+    );
+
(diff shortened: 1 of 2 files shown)


tmooney:
The double-quotes are fine.  (If you want single-quoting, you could use the generic quoting, like `q{my string}`.  See also [Quote and Quote-like Operators](https://perldoc.perl.org/perlop.html#Quote-and-Quote-like-Operators) in the perldoc.) 

Your subtest counts as a test, so you'll need to add one to the overall test count here:
https://github.com/genome/genome/blob/4ffc4eea548c9acb8a84ebdf4984fa0835c3ae2b/lib/perl/Genome/Library.t#L12

apaul7:
done! I also figured out how to run the tests locally `genome-perl Library.t`

lib/perl/Genome/Library.pm
```diff
@@ -154,6 +154,45 @@ sub is_rna {
...
+}
+
+sub is_10x_gex {
+    my $self = shift;
+
+    return 1 if Genome::Utility::List::in(
+        $self->protocol, (
+            "10x_SC-5'GEX",
+            "10x_SC-3'GEX V3",
+            "10x_SC-5'GEX"
```
tmooney: This list includes the same item, `10x_SC-5'GEX`, twice.

apaul7: good catch! One of them was supposed to be `10x_SC-3'GEX`. Just pushed the correction.

PR #1944, opened 2019-08-13
adding simple cram handling to readcount tools

chrisamiller:
Obviously it's not awesome to have it linked to samtools in my home dir, but I'm not even sure if we can compile a more recent version of samtools in the ancient container.  This is meant to be a band-aid fix that will hold us over until someone hacks the new version of htslib into bam-readcount proper.

### DIFF
--- a/lib/perl/Genome/Model/Tools/Sam/Readcount.pm
+++ b/lib/perl/Genome/Model/Tools/Sam/Readcount.pm
@@ -119,7 +119,31 @@ sub command {
 
     my $version = $self->use_version;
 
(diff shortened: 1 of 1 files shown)


tmooney:
What do you think of incorporating the existing solution for BAM/CRAM conversion?
https://github.com/genome/genome/blob/6abac49d0581024c2648aee6a8b8822e3c8fba3f/lib/perl/Genome/InstrumentData/Command/AlignmentResult/Convert/Base.pm#L85-L94

chrisamiller:
Yeah, that is substantially more elegant.  I'll try to refactor around that

jasonwalker80:
Closing since bam-readcount is a) run in CWL builds b) will support CRAM natively in future releases

PR #1945, opened 2019-09-17
Command to clean up Scratch Directory for a Build

tmooney:
Some users have requested a way to clean up the scratch space without abandoning a build.  (In this case, so they could continue to use the build logs while otherwise reclaiming the disk space.)

This command could also possibly be automated to reclaim space after a certain amount of time without necessarily removing a Failed build from someone's radar.


chrisamiller:
+1

PR #1946, opened 2019-10-17
Allow chains in instrument data properties for model inputs.

tmooney:
This does require each step in the chain to only return a single object, so no crazy multiplicative effects!

(Like the listers and the rules, putting "delete" in would be amusing, so don't do that either!)

### DIFF
--- a/lib/perl/Genome/Config/Profile.pm
+++ b/lib/perl/Genome/Config/Profile.pm
@@ -4,6 +4,7 @@ use strict;
 use warnings;
 
 use List::MoreUtils qw/ any uniq /;
+use Try::Tiny qw/ try catch /;
 
 use Genome;
 use UR::Util;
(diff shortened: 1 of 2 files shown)


lib/perl/Genome/Config/Profile.pm
```diff
@@ -188,14 +189,35 @@ sub _value_for_instrument_data_property {
...
+    unless ($instrument_data_property) {
+        die 'bad call to _apply_method_chain_for_instrument_data_property: missing instrument_data_property';
+    }
+    my @chain = split('->', $instrument_data_property);
+
+    my $current_result = $instrument_data;
+
+    try {
+        for my $next_property (@chain) {
+            $current_result = $current_result->$next_property;
```
chrisamiller: I didn't know about the delete thing!  Is that something we should guard against here?  

    die('"delete" is not a valid property') if $next_property eq "delete";

tmooney: If it's likely someone will accidentally specify `delete` as the property they want in a configuration, then absolutely we could add that.

If it's intended to stop malicious behavior, this would be a lot more roundabout of a strategy than any of the other myriad available ways of deleting bunches of data.

PR #1947, opened 2019-11-18
Use job group for Cromwell workers.

tmooney:
For systems that require a job group for all job submissions, this might come in handy!


chrisamiller:
+1

PR #1951, opened 2019-12-12
Convert protocol lists to regexes to account for lack of LIMS normalization.

tmooney:
An attempt to handle the surprising variation in similarly named protocols!


chrisamiller:
+1

PR #1954, opened 2020-01-22
allow build exports to use the same disk group

chrisamiller:


lib/perl/Genome/Disk/Detail/Allocation/Copier.pm
```diff
@@ -104,7 +104,7 @@ sub _get_copy_shadow_params {
         kilobytes_requested => $allocation->kilobytes_requested,
         owner_class_name => "UR::Value",
         owner_id => "shadow_allocation",
-        exclude_mount_path => $allocation->mount_path,
+        #exclude_mount_path => $allocation->mount_path,
```
tmooney: We could probably delete the line entirely.

```suggestion
```

PR #1955, opened 2020-02-05
CWLPipeline: Accept WDL and custom mount paths

tmooney:
* Cromwell can run WDLs, so if someone tries to give us one, try to run it.
* In the "string" version of pipelines, we don't localize references and instead assume they're mounted in a standard location.  Now there's a config variable to help mount that standard location.  (This isn't strictly required for the File version of workflows to work, but we like avoiding copies!)


tmooney:
We're sure fast approaching a point where having a template for the Cromwell config might be better than our inline block strings!

PR #1958, opened 2020-02-12
add useful message about AnP hold state

chrisamiller:
This seems like a useful "heads up" message  (that would have gotten my builds rolling a few hours sooner today 😄)

### DIFF
--- a/lib/perl/Genome/Config/AnalysisProject/Command/AddInstrumentData.pm
+++ b/lib/perl/Genome/Config/AnalysisProject/Command/AddInstrumentData.pm
@@ -55,7 +55,9 @@ sub execute {
             instrument_data => $instrument_data,
         );
     }
-
+    if($analysis_project->status eq "Hold"){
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Config/AnalysisProject/Command/AddInstrumentData.pm
```diff
@@ -55,7 +55,9 @@ sub execute {
             instrument_data => $instrument_data,
         );
     }
-
+    if($analysis_project->status eq "Hold"){
+        $self->status_message("Analysis project is in a \"Hold\" state. Instrument data will not be processed until it is released");
```
tmooney: I think if we're going to include this message it should also report on `Pending` projects, perhaps:
```perl
if ($analysis_project->status ne 'In Progress') {
    $self->status_message('Analysis Project is in a "%s" state. Assigned Instrument Data will not be processed until it has been released.', $analysis_project->status);
}
```

PR #1960, opened 2020-02-14
adding tarball option to build export

chrisamiller:
Globus eats symlinks, so tarballs seem like a reasonable, if somewhat I/O inefficient way to transfer build directories.  It also simplifies verifying the transfer, since we can compare file sizes or md5s of the tarballs. (Globus does this internally as well, but "trust and verify"...)

### DIFF
--- a/lib/perl/Genome/Model/Build/Command/Export.pm
+++ b/lib/perl/Genome/Model/Build/Command/Export.pm
@@ -117,7 +117,7 @@ sub execute {
 
     if($self->create_tarball){
         my $tarball = File::Spec->join($self->target_export_directory, "build" . $self->build->id . ".tar");   
-        my $rv = Genome::Sys->shellcmd(cmd => ["tar","-cf",$tarball,"-C",$tempdir,"build" . $self->build->id]);
+        my $rv = Genome::Sys->shellcmd(cmd => ["tar","-cf",$tarball,"-C",$tempdir,"build" . $self->build->id], output_files => [ $tarball ]);
         unless ($rv) {
             $self->fatal_message("Could not create tar file");
         }
--- a/lib/perl/Genome/Model/Build/Command/Export.pm
+++ b/lib/perl/Genome/Model/Build/Command/Export.pm
@@ -42,8 +42,7 @@ sub execute {
         #create temp directory for pre-tarball files
         $tempdir = Genome::Sys->create_temp_directory();
         unless($tempdir) {
-            $self->error_message("Unable to create temporary file $!");
-            die;
+            $self->fatal_message("Unable to create temporary directory $!");
         }
         $export_directory = File::Spec->join($tempdir, "build" . $self->build->id);
     }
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Model/Build/Command/Export.pm
```diff
@@ -97,6 +120,14 @@ sub execute {
 
     $find->($export_directory);
 
+    if($self->create_tarball){
+        my $tarball = File::Spec->join($self->target_export_directory, "build" . $self->build->id . ".tar");   
+        my $cmd = "tar -cf $tarball -C $tempdir build" . $self->build->id;
+        system("$cmd") and do {
```
tmooney: I'd prefer to see `Genome::Sys->shellcmd` over `system`.  Or maybe `Genome::Sys->tar`:
https://github.com/genome/genome/blob/cd160fd03c34189ea7a362a8afc8e395a9843fc4/lib/perl/Genome/Sys.pm#L740

(Using `-C` could be a nice improvement to that implementation, and it (or here) could use the array form of specifying the command!)

lib/perl/Genome/Model/Build/Command/Export.pm
```diff
@@ -44,7 +63,9 @@ sub execute {
     $resolve_symlinks = sub {
         my $path = $File::Find::name;
         if (-l $path) {
+            #print STDERR "path: $path\n";
```
tmooney: I'd either remove these commented-out `print`s or convert them to `debug_message`s if they might be useful more generally.

lib/perl/Genome/Model/Build/Command/Export.pm
```diff
@@ -30,6 +37,18 @@ sub execute {
     $self->check_available_space($allocation);
 
     my $export_directory = $self->export_directory;
+    my $tempdir;
+    if($self->create_tarball){
+        #create temp directory for pre-tarball files
+        $tempdir = Genome::Sys->create_temp_directory();
+        unless($tempdir) {
+            $self->error_message("Unable to create temporary file $!");
+            die;
```
tmooney: * This would be better as `die $self->error_message` or even better as `$self->fatal_message`.
* Note that `$!` is examined in `Genome::Sys`, so this is unlikely to ever be reached:
https://github.com/genome/genome/blob/cd160fd03c34189ea7a362a8afc8e395a9843fc4/lib/perl/Genome/Sys.pm#L941-L954

chrisamiller:
All good suggestions. Should be ready for re-review

lib/perl/Genome/Model/Build/Command/Export.pm
```diff
@@ -97,6 +115,14 @@ sub execute {
 
     $find->($export_directory);
 
+    if($self->create_tarball){
+        my $tarball = File::Spec->join($self->target_export_directory, "build" . $self->build->id . ".tar");   
+        my $rv = Genome::Sys->shellcmd(cmd => ["tar","-cf",$tarball,"-C",$tempdir,"build" . $self->build->id]);
```
tmooney: Could also add a this key to `shellcmd`, and it'll double-check the file was created with size:
```perl
   output_files => [ $tarball ],
````

chrisamiller: did this as well

PR #1961, opened 2020-02-18
Clone of PR#1960

tmooney:
Repeat of #1960 to apply to restored master branch.


lib/perl/Genome/Model/Build/Command/Export.pm
```diff
@@ -30,6 +37,17 @@ sub execute {
     $self->check_available_space($allocation);
 
     my $export_directory = $self->export_directory;
+    my $tempdir;
+    if($self->create_tarball){
+        #create temp directory for pre-tarball files
+        $tempdir = Genome::Sys->create_temp_directory();
+        unless($tempdir) {
+            $self->fatal_message("Unable to create temporary directory $!");
```
tmooney:
```suggestion
            $self->fatal_message("Unable to create temporary directory: $!");
```

PR #1963, opened 2020-02-20
Use the current interpreter rather than assuming genome-perl.

tmooney:
This usually doesn't matter at all, but if someone is trying to test different perl versions they'll be glad it's here :)

### DIFF
--- a/lib/perl/Genome/Disk/Allocation.pm
+++ b/lib/perl/Genome/Disk/Allocation.pm
@@ -605,7 +605,7 @@ sub _execute_system_command {
         my $perl_program_string = join('; ', @statements);
 
         my @cmd = (
-            'genome-perl',
+            $^X,
             @includes,
             @_execute_system_command_perl5opt,
             '-e',


lib/perl/Genome/Disk/Allocation.pm
```diff
@@ -605,7 +605,7 @@ sub _execute_system_command {
         my $perl_program_string = join('; ', @statements);
 
         my @cmd = (
-            'genome-perl',
+            $^X,
```
chrisamiller: I don't know this perl convention and it's not easily googl-able...

tmooney: There's a listing of Perl magic variables here: https://perldoc.perl.org/perlvar.html (I typically search "perlvar" to find it.)

chrisamiller: `$^X - The name used to execute the current copy of Perl, from C's argv[0] or (where supported) /proc/self/exe.`

TIL

PR #1964, opened 2020-02-24
Use UUIDs for new disk volumes.

tmooney:
We don't have any sequence generators, anyway 😄


tmooney:
Existing volumes do have IDs that happen to be integers, but the underlying column in the database is already text.

chrisamiller:
+1

PR #1973, opened 2020-05-20
Use UUID generator for creating disk groups.

tmooney:
We've heretofore been importing them from an external canonical source that determines their IDs, but now we need to be able to make them directly.

### DIFF
--- a/lib/perl/Genome/Disk/Assignment.pm
+++ b/lib/perl/Genome/Disk/Assignment.pm
@@ -7,11 +7,11 @@ class Genome::Disk::Assignment {
     table_name => 'disk.volume_group_bridge',
     id_by => [
         group_id => {
-            is => 'Number',
+            is => 'Text',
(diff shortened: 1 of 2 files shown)


tmooney:
Yeah--those copycat tests fail intermittently seemingly because of floating point differences.  We added some tolerances in there, but they're still too strict.  Maybe it's time to stop carrying them along?

chrisamiller:
The CBS segmentation algorithm inside copycat may also be failing tests because it is non-deterministic, in that it starts with randomly seeded locations.  It would in theory be possible to pass a seed for testing, but would require refactoring the underlying code, etc, and has never seemed worth it. I think it'd be fine to drop those tests right now

tmooney:
Opened #1975 for the copycat tests.  Though it passed the most recent attempt, so I'll go ahead and merge 😄

PR #1976, opened 2020-05-21
New command to list docker images for a CWL pipeline build.

tmooney:
Someone was asking me what versions of things ran in a given model, and it seemed like there should be an easier way to answer this question!


zlskidmore:
tested it out works/looks fine to me +1

PR #1979, opened 2020-06-17
Allow imported BAMs with arbitrary filenames

tmooney:
We heretofore have expected `all_sequences.bam` to be the name for all imported BAMs.  The new `trusted-data` importer makes it easier to not follow this convention.  As long as there's only one BAM in the directory, we should be fine to move along using it.

(Plus a bonus bugfix in an existing error message!)

### DIFF
--- a/lib/perl/Genome/InstrumentData/Imported.pm
+++ b/lib/perl/Genome/InstrumentData/Imported.pm
@@ -422,9 +422,16 @@ sub bam_path {
 
     my $bam_file = $allocation->absolute_path . "/all_sequences.bam";
 
-    return if not -e $bam_file;
+    return $bam_file if -e $bam_file;
 
-    return $bam_file;
+    my @possible_bams = glob($allocation->absolute_path . "/*.bam");
+    if (@possible_bams == 1) {
+        return $possible_bams[0];
+    } elsif (@possible_bams > 1) {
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/InstrumentData/Imported.pm
```diff
@@ -417,15 +417,21 @@ sub bam_path {
...
 
-    return if not -e $bam_file;
+    return $bam_file if -e $bam_file;
+
+    my @possible_bams = glob($allocation->absolute_path . "/*.bam");
+    if (@possible_bams == 1) {
+        return $possible_bams[0];
+    } elsif (@possible_bams > 1) {
+        $self->fatal_message('Multiple BAMs found in allocation for imported instrument data: %s.  Name one "all_sequences.bam" to have it used here.', $self->__display_name__);
+    }
```
chrisamiller, line 433: What if the glob returns nothing because there is no bam file?

tmooney: We fall the to empty `return;` at the end and assume it's something else

chrisamiller: okay - fair enough!

tmooney: (This still doesn't handle the issue of running from somewhere where the path isn't mounted... but that isn't a new problem and it extends far beyond this one accessor.)

PR #1981, opened 2020-06-29
New command `genome disk allocation transfer`.

tmooney:
This is a command for transferring data to a remote volume.  Having an appropriate endpoint for the transfer is left as an exercise for the user.  (Thankfully there are existing docs for that :smile:)

This command has not been tested prior to the PR creation(!) We should identify some potential tests cases before merging.


tmooney:
@jasonwalker80 Filled out the tests a bit more.  (Also the command itself has been updated after some real-world use on my part 😄)

PR #1982, opened 2020-07-24
Remove modifications of pre-provided LSF temp directory.

tmooney:
On the new cluster this directory is locked down so we lack permissions to change it.


chrisamiller:
makes sense +1

PR #1984, opened 2020-09-17
Support importing LIMS data into the system directly onto a remote volume.

tmooney:
This should facilitate transfers with slightly fewer manual steps and without intermediate staging on a local volume.


chrisamiller (approved):
TIL that `mkdir -p` will succeed even if the dir exists - nice
+1

zlskidmore:
+1 from me, I have tested this and it is running as expected, does seem to be a bit slow but I don't think anything can be done about that.

zlskidmore (approved):

PR #1986, opened 2020-09-21
Use HTTPS to connect to dgidb.

tmooney:
The new release of dgidb requires SSL. The plain HTTP version redirects and this code can't handle the 301 response.  This is likely a good candidate for outright removal, but these changes make the tests pass again (albeit mostly in a lame way).


johnegarza (approved):

jasonwalker80 (approved):
+1

PR #1987, opened 2020-09-28
Use explicit units for memory reservations (LSF10 feature).

tmooney:
The historic default for units in LSF `bsub`s are confusing (and depending on cluster configuration may be differently inconsistent). Now it's possible to specify a unit suffix, so let's set it to the same precision as CWL affords.

[Marking this as a draft PR because we can't merge it so long as we need to support older clusters.  But it's a change we likely want to patch in on new enough versions.]

### DIFF
--- a/lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
+++ b/lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
@@ -234,7 +234,6 @@ backend {
       config {
         runtime-attributes = """
         Int cpu = 1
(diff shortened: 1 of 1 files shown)


jasonwalker80:
Is this one still a WIP?

tmooney:
@jasonwalker80 not really, but this doesn't work on LSF9... so we can't merge it as long as we have to support that.  It's possible we could patch it in on the new cluster as part of the deploy process--we already have to do that for one other thing...

Maybe it's worth maintaining a second branch for?

PR #1988, opened 2020-10-08
Auto-connect production imports to remote volumes.

tmooney:
Add some very site specific hard-coded stuff to help make imports from LIMS easier.

PR #1989, opened 2020-10-13
New command to help identify which volumes to mount.

tmooney:
This command lists the possible volumes configured for an AnP.  This could be used for a safe list of volumes to mount when launching a container to work with the AnP.

After this and #1988 we'll have the building blocks to possibly support BQM launching remote builds.


jasonwalker80 (approved):
+2, seems like a useful utility.

PR #1991, opened 2020-10-15
Make the "prod-builder" run_as configurable.

tmooney:
Now that we're operating on multiple clusters we've run into a situation where it'd be nice to distinguish between then with different "run_as" indications.  This will let us change it when run on the new cluster while leaving it the same on the old.  (Or someone enterprising could set the appropriate environment variable to setup for the "other" cluster for the one they're on!)


zlskidmore:
I tried out this PR with AnP: 9df6a1f2f94b49c69702ff1899079da8

Everything appears to work as expected :)

PR #1993, opened 2020-11-09
Use more UR-centric approach to test for owner class existence.

tmooney:
UR classes don't necessarily have defined perl module files, and we don't support auto-updating any non-UR classes, so this should be good enough!


zlskidmore:
+1, I used this to update model_data allocations, that process works for this PR when the UR-class is CWL, wheras before it didn't :)

PR #1994, opened 2021-01-26
Create commands for disk groups and volumes

tmooney:
Now that we're mostly managing volumes manually, here are two new commands: `genome disk group create` and `genome disk volume create`.  This should help make the process easier!


jasonwalker80 (approved):
+2

PR #1995, opened 2021-03-15
Add ability to queue rebuilds for CwlPipeline builds.

tmooney:
This functions like `genome model build queue` but for restarting builds rather than creating new ones.  Since we're moving away from supporting the queue for arbitrary users, this is set up to only allowing the queue of failed "production" builds.

To lend functionality to this, there'll need to be a task that runs to process the queue, such as via
```bash
ur list objects --subject Genome::Model::Build::CwlPipeline --filter rebuild_requested=1 --show id
```

### DIFF
--- a/lib/perl/Genome/Model/CwlPipeline.pm
+++ b/lib/perl/Genome/Model/CwlPipeline.pm
@@ -26,6 +26,13 @@ class Genome::Model::CwlPipeline {
             doc => 'short name for pipeline to include in default model names',
         },
     },
+    has_metric => {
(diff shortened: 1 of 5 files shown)


chrisamiller (approved):
+1 

I've always wondered why the "reason" was required for queuing builds, as I've never actually consulted those.  Nice to see it optional here.

tmooney:
I've looked at the reason folks put in sometimes! Since `requeue` only works on failed builds, I figured the basic reason will always be some variation of "hey, this failed" 😄

PR #1996, opened 2021-03-17
Use the apparent size for allocations.

tmooney:
With the fancy fancy cache layer we're using these days, the actual size of files that haven't been pulled into the cache is not detemrined correctly by `du`.  But the apparent size is good enough!


chrisamiller (approved):
+1

PR #1997, opened 2021-03-24
Enable ACL support for filetests.

tmooney:
Less performant, sure, but actually works with a lovely non-posix-compliant filesystem, too!

See also: https://perldoc.perl.org/filetest


chrisamiller (approved):
+1

PR #1998, opened 2021-04-14
When call-caching is enabled, check the health of jobs.

tmooney:
Before we were letting jobs fail and manually running `script.submit` for replacement jobs in the event of LSF failure due to the high cost of starting a workflow from the beginning over such a trivial failure. With call-caching that cost is reduced, so it's better to have quicker notification of issues.

(Ten minutes is an arbitrary choice on my part; I don't have any particular justification for it if someone has suggestions for alternate intervals.)


apaul7 (approved):
+1, should be super useful!

chrisamiller:
+1

PR #1999, opened 2021-04-22
Update connection string for usage log.

tmooney:
It makes me sad this string is hardcoded here, but I guess updating it once every several years isn't so bad 🙂


acoffman (approved):
+1

Issue #2000, opened 2021-04-26
AnP creation tweaks


chrisamiller:
1) Add "No Configuration" as an option in the AnP menu-items list you see when naively typing "genome analysis-projects create NAME"

2) If no menu item is specifically selected, don't add everything by default.

Both of these would prevent a lot of confusion (and cleanup) from inexperienced users.

This is the right file, I haven't had time to hack into it yet. This is a note for whoever finds time first:
https://github.com/genome/genome/blob/master/lib/perl/Genome/Config/AnalysisProject/Command/Create.pm

tmooney:
Implementing number 1 would require replacing the existing chooser with a custom selector.  (The chooser is limited to menu items and surely a `No Configuration` menu item would be a little strange and self-contradictory :smile:)  The chooser also doesn't have a way to choose nothing--that's why there's a `--no-config` option to use that circumvents the whole thing.

A few options we could implement:
* Add a duplicate "confirm" dialog if someone chooses everything.  (This is kind of silly, since the user already actively confirmed once...)
* Restrict the selection to only one menu item, so they're forced to choose which menu item they want and can't blindly hit `c`.  (This is inconvenient if someone really wanted a few menu items.)

If creating without a configuration is the common case, I'd suggest we switch the `--no-config` option to default on so someone has to explicitly specify they want to add configuration.  (Personally I might prefer to remove configuration adding from this command altogether and let folks use `add-menu-item` if they want to add menu item(s). I was previously told that combining things into single commands is handy, though, so I'd want to get a consensus on changing it.)

chrisamiller:
> Add a duplicate "confirm" dialog if someone chooses everything. (This is kind of silly, since the user already actively confirmed once...)
I really don't think the prompt is clear that you're choosing everything unless you read very carefully.  Maybe the solution is just changing that prompt's text to be more direct? :

`
Please confirm that you would like to add all of the above configurations 'analysis_menu_items' or modify your selection
`

Or maybe even adding an extra interactive step?  

```
$ genome analysis-project create "asdf" -e prod-builder
  Creating project "asdf".
  Would you like to choose from a menu of preset configurations, or start with no configs and add your own?
(m)enu of items  (n)o config

```

chrisamiller:
resolved by tmooney's prs

PR #2001, opened 2021-04-28
Replace --no-config option with --prompt-for-config option.

tmooney:
This is meant to address the concern raised in #2000 that users are intending to use no menu items and instead add all of them.


chrisamiller:
I'm not sure if I like this solution. I think having the menu pop up as a default is a good thing for new users, but I just wish the default wasn't to add all of them (or at least would make it really clear that they were all being added!)

PR #2002, opened 2021-04-28
AnP Create: Now with More Interactivity!

tmooney:
Here's an alternative proposal to address #2000 that adds a prompt and a bit more text about interacting with the chooser.

### DIFF
--- a/lib/perl/Genome/Config/AnalysisProject/Command/Create.pm
+++ b/lib/perl/Genome/Config/AnalysisProject/Command/Create.pm
@@ -125,6 +125,19 @@ sub _get_menu_items {
     } elsif ($self->no_config()) {
         return [];
     } else {
+
+        my $add_config = $self->_ask_user_question(
+            'Would you like to add configuration from the preset config menu items?',
+            60,
+            'y|n',
+            'y',
+            '(y)es, add config / (n)o, start with no config',
+        );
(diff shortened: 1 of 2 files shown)


lib/perl/Genome/Config/AnalysisProject/Command/Create.pm
```diff
@@ -125,12 +125,25 @@ sub _get_menu_items {
     } elsif ($self->no_config()) {
         return [];
     } else {
+
+        my $add_config = $self->_ask_user_question(
+            'Would you like to add configuration from the preset config menu items?',
+            60,
```
chrisamiller, line 131: what does the 60 here do?

chrisamiller (approved):
I like this a lot and think it makes things more user friendly

lib/perl/Genome/Config/AnalysisProject/Command/Create.pm
tmooney: It's the "timeout".  If you don't answer in 60 seconds, it `exit`s.  (That's actually the default in `_ask_user_question`, but made explicit here.)  Can also use `0` for unlimited time.

chrisamiller: got it! 👍

PR #2005, opened 2021-04-29
Basic config file changes

johnegarza:
Minimal changes to support adding environmental and model config files to AnPs. Instead of copying new files to the appropriate local directory, as compute1 permissions prevent this when run by regular users, the file is instead copied to a GCP bucket. Jenkins then periodically checks the bucket for new files, pushes them to the appropriate github repo, and uses git to sync storage0 and storage1.

### DIFF
--- a/lib/perl/Genome/Config/Profile/Item.pm
+++ b/lib/perl/Genome/Config/Profile/Item.pm
@@ -150,8 +150,22 @@ sub _copy_file_to_allocation {
 
     my $filename = File::Basename::basename($original_file_path);
     my $destination_file_path = File::Spec->catdir($allocation->absolute_path, $filename);
-    Genome::Sys->copy_file($original_file_path, $destination_file_path);
-    $allocation->reallocate();
+    if ($ENV{UR_DBI_NO_COMMIT}) {
+        Genome::Sys->copy_file($original_file_path, $destination_file_path);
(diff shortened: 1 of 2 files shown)


lib/perl/Genome/Config/Profile/Item.pm
```diff
@@ -150,8 +150,23 @@ sub _copy_file_to_allocation {
...
+        Genome::Sys->shellcmd(
+            cmd => [
+                '/usr/bin/gsutil/gsutil',
+                'cp',
+                $original_file_path,
+                'gs://gms_environment_config/' . $model_config,
+            ],
+            input_files => [$original_file_path],
+        );
+        $self->status_message('Model config file queued for installation')
```
tmooney, line 168: I don't think anything in the official documentation refers to these as "model config files", so I might advocate referring to it as a "config file" or a "config item file" or "config item" or somesuch.

tmooney (commented):
This generally looks good. I think we should be careful with the terminology--it's already a challenge to keep it all straight. 😄

lib/perl/Genome/Config/Profile/Item.pm
johnegarza: Done!

tmooney (approved):
+1 ☁️

PR #2006, opened 2021-05-03
Fix uploaded file name

johnegarza:
Add in a forgotten `.` to the constructed file name and make sure that it does not contain the original file name's suffix, following the naming convention expected by a downstream jenkins script used for syncing.

### DIFF
--- a/lib/perl/Genome/Config/Profile/Item.pm
+++ b/lib/perl/Genome/Config/Profile/Item.pm
@@ -156,7 +156,7 @@ sub _copy_file_to_allocation {
     }
     else {
         my ($file_name, $file_dirs, $file_suffix) = File::Basename::fileparse($config_path, qr/\.[^.]+/);
-        my $config_file = 'model.' . $self->id . '.' . $file_name . '.yaml';
+        my $config_file = join('.', 'model', $self->id, $file_name, 'yaml');
         Genome::Sys->shellcmd(
             cmd => [
                 '/usr/bin/gsutil/gsutil',
--- a/lib/perl/Genome/Config/Profile/Item.pm
+++ b/lib/perl/Genome/Config/Profile/Item.pm
@@ -155,7 +155,8 @@ sub _copy_file_to_allocation {
         $allocation->reallocate();
     }
     else {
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Config/Profile/Item.pm
```diff
@@ -155,7 +155,8 @@ sub _copy_file_to_allocation {
         $allocation->reallocate();
     }
     else {
-        my $config_file = 'model.' . $self->id . $filename . '.yaml';
+        my ($file_name, $file_dirs, $file_suffix) = File::Basename::fileparse($config_path, qr/\.[^.]+/);
+        my $config_file = 'model.' . $self->id . '.' . $file_name . '.yaml';
```
tmooney, line 159: This might make it easier to avoid missing `.`s if the name changes again later.
```suggestion
        my $config_file = join('.', 'model', $self->id, $file_name, 'yaml');
```

tmooney (approved):
+1, the suggestion is entirely optional if you like it!

lib/perl/Genome/Config/Profile/Item.pm
```diff
@@ -155,7 +155,8 @@ sub _copy_file_to_allocation {
         $allocation->reallocate();
     }
     else {
-        my $config_file = 'model.' . $self->id . $filename . '.yaml';
+        my ($file_name, $file_dirs, $file_suffix) = File::Basename::fileparse($config_path, qr/\.[^.]+/);
```
tmooney, line 158: If you don't need the dirs and suffix you can omit them.  (This suggestion also corrects the variable name from the nonexistent `$config_path`.)
```suggestion
        my ($file_name) = File::Basename::fileparse($original_file_path, qr/\.[^.]+/);
```

tmooney (commented):
Looks like this line is why tests are failing.

lib/perl/Genome/Config/Profile/Item.pm
johnegarza: Thanks, forgot to edit my copy/paste!

PR #2007, opened 2021-05-04
Skip allocation path creation for config files

johnegarza:
These changes retain the ability to write directly to the filesystem, for use in testing or by the service account running CQID, while otherwise skipping allocation path creation for new environment and config files. The paths are created by the jenkins task that syncs the new files.

### DIFF
--- a/lib/perl/Genome/Config/Profile/Item.pm
+++ b/lib/perl/Genome/Config/Profile/Item.pm
@@ -132,7 +132,7 @@ sub has_model_for {
 sub _create_allocation_for_file {
     my $self = shift;
     my $file_to_store = shift;
-    my $skip_allocation_path_creation = shift // 1;
+    my $skip_allocation_path_creation = (shift // 1) && !$ENV{UR_DBI_NO_COMMIT};
 
     my $allocation = Genome::Disk::Allocation->create(
         owner_id            => $self->id,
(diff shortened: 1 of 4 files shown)


lib/perl/Genome/Disk/Detail/Allocation/Creator.pm
```diff
@@ -29,7 +32,9 @@ sub create_allocation {
     my $allocation_object = $self->_get_allocation_without_lock(
         \@candidate_volumes);
 
-    $self->create_directory_or_delete_allocation($allocation_object);
+    if ($ENV{UR_DBI_NO_COMMIT} or not $skip_allocation_path_creation) {
```
tmooney, line 35: I kind of think this should just be
```perl
unless ($skip_allocation_path_creation) {
```
There's no particular reason why allocations themselves can't support skipping path creation under no-commit.  So I'd suggest we move this logic out and handle it in the cases that care.

As an added bonus, we could add a testcase to the `.t` that makes sure we don't create the path when the flag is enabled!

lib/perl/Genome/Config/Profile/Item.pm
```diff
@@ -131,26 +132,29 @@ sub has_model_for {
 sub _create_allocation_for_file {
     my $self = shift;
     my $file_to_store = shift;
+    my $skip_allocation_path_creation = shift // 1;
```
tmooney, line 135: Maybe this becomes:
```perl
shift // !$ENV{UR_DBI_NO_COMMIT}
```
or even:
```perl
(shift // 1) && !$ENV{UR_DBI_NO_COMMIT}
```
if we want no-commit to disable the option regardless of whatever was passed in.

tmooney (commented):
Generally makes sense to me!  I like the way you figured out for still making the concretization work.

tmooney (approved):
+1 🎿

PR #2009, opened 2021-05-06
Add an update env file command

johnegarza:


tmooney (approved):
+1, part of me wants to suggest having a common parent for the add and update commands that knows what the filename format is for the staged file in Google.

I do think in the long run we should make the bucket name a`genome config` item instead of hard-coding it.

tmooney (approved):
+1, knowing the no-commit case doesn't work.  I can make a follow-up PR for the config variable. (That will involve more commits to the docker repos to set said variable, too.)

PR #2010, opened 2021-05-10
Use env variables instead of hardcoding gcp bucket

johnegarza:


tmooney (approved):
+1.  This also opens the door to easily using a different bucket for testing if we so choose at some point.

PR #2012, opened 2021-05-13
Skip disk querying

johnegarza:
Following #2007, direct disk writes to create allocation paths are now skipped in most cases (when unprivileged users are running commands) in favor of queuing them for creation by a Jenkins task running as a service user with the proper permissions. However, disk queries to ensure that the volume has space still occur, which causes an error due to the way the core gms volume is mounted on compute1. This patch removes some of the direct disk queries, while still preserving checks that query the database to determine if there should be free space. Since the database is updated when allocations are created, even when the paths are not, this should be safe.

### DIFF
--- a/lib/perl/Genome/Disk/Detail/Allocation/Creator.pm
+++ b/lib/perl/Genome/Disk/Detail/Allocation/Creator.pm
@@ -28,10 +28,8 @@ sub create_allocation {
     $self->wait_for_database_pause;
 
     my @candidate_volumes = $self->candidate_volumes;
-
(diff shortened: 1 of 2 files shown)


tmooney (approved):
+1 assuming tests pass. If we go with this, we can replace these overrides with this new feature, too:
https://github.com/genome/genome/blob/c2e9669d39352367f4e327e09c9f70fe1c11bda5/lib/perl/Genome/Site/TGI/Command/ImportDataFromLims.pm#L250-L251

PR #2013, opened 2021-05-20
Support deleting a config profile item with linked tags.

tmooney:
This will still fail if the config was ever used for models (which prevents crazy cascading deletes!), but an unused config that happened to have a tag can now be cleaned up.


johnegarza (approved):

chrisamiller:
+1

PR #2014, opened 2021-05-25
Use AnP config for build creation/running in refseqs.

tmooney:
This should make defining references within someone's non-core AnP go a little more smoothly!


jasonwalker80 (approved):
+1

PR #2015, opened 2021-05-25
Add a "synonym file" converter algorithm.

tmooney:
This is an alternative that is similar to the `chr_md5` approach but with an explicit synonym file such as can be used with VEP.


jasonwalker80 (approved):
+1, I like this, 🥇

PR #2016, opened 2021-05-27
Tolerate the user specifying an explicit volume instead of a group.

tmooney:
The remote volume support regressed the tolerance added in #1866 in this particular case.  It could still be avoided by using the right type of data being requested, though!

### DIFF
--- a/lib/perl/Genome/Site/TGI/Command/ImportDataFromLims.pm
+++ b/lib/perl/Genome/Site/TGI/Command/ImportDataFromLims.pm
@@ -228,6 +228,10 @@ sub _create_allocation {
     );
 
     my @volumes = Genome::Disk::Detail::Allocation::Creator->get_candidate_volumes(disk_group_name => $params{disk_group_name});
+    unless (@volumes) {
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Site/TGI/Command/ImportDataFromLims.pm
```diff
@@ -228,6 +228,10 @@ sub _create_allocation {
     );
 
     my @volumes = Genome::Disk::Detail::Allocation::Creator->get_candidate_volumes(disk_group_name => $params{disk_group_name});
+    unless (@volumes) {
+        #sometimes users pass volumes when a group is needed.  The allocation system tolerates this, so let's tolerate it here, too.
+        push @volumes, Genome::Disk::Volume->get(mount_path => $params{disk_group_name});
+    }
```
chrisamiller, line 234: seems handy  +1

chrisamiller (approved):

jasonwalker80 (approved):
+1

PR #2017, opened 2021-06-01
Use a hardcoded lims docker image.

tmooney:
This purports to look it up in config, but the config is a lie.  The latest version of the config is itself stored inside the docker image, so it's not possible to look it up from the outside.  Rather than rely on an old version of the config, just put in the current latest value.


chrisamiller (approved):
+1

PR #2019, opened 2021-06-15
Handle not having automatic mounts in Cleanup Succeeded

tmooney:
This creates an option to submit sub-jobs for the abandon process to give a chance to mount the requisite volumes for the abandonment process.

This also includes a cleanup command to look for allocations for abandoned builds (because sometimes disk is unreliable...).


chrisamiller (approved):
+1  yeah, that oughta help

PR #2022, opened 2021-06-17
The help text claims it assigns all data, so let's do it.

tmooney:
This has been a minor annoyance for awhile.


jasonwalker80 (approved):
+2

PR #2024, opened 2021-06-23
Monkeypatch to get around longstanding bug in Logger.

tmooney:
I think we should switch users of this to use the UR logging methods instead, but this is a quick fix for now.


chrisamiller (approved):
+1  I don't really know the background here, but certainly doesn't seem malicious or likely to break anything!

tmooney:
(I should've put this in the original description:) There's a bug in a module we're using that's been around since 2017: https://rt.cpan.org/Public/Bug/Display.html?id=120329

Since UR has various log levels already built into it, I think we should eliminate our dependence here if we can.  But this bug was hitting me right now when I try `genome model build start`, so this was a quicker fix for that issue.

PR #2026, opened 2021-07-19
Only try to assign existing data when there is some.

tmooney:
This goes along with the just-merged #2022... Everything still mostly works without this tweak--it's just noisier than it needs to be 😄


chrisamiller (approved):
+1

PR #2027, opened 2021-07-21
Add a per-AnP limit for scheduled builds.

tmooney:
This way a single project won't monopolize the queues when it has a job group configured that causes its builds to PEND even when slots are available for other builds to run.

The existing version was being particularly careful with its DB queries, so we'll want to keep an eye on performance to make sure calling `$model->analysis_project->id` isn't slowing stuff down too much.


jasonwalker80 (approved):
+2, LGTM, unsure why a test is failing…

chrisamiller (approved):
+1

PR #2028, opened 2021-07-26
updating cwltool path

chrisamiller:
path changed in more recent docker versions


tmooney:
After having built the new image and tested it, we actually want to keep the existing path here.  Both paths have a version of cwltool.  `/usr/bin/cwltool` is the packaged version from Ubuntu, and `/usr/local/bin/cwltool` is the version from `pip` (which is the one we want to be using here).  The issue with recent versions is that the `pip` version wasn't actually getting installed like we wanted so only the Ubuntu version was visible.

chrisamiller:
good to know!

PR #2029, opened 2021-07-29
I believe this should be kill-docker like it's submit-docker.

tmooney:
It only matters if you have certain other settings enabled that we weren't using before.  But now it does matter 😄


chrisamiller (approved):
+1 seems correct based on here https://cromwell.readthedocs.io/en/stable/backends/SGE/

PR #2030, opened 2021-08-09
Some fixes to a log message.

tmooney:
The `::` in place of `->` was preventing this from proceeding if any
given AnP was at its cap.


chrisamiller (approved):
+1

PR #2033, opened 2021-09-01
A queue for requesting stop/abandon for production builds.

tmooney:
For CwlPipeline builds, now they can be requested to be stopped or abandoned in addition to being restarted.  Since stop and abandon are destructive, only the AnP owner (or an admin) can make these requests.

### DIFF
--- a/lib/perl/Genome/Model/Command/Admin/CleanupSucceeded.pm
+++ b/lib/perl/Genome/Model/Command/Admin/CleanupSucceeded.pm
@@ -2,6 +2,7 @@ package Genome::Model::Command::Admin::CleanupSucceeded;
 
 class Genome::Model::Command::Admin::CleanupSucceeded {
     is => 'Command::V2',
+    roles => ['Genome::Model::Command::Submittable'],
     doc => 'Abandon unsuccessful builds that have been superceded.',
     has => [
         models => {
@@ -22,8 +23,6 @@ class Genome::Model::Command::Admin::CleanupSucceeded {
 use strict;
 use warnings;
 use Genome;
-
-use Genome::Sys::LSF::bsub qw();
 
 sub help_detail {
     'Abandon unsuccessful builds that have been superceded.'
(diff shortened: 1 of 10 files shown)


lib/perl/Genome/Model/Command/Admin/CleanupSucceeded.pm
```diff
@@ -93,21 +92,9 @@ sub _submit_abandon_jobs {
...
         my $anp = Genome::Config::AnalysisProject->get($anp_id);
 
-        my @vol = $anp->possible_volumes;
-
-        my $anp_guard = $anp->set_env;
-        my $volume_guard = Genome::Config::set_env('docker_volumes', join(' ', map { "$_:$_" } @vol));
-        my $image_guard = Genome::Config::set_env('lsb_sub_additional', sprintf('docker0(%s)', $ENV{LSB_DOCKER_IMAGE})); #use the current image regardless of the AnP config
-        unless($ENV{LSF_DOCKER_NETWORK} eq 'host') {
-            $self->fatal_message('Parent container must have LSF_DOCKER_NETWORK=host set.');
-        }
```
chrisamiller, line 103: what does this do exactly?

lib/perl/Genome/Model/CwlPipeline/Command/ProcessActionQueue.pm
```diff
@@ -0,0 +1,74 @@
...
+Takes the actions requested for CWLPipeline builds.
+EOHELP
+}
+
+sub execute {
+    my $self = shift;
+
+    my @to_process = Genome::Model::Build::CwlPipeline->get(
+        action_requested => ['stop', 'abandon'],
+        -limit => $self->max_count,
```
chrisamiller, line 35: what does this leading dash do?

lib/perl/Genome/Model/CwlPipeline/Command/Requeue.pm
```diff
@@ -34,23 +40,38 @@ EOHELP
...
+            unless($build->run_by eq Genome::Config::get('builder_run_as_user')) {
+                $self->error_message('Can only %s builds for the configured builder user (%s). For your own builds, use `genome model build %s` directly.', $action, Genome::Config::get('builder_run_as_user'), $action);
+                next BUILD;
+            }
 
-        unless($build->run_by eq Genome::Config::get('builder_run_as_user')) {
-            $self->error_message('Can only requeue builds for the configured builder user (%s). Restart your own builds directly with `genome model cwl-pipeline restart`.', Genome::Config::get('builder_run_as_user'));
-            next BUILD;
+            $DB::single = 1;
+            unless($build->model->analysis_project->created_by eq Genome::Sys->username or Genome::Sys->current_user_is_admin) {
```
chrisamiller, line 65: which users are currently admins?  prod-builder and mgi-svc-bga-run?

lib/perl/Genome/Model/Command/Admin/CleanupSucceeded.pm
tmooney: `LSF_DOCKER_NETWORK=host` is necessary for the container to be able to communicate with child interactive LSF jobs.  This allows the terminal stdout/stderr from the subjobs to be included directly in the stdout/stderr of the current process.

(Obviously this particular line now does nothing, since this is existing code that was relocated to the helper role in this PR 😄.)

lib/perl/Genome/Model/CwlPipeline/Command/Requeue.pm
tmooney: Try: `genome sys user list -f user_roles.name=admin`
If you're `mgi-svc-bga-run`, you can stop and abandon builds directly.  But we could make that user an admin, too.

lib/perl/Genome/Model/CwlPipeline/Command/ProcessActionQueue.pm
tmooney: It's part of the parameter name.  It's a UR-internal parameter so it uses a leading dash to avoid conflicting with any potential property names on the object itself.

chrisamiller (approved):
+1

PR #2034, opened 2021-11-02
Enable GCP workflows via environment configuration options

johnmaruska:
Enable GCP workflows by setting `cwl_runner` to `cromwell_gcp`

Additional settings required are:
- `workflow_deps_zip`, path to a .zip file containing WDL dependencies, likely pointing at analysis-wdls repository
- `cromwell_workflow_options`, path to a .json file for Cromwell settings, different for each lab at minimum
- `cromwell_gcp_bucket` which bucket to place Cromwell generated files like outputs and logs

gcloud uses user permissions, so the user must `gcloud auth login` before sending off jobs

### DIFF
--- a/lib/perl/Genome/Cromwell.pm
+++ b/lib/perl/Genome/Cromwell.pm
@@ -132,17 +132,17 @@ sub _send_request {
     my $req = shift;
 
     my $self = $class->_singleton_object;
-  ATTEMPT: for (1..5) {
-      my $response = $self->user_agent->request($req);
+    ATTEMPT: for (1..5) {
+        my $response = $self->user_agent->request($req);
 
-      unless ($response->is_success) {
-          $self->fatal_message('Error querying server: %s', $response->status_line);
-          sleep 5;
-          next ATTEMPT;
-      }
+        unless ($response->is_success) {
+            $self->fatal_message('Error querying server: %s', $response->status_line);
+            sleep 5;
+            next ATTEMPT;
+        }
 
-      return $response->decoded_content;
-  }
+        return $response->decoded_content;
+    }
 
     $self->fatal_message('Failed to query server after serveral attempts.');
 }
--- a/lib/perl/Genome/Cromwell.pm
+++ b/lib/perl/Genome/Cromwell.pm
@@ -35,7 +35,7 @@ class Genome::Cromwell {
                 return $ua;
             },
         },
-        ],
+    ],
 };
 
 sub status {
--- a/lib/perl/Genome/Cromwell.pm
+++ b/lib/perl/Genome/Cromwell.pm
@@ -5,6 +5,7 @@ use warnings;
 
 use JSON qw(to_json from_json);
 use HTTP::Request;
+use HTTP::Request::Common qw(POST);
 use LWP::UserAgent;
 use IO::Socket::SSL qw();
 use IPC::Run qw();
@@ -34,9 +35,18 @@ class Genome::Cromwell {
                 return $ua;
             },
         },
-    ],
+        ],
 };
 
+sub status {
+    my $class = shift;
+    my $workflow_id = shift;
+
+    my $self = $class->_singleton_object;
+    my $url = $self->_request_url($workflow_id, 'status');
+
+    return $self->_make_json_request('GET', $url);
+}
 
 sub outputs {
     my $class = shift;
@@ -80,6 +90,24 @@ sub timing {
     return $content;
 }
 
+sub submit_workflow {
+    my $class = shift;
+    my ($definition, $inputs, $dependencies, $options) = @_;
+
+    my $self = $class->_singleton_object;
+    my $url = $self->_request_url();
+    my $content = [ workflowSource  => [$definition],
+                    workflowInputs => [$inputs],
+                    workflowDependencies => [$dependencies],
+                    workflowOptions => [$options] ];
+    my $req = POST( $url,
+                    Content_Type => 'multipart/form-data',
+                    Content => $content );
+
+    return from_json($self->_send_request($req));
+}
+
+
 sub _request_url {
     my $class = shift;
     my @parts = @_;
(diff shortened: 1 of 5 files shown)


apipe-tester:

tmooney:
Would the CI test this please? Thanks!

lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
```diff
@@ -177,6 +179,97 @@ sub run_cromwell {
...
+            wait_for_completion => 1,
+            log_file => "$logdir/pull_outputs.log",
+            cmd => [
+                "python3", "/opt/scripts/pull_outputs.py",
+                $workflow_id, "--output=$results_dir", "--cromwell-url=$cromwell_url"
+            ]
+            );
+        $self->_generate_timing_report($workflow_id);
+    } else {
+        $self->fatal_message("Workflow $workflow_id has non-succeeded status $status using workflow definition $main_workflow_file. Logs at $logdir");
```
tmooney, line 247: Is it true that the logs are at `$logdir`?  Looks like the log gets put there in the step that would happen after this `fatal_message`.

lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
```diff
@@ -177,6 +179,97 @@ sub run_cromwell {
...
+    }
+
+    $self->_fetch_cromwell_log($workflow_id, $workflow_options, $logdir);
+
+    $guard = Genome::Config::set_env('lsb_sub_additional', $lsb_sub_additional);
```
tmooney, line 252: `Genome::Config::set_env` is a `Scope::Guard` object that resets the value when the reference goes out of scope, so this particular line doesn't do too much.  The old value was about to be restored anyway.  If you do want to be explicit about resetting the previous value, this could also be `undef $guard` instead.

lib/perl/Genome/Cromwell.pm
```diff
@@ -99,6 +127,26 @@ sub _make_json_request {
     return from_json($content);
 }
 
+sub _send_request {
+    my $class = shift;
+    my $req = shift;
+
+    my $self = $class->_singleton_object;
+  ATTEMPT: for (1..5) {
```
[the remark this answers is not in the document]
johnmaruska: ugh that must have been emacs aggressive-indent-mode

lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
johnmaruska: I'm pretty sure I saw the logs populate after this message but that may be unsafe behavior. I'll switch to inside the branches just to be sure

lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
johnmaruska: I didn't know that it automatically reset when it goes out of scope, neat

lib/perl/Genome/Cromwell.pm
```diff
@@ -34,9 +35,18 @@ class Genome::Cromwell {
                 return $ua;
             },
         },
-    ],
+        ],
```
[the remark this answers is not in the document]
johnmaruska: aggressive-indent strikes again, thought i avoided these

lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
```diff
@@ -177,6 +179,95 @@ sub run_cromwell {
     $self->_stage_cromwell_outputs($results_dir);
 }
 
+sub run_cromwell_gcp {
+    my $self = shift;
+    my $yaml = shift;
+    my $results_dir = shift;
+
+    my $zip_deps = Genome::Config::get('workflow_deps_zip');
```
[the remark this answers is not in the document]
johnmaruska: It can be generated from the analysis-wdls directory that it contains. Outside of GMS I do that on submit
https://github.com/griffithlab/cloud-workflows/blob/main/scripts/submit_workflow.sh#L66-L76

lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
```diff
@@ -177,6 +181,103 @@ sub run_cromwell {
...
+    my $cloud_yaml = $yaml; $cloud_yaml =~ s/.ya?ml/_cloud.yaml/;
+    my $guard = Genome::Config::set_env('lsb_sub_additional', 'docker(jackmaruska/cloudize-workflow:1.0.0)');
+    delete local $ENV{BOTO_CONFIG};
+
+    Genome::Sys::LSF::bsub::bsub(
+        queue => $queue,
+        user_group => $user_group,
+        resource_string => 'rusage[mem=512M:internet2_upload_mbps=500]',
+        wait_for_completion => 1,
+        log_file => "$logdir/cloudize_workflow.log",
```
tmooney, line 211: We tend to use `File::Spec->join` to build these paths.  You can see a few examples of it elsewhere in this file.

lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
```diff
@@ -177,6 +181,103 @@ sub run_cromwell {
...
+            $yaml,
+            "--output=$cloud_yaml"] );
+
+    $self->debug_message('input yaml: %s', $yaml);
+    $self->debug_message('cloud yaml: %s', $cloud_yaml);
+
+    # Zip dependencies
+    my $deps_zip = "$data_dir/deps.zip";
+    my $prev_dir = getcwd;
+    chdir(dirname($main_workflow_file));
```
tmooney, line 226: It might be my preference to use `File::Spec`'s `splitpath` instead of relying on `dirname`, but I concede that this is probably fine here 😄

tmooney (approved):
+1 ☁️

chrisamiller (approved):
+1  I initially wondered whether `jackmaruska/cloudize-workflow:1.0.0` should be a config var, but on second reflection, I think it's likely to remain fairly stable and hardcoding it makes sense - a change/version bump requires a new version of genome, but that's probably okay

PR #2036, opened 2021-12-14
Updates to support per-job containers when running with toil.

tmooney:
Any container used in the workflow must be capable of running python3 with glibc version of at least 2.16 so that toil's wrapper can run.  (This is similar to how Cromwell requires every image to have a bash shell so its wrapper can run.)  Python doesn't actually need to be installed--we have a copy we mount in the container so long as the shared libraries are compatible.


chrisamiller (approved):
+1 - seems reasonable

tmooney:
Rebased to resolve conflict with #2035.

PR #2037, opened 2021-12-21
Add config for specifying the queue for interactive jobs.

tmooney:
We were relying on the fact that the default queue is interactive, which is not ideal!


chrisamiller (approved):
+1

PR #2038, opened 2021-12-21
Commands to back BQM and RQB on the new CI server

tmooney:
The old versions of these were in scripts lying about in different places.  By exposing them this way hopefully they're easier to maintain (and also to run manually if needed for some reason).


chrisamiller (approved):
+1  seems useful to have these more exposed

PR #2039, opened 2022-01-06
Propagate cromwell_gcp_project setting to gms/start.sh script

johnmaruska:
During my tests with this, the `--project` setting in gcloud commands called by gms/start.sh was implicitly pulling the setting from the environment, set by `gcloud config set project PROJECT` instead of using the value set in genome config. 

This commit fixes this by propagating the value into gms/start.sh which has been updated to accept that argument and use it in the creation of the cromwell VM. 
[cloud-workflows/gms/start.sh updates here](https://github.com/griffithlab/cloud-workflows/commit/49b62f602febaf76459af46ece717fa49bd364e3)


apipe-tester:

chrisamiller (approved):
+1

tmooney:
Would the CI test this please?  (I thought it was supposed to trust you more by now...)

tmooney (approved):
+1 🫓

PR #2041, opened 2022-01-11
Only modify tmp basedir if new location is writable.

tmooney:
For some reason, some servers we're using create a doubly-nested LSF
tmpdir structure, and the inner one is r/w only by root.  In this case
they're already giving us the first level as $ENV{TMPDIR} so we can just
trust it anyway.  This code is still useful on our older cluster,
though, so not just removing it outright yet.

Is this the first instance of using perl 5.10 filetest stacking in our codebase? It might be :smile:


chrisamiller (approved):
+1 I've encountered this behavior and always had to hack around it. Fix is welcome

PR #2042, opened 2022-01-12
Use fully qualified domain name for database server.

tmooney:
Some hosts don't have a friendly search path configured, so this is a little more certain to work as expected.  This should help with some intermittent failures.


chrisamiller (approved):
+1

PR #2043, opened 2022-01-13
Consistently use Genome::Sys to test readability of files.

tmooney:
To handle non-POSIX permissions we need to not rely on the default behavior of -r.


chrisamiller (approved):
+1

PR #2044, opened 2022-01-13
Revise check for matching hosts.

tmooney:
We don't care if they're "blades". In fact, we only care that the
command tries to return any results at all, so check for the header row!
When there aren't any hosts, it'll return this instead:

`Not enough host(s) currently eligible`

So looking for the line to start with `HOST` is good enough.


chrisamiller (approved):
+1 appreciate that you changed the commented out ones too 😆   Good ol' find and replace

tmooney:
Haha yeah, I almost just removed those lines instead, but you never know if maybe we'll update those to have variable memory requests... any day now!

PR #2045, opened 2022-01-18
Use the Genome::Sys method to get the java executable path.

tmooney:
There is no `/gapp` on the new cluster, so use our config instead since we already have that mechanism for controlling where to look!


chrisamiller (approved):
+1

PR #2046, opened 2022-01-18
Claim that .t test per-lane BAMs are all 1024KB.

tmooney:
The tests try to allocate space when revivifying the per-lane BAMs. When the bam size is unknown, the default size request is 100,000,000KB. When temp on the host running the test has less than this amount of space free, the test can fail.

(If temp has less than 1024KB free, we're probably going to have other trouble, anyway!)


chrisamiller (approved):
+1 seems reasonable

PR #2047, opened 2022-01-20
Some straggling test fixes

tmooney:
* Disable DV2 MethRatio test (by popular consensus that it isn't being used).
* FASTQ splitter properly checks it can write to the output directory it will use.


chrisamiller (approved):
+1

PR #2048, opened 2022-01-21
GCP subnet and separate buckets

johnmaruska:
Two changes to reflect different infrastructure for GCP runs
- specify a subnet to run the compute VM. initial setup for GMS+GCP runs creates a subnet with a firewall to only allow access inside of WashU network
- ~~separate buckets for inputs and executions files. also handled by initial setup, different buckets have different lifecycle rules so execution files which are no longer needed can be expired out without losing inputs which may be reused consistently. output files expected to be pulled to local storage, not persisted between buckets~~


johnmaruska:
I did a test run with somatic_exome that went all the way through as expected

johnmaruska:
Remove separate buckets, @chrisamiller and I talked a bit and decided on single bucket with expiry rule. Example data and large static files (e.g. vep_cache) handled separately

lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
```diff
@@ -388,7 +390,7 @@ sub _generate_workflow_options_gcp {
 
     my $contents = <<"EOCONFIG"
 {
-    "final_workflow_log_dir": "gs://$cromwell_gcp_bucket/build.$build_id/logs",
+    "final_workflow_log_dir": "gs://$cromwell_gcp_bucket/build.$build_id"/logs,
```
chrisamiller, line 393: is this quotation mark supposed to be inside the slash or at the end?

johnmaruska: This quote is haunting me. I have no idea why it keeps changing back but this will be the fourth time I've fixed it?

johnmaruska: Ugh it was just my last push didn't go through

tmooney (approved):
+1 🚢

PR #2049, opened 2022-01-24
Add a tiebreaker in the sorting algorithm.

tmooney:
If perchance both items have identical timestamps, use ID as a tiebreaker so the ordering is still consistent between them.

This test has been failing intermittently so hopefully this helps with that!


chrisamiller (approved):
+1

PR #2050, opened 2022-01-24
Don't worry about if we have space for allocations in this test.

tmooney:
Currently CWL scratch directories reserve 50GB, but the test uses local temp and doesn't require nigh that much space anyway!

This is another intermittent failure that came from landing on a blade with <50GB `/tmp` space.


chrisamiller (approved):
+1

PR #2051, opened 2022-01-25
For test stability, have each build created at a different time.

tmooney:
It is true that some of our commands will fail if multiple builds were
created within the same second.  There's locking in place that prevents
that from happening in practice, so this test was failing for an edge
condition we don't normall encounter.  (If we truly have multiple
simultaneous builds in the same status, picking one arbitrarily to
execute is the "right" behavior here, anyway.)


chrisamiller (approved):
+1

PR #2052, opened 2022-01-26
If there are multiple Cromwell workflows, use the most recent.

tmooney:
If a build fails after the Cromwell workflow has succeeded, then a
build `restart` may create a new workflow in which case there may
be multiple succeeded workflows.  Instead of crashing, we'll trust
that the most recent one is the one we care about.


chrisamiller (approved):
seems reasonable +1

PR #2053, opened 2022-01-26
Don't use "last_complete_build" in this test.

tmooney:
Our command uses the creation date to determine which build is most recent for purposes of abandoning prior attempts.  However if an older build completed after a later build, then our test would be grabbing the wrong one for being "saved".  (Instead let's keep tabs on the last one we made, since it should be the most-recently-created and thus the one to be saved.

(This is a followup to #2051 which made the test builds actually distinguishable by the command to begin with.)


chrisamiller (approved):
+1

PR #2054, opened 2022-01-26
Rename "prod-builder" env option to "automated".

tmooney:
Since "prod-builder" isn't really descriptive and is also the name of the default runner user, it was suggested we use a neutral term instead, especially for those circumstances where a different runner is being used.


chrisamiller (approved):
+1

PR #2055, opened 2022-02-03
Update repo version of class browser cache and tab-completion spec.

tmooney:
We already update these routinely as part of the snapshot process, but
the job that was keeping the repo updated has been disabled for awhile.
This is just prep for turning it back on.


chrisamiller (approved):
+1

PR #2056, opened 2022-02-16
Fix enumeration of valid values for --action parameter.

tmooney:
They're actually defined on the model class and not the derived build class.  Also, a little explanation for "0" seems helpful!


chrisamiller (approved):
+1

PR #2057, opened 2022-03-08
Only return true if we succeeded to process all instrument data.

tmooney:
The motivation behind this is so that CQID will decline to process data that fails to import.


chrisamiller (approved):
looks good to me!

PR #2058, opened 2022-04-05
Use mgibio image for cloudize-workflows in cromwell_gcp workflows

johnmaruska:
We're moving cloudize-workflows image from my personal dockerhub to the mgibio dockerhub. The contents are still the same repo here: https://github.com/griffithlab/cloud-workflows/

### DIFF
--- a/lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
+++ b/lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
@@ -231,7 +231,7 @@ sub run_cromwell_gcp {
     # Cloudize workflow
     #
     my $cloud_yaml = $yaml; $cloud_yaml =~ s/.ya?ml/_cloud.json/;
-    my $lsb_sub_guard = Genome::Config::set_env('lsb_sub_additional', 'docker(jackmaruska/cloudize-workflow:1.3.1)');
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
```diff
@@ -231,7 +231,7 @@ sub run_cromwell_gcp {
     # Cloudize workflow
     #
     my $cloud_yaml = $yaml; $cloud_yaml =~ s/.ya?ml/_cloud.json/;
-    my $lsb_sub_guard = Genome::Config::set_env('lsb_sub_additional', 'docker(jackmaruska/cloudize-workflow:1.3.1)');
+    my $lsb_sub_guard = Genome::Config::set_env('lsb_sub_additional', 'docker(mgibio/cloudize-workflow:1.1.0)');
```
tmooney, line 234: Does the new image's 1.1.0 correspond to the old 1.3.1 or is this a different version?

johnmaruska: Newer but just adds scripts unrelated to running in GMS. Versioning was very wild-west and I only recently started doing real releases so there isn't really a way to get the old one without detective work. I can push through a test run to confirm behavior if you'd like

tmooney (approved):
+1 🚢

PR #2061, opened 2022-04-16
updating cloudize-workflow image

chrisamiller:
updating to 1.2.2


tmooney (approved):
+1, though part of me thinks we could make this a config entry so it can be changed in the AnP config.

chrisamiller:
Agreed. Making an issue for that now so I don't forget to do it later!

tmooney:
Would the CI be so kind as to test this please?

PR #2064, opened 2022-05-30
Add multithreading and suppress output

johnegarza:
Add the `-m` flag to run the upload in parallel and reduce upload time, and add the `-q` flag to prevent listing every uploaded file (currently over 250).


tmooney (approved):
+1 🎏

PR #2065, opened 2022-08-04
Support restarting if our local build fails but the cloud VM is still running.

tmooney:
Basically if a cloud VM already exists for our build then we'll treat it as our own and monitor it the usual way rather than start from scratch.

This is an incremental improvement, but it'd be good to also handle the case where the cloud VM has already completed before we restarted and we need to resume with pulling down the artifacts and outputs.

### DIFF
--- a/lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
+++ b/lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
@@ -227,84 +227,90 @@ sub run_cromwell_gcp {
 
     my $poll_interval_seconds = 300;
 
-    #
-    # Cloudize workflow
-    #
-    my $cloud_yaml = $yaml; $cloud_yaml =~ s/.ya?ml/_cloud.json/;
     my $lsb_sub_guard = Genome::Config::set_env('lsb_sub_additional', 'docker(mgibio/cloudize-workflow:1.2.2)');
     delete local $ENV{BOTO_CONFIG};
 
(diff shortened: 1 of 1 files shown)


lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
```diff
@@ -227,90 +227,93 @@ sub run_cromwell_gcp {
...
-            $yaml,
-            "--output=$cloud_yaml"] );
-
-    # Zip dependencies
-    my $deps_zip_path = File::Spec->join($data_dir, 'deps.zip');
-    my $deps_zip_url = "gs://$bucket/build.$build_id/deps.zip";
-    my $prev_dir = getcwd;
-    my(undef, $deps_dir, undef) = File::Spec->splitpath($main_workflow_file);
+    if ($self->_cloud_instance_is_running) {
+        $self->status_message("Detected already running cloud VM for job.  Will monitor for completioni every $poll_interval_seconds seconds.");
```
chrisamiller, line 234: s/completioni/completion/

chrisamiller (approved):
minor typo, otherwise looks good!

lib/perl/Genome/Model/CwlPipeline/Command/Run.pm
tmooney: Oops, good old "typing `i` while already in insert mode"...

PR #2066, opened 2023-03-27
Fix typo

johnegarza:
Instead of silently defaulting to md5 hashing, this should hopefully speed up workflows by actually using `xxh64` instead.


tmooney (approved):
+1,234,567,890 (sigh)

PR #2067, opened 2023-04-11
Centralize checking for the existence of eligible LSF hosts.

tmooney:
There were a few competing implementations around for this; better that we have one that we can fix up when/if issues arise.

(A lot of this seems like a good candidate for outright deprecation and removal at this point... but this PR is a quick change to get tests to pass again for now.)


chrisamiller (approved):
+1

chrisamiller (approved):
+1

PR #2068, opened 2023-04-11
Skip Memcache.t that depends on production memcache.

tmooney:
It's also not really safe for concurrent test runs!

Realistically we're more likely to remove this than to fix it, since memcache was used for the perl-generated webviews but we have moved on to genome/spectacle now.


chrisamiller (approved):
+1

PR #2070, opened 2023-09-11
Don't crash if `bsub` for inner job fails.

tmooney:
A misconfigured AnP could result in the bsub failing for various reasons.  Better to skip it and move on than to prevent any other queue processing.

This isn't a perfect solution--enough misconfigured AnPs could still clog up the queue--but it's better than a single one bringing it down.  It'd be nicer if we ended up with an `Unstartable` build instead of still failing out, but that would require some bigger changes to how this all works that I'm not convinced are worth it.


chrisamiller (approved):
+1 seems reasonable

PR #2071, opened 2023-10-18
Only tie read-only since we don't need to write this file.

tmooney:
Also, report the error when the `tie` fails instead of mysteriously failing to parse the header later.

(The test files are mounted read-only now, so the tests that covered this would fail on the `tie` with a `Read-only file system` message.  Users should be able to import instrument data with a read-only input TSV anyway.)


chrisamiller (approved):
seems reasonable +1

PR #2072, opened 2023-10-26
Only cleanup if we started to try to do work.

tmooney:
We shouldn't try to clean up if we're erroring out before attempting the archive-especially if the data was already archived!

This is another PR that's easier to read with the `?w=1` URL parameter 🙂


chrisamiller (approved):
+1

tmooney:
There were some shortcomings of the first version of this PR so I wrote a few tests and redid it.

acoffman (approved):

PR #2073, opened 2023-11-13
Remove LIMS-specific extensions.

tmooney:
With the retirement of the legacy LIMS system these are no longer functional.  (Many of the classes under `Genome::Site::TGI::Synchronize` are also likely candidates for removal, but there will be a final database snapshot we could theoretically query for a time as needed.)


chrisamiller (approved):
+1   👋

PR #2074, opened 2023-11-15
Define a "queue" runtime attribute for WDL tasks.

tmooney:
This will let certain jobs be dispatched to different queues.


chrisamiller (approved):
+1

PR #2075, opened 2023-11-20
Avoid attempting to connect to OLTP database

tmooney:
This fixes up `Sys.pm` to be more judicious about which database sources it tries to disconnect to avoid inadvertently processing connection info to a datasource that wasn't connected.

This also disables tests that depend on an OLTP connection.  Still holding off on outright removing modules that depend on it in case we need to refer back to the database dump later.


chrisamiller (approved):
+1

PR #2076, opened 2023-11-21
Use a smaller allocation size for testing.

tmooney:
The test data is about 50MB so a 100MB allocation provides plenty of room.  This prevents failures when running tests on a blade with <32GB of free temp space, since we definitely don't need that much!

(Probably should've done this years ago 🙂)


chrisamiller (approved):
+1

PR #2077, opened 2024-03-12
Retry postgres connections for up to 90 seconds.

tmooney:
We sometimes get our TCP connection failing trying to connect, so let's try harder to successfully connect.  (This depends on genome/docker_genome_perl_environment#39 to work on our cluster.  Which is to say that after this is merged, any builds that want to take advantage will need to update their docker images.)


chrisamiller (approved):
+1 here's hoping this works!

